PR #8, opened 2019-11-06
Add thriftpool package

fishy:
This package defines an thrift client pool interface, and comes with a
channel based implementation.

The benchmark result shows that the channel based implementation takes
~100ns for a Get/Release pair calls.

    $ go test -bench=. -benchmem
    goos: darwin
    goarch: amd64
    pkg: github.com/reddit/baseplate.go/thriftpool
    BenchmarkPoolGetRelease/channel-8         	12180511	        95.9 ns/op	       0 B/op	       0 allocs/op
    PASS
    ok  	github.com/reddit/baseplate.go/thriftpool	1.288s

### DIFF
--- a/thriftpool/channel.go
+++ b/thriftpool/channel.go
@@ -1,0 +1,120 @@
+package thriftpool
+
+import (
+	"sync/atomic"
+)
+
+type channelPool struct {
+	pool           chan Client
+	opener         ClientOpener
+	numActive      int32
+	minConnections int
+	maxConnections int
+}
+
+// Make sure channelPool implements Pool interface.
+var _ Pool = (*channelPool)(nil)
+
+// NewChannelPool creates a new thrift client pool implemented via channel.
+func NewChannelPool(minConnections, maxConnections int, opener ClientOpener) (Pool, error) {
+	if minConnections > maxConnections {
+		return nil, &ConfigError{
+			MinConnections: minConnections,
+			MaxConnections: maxConnections,
+		}
+	}
+
+	pool := make(chan Client, maxConnections)
+	for i := 0; i < minConnections; i++ {
+		c, err := opener()
+		if err != nil {
+			return nil, err
+		}
+		pool <- c
+	}
+
+	return &channelPool{
+		pool:           pool,
+		opener:         opener,
+		minConnections: minConnections,
+		maxConnections: maxConnections,
+	}, nil
+}
+
+// Get returns a thrift client from the pool.
+func (cp *channelPool) Get() (client Client, err error) {
+	defer func() {
+		if err == nil {
+			atomic.AddInt32(&cp.numActive, 1)
+		}
+	}()
+
+	select {
+	case c := <-cp.pool:
+		if c.IsOpen() {
+			return c, nil
+		}
+	default:
+	}
+
+	if cp.IsExhausted() {
+		err = ErrExhausted
+		return
+	}
+	return cp.opener()
+}
+
+// Release releases a client back to the pool.
+//
+// If the pool is full, the client will be closed instead.
+//
+// Calling Release after Close will cause panic.
+func (cp *channelPool) Release(c Client) error {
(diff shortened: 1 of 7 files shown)


fishy:
I have two open questions here:

1. For the package path, do you guys prefer `thriftpool` or `thriftbp/pool`?
2. Do we want to leave the list implementation in, or just remove it and only provide the channel implementation?

thriftpool/list.go
```diff
@@ -0,0 +1,128 @@
...
+	pool           []Client
+	opener         ClientOpener
+	numActive      int32
+	minConnections int
+	maxConnections int
+}
+
+var _ Pool = (*listPool)(nil)
+
+// NewListPool creates a new thrift client pool implemented via lise/slice.
```
pacejackson: typo:

```suggestion
// NewListPool creates a new thrift client pool implemented via list/slice.
```

thriftpool/list.go
```diff
@@ -0,0 +1,129 @@
...
+	minConnections int
+	maxConnections int
+}
+
+// Make sure listPool implements Pool interface.
+var _ Pool = (*listPool)(nil)
+
+// NewListPool creates a new thrift client pool implemented via list (slice).
+//
+// In most cases you should use the channel implementation instead.
```
pacejackson: Is there a reason someone would use it?  It might make sense to only do the `channelPool` implementation if not, and if so a comment with some guidance about when you would use this might be nice.

thriftpool/channel.go
```diff
@@ -0,0 +1,120 @@
...
+			atomic.AddInt32(&cp.numActive, 1)
+		}
+	}()
+
+	select {
+	case c := <-cp.pool:
+		if c.IsOpen() {
+			return c, nil
+		}
+	default:
```
pacejackson: Am I correct in understanding that this means that if the pool is exhausted, we don't block at all to wait for a connection to be released?  That seems like something we might want to support rather than immediately returning an error?

thriftpool/channel.go
```diff
@@ -0,0 +1,120 @@
...
+		c = newC
+	}
+
+	select {
+	case cp.pool <- c:
+		atomic.AddInt32(&cp.numActive, -1)
+		return nil
+	default:
+		// Pool is full, just close it instead.
+		return c.Close()
```
pacejackson: 🔕 I was trying to think through _how_ this can happen, is it because the following can happen:

1. 2 goroutines call `Get`
2. Both try to get a connection while the pool is empty
3. _Before_ they check `IsExhausted`, a connection is released back into the pool
4. Both goroutines open new connections and return those

thriftpool/channel.go
pacejackson: Ah, I forgot that we don’t necessarily start with a full pool, this makes more sense given that.

fishy: Yes :)

thriftpool/channel.go
pacejackson: Note, this can also happen because the pool doesn’t necessarily start full, so there can be 1 slot left and 2 goroutines calling “get” at the same time, so they both can pass the exhausted check and make 2 new connections

thriftpool/list.go
fishy: In the case that they might call `Release` after `Close`, maybe? But yea I don't feel strongly to keep this implementation.

fishy:
:haircut:

thriftpool/channel.go
```diff
@@ -0,0 +1,120 @@
+package thriftpool
```
konradreiche: I like `thriftpool` over `thriftbp/pool` as it keeps the package depth low.

fishy:
> This looks good, I'd maybe suggest adding a test which tests the `thriftpool` in a concurrent scenario to make sure we catch and test for regression of any data races.

Done. PTAL.

PR #10, opened 2019-11-09
Update README.md

scravy:
Link existing baseplate.py project


README.md
```diff
@@ -1,5 +1,5 @@
 # baseplate.go
-Baseplate implemented in go
+[Baseplate](https://github.com/reddit/baseplate.py) implemented in go
```
fishy: I don't think this is the correct change, although it indeed provides more context than before.

PR #12, opened 2019-11-11
filewatcher: Use atomic.Value instead of lock

fishy:
Also fix a bug of storing nil value when parser fails, and add a test
case for that.

### DIFF
--- a/filewatcher/filewatcher_test.go
+++ b/filewatcher/filewatcher_test.go
@@ -2,10 +2,12 @@ package filewatcher_test
 
 import (
 	"context"
+	"errors"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


filewatcher/filewatcher_test.go
```diff
@@ -95,8 +95,8 @@ func TestFileWatcher(t *testing.T) {
 func TestFileWatcherTimeout(t *testing.T) {
 	interval := time.Millisecond
 	filewatcher.InitialReadInterval = interval
-	round := interval * 10
-	timeout := round * 8
+	round := interval * 20
```
konradreiche: What is this change about?

fishy: on my local machine it takes right above 85ms for this one (it would pass if it's <85ms without any changes)

PR #14, opened 2019-11-25
Set up Zipkin Tracer

lbelaus:
**Description**
Set up Zipkin tracer and OpenTracing bridge for using using Zipkin as the backend.


tracing/tracer.go
```diff
@@ -0,0 +1,94 @@
...
+	if addr == "" {
+		// TODO: log info level
+		return
+	}
+
+	zipkinReporter := zipkinhttp.NewReporter(addr)
+
+	endpoint, err := zipkin.NewEndpoint(serviceName, "0.0.0.0:0")
+	if err != nil {
+		log.Fatalf("unable to create local endpoint: %+v\n", err)
```
fishy: Let's not use the stdlib log in library code. Instead, pass in a log [`Wrapper`](https://godoc.org/github.com/reddit/baseplate.go/log#Wrapper), use that to do logging (example: https://github.com/reddit/baseplate.go/blob/903bbee5dcd49766b720b2bbd585dfa88f9f7195/filewatcher/filewatcher.go#L71-L73), and return an error here.

Or we can just return the error back to the caller and let the caller to decide what to do (logging and/or panic).

Same as the other log usages.

tracing/tracer.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"fmt"
+
+	"github.com/opentracing/opentracing-go"
+	zipkinot "github.com/openzipkin-contrib/zipkin-go-opentracing"
+	zipkin "github.com/openzipkin/zipkin-go"
+	"github.com/openzipkin/zipkin-go/reporter"
+	zipkinhttp "github.com/openzipkin/zipkin-go/reporter/http"
+	"github.com/reddit/baseplate.go/log"
```
fishy: nit: Please add a blank line before this line, so that we have 3 import groups: stdlibs, third party packages, and packages from the same repo/go module.

tracing/tracer.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Tracer is exported to support start tracible server etc
+var Tracer = opentracing.GlobalTracer()
+
+// InitTracer configure the global tracer
+func InitTracer(
+	serviceName, addr string,
+	sampleRate float64,
+	logger log.Wrapper,
+) {
```
fishy: The old `log.Fatalf` will also call panic, so now without that we do need to return an error from this function when something bad happens, so the caller would know to call panic.

tracing/tracer.go
```diff
@@ -0,0 +1,103 @@
...
+		zipkin.WithSampler(sampler),
+		zipkin.WithLocalEndpoint(endpoint),
+	)
+
+	tracer := zipkinot.Wrap(nativeTracer)
+	opentracing.SetGlobalTracer(tracer)
+
+	if logger != nil {
+		logger("Global tracer created")
+	}
```
fishy: I don't think this log is necessary. As a principal of library and logging, library code should only do logging when it's something bad and it cannot return the error to the caller (in most cases, it should try avoid logging as best as possible, only return the error to the caller and let the caller to decide how to handle that).

Also the other logger usages should have the nil check :)

fishy:
You can try run `go mod tidy` and commit the changes it made to fix the drone problem. If it doesn't fix the drone problem, probably your go is not the latest version.

tracing/tracer.go
lbelaus: Yea, I was hoping we will extend existing logger.Wrapper to panic rather than return error every time we call logger.

fishy: not all log messages should cause panic (e.g. the info log you are adding at the end, or the logging you added in `EndSpan` call), so that really should be handled by the caller instead of the library code.

lbelaus: What I'm trying to say is that we need to add different log levels to the logger, so we can log info, debug, error, fatal.

This specific function is called during start up, and I think it makes sense to fail with non-zero status code.

fishy: Yes but that means:
1. You need to proper document it (in the function comment) that the passed in log wrapper is supposed to call panic
2. That's kinda anti-pattern and error-prone (user who don't read the documentation carefully will likely be unaware of that)

Alternatively, you can just call panic after calling `logger`.

Also as a library code we really don't need info and debug level logs in most cases. Those are more suitable for service/application code.

In this case (`InitTracer` function), I don't really see why we can't just return the err to the caller without any logging, and the caller just do:

```
if err := tracing.InitLogger(...); err != nil {
  log.Fatalf(err)
}
```

tracing/tracer.go
```diff
@@ -17,26 +17,22 @@ var zipkinReporter reporter.Reporter
...
+		return errors.New("Endpoint to send spans to is undefined")
 	}
 
 	zipkinReporter := zipkinhttp.NewReporter(addr)
 
 	endpoint, err := zipkin.NewEndpoint(serviceName, "0.0.0.0:0")
-	if err != nil && logger != nil {
-		logger(fmt.Sprintf("Unable to create local endpoint: %+v\n", err))
+	if err != nil {
+		return errors.Wrap(err, "Unable to create local endpoint")
```
fishy: We are on go1.13, so we don't really need third party library (`github.com/pkg/errors`) for that :) We can just use `fmt.Errorf("Unable to create local endpoint: %w", err)` to wrap them.

lbelaus: go1.13 error wrapping doesn't contain stack traces for where the original error occurred

lbelaus:
I think we cannot expect checksums for `golint` to be the same, because we do `go get -u golang.org/x/lint/golint` as part of the lint step of drone workflow.

If we add the dependency to go.mod, `go mod tidy` removes it, because it's unused dependency. I think swapping lint and check-go-modules should work.

fishy:
please remove the last two commits, none of the code depends on golint. if you run `go get .../golint` outside of this directory it won't affect your go.mod and go.sum files

fishy:
^ sorry, never mind

tracing/tracer.go
fishy: nit: Since those wrapped errors all have different prefixes, it would be easy to figure out which return it is, I don't think the call stack actually brings a lot of extra value here. If we really want the call stacks I'd prefer to pass in the logger which prints the call stack while logging, instead of wrap the call stack into the error (both because we need to use `%+v` to get them, and if the logging itself also prints the call stack the two call stacks will get mixed together).

But if you really prefer this one, I'm fine with it.

tracing/tracer.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+var zipkinReporter reporter.Reporter
+
+// Tracer is exported to support start tracible server etc
+var Tracer = opentracing.GlobalTracer()
+
+// InitTracer configure the global tracer
+func InitTracer(serviceName, addr string, sampleRate float64,
+) error {
```
fishy: nit: this line break is kinda weird.

PR #17, opened 2019-11-27
tracing: Add thrift integration

fishy:
Add helper functions to create span from thrift context and to create
context from span.

Please note that this is a breaking change as we move away from
opentracing abstraction to use zipkin-go directly.


fishy:
:haircut: @lbelaus 

As discussed on slack, I made changes to move away from opentracing abstraction to use zipkin library directly. Also added tests and moved string set to its own package. PTAL.

fishy:
👓 @andizzle

set/string_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+func TestStringSet(t *testing.T) {
+	const (
+		key1 = "key1"
+		key2 = "key2"
+	)
+
+	var s set.String
+
+	t.Run(
```
konradreiche: Is that an alternative to table-driven testing to make debugging of tests easier?

fishy: This is more like control the order of sequential (sub-)tests, and also share state between them.

PR #18, opened 2019-12-09
use TempDir to create watchable dir prior to tmp file creation

andizzle:
dir created by `ioutil.TempFile` is not watchable by fsnotify (tested in OS X).

Use `ioutil.TempDir` to create a watchable dir prior the tmp file.


konradreiche:
Great catch, does that fix the too many open files error by any chance?

fishy:
> Great catch, does that fix the too many open files error by any chance?

Yes :)

Issue #16, opened 2019-11-27
Secret callback


andizzle:
When fetched secret is used for creating a downstream client, it's possible that we need to create a new client once the secret is updated. Add capability of being notified of updated secret via secret store will allow us to renew a client as needed.

Maybe something like 
```go
func NewStore(ctx context.Context, path string, logger log.Wrapper, notify chan interface) (*Store, error) {
	parser := func(r io.Reader) (interface{}, error) {
		secrets, err := NewSecrets(r)
		if err != nil {
			return nil, err
		}
                notify <- secrets
		return secrets, nil
	}
        ...
```

Also open for other suggestions.

fishy:
I like the channel idea. Just two things to note:

1. We shouldn't change `NewStore` because most of the use cases doesn't need the callback. We can create a new one like `NewStoreWithCallback`
2. The `notify <- secrets` should probably be in its own goroutine, otherwise if the caller didn't read from the channel timely, this would block the parser.

konradreiche:
What does a downstream client mean? Is that a client implementation you are providing yourself outside of baseplate.go?

fishy:
> What does a downstream client mean? Is that a client implementation you are providing yourself outside of baseplate.go?

Yes.

andizzle:
This is addressed by #20

fishy:
We still need to allow registering of middlewares after initialization, but that can be in its own ticket.

PR #19, opened 2019-12-09
add in store with callback

andizzle:
In respond to #16

### DIFF
--- a/secrets/store.go
+++ b/secrets/store.go
@@ -44,6 +44,36 @@ func NewStore(ctx context.Context, path string, logger log.Wrapper) (*Store, err
 	}, nil
 }
 
+// NewStoreWithNotify returns a new instance of Store by configuring it
+// with a filewatcher to watch the file in path for changes ensuring secrets
+// store will always return up to date secrets.
+//
+// In addition to NewStore, a channel is passed in to allow updated secret
+// be picked up outside of the store. This is mainly used as a signal to trigger
+// subsequent actions with the secret.
+//
+// Context should come with a timeout otherwise this might block forever, i.e.
+// if the path never becomes available.
+func NewStoreWithNotify(ctx context.Context, path string, logger log.Wrapper, notify chan *Secrets) (*Store, error) {
+	parser := func(r io.Reader) (interface{}, error) {
+		secrets, err := NewSecrets(r)
+		if err != nil {
+			return nil, err
+		}
+		go func() {
+			notify <- secrets
+		}()
+		return secrets, nil
+	}
+	result, err := filewatcher.New(ctx, path, parser, logger)
+	if err != nil {
+		return nil, err
+	}
+	return &Store{
+		watcher: result,
+	}, nil
+}
+
 // GetSimpleSecret fetches a simple secret or error if the key is not present.
 func (s *Store) GetSimpleSecret(path string) (SimpleSecret, error) {
 	var secret SimpleSecret
--- a/secrets/store_test.go
+++ b/secrets/store_test.go
@@ -95,6 +95,63 @@ func TestNewStore(t *testing.T) {
 	}
 }
 
+func TestNewStoreWithNotify(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected *Secrets
+	}{
+		{
+			name: "specification example",
+			input: `
+					{
+						"secrets": {
+							"secret/myservice/some-api-key": {
+								"type": "simple",
+								"value": "Y2RvVXhNMVdsTXJma3BDaHRGZ0dPYkVGSg==",
+								"encoding": ""
+							}
+						}
+					}
+			`,
+			expected: &Secrets{
+				simpleSecrets: map[string]SimpleSecret{
+					"secret/myservice/some-api-key": SimpleSecret{"Y2RvVXhNMVdsTXJma3BDaHRGZ0dPYkVGSg=="},
+				},
+				versionedSecrets:  map[string]VersionedSecret{},
+				credentialSecrets: map[string]CredentialSecret{},
+			},
+		},
+	}
+
+	dir, err := ioutil.TempDir("", "secret_test_")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, tt := range tests {
+		tt := tt // capture range variable for parallel testing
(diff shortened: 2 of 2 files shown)


secrets/store.go
```diff
@@ -44,6 +44,36 @@ func NewStore(ctx context.Context, path string, logger log.Wrapper) (*Store, err
...
+// Context should come with a timeout otherwise this might block forever, i.e.
+// if the path never becomes available.
+func NewStoreWithNotify(ctx context.Context, path string, logger log.Wrapper, notify chan *Secrets) (*Store, error) {
+	parser := func(r io.Reader) (interface{}, error) {
+		secrets, err := NewSecrets(r)
+		if err != nil {
+			return nil, err
+		}
+		go func() {
+			notify <- secrets
```
fishy: On second thought, put this in a goroutine might not be such a good idea. If the enqueue blocks, and then the secret changed again and we have another enqueue blocked, then later the channel get unblocked, I'm not sure we guarantee that the two blocked enqueue will be guaranteed the order we intended.

We probably should do that in the same goroutine, and document it in the function's comment (and probably also provide an example) that the caller is supposed to unblock it timely.

secrets/store.go
```diff
@@ -44,6 +44,36 @@ func NewStore(ctx context.Context, path string, logger log.Wrapper) (*Store, err
...
+// with a filewatcher to watch the file in path for changes ensuring secrets
+// store will always return up to date secrets.
+//
+// In addition to NewStore, a channel is passed in to allow updated secret
+// be picked up outside of the store. This is mainly used as a signal to trigger
+// subsequent actions with the secret.
+//
+// Context should come with a timeout otherwise this might block forever, i.e.
+// if the path never becomes available.
+func NewStoreWithNotify(ctx context.Context, path string, logger log.Wrapper, notify chan *Secrets) (*Store, error) {
```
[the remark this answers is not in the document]
andizzle: huh didn't realize that. A public struct with only private properties and functions doesn't have a wide use case. We probably should at leave give it a `GetType()` or open up its properties.

External testing is a good idea.

secrets/store.go
andizzle: Only the last change matters. Instead of accepting a channel, what if we create and return a channel from this function, this way we control how it's created. Give it a buffer of 1, and check if there's a message in the channel prior to push one, if so we pop the old message first.

```go
func NewStoreWithNotify(ctx context.Context, path string, logger log.Wrapper) (*Store, chan *Secrets, error) {
        notify := make(chan *Secrets, 1)
	parser := func(r io.Reader) (interface{}, error) {
		secrets, err := NewSecrets(r)
		if err != nil {
			return nil, err
		}
                 if len(notify) > 0 {
                        <- notify
                }
		notify <- secrets
		return secrets, nil
	}
	result, err := filewatcher.New(ctx, path, parser, logger)
	if err != nil {
		return nil, err
	}
	return &Store{
		watcher: result,
	}, notify, nil
}
```

secrets/store.go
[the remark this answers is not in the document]
konradreiche: Agreed, `Secret` should become private in a sense but I guess that won't be the case when I understood your issue properly, then it definitely should be opened up. I generally do like giving more access to underlying functionality for better library extension.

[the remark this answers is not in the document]
konradreiche: If we use channels as an argument, it should be locked down to its directional type. So I think we only want to receive from the channel so it should be `<-chan *Secrets`.

secrets/store.go
konradreiche: I wonder if there is an implementation here that would let the using end implement the concurrency, I am very wary of adding channels to an a public API. The reasoning is basically there are a lot of principals that need to be followed to make that work.

fishy: @andizzle in your example I don't think there's a way to guarantee atomicy between `if len(notify) > 0` and the next line, so it's possible that at `if` it's larger than 0, but on the next line the caller's consumer already consumed it, which makes the next line a dead lock. Also this is just too complex.

@konradreiche I think with proper documentation to clear the responsibilities of the caller, use channel in a public API is fine. But if we don't like it, the alternative is probably create a mapping of secret path to callbacks, something like `map[string]func(path strnig, secret GenericSecret)`.

konradreiche: @fishy I was thinking of callbacks, too which would be the only alternative being fairly safe but I'd think it would be a shame not to utilize channels here.

This seems like a good list of things to watch out for when adding channels to public APIs https://inconshreveable.com/07-08-2014/principles-of-designing-go-apis-with-channels/

fishy: The link is good. I almost forgot that we can make it non-blocking (select with default), but we certainly don't want to drop the callback with slow responders. Maybe we can allow the caller to pass in a `time.Duration` as the max time we would block for, and drop the callback after that? Not sure if that's a worse idea than just blocking forever.

I don't think blocking forever is a bad idea, as long as we proper document it.

konradreiche: I think the callback idea is not the worst either, it could be part of `Secrets` and accessible to the client-side to set it. It has the advantage that the client-side is responsible for implementing the concurrency and there are less things to worry about on our end.

The question is why we'd want to use a channel, channels are for communication but also synchronization but if we don't necessarily need to take care of the synchronization on our end I am not so sure about it.

andizzle: I don't mind prototyping the callback idea. I think it's the easiest for the caller (and safest for us), it's just not as fancy as channel. I picked channel initially because the use case I want to cover was relatively simple. I do not know if we have a case where the secret file changes very frequently. This callback extension is meant to handle third party API client recreate, secret decode etc, where it almost certain that it won't scale well if the secret changes too often, even with straight implementation outside of this lib.

fishy: Taking a step back, we don't really need to send the actual data in this callback. It's possible we just notify the caller that the data changed, and they can just retrieve the data normally as they would do. The problem with that approach right now is that filewatcher only set the data after parser returned, so there's no way to actually get a notification *after* the data is modified. I won't mind adding a post parser callback to filewatcher if we decided that's a good idea to support it.

fishy:
Should we close this one now that https://github.com/reddit/baseplate.go/pull/20 is merged?

PR #20, opened 2019-12-12
Use middleware approach to handle callbacks

andizzle:
Middleware allows caller to add custom functions to work with Secrets as they see fit. This option allows multiple layers callback and also leaves concurrent and blocking problem to the caller.

~We still need to open up Secrets to allow the caller to interact with it.~

This is entirely open for discussion. I just find it easier to express the idea with actual code :)

Middleware chaining benchmark result:
```
go test -bench=. -benchmem * -test.v -run=BenchmarkStoreMiddlewares
goos: darwin
goarch: amd64
BenchmarkStoreMiddlewares/number_of_middlewares_1-12               52531             22815 ns/op             288 B/op          8 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_2-12               51770             23652 ns/op             304 B/op          9 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_4-12               52406             23373 ns/op             336 B/op         11 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_8-12               51429             23478 ns/op             400 B/op         15 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_16-12              45654             23951 ns/op             528 B/op         23 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_32-12              49756             24492 ns/op             784 B/op         39 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_64-12              44978             26391 ns/op            1296 B/op         71 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_128-12             42652             27812 ns/op            2320 B/op        135 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_256-12             36979             30805 ns/op            4368 B/op        263 allocs/op
BenchmarkStoreMiddlewares/number_of_middlewares_512-12             28684             38541 ns/op            8464 B/op        519 allocs/op
PASS
ok      command-line-arguments  15.572s
```

### DIFF
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -31,6 +31,37 @@ type Secrets struct {
 	vault             Vault
 }
 
+// GetSimpleSecret fetches a simple secret or error if the key is not present.
+func (s *Secrets) GetSimpleSecret(path string) (SimpleSecret, error) {
+	secret, ok := s.simpleSecrets[path]
+	if !ok {
+		return secret, ErrorSecretNotFound(path)
+	}
+
+	return secret, nil
+}
+
+// GetVersionedSecret fetches a versioned secret or error if the key is not present.
+func (s *Secrets) GetVersionedSecret(path string) (VersionedSecret, error) {
+	secret, ok := s.versionedSecrets[path]
+	if !ok {
+		return secret, ErrorSecretNotFound(path)
+	}
+
+	return secret, nil
+}
+
+// GetCredentialSecret fetches a credential secret or error if the key is not
+// present.
+func (s *Secrets) GetCredentialSecret(path string) (CredentialSecret, error) {
+	secret, ok := s.credentialSecrets[path]
+	if !ok {
+		return secret, ErrorSecretNotFound(path)
+	}
+
+	return secret, nil
+}
+
 // SimpleSecret represent basic secrets.
 type SimpleSecret struct {
 	Value Secret
@@ -287,4 +318,12 @@ func (e encoding) decodeValue(value string) (Secret, error) {
 		}
 		return Secret(data), nil
 	}
+}
+
+// ErrorSecretNotFound is returned when the key for a secret is not present in
+// the secret store.
+type ErrorSecretNotFound string
+
+func (path ErrorSecretNotFound) Error() string {
+	return "no secret has been found for " + string(path)
 }
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -35,7 +35,7 @@ type Secrets struct {
 func (s *Secrets) GetSimpleSecret(path string) (SimpleSecret, error) {
 	secret, ok := s.simpleSecrets[path]
 	if !ok {
-		return secret, ErrorSecretNotFound(path)
+		return secret, SecretNotFoundError(path)
 	}
 
 	return secret, nil
@@ -45,7 +45,7 @@ func (s *Secrets) GetSimpleSecret(path string) (SimpleSecret, error) {
 func (s *Secrets) GetVersionedSecret(path string) (VersionedSecret, error) {
 	secret, ok := s.versionedSecrets[path]
 	if !ok {
-		return secret, ErrorSecretNotFound(path)
+		return secret, SecretNotFoundError(path)
 	}
 
 	return secret, nil
@@ -56,7 +56,7 @@ func (s *Secrets) GetVersionedSecret(path string) (VersionedSecret, error) {
 func (s *Secrets) GetCredentialSecret(path string) (CredentialSecret, error) {
 	secret, ok := s.credentialSecrets[path]
 	if !ok {
-		return secret, ErrorSecretNotFound(path)
+		return secret, SecretNotFoundError(path)
 	}
 
 	return secret, nil
@@ -320,10 +320,10 @@ func (e encoding) decodeValue(value string) (Secret, error) {
 	}
 }
 
-// ErrorSecretNotFound is returned when the key for a secret is not present in
+// SecretNotFoundError is returned when the key for a secret is not present in
 // the secret store.
-type ErrorSecretNotFound string
+type SecretNotFoundError string
 
-func (path ErrorSecretNotFound) Error() string {
+func (path SecretNotFoundError) Error() string {
 	return "no secret has been found for " + string(path)
 }
--- a/secrets/store.go
+++ b/secrets/store.go
@@ -9,6 +9,14 @@ import (
 	"github.com/reddit/baseplate.go/log"
 )
 
+// Middleware functions
+type (
+	SecretHandlerFunc func(sec *Secrets)
+	SecretMiddleware  func(next SecretHandlerFunc) SecretHandlerFunc
+)
+
+func noOpSecretHandlerFunc(sec *Secrets) {}
+
 // Store gives access to secret tokens with automatic refresh on change.
 //
 // This local vault allows access to the secrets cached on disk by the fetcher
@@ -19,6 +27,8 @@ import (
 // to always have the current version in the face of key rotation etc.
 type Store struct {
 	watcher *filewatcher.Result
+
+	secretHandlerFunc SecretHandlerFunc
 }
 
 // NewStore returns a new instance of Store by configuring it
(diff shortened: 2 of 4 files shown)


secrets/store_test.go
```diff
@@ -95,6 +96,61 @@ func TestNewStore(t *testing.T) {
...
+
+	wg.Add(expectedMiddlewareCalls)
+	var middleware = func(next SecretHandlerFunc) SecretHandlerFunc {
+		return func(sec *Secrets) {
+			middlewareCall++
+			wg.Done()
+			next(sec)
+		}
+	}
+	store.SecretHandler(middleware, middleware)
```
andizzle: This is an example of multi-layered middleware. This can be used to handle callbacks for different secrets, the actual implementation of one or more middleware is up to the caller.

secrets/store.go
```diff
@@ -27,21 +38,43 @@ type Store struct {
...
+
+	if s.secretHandlerFunc != nil {
+		s.secretHandlerFunc(secrets)
+	}
+
+	return secrets, nil
+}
+
+// SecretHandler creates the middleware chain.
+func (s *Store) SecretHandler(middlewares ...SecretMiddleware) {
```
fishy: Is there a reason to export this function? I thought just allow the callers to specify middlewares in `NewStore` should be enough?

secrets/store_test.go
```diff
@@ -95,6 +96,61 @@ func TestNewStore(t *testing.T) {
...
+			"previous": "aHVudGVyMg=="
+		}
+	},
+	"vault": {
+		"url": "vault.net",
+		"token": "17213328-36d4-11e7-8459-525400f56d04"
+	}
+}`
+	tmpFile.Write([]byte(specificationModification))
+	wg.Wait()
```
fishy: This works when the test passes. But when there's a bug and this test doesn't pass, it could actually cause a deadlock.

andizzle:
> The middleware part looks good to me in general, but we still need to export enough fields to `Secret` to make it useful to the callers.

Thought I would have a different PR for that as we haven't decide what/how to open up Secrets.

secrets/store_test.go
andizzle: Yes. Actually I could just test it with NewStore function, no need to make an actual file change to trigger this.

secrets/store.go
andizzle: My mind isn't fully made up. I think in general it's a good idea to allow additional middleware gets attached for some unforeseen reason, but I'm happy to make it private until we have an actual use case.

andizzle:
@konradreiche what do you think of this approach? Also what's your opinion on how to open up Secrets? Adding accessors to secret maps could be one way, we just need to make sure it's read only with mutex.

Also, what's the value in `vault` attribute?
cc @fishy

andizzle:
Added Secrets accessor, PTAL

secrets/store.go
```diff
@@ -27,65 +38,73 @@ type Store struct {
...
+		s.secretHandlerFunc = nopSecretHandlerFunc
+	}
+
+	for _, m := range middlewares {
+		s.secretHandlerFunc = m(s.secretHandlerFunc)
+	}
+}
+
+// GetSimpleSecret loads secrets from watcher, and fetches a simple secret from secrets
+func (s *Store) GetSimpleSecret(path string) (SimpleSecret, error) {
```
fishy: Should we make a breaking change so that `Store` no longer provides `Get*Secret` functions directly, but return the `Secret` instead and user can use `Get*Secret` from `Secret`?

andizzle: I'm very hesitated about it. Ideally we should be introducing v2 if there's breaking change, but is there anything using this at the moment? If not, it's not a breaking change, if there is, how much work are involved to cascade the change?

fishy: We are at v0. You can make breaking changes on v0 (but should be careful about them).

secrets/secrets.go
```diff
@@ -23,6 +23,37 @@ type Secrets struct {
 	vault             Vault
 }
 
+// GetSimpleSecret fetches a simple secret or error if the key is not present.
+func (s *Secrets) GetSimpleSecret(path string) (SimpleSecret, error) {
+	secret, ok := s.simpleSecrets[path]
+	if !ok {
+		return secret, ErrorSecretNotFound(path)
```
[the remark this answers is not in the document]
andizzle: Personally I think returning an error is a more common pattern. It's also a bit easier for caller to log it.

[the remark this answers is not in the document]
fishy: Do we foresee any other errors could be returned here? If not this is an overkill. If we do have other errors to be returned here, do you want callers to actually inspect the error type to be `ErrorSecretNotFound` and handle it differently?

secrets/store.go
fishy: Also cc @konradreiche

andizzle: From a different perspective, `Store` is the primary entry point to secrets, it might be cleaner to keep these methods so that in most client code we can do
```go
        sec, err := Store.Get*Secret()
```

instead of
```go
        secrets, err := Store.GetSecrets()
        if err != nil {
                ...
        }
        sec, err := secrets.Get*Secret()
```

fishy: Is there a reason `GetSecrets` could return an error? If not this would just be `store.GetSecrets().Get*Secret()`.

secrets/secrets.go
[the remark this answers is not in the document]
andizzle: Is there any potential errors related to `s.vault`? Also why do we have this attribute? I wonder if there's any cases where secret validation should be tied into `s.vault`.

I don't want to make too much assumption here :)

secrets/store.go
andizzle: function chaining in Go is not very common and often leads to some problems. The original code does type assert so there's an error to check
```
	secrets, ok := s.watcher.Get().(*Secrets)
```

fishy: That type assertion would never fail: https://godoc.org/github.com/reddit/baseplate.go/filewatcher#Result.Get

andizzle: oh... right. haha.

OK gonna make one last argument here:
`Store` stores a `*Secrets`, every time there's a file change, `Store` creates a new `*Secrets` and put it into filewatch data. I think have a `GetSecrets() *Secrets` is an error prone API, user may thought it's ok to keep using returned `*Secrets` since it's a pointer to the value, but in fact they should call `GetSecrets()` every time. And if they do use the same `*Secrets`, there's no error to report the problem.

We need to document this well, or maybe have `GetSecrets` return a struct and call it out the object may _expire_ when there's a file change, either way it could still lead to problem unknowingly.

fishy: Yea I agree that we need to document this well, but this is similar to the whole "don't cache any secrets and always get them before using" document we already have.

andizzle: If we are happy with it I can make the change. To me the current way is less likely to lead to problems. It's a little redundant, but considering the functions attached to `Secrets` are pretty much only used when there's a callback need, I'm ok with it.

fishy: I'll wait for Konrad to chime in :)

secrets/secrets.go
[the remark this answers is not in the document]
konradreiche: I'd always try to imagine these questions from the caller side. What is the common scenario on the usage side? I think errors are for exceptional behavior, would it be common to access a secret and it wouldn't be there? Then returning a `bool` okay makes sense, if it's more an exceptional behavior I think we should stick to error.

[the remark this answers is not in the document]
fishy: Good point. I think it should remain as an error.

konradreiche:
What the are use cases for the middlewares here? @andizzle you wrote:

> This option allows multiple layers callback and also leaves concurrent and blocking problem to the caller.

What would be a concrete example that would justify adding this additional complexity?

andizzle:
> What would be a concrete example that would justify adding this additional complexity?

Considering the `Store` can hold multiple secrets, it is possible that the main application runs different secret consumers that needs their own callback logic. For example, an application could be using a kafka producer that requires authentication, a thridparty integration point that requires api token, and a message encryption key that encrypts outgoing messages.

Use multiple middlewares helps avoid a big callback function that has mixed secrets load/reload logic. Each secret consumer can hook up with the callback with their own middleware, it's much smaller footprint.

One can also create individual call functions and call then in a final callback functions that is hooked up in the Store, but that requires more organization and planning on the client side, which may not be very intuitive for the caller.

secrets/secrets.go
```diff
@@ -278,3 +309,11 @@ func (e encoding) decodeValue(value string) (string, error) {
 		return string(data), nil
 	}
 }
+
+// ErrorSecretNotFound is returned when the key for a secret is not present in
+// the secret store.
+type ErrorSecretNotFound string
```
fishy: style nit: The go convention is that if it's an error type, it's named `BlaBlaError`. If it's an error variable, it's named `ErrBlaBla`. So this should be `SecretNotFoundError` instead.

fishy: Just realized this is from old code, but you could still fix it :) (but it would be a breaking change)

andizzle:
To Allow register middleware from other lib, we just need to make this function public.

https://github.com/reddit/baseplate.go/pull/20/files#diff-1a3fa2e612f9f352390e03634cc256d0R70

fishy:
> To allow register middleware from other lib, we just need to make this function public.
> 
> https://github.com/reddit/baseplate.go/pull/20/files#diff-1a3fa2e612f9f352390e03634cc256d0R70

It will be slightly more complex than that. After `Store` is initialized, calling that function will not give you the initial secret values, instead, the new middleware will only be called when the next time secrets changes. So we need to call the new middleware with whatever we already have when a new middleware is registered.

andizzle:
I see. The middleware needs to be called upon registration in this scenario. I need to put more thoughts into this. This does make it a compelling reason to keep the middleware approach.

Issue #15, opened 2019-11-25
Add more sophisticated Logger


lbelaus:
Implement logging other that standard log package

fishy:
We use [zap](https://godoc.org/go.uber.org/zap) in our code, it has higher performance than the [go-kit logging library](https://godoc.org/github.com/go-kit/kit/log) we used before. We can port our zap integration into baseplate.go

That being said, this is still more for baseplate users to use, usage by baseplate itself still need to be careful.

fishy:
Closed with #21

PR #21, opened 2019-12-12
Add zap logger integration

sidorovroman:
The logger is requested in the #15. 
The PR ports ads-go zap integration.

~I removed the `type Wrapper func(msg string)` assuming we want to have only one standard of logging. I can return it back if we need it for some cases.~

Commits are logically separated.


filewatcher/filewatcher.go
```diff
@@ -68,9 +67,7 @@ func (r *Result) watcherLoop(
 			return
 
 		case err := <-watcher.Errors:
-			if logger != nil {
-				logger("watcher error: " + err.Error())
-			}
+			log.Errorw("watcher error", "err", err)
```
sidorovroman: do we have a preferable way to log errors? does it seems fine?

filewatcher/filewatcher_test.go
```diff
@@ -255,9 +245,6 @@ func TestParserFailure(t *testing.T) {
 	}
 	// Give it some time to handle the file content change
 	time.Sleep(interval * 10)
-	if atomic.LoadInt64(&loggerCalled) == 0 {
```
sidorovroman: I'm not sure how we can test the logger call without injecting the logger. Is it ok to remove the check or could you recommend a way to test it without injection?

filewatcher/filewatcher.go
fishy: This is fine (as an example). But we do not want to remove the `Wrapper` interface.

There are two parts of the log package:

1. A logger to be used by baseplate.go users (microservices), to log their stuff
2. A way to log some of the library code (other baseplate.go packages), which should be kept to an absolute minimal, like in this case that it's a background goroutine that there's no other way the library can communicate this error to the caller.

log/log.go
```diff
@@ -0,0 +1,220 @@
...
+	case ErrorLevel:
+		lvl = zapcore.ErrorLevel
+	case PanicLevel:
+		lvl = zapcore.PanicLevel
+	case FatalLevel:
+		lvl = zapcore.FatalLevel
+	}
+
+	return lvl
+}
```
fishy: nit: I think this is the same as:
```go
switch loglevel {
default:
	return ZapNopLevel
case DebugLevel:
	return zapcore.DebugLevel
case InfoLevel:
	return zapcore.InfoLevel
case WarnLevel:
	return zapcore.WarnLevel
case ErrorLevel:
	return zapcore.ErrorLevel
case PanicLevel:
	return zapcore.PanicLevel
case FatalLevel:
	return zapcore.FatalLevel
}
```
which is slightly clearer.

filewatcher/filewatcher.go
sidorovroman: I returned the `Wrapper` interface back. Thanks!

log/zap_wrapper.go
```diff
@@ -0,0 +1,41 @@
...
+
+// KitLogger returns a go-kit compatible logger.
+func KitLogger(level Level) KitWrapper {
+	return KitWrapper(StringToAtomicLevel(level))
+}
+
+// SimpleLog is a baseplate.go and apache thrift compatible log wrapper.
+func SimpleLog(msg string) {
+	Error(msg)
+}
```
fishy: This function should be moved to wrapper.go, then it should be made to support levels (similar to the go-kit wrapper):

```go
func ZapWrapper(level zapcore.Level) Wrapper {
	return func(msg) {
		switch zapcore.Level(w) {
		default:
			// for unknown values, fallback to info level.
			fallthrough
		case zapcore.InfoLevel:
			Info(msg)
		case zapcore.DebugLevel:
			Debug(msg)
		case zapcore.ErrorLevel:
			Error(msg)
		case zapcore.PanicLevel:
			Panic(msg)
		case zapcore.FatalLevel:
			Fatal(msg)
		case ZapNopLevel:
			// do nothing
		}
	}
}
```

After you moved this function out, this file can be renamed to kit_wrapper.go as that's what it does.

log/log.go
```diff
@@ -0,0 +1,220 @@
...
+func InitLogger(logLevel Level, cfg ...zap.Config) {
+	if logLevel == NopLevel {
+		logger = zap.NewNop().Sugar()
+		return
+	}
+
+	var config zap.Config
+
+	if len(cfg) > 0 {
+		config = cfg[0]
```
fishy: It's kind of weird that we accept vararg but drop everything except the first one. Maybe we should have a `InitLoggerWithConfig(logLevel Lever, cfg zap.Config)` instead?

cc @andizzle

log/log_bench_test.go
```diff
@@ -0,0 +1,26 @@
+package log
```
fishy: nit: I don't think we really need this file.

fishy:
@sidorovroman Can you also move the package documentation (https://github.com/reddit/baseplate.go/blob/55f6eaf6ee5fa000b68117af98241a7c4d4600d2/log/wrapper.go#L1-L2) into its own file (`doc.go`), and update the content to reflect the changes in this PR?

log/log.go
andizzle: Be more explicit in a lib is a good idea. 👍

log/log.go
andizzle: maybe put `default` at the bottom? In most cases Nop is least used case.

sidorovroman: refactored and moved the default case to the bottom.

log/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package log provides common logging interface used either by baseplate.go users
+// or by other baseplate.go packages.
```
fishy: Package log provides a wrapped zap logger interface for microservices to use, and also a simple Wrapper interface to be used by other baseplate.go packages.

log/log.go
fishy: 🔕 I don't think `default` being at the top or bottom will actually have a performance impact. When it's at the top we still need to compare against all other cases before falling back to `default`. You can check the assembly code it generated to verify that :)

I personally slightly prefer to have `default` at the top in go code as a way to make it more obvious that this `switch...case` block do have a `default` branch, but I don't feel strongly about that.

log/wrapper.go
```diff
@@ -1,8 +1,7 @@
-// Package log provides common logging interface used by other baseplate
-// packages.
 package log
 
 import (
+	"go.uber.org/zap/zapcore"
 	stdlog "log"
 	"testing"
```
fishy: style nit: Please put third party imports at the bottom, with a blank line separating them from stdlib imports.

log/wrapper.go
```diff
@@ -36,3 +35,28 @@ func TestWrapper(tb testing.TB) Wrapper {
...
+			Error(msg)
+		case zapcore.PanicLevel:
+			Panic(msg)
+		case zapcore.FatalLevel:
+			Fatal(msg)
+		case ZapNopLevel:
+			// do nothing
+		}
+	}
+}
```
fishy: On second thought, for performance (it's not that performance in this function really matters that much, but ¯\\\_(ツ)\_/¯), let's do it this way instead:

```go
	// For unknown values, fallback to info level.
	f := Info
	switch logLevel {
	case zapcore.DebugLevel:
		f = Debug
	case zapcore.WarnLevel:
		f = Warn
	case zapcore.ErrorLevel:
		f = Error
	case zapcore.PanicLevel:
		f = Panic
	case zapcore.FatalLevel:
		f = Fatal
	case ZapNopLevel:
		return NopWrapper
	}
	return func(msg string) {
		f(msg)
	}
```
This way we only do the switch once, instead of every time the wrapper is called.

log/log.go
```diff
@@ -0,0 +1,221 @@
...
+	config.EncoderConfig.EncodeLevel = CapitalLevelEncoder
+	config.Level = zap.NewAtomicLevelAt(lvl)
+
+	if lvl == zap.DebugLevel {
+		config.Development = true
+	}
+	// TODO: error handling
+	l, _ := config.Build(zap.AddCallerSkip(2))
+	logger = l.Sugar()
+}
```
fishy: In order to reduce duplicated code, I'd prefer we call `InitLoggerWithConfig` from `InitLogger`. Basically in `InitLogger` we just create the config and then call `InitLoggerWithConfig`.

log/wrapper.go
sidorovroman: looks good! thanks!

log/log.go
```diff
@@ -0,0 +1,214 @@
...
+	FatalLevel Level = "fatal"
+
+	// This will have the same effect as nop but slower
+	ZapNopLevel zapcore.Level = zapcore.FatalLevel + 1
+)
+
+// InitLogger provides a quick way to start or replace a logger
+// Pass in debug as log level enables development mode (which makes DPanicLevel logs panic).
+func InitLogger(logLevel Level) {
+	var config = zap.NewProductionConfig()
```
fishy: style nit: Prefer `config := ...` over `var config = ...`. (https://github.com/uber-go/guide/blob/master/style.md#local-variable-declarations)

sidorovroman: thanks for the link! I updated the code.

PR #29, opened 2019-12-23
Add events package

fishy:


events/events.go
```diff
@@ -0,0 +1,104 @@
...
+	return q.queue.Close()
+}
+
+// Put serializes and puts an event into the event queue.
+func (q Queue) Put(ctx context.Context, event thrift.TStruct) error {
+	ctx, cancel := context.WithTimeout(ctx, q.maxTimeout)
+	defer cancel()
+
+	serializer := thrift.NewTSerializer()
+	serializer.Protocol = thrift.NewTJSONProtocol(serializer.Transport)
```
pacejackson: 🔕 Should we build the serializer on the `Queue` itself (or make it global) rather than rebuilding each time?

fishy: `thrift.TSerializer` is not thread safe (we were bitten by it before :) )

It might make sense to use a resource pool for it, though.

pacejackson: > thrift.TSerializer is not thread safe (we were bitten by it before :) )

What happened?

> It might make sense to use a resource pool for it, though.

I wonder if that might be a premature optimization, a resource pool is a more complicated to implement than just a global/class var and I'm not sure this actually causes problems.

fishy: We tried to use a global serializer for everything and ended up with corrupt data everywhere.

This is actually the perfect use case of a resource pool. The main part we will save from using a resource pool is the initialization and gc time.

fishy: Added resource pool and also tests. PTAL.

events/events_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func TestV2Put(t *testing.T) {
+	const queueSize = 100
+	const doubleTime = DefaultMaxPutTimeout * 2
+	const tripleTime = DefaultMaxPutTimeout * 3
+
+	var v2 *Queue
+	var queue *mqsend.MockMessageQueue
+	t.Run(
```
pacejackson: 🔕 why run setup code like this under `t.Run`?

events/events.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// The default message queue name for v2 events.
+	DefaultV2Name = "v2"
+)
+
+// A Queue is an event queue.
+type Queue struct {
+	queue      mqsend.MessageQueue
+	maxTimeout time.Duration
+	pool       sync.Pool
```
pacejackson: 🔕 would it be better to call this `serializerPool`?

fishy: Done.

events/events_test.go
fishy: Subtests are good because:

1. When a test fails, it shows which subtest failed, so you automatically gets more context info about the failure
2. It limits the scope of `t.Fatal`.

You can think about it as a way to connect a few tests in sequence, and allow state sharing between them.

fishy: But actually it makes less sense in this particular case so I changed it back to just a single test.

PR #30, opened 2019-12-26
secrets: Define Secret type

fishy:
This is a breaking change.

Migrate SimpleSecret and VersionedSecret from string to Secret.

Also some test improvements:

- Close tmp files after writing to make sure the writes are flushed to
  the disk
- Delete the tmp directories after tests
- Reduce the sleep time on TestSecretFileIsUpdated as we flushed the
  writes now

### DIFF
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -15,6 +15,14 @@ const (
 	credentialSecret = "credential"
 )
 
+// A Secret is the base type of secrets.
+type Secret []byte
+
+// IsEmpty returns true if the secret is empty.
+func (s Secret) IsEmpty() bool {
+	return len(s) == 0
+}
+
 // Secrets allows to access secrets based on their different type.
 type Secrets struct {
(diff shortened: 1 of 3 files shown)


fishy:
cc @andizzle

andizzle:
LGTM!

secrets/secrets.go
```diff
@@ -15,6 +15,14 @@ const (
 	credentialSecret = "credential"
 )
 
+// A Secret is the base type of secrets.
+type Secret []byte
+
+// IsEmpty returns true if the secret is empty.
+func (s Secret) IsEmpty() bool {
+	return len(s) == 0
```
andizzle: should we check if it's nil as well?

fishy: nil slice is empty slice (`len(nil)` will return 0).

andizzle: nope we don't need to :)

Issue #6, opened 2019-11-05
Baseplate Message Signature


fishy:
[Baseplate spec link](https://pages.github.snooguts.net/reddit/baseplate.spec/component-apis/signatures). It should be trivial to implement once we had secrets store implementation.

fishy:
Closed with 5b6d028fef9573c95273b729d9a571579dcea970.

PR #31, opened 2019-12-27
Add signing package

fishy:
https://github.com/reddit/baseplate.go/issues/6

~This PR depends on https://github.com/reddit/baseplate.go/pull/30, reviewers can skip the first commit.~

### DIFF
--- a/signing/doc_test.go
+++ b/signing/doc_test.go
@@ -1,0 +1,43 @@
+package signing_test
+
+import (
+	"errors"
+	"time"
+
+	"github.com/reddit/baseplate.go/metricsbp"
+	"github.com/reddit/baseplate.go/secrets"
+	"github.com/reddit/baseplate.go/signing"
+)
+
+func Example() {
+	// Should be properly initialized in production code.
+	var (
+		store      *secrets.Store
+		secretPath string
+		metrics    metricsbp.Statsd
+	)
+
+	secret, _ := store.GetVersionedSecret(secretPath)
+
+	const msg = "Hello, world!"
+
+	// Sign
+	signature, _ := signing.Sign(signing.SignArgs{
+		Message:   []byte(msg),
+		Key:       secret.Current,
+		ExpiresIn: time.Hour,
+	})
+
+	// Verify
+	err := signing.Verify([]byte(msg), signature, secret.GetAll()...)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


signing/doc_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"errors"
+	"time"
+
+	"github.com/reddit/baseplate.go/metricsbp"
+	"github.com/reddit/baseplate.go/secrets"
+	"github.com/reddit/baseplate.go/signing"
+)
+
+func Example() {
+	// Should be proper initialized in prod code.
```
pacejackson:
```suggestion
	// Should be properly initialized in production code.
```

signing/interface.go
```diff
@@ -0,0 +1,116 @@
...
+	// V1: If ExpiresAt is non-zero, it will be used.
+	// Otherwise time.Now().Add(ExpiresIn) will be used.
+	// Note that V1 only defined second precision,
+	// any sub-second precision in ExpiresIn or ExpiresAt will be dropped and
+	// rounded down.
+	ExpiresAt time.Time
+	ExpiresIn time.Duration
+}
+
+// The Interface of a concret version of Baseplate message signing spec.
```
pacejackson:
```suggestion
// The Interface of a concrete version of Baseplate message signing spec.
```

signing/interface.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Signature expiring time.
+	//
+	// V1: If ExpiresAt is non-zero, it will be used.
+	// Otherwise time.Now().Add(ExpiresIn) will be used.
+	// Note that V1 only defined second precision,
+	// any sub-second precision in ExpiresIn or ExpiresAt will be dropped and
+	// rounded down.
+	ExpiresAt time.Time
+	ExpiresIn time.Duration
```
pacejackson: Why is there both `ExpiresAt` and `ExpiresIn`?  Wouldn't it be clearer to pick one?

fishy: So it's easier to use :) I think the documentation clearly explained the priority when both are set, so that should be good enough?

pacejackson: It's up to you ultimately, personally I find it better to only have one option rather than two that do the same thing.  It makes it easier to understand the API just by browsing it rather than having to dive into the documentation and it's one less thing to have to check and remember when using it.

PR #34, opened 2020-01-06
metricsbp: Use influxstatsd instead of statsd implementation

fishy:
go-kit's influxstatsd implementation gives us the ability to use
tags/labels in statsd metrics.

This change changed the underlying type but should be mostly
backward-compatible.

### DIFF
--- a/metricsbp/statsd.go
+++ b/metricsbp/statsd.go
@@ -9,21 +9,22 @@ import (
 	"github.com/reddit/baseplate.go/log"
 
 	"github.com/go-kit/kit/metrics"
-	"github.com/go-kit/kit/metrics/statsd"
+	"github.com/go-kit/kit/metrics/influxstatsd"
 )
 
 // ReporterTickerInterval is the interval the reporter sends data to statsd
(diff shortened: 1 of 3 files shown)


fishy:
cc: @isugimpy

metricsbp/statsd.go
```diff
@@ -59,16 +60,27 @@ type StatsdConfig struct {
...
 	// The log level used by the reporting goroutine.
 	LogLevel log.Level
+
+	// LabelValues are the labels/tags to be attached to every metrics created
+	// from this Statsd object. For labels/tags only needed by some metrics,
+	// use Counter/Gauge/Timing.With() instead.
+	//
+	// LabelValues must be in key value pairs, or NewStatsd will panic:
+	//
+	//     LabelValues: []string{"key1", "value1", "key2", "value2"}
```
pacejackson: 🔕 Can we make the interface here a `map` and translate it into a list before passing to statsd?

fishy: Good point. Done.

PR #36, opened 2020-01-15
runtimebp: Allow set GOMAXPROCS with a customized formula

fishy:
With lower NumCPU reported now, the old formula of NumCPU*2-1 no longer
works. This change will help us to test new formulas, before we can
change to the default formula to a better one later.

### DIFF
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -74,12 +74,26 @@ func readNumberFromFile(path string, buf []byte) (float64, error) {
 	return float64(f), nil
 }
 
-// GOMAXPROCS sets runtime.GOMAXPROCS.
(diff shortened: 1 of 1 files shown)


fishy:
@konradreiche So we used to use `runtime.GOMAXPROCS(runtime.NumCPU()*2 - 1)`, note that that's stdlib `runtime` so it uses the number of CPUs on the physical machine. With the current default `runtimebp` package behavior, we still use the same formula, but changed the number of CPUs to the number assigned to the container.

On the case we deployed the code, the machine has 16 cpus (so we set GOMAXPROCS to 31), while the container is only assigned 2 CPUs (so it's 3). The much lower number set in GOMAXPROCS caused goroutines to be starved.

PR #37, opened 2020-01-17
runtimebp: GOMAXPROCS return two values instead of one

fishy:
I thought runtime.GOMAXPROCS returns the new value set, but apparently I
was wrong and it actually returns the previous value. So make
runtimebp.GOMAXPROCS to return both old value (from the return value of
runtime.GOMAXPROCS) and new value to make it more useful for debugging.

This is a breaking change, but the impact should be minimal.

### DIFF
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -86,14 +86,16 @@ func defaultMaxProcsFormula(n float64) int {
 // in bound of [min, max].
 //
 // Currently the default formula is NumCPU()*2-1 rounding up.
-func GOMAXPROCS(min, max int) int {
+func GOMAXPROCS(min, max int) (oldVal, newVal int) {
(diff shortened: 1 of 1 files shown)


runtimebp/cpu.go
```diff
@@ -86,14 +86,16 @@ func defaultMaxProcsFormula(n float64) int {
 // in bound of [min, max].
 //
 // Currently the default formula is NumCPU()*2-1 rounding up.
-func GOMAXPROCS(min, max int) int {
+func GOMAXPROCS(min, max int) (oldVal, newVal int) {
```
pacejackson: 🔕 

```suggestion
func GOMAXPROCS(min, max int) (old, new int) {
```

😅

fishy: `new` is a keyword, otherwise I would not add `Val` prefix.

pacejackson: Keywords, my old nemesis...

PR #38, opened 2020-01-17
Add Span integration for go-redis

pacejackson:
Adds a `redis.Hook` that allows us to add spans around Redis calls when using [go-redis](https://github.com/go-redis/redis).


tracing/redis.go
```diff
@@ -0,0 +1,60 @@
+package tracing
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-redis/redis/v7"
```
konradreiche: I wonder if there are cases where we want to allow for use of https://github.com/gomodule/redigo, too.

tracing/span.go
```diff
@@ -16,6 +16,13 @@ const (
 	SpanTypeServer
 )
 
+type contextKey int
+
+const (
+	serverSpanKey contextKey = iota
```
konradreiche: Is `contextKey iota` a good idea? I remember starting out like this at a previous project and then I would run into context key conflicts across services pretty soon, after that I switched to strings. This might not apply here.

tracing/redis.go
fishy: I think baseplate.go should pick and stick with a single third party redis library.

tracing/span.go
fishy: As long as they are different types, this shouldn't be a problem. Say you have:

```go
type foo int
type bar int
f = foo(0)
b = bar(0)
```

`f` & `b` do not equal because they have different types, that's how context keys work.

integrations/redis.go
```diff
@@ -0,0 +1,154 @@
...
+	clientName string
+	metrics    metricsbp.Statsd
+}
+
+var _ redis.Hook = RedisBaseplateHook{}
+
+// NewRedisBaseplateHook creates a new RedisBaseplateHook to integrate a Redis
+// client from `go-redis` with Baseplate monitoring and tracing.
+func NewRedisBaseplateHook(ctx context.Context, clientName string, metricsConfig metricsbp.StatsdConfig) RedisBaseplateHook {
+	metrics := metricsbp.NewStatsd(ctx, metricsConfig)
```
fishy: Every time you call `NewStatsd` a new background reporting goroutine will be started, so everything inside a process (micro service) should use a single `metricsbp.Statsd`. Here in `NewRedisBaseplateHook` we should just pass that in, instead of creating a new one.

For that reason we might not really need a constructor. We could just export the 2 fields so callers can just do:

```go
hook := integrations.RedisBaseplateHook{
  ClientName: "foo",
  Metrics: statsd,
}
```

integrations/redis.go
```diff
@@ -0,0 +1,154 @@
...
+		h.recordStatus(cmd, err)
+	}
+	return span.End(ctx, err)
+}
+
+func (h RedisBaseplateHook) recordTime(ctx context.Context, name string) {
+	start := getRedisCommandStartTime(ctx)
+	// The default value for time.Time has a value of "0", just return if that
+	// is returned rather than comparing to the current time.
+	if start.Nanosecond() == 0 {
```
fishy: You should use `start.IsZero()` here instead.

integrations/redis.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+
+	duration := time.Now().UTC().Sub(start).Nanoseconds()
```
fishy: The `.UTC()` part is unnecessary.

go-kit metrics package actually have a timer implementation. Basically in `startChildSpan` instead of storing the `time.Time` in the context, you store a [`metrics.Timer`](https://godoc.org/github.com/go-kit/kit/metrics#NewTimer) instead. Then here, if you get a non-nil timer out of the context, you can just call its `ObserveDuration` here. We can also add our own wrapper to make sure it uses millisecond as the unit in `metricsbp` package.

integrations/redis.go
```diff
@@ -0,0 +1,154 @@
...
+	redisCommandStartTime
+)
+
+func setRedisCommandStartTime(ctx context.Context) context.Context {
+	return context.WithValue(ctx, redisCommandStartTime, time.Now().UTC())
+}
+
+func getRedisCommandStartTime(ctx context.Context) (start time.Time) {
+	if t := ctx.Value(redisCommandStartTime); t != nil {
+		if startTime, ok := t.(time.Time); ok {
```
fishy: Also nit, I think the nil check could be omitted here. `nil.(time.Time)` will just return (zero value, false) so the ok check should be sufficient. (for types that could be nil, for example `*time.Time`, it's possible that it's actually `nil` with type `*time.Time` so the type cast would return (nil, true), but since there's no code to set it that way we won't need to worry about that case).

integrations/redis.go
fishy: https://github.com/reddit/baseplate.go/pull/40

integrations/redis.go
```diff
@@ -0,0 +1,125 @@
+package integrations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-redis/redis/v7"
```
fishy: style nit: we usually put the imports in 3 groups: stdlibs, libraries from the same git repo, and everything else. so add a blank line after this line to separate the two groups.

integrations/redis.go
```diff
@@ -0,0 +1,125 @@
...
+		if err != nil {
+			break
+		}
+	}
+	return h.endChildSpan(ctx, err)
+}
+
+func (h RedisBaseplateHook) startChildSpan(ctx context.Context, cmdName string) context.Context {
+	span := tracing.GetServerSpan(ctx)
+	if span == nil {
```
fishy: when there's no server span, do we still want the timer metrics?

integrations/example_redis_test.go
```diff
@@ -0,0 +1,31 @@
+package integrations_test
+
+import (
+	"context"
+	"github.com/go-redis/redis/v7"
```
fishy: style nit: Add a blank line between L4&5.

integrations/redis.go
pacejackson: Thanks to #41 we don't even have to do this in here anymore, we can just rely on the metrics from the baseplate span hook. :)

integrations/redis.go
```diff
@@ -0,0 +1,68 @@
...
+// AfterProcessPipeline ends the client span started by BeforeProcessPipeline,
+// publishes the time the Redis pipeline took to complete, and a metric
+// indicating whether the pipeline was a "success" or "fail"
+func (h RedisSpanHook) AfterProcessPipeline(ctx context.Context, cmds []redis.Cmder) error {
+	var err error
+	for _, cmd := range cmds {
+		err = cmd.Err()
+		if err != nil {
+			break
+		}
```
fishy: :no_bell: Currently we just return the first error, I wonder if we should combine all the errors instead. Konrad did [this in secrets package](https://github.com/reddit/baseplate.go/blob/ada824d60917800f49abc204a1d3a3733fb4549d/secrets/secrets.go#L167-L177), I also wrote [a package for combining errors before](https://godoc.org/github.com/fishy/errbatch).

integrations/redis_test.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			if err = hooks.AfterProcess(ctx, cmd); err != nil {
+				t.Fatalf("Unexpected error: %s", err)
+			}
+			childSpan = tracing.GetChildSpan(ctx)
+			if childSpan == nil {
+				t.Fatalf("'childSpan' is 'nil'")
+			}
+			if err = childSpan.End(ctx, nil); err != tracing.ErrSpanAlreadyEnded {
```
fishy: nit: Since go1.13, prefer to use `errors.Is(err, tracing.ErrSpanAlreadyEnded)` instead of direct comparison (`Is` also does unwrapping).

tracing/span.go
```diff
@@ -199,7 +204,13 @@ func (s *Span) ShouldSample() bool {
 // It also sends the span into tracer, if the span should be sampled.
 // The context object passed in is used to control the timeout of the sending,
 // see Tracer.MaxRecordTimeout for more info.
+//
+// Returns ErrSpanAlreadyEnded if called more than once.
 func (s *Span) End(ctx context.Context, err error) error {
+	if !s.end.IsZero() {
+		return ErrSpanAlreadyEnded
```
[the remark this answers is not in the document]
konradreiche: I don't have much experiencing with setting up tracing, I wonder if reporting an error could indeed prevent bugs on the calling side, couldn't it? If it doesn't create an actual harmful scenario than I'd also side on leaving it out otherwise it seems useful to catch misbehavior early on.

integrations/redis.go
```diff
@@ -0,0 +1,83 @@
...
+	return h.startChildSpan(ctx, "pipeline"), nil
+}
+
+// AfterProcessPipeline ends the client span started by BeforeProcessPipeline,
+// publishes the time the Redis pipeline took to complete, and a metric
+// indicating whether the pipeline was a "success" or "fail"
+func (h RedisSpanHook) AfterProcessPipeline(ctx context.Context, cmds []redis.Cmder) error {
+	var errs batcherror.BatchError
+	for _, cmd := range cmds {
+		if err := cmd.Err(); err != nil {
```
fishy: nit: You don't even need the `!= nil` check with BatchError, you can just do `errs.Add(cmd.Err())` :)

fishy:
The test failure seems like it's just flaky, can you try to rekick drone?

pacejackson:
I don't have permissions to re-kick drone

integrations/redis.go
```diff
@@ -0,0 +1,81 @@
...
+	"strings"
+
+	"github.com/go-redis/redis/v7"
+
+	"github.com/reddit/baseplate.go/batcherror"
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+// PipelineErrors is a set of errors from the commands that run in a Pipeline.
+type PipelineErrors []error
```
fishy: I think this is unused now

Issue #33, opened 2019-12-30
Edge request context


fishy:
[spec link](https://pages.github.snooguts.net/reddit/baseplate.spec/component-apis/edge-request-context), [baseplate.py implementation link](https://github.com/reddit/baseplate.py/blob/master/baseplate/lib/edge_context.py)

pacejackson:
I think we can close this?

fishy:
The event fields part are not done yet (depending on experiments package).

PR #39, opened 2020-01-17
Add edgecontext package

fishy:
https://github.com/reddit/baseplate.go/issues/33

Note that event fields are not implemented yet. They depend on the experiment implementation.


edgecontext/edgecontext.go
```diff
@@ -0,0 +1,147 @@
...
+func FromThriftContext(ctx context.Context) (*EdgeRequestContext, error) {
+	header, ok := thrift.GetHeader(ctx, thriftbp.HeaderEdgeRequest)
+	if !ok {
+		return nil, ErrNoHeader
+	}
+
+	request := baseplate.NewRequest()
+	deserializer := deserializerPool.Get().(*thrift.TDeserializer)
+	defer func() {
+		// See also: https://github.com/apache/thrift/pull/1987
```
fishy: This is merged in upstream, but I'll do the cleanup in a follow up PR instead, as we have other places that also need to migrate to the new `TSerializerPool`.

edgecontext/timestamp.go
```diff
@@ -0,0 +1,74 @@
...
+const millisecondsPerSecond = int64(time.Second / time.Millisecond)
+
+// TimestampMillisecond implements json encoding/decoding using milliseconds
+// since EPOCH.
+type TimestampMillisecond time.Time
+
+var (
+	_ json.Unmarshaler = (*TimestampMillisecond)(nil)
+	_ json.Marshaler   = TimestampMillisecond{}
+)
```
pacejackson: 🔕 Should we have a `lib` or `utils` package for more generic things like this?  There's not really a reason for it to be in `edgecontext`.

edgecontext/edgecontext.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/apache/thrift/lib/go/thrift"
+)
+
+// ErrLoIDWrongPrefix is an error could be returned by New() when passed in LoID
+// does not have the correct prefix.
+var ErrLoIDWrongPrefix = errors.New("edgecontext: loid should have t2_ prefix")
+
+// ErrNoHeader is an error could be returned by FromThriftContext() when passed
+// in context does not have Edge-Request header set.
+var ErrNoHeader = errors.New("edgecontext: no Edge-Request header found")
```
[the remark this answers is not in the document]
fishy: I usually only have an `errors.go` for error types. For variables that's an overkill.

edgecontext/timestamp.go
fishy: Agreed, but I'll do that in a followup PR instead. We also have the microsecond types used in tracing that should be moved out as well.

konradreiche: @pacejackson Not a big fan of generic packages, they become bit a dumping ground

fishy: The package certainly shouldn't be named something too generic like `lib` or `utils`. I'm thinking about `timebp`.

konradreiche: Yes. By the way, it seems the suffixing of `bp` is spreading across the repository. I find it makes the folder structure a lot harder to read. Wouldn't it be possible to just name those things what they are, i.e. `time` and if there's an import conflict, let the user rename it if necessary.

fishy: most of those are the cases that importing conflict is bound to happen. for example metricsbp itself imported go-kit's metrics package, and timebp will certainly import stdlibs time library. I actually like the -bp suffix :)

PR #41, opened 2020-01-19
Span hooks

pacejackson:
This PR adds "Hooks" into the tracing lifecycle like the [Observers](https://github.com/reddit/baseplate.py/blob/master/baseplate/__init__.py#L32-L86) in `baseplate.py`.  This allows us to hook custom functionality into the Baseplate request lifecycle and the Span lifecyle.  

This also includes hooks for Metrics which allow us to automatically add timers around Spans and success/fail metrics.

### DIFF
--- a/tracing/span.go
+++ b/tracing/span.go
@@ -16,6 +16,13 @@ const (
 	SpanTypeServer
 )
 
+type contextKey int
+
+const (
+	serverSpanKey contextKey = iota
+	childSpanKey
+)
+
 // FlagMask values.
 //
 // Reference: https://github.com/reddit/baseplate.py/blob/1ca8488bcd42c8786e6a3db35b2a99517fd07a99/baseplate/observers/tracing.py#L60-L64
@@ -213,6 +220,21 @@ func (s *Span) CreateLocalChild(name string, component string) *Span {
 	return s.createChild(name, SpanTypeLocal, component)
 }
 
+// CreateLocalChildForContext creates a local child-span with given name and
+// component and a context that can be used by the client you are creating the
+// span for.
+//
+// component is optional local component name and could be empty string.
+//
+// Timestamps, counters, and tags won't be inherited.
+// Parent id will be inherited from the span id,
+// and span id will be randomly generated.
+// Trace id, sampled, and flags will be copied over.
+func (s *Span) CreateLocalChildForContext(ctx context.Context, name string, component string) (context.Context, *Span) {
+	child := s.CreateLocalChild(name, component)
+	return context.WithValue(ctx, childSpanKey, child), child
+}
+
 // CreateClientChild creates a client child-span with given name.
 //
 // A client child-span should be used to make requests to other upstream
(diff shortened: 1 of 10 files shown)


tracing/span.go
```diff
@@ -33,7 +40,7 @@ const (
 type Span struct {
 	tracer *Tracer
 
-	name     string
+	Name     string
```
fishy: style nit: move the exported fields to the top, before all the unexported fields.

tracing/span.go
```diff
@@ -66,6 +75,22 @@ func newSpan(tracer *Tracer, spanType SpanType) *Span {
...
+	if spanType == SpanTypeServer {
+		onServerSpanCreate(span)
+		span.startSpan()
+	}
+	return span
+}
+
+func (s *Span) startSpan() {
+	for _, hook := range s.hooks {
+		hook.OnStart(s)
```
fishy: Should we check the error here? If not, do we need the interface to actually return an error?

tracing/span.go
```diff
@@ -175,11 +200,12 @@ func (s *Span) End(ctx context.Context, err error) error {
 	if s.spanType == SpanTypeLocal && s.component != "" {
 		s.SetTag(ZipkinBinaryAnnotationKeyLocalComponent, s.component)
 	}
-
+	for _, hook := range s.hooks {
+		hook.OnEnd(s, err)
```
fishy: Same question here.

tracing/tracer.go
```diff
@@ -111,9 +111,8 @@ func (t *Tracer) Close() error {
 
 // NewTrace creates a new trace (top level local span).
 func (t Tracer) NewTrace(name string) *Span {
-	span := newSpan(&t, SpanTypeLocal)
-	span.name = name
-	span.traceID = rand.Uint64()
+	span := newSpan(&t, name, SpanTypeLocal)
+	span.NameeID = rand.Uint64()
```
fishy: What is `NameeID`? It's not defined.

tracing/span.go
pacejackson: I was thinking I would use `s.tracer.Logger` to log an error message?

tracing/tracer.go
pacejackson: I think this was just a wonky delete, it was supposed to delete one line but it looks like instead I deleted 1/2 of each line.

tracing/span.go
```diff
@@ -66,6 +79,29 @@ func newSpan(tracer *Tracer, spanType SpanType) *Span {
...
+	span := newSpan(tracer, name, SpanTypeServer)
+	onServerSpanCreate(span)
+	span.startSpan()
+	return span
+}
+
+func (s *Span) startSpan() {
+	for _, hook := range s.hooks {
+		if err := hook.OnStart(s); err != nil {
+			log.Error("OnStart hook error: " + err.Error())
```
fishy: nit: When using the zap logger, prefer to do this instead: `log.Errorw("OnStart hook error", "err", err)`. Or we can use the wrapped simple logger in tracer:

```
if s.tracer.Logger != nil {
  s.tracer.Logger("OnStart hook error: " + err.Error())
}
```

PR #45, opened 2020-01-22
Add support for experiments

konradreiche:
I wanted to start sharing progress on the experiments branch. I'll have to figure out what would count as first pass in terms of features. Until then I'll try to replicate unit tests from the baseplate.py project as good as possible.

Roadmap
- [x] experiment interface
- [x] simple experiments
- [x] variant sets
  - [x] interface
  - [x] single variant set
  - [x] multi variant set
  - [x] range variant set
  - [x] rollout variant set
- [x] overrides
- [x] targeting


experiments/experiments.go
```diff
@@ -0,0 +1,195 @@
...
+	now := time.Now()
+	return e.enabled && !now.Before(e.startTime) && now.Before(e.endTime)
+}
+
+type Variant struct {
+	Name string  `json:"name"`
+	Size float64 `json:"size"`
+}
+
+type ErrorUnknownExperiment string
```
fishy: style nit: For error types, use `NameError` instead of `ErrorName`.

experiments/experiments.go
```diff
@@ -0,0 +1,195 @@
...
+	// ID is the experiment identifier and should be unique for each experiment.
+	ID int `json:"id`
+	// Name is the experiment name and should be unique for each experiment.
+	Name           string           `json:"name"`
+	Owner          string           `json:"owner"`
+	Enabled        *bool            `json:"enabled"`
+	Version        string           `json:"version"`
+	Type           string           `json:"type"`
+	StartTimestamp int64            `json:"start_ts"`
+	StopTimestamp  int64            `json:"stop_ts"`
```
fishy: We should define a custom type in `timebp`, similar to `TimestampMillisecond` there.

experiments/experiments.go
```diff
@@ -0,0 +1,195 @@
...
+		return "", nil
+	}
+	bucket := e.calculateBucket(bucketKey)
+	return e.variantSet.ChooseVariant(bucket), nil
+}
+
+func (e *SimpleExperiment) calculateBucket(bucketKey string) int {
+	target := new(big.Int)
+	bucket := new(big.Int)
+	hashed := sha1.Sum([]byte(fmt.Sprintf("%s%s", e.seed, bucketKey)))
```
fishy: I think you want to use `e.hash` instead of hardcoded `sha1.Sum` here?

konradreiche: Do you mean allocating the hash in the constructor, i.e. `sha1.New()`?

fishy: No you already did that before.

fishy: on line 134

konradreiche: Oh I see that needs to be removed because it seems to create a buffer that you can write to so it doesn't seem to be of purpose for reusing it, i.e.

```
h := sha1.New()
io.WriteString(h, "His money is twice tainted:")
io.WriteString(h, " 'taint yours and 'taint mine.")
fmt.Printf("% x", h.Sum(nil))
```

fishy: You can call `Reset` if you need to reuse it, but that's not thread safe and cannot be used concurrently. If you need to reuse them concurrently, then the type should be `func() hash.Hash` (which `sha1.New` satisfies) instead.

Also if you don't think we'll ever use any hash other than sha1, then this is unnecessary and we can just hardcode sha1 instead.

experiments/experiments.go
```diff
@@ -0,0 +1,219 @@
...
+	Variant(args map[string]string) (string, error)
+	LogBucketing() bool
+}
+
+func (e *Experiments) Variant(name string, args map[string]string, bucketingEventOverride bool) (string, error) {
+	experiment, err := e.experiment(name)
+	if err != nil {
+		return "", err
+	}
+	return experiment.Variant(args)
```
pacejackson: Shouldn't `Variant` also log an event?

experiments/targeting.go
```diff
@@ -0,0 +1,298 @@
...
+	values, valuesOK := inputNodes["values"]
+	if !valueOK && !valuesOK {
+		return nil, TargetingNodeError("EqualNode expects input key 'value' or 'values'.")
+	}
+
+	var acceptedValues []interface{}
+	if valueOK {
+		acceptedValues = []interface{}{value}
+	}
+	if valuesOK {
```
pacejackson: it seems like the intention here is to favor `values` over `value`?  If that is the case, I think it's clearer to write that as:

```go
if valuesOK {
	acceptedValues = values.([]interface{})
} else if valueOK {
	acceptedValues = []interface{}{value}
}
```

experiments/targeting.go
```diff
@@ -0,0 +1,298 @@
...
+	switch cv := candidateValue.(type) {
+	case int:
+		return n.less(float64(cv), value)
+	case float64:
+		n.less(cv, value)
+	}
+	return false
+}
+
+type Targeting interface {
```
pacejackson: style nit: can you move this to the top of the file so it's easier to tell that the `Node`s implement this?

experiments/targeting.go
```diff
@@ -0,0 +1,298 @@
...
+	switch n := node.(type) {
+	case []interface{}:
+	case map[interface{}]interface{}:
+		key := firstKey(n)
+		return mapOperatorNode(strings.ToLower(key), n[key])
+	case map[string]interface{}:
+		key := firstKeyStr(n)
+		return mapOperatorNode(strings.ToLower(key), n[key])
+	default:
+		panic(n)
```
pacejackson: should this just return an error rather than panicking?

experiments/experiments.go
konradreiche: Yes, I just saw that this branch is missing the event logging. Maybe it would make sense to add this in another PR?

experiments/targeting.go
konradreiche: Yes, that's what I got from reading the baseplate.py code, that looks good.

experiments/targeting.go
konradreiche: Yes, never panic. Still got some cruft to clean up.

experiments/experiments.go
pacejackson: Ya this PR is getting pretty big, I think it's fine to do multiple PRs to get this out in manageable chunks rather than trying to get it all in one big PR.

experiments/experiments.go
konradreiche: Can we reuse `timebp.TimestampMillisecond` here?

fishy: but here it's seconds, not milliseconds?

experiments/experiments.go
```diff
@@ -0,0 +1,273 @@
...
+	for key, value := range args {
+		lowered[strings.ToLower(key)] = value
+	}
+	return lowered
+}
+
+func (e *SimpleExperiment) calculateBucket(bucketKey string) int {
+	target := new(big.Int)
+	bucket := new(big.Int)
+	hashed := sha1.Sum([]byte(fmt.Sprintf("%s%s", e.bucketSeed, bucketKey)))
```
fishy: nit: The performance of `fmt.Sprintf` is not very good (because the unknown of the length of the final result and the allocations happened), for this case `e.bucketSed + bucketKey` would be faster and you don't lose too much readability (for most of the other cases in this package `fmt.Sprintf` is still much more readable and convenient)

experiments/targeting.go
```diff
@@ -0,0 +1,345 @@
...
+		key := firstKey(n)
+		return mapOperatorNode(strings.ToLower(key), n[key])
+	case map[string]interface{}:
+		key := firstKeyStr(n)
+		return mapOperatorNode(strings.ToLower(key), n[key])
+	}
+	return nil, TargetingNodeError(fmt.Sprintf("node type %T unknown", node))
+}
+
+func firstKeyStr(m map[string]interface{}) string {
```
fishy: Is it that when we use `firstKeyStr` and `firstKey`, the map always contain exactly one key-value pair? If that's not the case, the iteration order of maps are intentionally randomized in go runtime so that you will usually get different `firstKeyStr`/`firstKey` results for the same map.

konradreiche: Yes, the maps will always only have one key of the way experiments configurations are structured. Here is an example

fishy: It's possible to define a struct with `Key` and `Value` (or whatever names), then in it's `UnmarshalJSON` implementation, try to unmarshal to a map, then check that map has len of 1, then get the key and value out.

fishy: Something like this: https://play.golang.com/p/LCGwucidXXV

konradreiche: Oh that's good. Do you think it's worthwhile changing it to use? I think the only problem is that it would still need to be decided at the stage of looking at the key which concrete struct to set it up with, i.e. `AnyNode`, `AllNode`, etc.

experiments/variants.go
```diff
@@ -0,0 +1,244 @@
+package experiments
+
+import "fmt"
+
+// VariantSet is the base interface for variant sets. A variant set contains a
+// set of experimental variants, as well aas their distributions. It is used by
```
fishy: typo: s/aas/as/

experiments/targeting.go
fishy: I think it makes more sense to use that if we get rid of the `map[interface{}]interface{}` case, but since this PR is already getting big we can do that in a followup PR instead.

experiments/targeting.go
```diff
@@ -0,0 +1,357 @@
...
+		ReturnValue: returnValue,
+	}
+}
+
+// Evaluate returns the configured boolean return value.
+func (n *OverrideNode) Evaluate(inputs map[string]interface{}) bool {
+	return n.ReturnValue
+}
+
+// ComparisonNode is a non-equality comparison operators (gt, ge, lt, le).
```
fishy: 🔕 Does it make more sense to merge `EqualNode` into `ComparisonNode` as well? `ne` is already handled here.

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/go-logfmt/logfmt v0.4.0 // indirect
 	github.com/go-redis/redis/v7 v7.0.0-beta.5
 	github.com/go-stack/stack v1.8.0 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
```
fishy: I _think_ the changes in `go.mod` and `go.sum` are accidental and could be reverted.

experiments/targeting.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	if len(config) != 1 {
+		return nil, TargetingNodeError("call to create targeting tree expects single input key")
+	}
+	return parseNode(config)
+}
+
+func parseNode(node interface{}) (Targeting, error) {
+	switch n := node.(type) {
+	case map[interface{}]interface{}:
```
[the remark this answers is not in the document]
konradreiche: `map[interface{}]interface{}` is encountered during parsing for the lower level JSON objects. The first level is detected by the JSONUnmarshaller as `map[string]interface{}`, every subsequent `map[interface{}][interface{}`.

experiments/targeting.go
konradreiche: I think the `EqualNode` allows for a wider variety of comparisons, string, integer, bool fields etc. where as comparison seems to be reserved for numbers, but I actually need to verify this. If that's not the case then `ComparisonNode` needs to become more flexible, too.

go.mod
konradreiche: Good catch, indeed.

experiments/targeting.go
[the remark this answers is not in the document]
fishy: do you have a concrete example? I think here we always try to convert the key to string anyways, which kind of implies that the key is supposed to be string? I also don't think I've seen a map in json that the key is not string.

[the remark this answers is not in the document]
konradreiche: You're right about that, that is the only case. I am just saying, that when the JSON document gets unmarshalled by Go into `map[string]interface{}]` then every value of that map is typed `map[interface{}][interface{}]`, because that's what the default JSON unmarshaller does even though every key is indeed a string. We can probably work around that behavior but that's just me dealing with what I received.

experiments/targeting.go
```diff
@@ -0,0 +1,357 @@
...
+// Targeting is the common interface to implement experiment targeting.
+// Evaluated whether the provided input matches the expected values.
+type Targeting interface {
+	Evaluate(inputs map[string]interface{}) bool
+}
+
+// NewTargeting parses the given targeting configuration into a Targeting.
+func NewTargeting(targetingConfig []byte) (Targeting, error) {
+	var config map[string]interface{}
+	decoder := json.NewDecoder(bytes.NewBuffer(targetingConfig))
```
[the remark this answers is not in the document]
konradreiche: Glad you reminded me of this. I forgot to integrate targeting and overrides into the code that determines the bucketing :man_facepalming: My plan would be to delay parsing for this, for instance:

```
type ParsedExperiment struct {
	ExperimentVersion int             `json:"experiment_version"`
	ShuffleVersion    string          `json:"shuffle_version"`
	BucketVal         string          `json:"bucket_val"`
	LogBucketing      bool            `json:"log_bucketing"`
	Variants          []Variant       `json:"variants"`
	BucketSeed        string          `json:"bucket_seed"`
	Targeting         json.RawMessage `json:"targeting"`
	Overrides         json.RawMessage `json:"overrides"`
}
```

and then pass the content of Targeting and Overrides to their respective constructor.

konradreiche:
I've integrated overrides and targeting into the `Variant` method so that they are actually applied. From this pull request still missing in the the event logging but that's definitely for another pull request.

experiments/experiments.go
```diff
@@ -0,0 +1,323 @@
...
+		bucketSeed = fmt.Sprintf("%d.%s.%s", experiment.ID, experiment.Name, shuffleVersion)
+	}
+	variantSet, err := FromExperimentType(experiment.Type, experiment.Experiment.Variants, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	targetingConfig := experiment.Experiment.Targeting
+	if len(targetingConfig) == 0 {
+		targetingConfig = []byte(`{"OVERRIDE": true}`)
```
konradreiche: I guess I'd rather separate parsing and pass in of target structure so that we can actually pass `map[string]interface{}{"OVERRIDE": true}` in here instead.

experiments/experiments.go
```diff
@@ -0,0 +1,323 @@
...
+	return int(bucket.Int64())
+}
+
+// UniqueID returns a unique ID for the experiment.
+func (e *SimpleExperiment) UniqueID(bucketVals map[string]string) string {
+	bucketVal, ok := bucketVals[e.bucketVal]
+	if !ok {
+		return ""
+	}
+	return fmt.Sprintf("%s:%s:%s", e.name, e.bucketVal, bucketVal)
```
pacejackson: 🔕 would `join` be preferable since these are all strings?

experiments/experiments.go
```diff
@@ -0,0 +1,323 @@
...
+	bucketVal string
+	// variantSet contains a set of experimental variants as well as their
+	// distributions. It is used by experiments to track which bucket a variant
+	// is assigned to.
+	variantSet VariantSet
+	// targeting allows to target users with multiple parameters supporting
+	// both AND and OR based logical grouping.
+	targeting Targeting
+	// overrides if matched allow to force a particular variant.
+	overrides []map[string]Targeting
```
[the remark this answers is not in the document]
konradreiche: It would be nice to get the documentation. I was considering exporting those fields just for that reason but from the specification it might be better to keep those private.

PR #46, opened 2020-01-22
Add hook for sending span errors to sentry

pacejackson:


tracing/error_reporter_hooks.go
```diff
@@ -0,0 +1,44 @@
+package tracing
+
+import "github.com/reddit/baseplate.go/log"
+
+const (
+	success = "success"
+	fail    = "fail"
+)
```
fishy: these seem unused here?

tracing/error_reporter_hooks.go
```diff
@@ -0,0 +1,44 @@
...
+
+// OnStart is a nop
+func (h ErrorReporterSpanHook) OnStart(span *Span) error {
+	return nil
+}
+
+// OnEnd logs a message and sends err to Sentry if err is non-nil.
+func (h ErrorReporterSpanHook) OnEnd(span *Span, err error) error {
+	if err != nil {
+		log.ErrorWithRaven("Span ended with an error", err)
```
fishy: 🔕 `log.Error*` will also print call stacks, but call stacks from `hook.OnEnd` might not be that useful. We might want to omit the logging part and only send to sentry here instead?

But if we do want to keep the logging part, we might want to also add span info here, so you can add either `"span", *span` or just `"spanname", span.Name` (as that's the only exported field of `Span`) to the args of `ErrorWithRaven`. You can also implement `Span.String()` so that adding `"span", *span` will output something more useful to the log.

tracing/error_reporter_hooks.go
```diff
@@ -1,11 +1,6 @@
 package tracing
 
-import "github.com/reddit/baseplate.go/log"
-
-const (
-	success = "success"
-	fail    = "fail"
-)
+import "github.com/getsentry/raven-go"
```
fishy: style nit: for single import, still use `()`:

```
import (
  "github.com/getsentry/raven-go"
)
```

as this makes future diff cleaner.

PR #47, opened 2020-01-22
Add 'Type' function to 'Span' to return a string of the SpanType

pacejackson:


pacejackson:
Another option for how to do this would be to export `spanType` (probably as `Type`?) and let the Hooks decide what to make of that value.  I don't really feel strongly about either approach, is one preferred over the other?

fishy:
> Another option for how to do this would be to export `spanType` (probably as `Type`?) and let the Hooks decide what to make of that value. I don't really feel strongly about either approach, is one preferred over the other?

export `SpanType` field directly might be bad as that allows caller to change its type, but export `SpanType()` as a read only getter, then implement `SpanType.String()` would be better, as that way user can inspect the type of the span, without relying on the comparison between arbitrary strings (as `SpanType` itself is an enum type).

metricsbp/baseplate_hooks.go
```diff
@@ -35,7 +35,7 @@ type SpanHook struct {
 }
 
 func newSpanHook(prefix string, metrics Statsd, span *tracing.Span) SpanHook {
-	name := fmt.Sprintf("%s.%s.%s", prefix, span.Type(), span.Name)
+	name := fmt.Sprintf("%s.%s.%s", prefix, span.Type().String(), span.Name)
```
fishy: nit: If the format `%v` will auto call the value's `.String()` (if available) so you can do `Sprintf("%s.%v.%s", prefix, span.Type(), span.Name)` instead.

tracing/span.go
```diff
@@ -18,6 +18,20 @@ const (
 	SpanTypeServer
 )
 
+// String returns a string representing the SpanType.
```
fishy: 🔕 golint actually don't force you to add doc comment for `type.String() string` (also `type.Error() string`) as that's part of the builtin type magic.

tracing/span_test.go
```diff
@@ -292,7 +292,7 @@ func TestSpanTypes(t *testing.T) {
 			func(t *testing.T) {
 				t.Parallel()
 				span := Span{Name: "test", spanType: c.spanType}
-				if span.Type() != c.expected {
+				if span.Type().String() != c.expected {
```
fishy: Here we should make `c.expected` be of type `SpanType` instead of string, and compare `span.Type() != c.expected` instead (and in `Errorf` below use `%v` instead of `%s`).

pacejackson: In this case, what we really want to test is the switch statement in `SpanType.String`, so that's not quite the test we want.  I think we can just take the `Span` out of the test entirely though and do this:

```golang
func(t *testing.T) {
	t.Parallel()
	if c.spanType.String() != c.expected {
		t.Errorf(
			"Expected SpanType.String to be %s, got %v",
			c.expected,
			c.spanType,
		)
	}
}
```

PR #49, opened 2020-01-22
tracing: Document thread unsafe functions

fishy:


fishy:
I think in real world scenario these functions are very unlikely to be called concurrently, so using a lock is likely an overkill and document should be sufficient. At least I'm fairly confident that should be the case for the global one, I'm not so sure about the span one.

pacejackson:
Ya, the expectation is that you'll call the global one when setting up your app and the span one should only get called when the span is created.

PR #52, opened 2020-01-22
metricsbp: Add Statsd.RunSysStats

fishy:


metricsbp/sys_stats.go
```diff
@@ -0,0 +1,113 @@
...
+		defer st.wg.Done()
+
+		ticker := time.NewTicker(SysStatsTickerInterval)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-st.ctx.Done():
+				return
+			case <-ticker.C:
```
konradreiche: What is the behavior here if we update the ticker interval, does that get pulled in or would the goroutine need to be restarted?

fishy: Updating the interval will have no effect here. You need to do that before calling `RunSysStats`.

fishy:
Regarding how to use it: basically after initialized Statsd you just call this function and forget about it :)

PR #53, opened 2020-01-23
Span enhancements and cleanup

pacejackson:
* Remove 'Prefix' from metricsbp.BaseplateHook
  - This should be set by the metrics client and/or whatever backend
    is collecting your metrics

* Add 'OnSetTag' and 'OnAddCounter' functions to the `tracing.SpanHook`
  interface.
  - This brings us close to par with the span Observers in baseplate.py

### DIFF
--- a/metricsbp/baseplate_hooks.go
+++ b/metricsbp/baseplate_hooks.go
@@ -18,7 +18,7 @@ type BaseplateHook struct {
 
 // OnServerSpanCreate registers MetricSpanHooks on a Server Span.
 func (h BaseplateHook) OnServerSpanCreate(span *tracing.Span) error {
-	serverSpanHook := newSpanHook("", h.Metrics, span)
+	serverSpanHook := newSpanHook(h.Metrics, span)
 	span.RegisterHook(serverSpanHook)
 	return nil
 }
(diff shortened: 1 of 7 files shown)


metricsbp/baseplate_hooks.go
```diff
@@ -35,7 +34,10 @@ type SpanHook struct {
 }
 
 func newSpanHook(prefix string, metrics Statsd, span *tracing.Span) SpanHook {
```
fishy: Oh so originally `BaseplateHook.Prefix` is supposed to be whatever we set in `metricsbp.Statsd` (via `StatsdConfig.Prefix`)? In that case we definitely don't need that in `BaseplateHook`.

Now do we still need prefix for this function? Is it still used anywhere?

tracing/hooks.go
```diff
@@ -12,6 +12,8 @@ type SpanHook interface {
 	OnCreateChild(child *Span) error
 	OnStart(span *Span) error
 	OnEnd(span *Span, err error) error
+	OnSetTag(span *Span, key string, value interface{}) error
+	OnAddCounter(span *Span, key string, delta float64) error
 }
```
fishy: Now that this interface is having more and more functions, I'd suggest provide an implementation, `type NopSpanHook struct{}`, which implements every function as a nop. This way the actual implementations can embed it and only implement the non-nop ones, for example for `ErrorReporterSpanHook`, it can just do:

```
type ErrorReporterSpanHook struct{
  tracing.NopSpanHook
}

func (h ErrorReporterSpanHook) OnEnd(span *Span, err error) error {
  ...
}
// It does not need to implement anything that's nop
```

This way we can also make it easier to add more functions to this interface in the future (otherwise every time you do that you break every implementations until they added the new function implementations).

fishy: But at the same time providing the base, nop implementation will make it harder to test implementations, as even if they misspell the function (e.g. `OnEdn` instead of `OnEnd`) the implementation still satisfy the interface, so it might make sense to provide better ways to test them (e.g. provide a mock span so they can write tests and register the hooks on the mock span, or something like that?)

metricsbp/baseplate_hooks.go
pacejackson: We do still use it, but I'm not sure we actually want to, I'm going to remove it.

tracing/hooks.go
pacejackson: Hmm, where would be a good place to put something like that?  Would we just export some `Test` functions in this file?  Or maybe a `testing` package?

fishy: It can be in the same package, called `MockSpan` (see [`MockMessageQueue`](https://godoc.org/github.com/reddit/baseplate.go/mqsend#MockMessageQueue) for an example), or in a `mock` subdirectory/subpackage (see https://godoc.org/github.com/Shopify/sarama/mocks as an example). I think in most cases the same package makes more sense.

PR #56, opened 2020-01-24
runtimbp: Add shares fallback

fishy:


runtimebp/cpu.go
```diff
@@ -55,6 +56,40 @@ func NumCPU() (n float64) {
...
+				"NumCPU: falling back to use runtime.NumCPU(): %v, %v\n",
+				n,
+				err,
+			)
+			n = float64(runtime.NumCPU())
+		}
+	}()
+
+	// Big enough buffer to read the number in the file wholly into memory.
+	buf := make([]byte, 1024)
```
konradreiche: Why not just do a `ioutil.ReadFile` and convert that?

fishy: the file could be huge, we don't want to blow up memory.

konradreiche: Doesn't that file only ever have a number in there? Well, doesn't matter, was just curious :)

fishy: It could be misconfigured somewhere.

PR #57, opened 2020-01-24
Span refactor

pacejackson:


tracing/hooks.go
```diff
@@ -3,17 +3,31 @@ package tracing
 // BaseplateHook allows you to inject functionality into the lifecycle of a
 // Baseplate request.
 type BaseplateHook interface {
-	OnServerSpanCreate(span *Span) error
+	OnServerSpanCreate(span ServerSpan) error
```
fishy: `ServerSpan` should be a pointer type here?

tracing/error_reporter_hooks.go
```diff
@@ -20,7 +20,7 @@ type ErrorReporterSpanHook struct {
 }
 
 // OnEnd logs a message and sends err to Sentry if err is non-nil.
-func (h ErrorReporterSpanHook) OnEnd(span *Span, err error) error {
+func (h ErrorReporterSpanHook) OnEnd(span *ServerSpan, err error) error {
```
fishy: `*ServerSpan` should be `Span` instead?

tracing/hooks.go
```diff
@@ -3,17 +3,31 @@ package tracing
...
+}
+
+// AddHooks adds hooks into the SpanHooks container.
+func (h *SpanHooks) AddHooks(hooks ...SpanHook) {
+	h.hooks = append(h.hooks, hooks...)
+}
+
+// Hooks returns a copy of the internal list of SpanHooks.
+func (h *SpanHooks) Hooks() []SpanHook {
+	return h.hooks
```
fishy: This is not actually a copy. For example the follow code could change your 2nd hook:

```
foo := hooks.Hooks()
foo[1] = newHook
```

fishy: I wonder if it's useful to define `SpanHooks` type, or maybe just typedef it as `type SpanHooks []SpanHook`.

tracing/span.go
```diff
@@ -366,35 +342,138 @@ func (s *Span) createChild(name string, spanType SpanType, component string) *Sp
...
+)
+
+const (
+	counterKeyPrefix   = "counter."
+	baseplateComponent = "baseplate"
+)
+
+type baseSpan struct {
+	trace *Trace
+	hooks *SpanHooks
```
fishy: nit: this does not need to be a pointer (all slices are actually pointers anyway).

tracing/span.go
```diff
@@ -366,35 +342,138 @@ func (s *Span) createChild(name string, spanType SpanType, component string) *Sp
...
+func GetServerSpan(ctx context.Context) *ServerSpan {
+	if span, ok := ctx.Value(serverSpanKey).(*ServerSpan); ok {
+		return span
+	}
+	return nil
 }
 
-// GetChildSpan gets the ChildSpan from the given context
+// GetChildSpan gets the child span from the given context
 func GetChildSpan(ctx context.Context) *Span {
```
fishy: Since `Span` is an interface now, it shouldn't be a pointer.

tracing/span.go
```diff
@@ -39,153 +40,89 @@ func (st SpanType) String() string {
...
-	counterKeyPrefix   = "counter."
-	baseplateComponent = "baseplate"
-)
-
-// Span defines a tracing span.
-type Span struct {
-	Name string
-
+// Trace holds the trace data for a Span and publishes ZipkinSpans.
+type Trace struct {
```
fishy: Do we use `Trace` as the top level `Span`, or do we only use it to hold internal state? If it's only for internal state then I don't think we want to export it.

tracing/span.go
pacejackson: Probably not, I can also do the simpler `type` suggestion.

tracing/span.go
pacejackson: I wasn't sure about how to handle that since I can't return `nil` if it's _not_ a pointer?  Do I just return an empty one?  Or maybe make a "null" span that doesn't do anything?

fishy: interfaces are pointers. you can return nil for an interface.

tracing/error_reporter_hooks.go
pacejackson: Ya, I just haven't gotten around to it.  All I've _really_ worked on is `span.go`

tracing/span.go
pacejackson: Lemme fiddle around with this more, maybe we can just not export it.

pacejackson:
baseplate.go as a whole needs to be updated to account for the changes, but the `tracing` package's tests are passing and I'm reasonably happy with the implementation if you have another chance to take a look.

tracing/span.go
```diff
@@ -209,91 +162,162 @@ func (s *Span) ToZipkinSpan() ZipkinSpan {
...
+	Start()
+	Stop(err error)
+	Publish(ctx context.Context, err error) error
+	SetTag(key string, value interface{})
+	AddCounter(key string, delta float64)
+	CreateLocalChild(name, component string) *LocalSpan
+	CreateClientChild(name string) *ClientSpan
+	LogError(msg string, err error)
+	SetDebug(v bool)
+}
```
fishy: Same here, please add comments to the functions, especially for functions not obvious, e.g. `Stop` vs. `Publish`.

tracing/span.go
```diff
@@ -209,91 +162,162 @@ func (s *Span) ToZipkinSpan() ZipkinSpan {
...
+
+// StartSpan starts the Span and calls all OnStart Hooks registered to the Span.
+func StartSpan(s Span) {
```
fishy: Is there a reason we need to use this function over `s.Start()`? Same as other similar functions below.

tracing/span.go
```diff
@@ -306,11 +330,92 @@ func (s *Span) CreateLocalChild(name string, component string) *Span {
...
+	ctx = CreateThriftContextFromSpan(ctx, child)
+	return ctx, child
+}
+
+func initChildSpan(parent Span, child Span) {
+	for _, hook := range parent.Hooks() {
+		if err := hook.OnCreateChild(child); err != nil {
+			parent.LogError("OnCreateChild hook error: ", err)
+		}
+	}
```
fishy: Should we call the hooks before or after we called `child.Start`? This is an example of things should be explained in the interface comments.

tracing/span.go
```diff
@@ -306,11 +330,92 @@ func (s *Span) CreateLocalChild(name string, component string) *Span {
...
+
+type baseSpan struct {
+	trace *trace
+	hooks []SpanHook
+
+	name string
+}
+
+// Name returns the name of the Span.
+func (s *baseSpan) Name() string {
```
fishy: Since `baseSpan` is an unexported type, I think golint won't force you to write doc comments on its exported functions. These comments should be moved to the interface instead.

tracing/span.go
```diff
@@ -320,81 +425,116 @@ func (s *Span) CreateLocalChildForContext(ctx context.Context, name string, comp
...
+	child.trace.flags = s.trace.flags
+	return child
 }
 
-func (s *Span) createChild(name string, spanType SpanType, component string) *Span {
-	span := newSpan(s.tracer, name, spanType)
-	span.spanType = spanType
-	span.component = component
+// ServerSpan defines a tracing span over an entire request to a server.
+type ServerSpan struct {
```
fishy: It should be mentioned in doc comment that it implements Span. Same as the other exported concrete types.

tracing/span.go
fishy: Also `baseSpan` is basically just 3 pointers, so there's no need to really use pointer as the receiver type. Usually we use non-pointer for immutable functions, and pointer for mutable functions, unless the struct is huge or there are some fields not copyable.

tracing/span.go
```diff
@@ -320,81 +425,116 @@ func (s *Span) CreateLocalChildForContext(ctx context.Context, name string, comp
...
-//
-//     clientCtx, span := parentSpan.ChildAndThriftContext(ctx, "myCall")
-//     result, err := client.MyCall(clientCtx, arg1, arg2)
-//     span.End(ctx, err)
-func (s *Span) ChildAndThriftContext(ctx context.Context, name string) (context.Context, *Span) {
-	ctx, span := s.CreateClientChildForContext(ctx, name)
-	ctx = CreateThriftContextFromSpan(ctx, span)
-	return ctx, span
+// SetServerSpan sets the Span as the ServerSpan on the given context
+func (s *ServerSpan) SetServerSpan(ctx context.Context) context.Context {
```
fishy: nit: Since this is a function of type `ServerSpan`, the name `SetServerSpan` feels redundant, maybe call it `InjectContext` or something?

tracing/span.go
pacejackson: These functions are what actually call the `Hooks` so you don't have to remember to do that if you want to do custom things in your implementation.

tracing/span.go
pacejackson: After, because `OnCreateChild` can/is often used to register Hooks on the child span.

tracing/span.go
pacejackson: That's true for everything but `AddHooks`, it appears that `s.hooks` is copied when I don't use a pointer because the Hooks are not actually added to the Span when the receiver is not a pointer.

fishy: Yes, AddHooks is a mutate function, others are not.

tracing/span.go
fishy: That sounds error prone, as the user is very likely to accidentally call `s.Span` instead of `StartSpan(s)`.

The interface here is slightly different from hooks. For hooks we provide an interface for the users to do their own implementations. Here we provide an interface because we provided several implementations with the same interface. So here I think we can just make sure the type functions do all the things (including calling the hooks) and get rid of these global functions.

tracing/span.go
fishy: But you called the hooks before you called `StartSpan` :)

tracing/span.go
fishy: See https://godoc.org/github.com/reddit/baseplate.go/batcherror#BatchError for an example, only `Add` and `Clear` uses pointers because these are the only functions that will mutate the state.

tracing/span.go
```diff
@@ -306,11 +330,90 @@ func (s *Span) CreateLocalChild(name string, component string) *Span {
...
+
+// SetDebug sets or unsets the debug flag of this Span.
+func (s *baseSpan) SetDebug(v bool) {
+	s.trace.setDebug(v)
+}
+
+// Start starts the Span.
+func (s *baseSpan) Start() {
+	return
+}
```
fishy: I think we don't need `Start` and `Stop` in `baseSpan`.

tracing/span.go
fishy: `SetDebug`, `SetTag` and `AddCounter` are other mutate functions (they don't need pointer receivers to work because you are using pointer fields, but for documentation reasons you should also use pointer receivers for them, but then `baseSpan` is an unexported type so that probably doesn't matter).

tracing/hooks.go
```diff
@@ -3,17 +3,17 @@ package tracing
...
-	OnStart(span *Span) error
-	OnEnd(span *Span, err error) error
-	OnSetTag(span *Span, key string, value interface{}) error
-	OnAddCounter(span *Span, key string, delta float64) error
+	OnCreateChild(child Span) error
+	OnStart(span Span) error
+	OnStop(span Span, err error) error
+	OnSetTag(span Span, key string, value interface{}) error
+	OnAddCounter(span Span, key string, delta float64) error
 }
```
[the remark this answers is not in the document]
pacejackson: Adding comments now 😄 

For `OnCreateChild` I based this on the Observer in baseplate.py which [does not pass in the parent](https://github.com/reddit/baseplate.py/blob/89475dd76bb622730044a169c7e2d07a44e77c97/baseplate/__init__.py#L74).  Originally I _did_ include the parent span but we weren't using it anywhere, I'm also not opposed to including it though.

tracing/span.go
pacejackson: Fair, I can change that.

tracing/span.go
```diff
@@ -201,24 +201,89 @@ func (t *trace) stopAndPublish(ctx context.Context, err error) error {
...
+	// combines creating the child span with calling any OnCreateChild hooks
+	// registered to the Span.
 	CreateClientChild(name string) *ClientSpan
+
+	// LogError is a helper method to log an error plus a message.
 	LogError(msg string, err error)
+
+	// SetDebug sets or unsets the debug flag of this Span.
 	SetDebug(v bool)
 }
```
fishy: I think all the "should not be called directly, use tracing.SameFunction instead" are unnecessary, unless there's a reason the Span itself cannot call those hooks (I don't think so).

Separating `Stop` and `Publish` probably makes sense, but the `Stop` implementation is still supposed to call all the hooks, `tracing. StopSpan` should simply just call `Stop` and `Publish`, nothing else.

Also I don't think `Publish` should take the `err` arg (since it's already taken in `Stop`)?

tracing/span.go
```diff
@@ -455,39 +517,49 @@ func (s *baseSpan) CreateLocalChild(name, component string) *LocalSpan {
...
 
-// Stop ends the span
-func (s *ServerSpan) Stop(err error) {
+// Stop stops the Span and is used to set custom Zipkin tags depending on the
+// error that it recieves.
+//
+// Stop should not be called directly, instead you should call
+// tracing.StopSpan() which combines calling span.Stop, span.Publish, and
+// calling all OnStop Hooks registered to the Span.
+func (s ServerSpan) Stop(err error) {
```
fishy: Stop is actually a mutation (you should also set `s.trace.stop`).

tracing/span.go
```diff
@@ -209,192 +163,463 @@ func (s *Span) ToZipkinSpan() ZipkinSpan {
...
+// Span to a back-end that records the Span.
+func (s *baseSpan) PreStop(err error) {
+	return
+}
+
+// Stop calls span.PreStop, calls all registered OnStop Hooks, serializes
+// the Span and sends the serialized Span to a back-end that records the
+// Span.
+func (s *baseSpan) Stop(ctx context.Context, err error) error {
+	s.PreStop(err)
```
fishy: if you expect this calls PreStop function from the concert type embedding baseSpan (for example, ServerSpan.PreStop), I don't think this works.

pacejackson: :doh:

fishy: I think having a global function is the only way to do that.

tracing/span.go
```diff
@@ -209,192 +163,492 @@ func (s *Span) ToZipkinSpan() ZipkinSpan {
...
+
+// Name returns the name of the Span.
+func (s baseSpan) Name() string {
+	return s.trace.name
+}
+
+// SpanType returns the SpanType of the Span.
+func (s baseSpan) SpanType() SpanType {
+	return SpanTypeUnknown
+}
```
fishy: since `baseSpan` is only used to be embedded into concrete types for code sharing, there's really no need to implement the functions that will be overridden by every concrete types, which includes `SpanType`, `PreStop` and `Stop`.

tracing/hooks.go
```diff
@@ -3,17 +3,35 @@ package tracing
...
+	// OnStart Hooks are called.
+	OnCreateChild(child Span) error
+
+	// OnStart is called after a child Span is created and the OnCreateChild
+	// Hooks on the parent Span are called.
+	OnStart(span Span) error
+
+	// OnStop is called by Span.Stop, after calling span.PreStop but before
+	// serializing and publishing the Span.
+	OnStop(span Span, err error) error
```
fishy: 🔕 Should we update `trace.stop` in `PreStop` based on this description? But I think the hook implementation getting span in this function has no real access to the stop time so it probably doesn't matter.

tracing/span.go
```diff
@@ -54,138 +57,89 @@ const (
...
-const (
-	counterKeyPrefix   = "counter."
-	baseplateComponent = "baseplate"
-)
-
-// Span defines a tracing span.
-type Span struct {
-	Name string
-
+type trace struct {
```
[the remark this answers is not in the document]
pacejackson: I was thinking the same thing

tracing/span.go
pacejackson: Unfortunately, we have to provide some sort of implementation of we want to actually call into the `Hooks` in the implementations provided here because then `baseSpan` has to adhere to the `Span` interface.

fishy: oh I see what's happening here now. I think calling the hooks with `baseSpan` is actually incorrect as the hook implementation cannot cast that back into the actual `ServerSpan`/`ClientSpan`/`LocalSpan`.

tracing/hooks.go
pacejackson: Ya, I think we want to leave `trace.stop` out of pre-stop since it's something we always want to update.

fishy: I think another point is, do we want to count the time hooks took into the span time? that can also affect when we record the stop (and start) time

fishy: Based on the answer to that question, we probably want to rename the hook functions to be more clear (e.g. `OnPreStop` if we count the hook times, or `OnPostStop` if we don't).

integrations/example_redis_test.go
```diff
@@ -17,12 +17,12 @@ func ExampleRedisSpanHook() {
 		// baseClient is not actually used to run commands, we register the Hook
 		// to it and use it to create clients for each Server Span.
 		baseClient redis.Client
-		span       *tracing.Span
+		span       *tracing.ServerSpan
```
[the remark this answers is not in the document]
pacejackson: This should be addressed now, within the context of the `Context`, I renamed childSpan to activeSpan and now in the Redis hook we check the `activeSpan` first, then fall back to the server span.

tracing/hooks.go
pacejackson: 🤔 

I think we do?  One issue we've ran into in the past is losing time data around things that happen outside of spans (like thrift serialization/deserialization time), so, as you mentioned in another place, calling this `OnPreStop` and running it before truly "stopping" might make sense.

pacejackson:
> Thinking about it more, I think make Span an interface probably doesn't make sense.

I'm inclined to agree 😢 

I still _like_ the interface approach and having different types, but it doesn't seem like there's a clean way to make it work.

tracing/span.go
```diff
@@ -43,358 +40,280 @@ type contextKey int
...
+	case SpanTypeServer:
+		if err != nil && errors.Is(err, context.DeadlineExceeded) {
+			// We intentionally don't use the top level span.SetTag function
+			// because we don't want to trigger any OnSetTag Hooks in this case.
+			s.trace.setTag(ZipkinBinaryAnnotationKeyTimeOut, true)
+		}
+	case SpanTypeLocal:
+		if s.component != "" {
+			// We intentionally don't use the top level span.SetTag function
+			// because we don't want to trigger any OnSetTag Hooks in this case.
```
fishy: Since it's consistent within this function and we only added this comment to 2 of the 4 places it's used, maybe just put it at the top of the function once instead :)

tracing/span_test.go
```diff
@@ -11,21 +11,21 @@ import (
...
 func TestDebugFlag(t *testing.T) {
-	var span Span
+	span := CreateServerSpan(nil, "test")
 
 	t.Run(
 		"set",
 		func(t *testing.T) {
 			span.SetDebug(true)
-			if span.flags != FlagMaskDebug {
+			if span.trace.flags != FlagMaskDebug {
```
fishy: 🔕 A few of those tests probably should also be moved into `trace_test.go`, but I don't feel strongly about that.

fishy:
There's also one linting error, though.

fishy:
🔕 This is optional if we think we are not gonna add more functions in the `SpanHook` interface in the future. Also this is probably more of to start a discussion.

In general an interface with too many functions is a design smell. It makes making non-breaking changes hard. Providing a Nop implementation for users to embed can solve part of that problem, but that comes with its own problems as well.

I would suggest that instead of a single, large interface, we split that into several, much smaller interfaces. This way in the future if we need to add new hooks, we just add a new interface, instead of adding functions to an existing interface (which will likely to cause breaking changes). Most of those interfaces should only have one function, but if there are two functions that has a strong correlation to each other we can put them into the same interface (maybe `OnPostStart` and `OnPreStop`?), like:

```go
type CreateChildSpanHook interface {
  OnCreateChild(parent, child *Span) error
}

type StartStopSpanHook interface {
  OnPostStart(span *Span) error
  OnPreStop(span *Span, err error) error
}

...
```

After that, there are two ways to go:

The first is that instead of a single `AddHooks` function, we just have an `AddFooHook` for each interface.

The second is that we still have a single `AddHooks` function, but that function accepts `interface{}` instead:

```go
func (s *Span) AddHooks(hooks ...interface{}) {
  ...
}
```

Then when calling the hooks, we cast them to the hook interface before actually calling them:

```go
func (s *Span) SetTag(key string, value interface{}) {
  ...
  for _, h := s.hooks {
    if hook, ok := h.(SetTagSpanHook); ok {
      if err := hook.OnSetTag(*s, key, value); err != nil {
        ...
      }
    }
  }
  ...
```

Either way we no longer need to provide `NopSpanHook` implementation, and implementations only need to implement the interfaces they needed, and they can still check the implementation on compilation time:

```go
// metricsbp.SpanHook
func (h SpanHook) OnCreateChild(parent, child *tracing.Span) error {
  ...
}

func (h SpanHook) OnAddCounter(span *tracing.Span, key string, delta float64) error {
  ...
}

var (
  _ tracing.AddCounterSpanHook = SpanHook{}
  _ tracing.CreateChildSpanHook = SpanHook{}
)
```

integrations/redis.go
```diff
@@ -49,7 +49,10 @@ func (h RedisSpanHook) AfterProcessPipeline(ctx context.Context, cmds []redis.Cm
 }
 
 func (h RedisSpanHook) startChildSpan(ctx context.Context, cmdName string) context.Context {
-	span := tracing.GetServerSpan(ctx)
+	span := tracing.GetActiveSpan(ctx)
+	if span == nil {
+		span = tracing.GetServerSpan(ctx)
+	}
```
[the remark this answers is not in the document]
fishy: I think that could result in people trying to stop the current span accidentally stop the server span instead.

metricsbp/baseplate_hooks.go
```diff
@@ -11,53 +11,57 @@ const (
...
 	serverSpanHook := newSpanHook(h.Metrics, span)
-	span.RegisterHook(serverSpanHook)
+	span.AddHooks(serverSpanHook)
 	return nil
 }
 
 // SpanHook wraps a Span in a Timer and records a "success" or "fail"
 // metric when the Span ends based on whether an error was passed to `span.End`
 // or not.
 type SpanHook struct {
```
fishy: Do we want to export this type at all? I think the intended use case is that user register the `CreateServerSpanHook`, which will auto register this hook whenever a new server span is created, so is there a use case that user want to register a `SpanHook` manually? If there's a use case for that, then we likely will also need to export `newSpanHook` function, as callers outside of this package has no way to setup the timer (if the caller just construct `SpanHook` without calling `newSpanHook`, `timer` will be nil, and then when `OnPostStart` is called it will panic).

tracing/errors.go
```diff
@@ -1,9 +1,14 @@
 package tracing
 
 import (
+	"errors"
 	"fmt"
 )
 
+// ErrNotSpanHook is the error logged when you pass an object to span.AddHooks
+// that does not conform to any of the span hook interfaces.
+var ErrNotSpanHook = errors.New("Attempting to add non-SpanHook object into Span's hook registry")
```
fishy: 🔕 This error is only used in logging and never returned to caller, so does not really need to be exported.

style nit: don't capitalize error messages (golint or govet is supposed to complain about that, not sure why it did not happen).

tracing/hooks.go
```diff
@@ -1,74 +1,103 @@
...
 	OnAddCounter(span *Span, key string, delta float64) error
 }
 
 var (
-	hooks []BaseplateHook
+	hooks []CreateServerSpanHook
 )
 
-// RegisterBaseplateHook registers a Hook into the Baseplate request lifecycle.
+func isSpanHook(hook interface{}) bool {
```
fishy: 🔕 This function is probably worth exporting so users implementing a hook can use it in their tests.

metricsbp/baseplate_hooks.go
pacejackson: Nope, I don't think so, still getting used to the idea of exporting vs not exporting 😄

tracing/span.go
```diff
@@ -43,358 +40,289 @@ type contextKey int
...
-			s.SetTag(ZipkinBinaryAnnotationKeyTimeOut, true)
+// It is recommended that you only call AddHooks on a Span within an
+// OnCreate Hook so the Span is set up with all of it's Hooks as a part of
+// it's creation.
+func (s *Span) AddHooks(hooks ...interface{}) {
+	for _, hook := range hooks {
+		if IsSpanHook(hook) {
+			s.hooks = append(s.hooks, hook)
+		} else {
+			s.LogError("AddHooks error: ", errNotSpanHook)
```
fishy: 🔕 It's probably worth it to add more info to the error message to make debugging easier (currently when this happens people looking at the log only knows that something messed up, but they probably have no idea which one). We can use `fmt.Errorf` with either `%T` (which shows the type of the hook) or `%#v`, or both. See https://godoc.org/fmt#hdr-Printing for more info.

tracing/errors.go
```diff
@@ -1,13 +1,24 @@
...
+
+var _ error = (*errNotSpanHook)(nil)
+
+func (e *errNotSpanHook) Error() string {
+	return fmt.Sprintf(
+		"attempting to add non-SpanHook object <%T> into span's hook registry\n%#v",
+		e.obj,
+		e.obj,
+	)
+}
```
fishy: This is overkill since we don't export it, just use `fmt.Errorf` in place (with `LogError` inside `span.go`) is enough. Also don't use `"\n"` inside error message, `": "` should be enough.

PR #63, opened 2020-01-26
metricsbp: Split histogram into histogram and timing

fishy:
We used to use go-kit's statsd implementation, which only provided
timing as histogram. We have since switched to influxstatsd
implementation, which actually provided both histogram and timing.

I looked into the code, looks like the only difference between them is
that for histograms it's writing to the wire with "|h" while for timing
it's writing to the wire with "|ms".

### DIFF
--- a/metricsbp/doc.go
+++ b/metricsbp/doc.go
@@ -15,15 +15,18 @@
 //     goos: darwin
 //     goarch: amd64
 //     pkg: github.com/reddit/baseplate.go/metricsbp
-//     BenchmarkStatsd/pre-create/histogram-8           9548755               121 ns/op              39 B/op          0 allocs/op
-//     BenchmarkStatsd/pre-create/counter-8            10924286               122 ns/op              44 B/op          0 allocs/op
-//     BenchmarkStatsd/pre-create/gauge-8              100000000               10.7 ns/op             0 B/op          0 allocs/op
-//     BenchmarkStatsd/on-the-fly/histogram-8           4342429               247 ns/op              94 B/op          2 allocs/op
-//     BenchmarkStatsd/on-the-fly/counter-8             4850746               263 ns/op             125 B/op          2 allocs/op
-//     BenchmarkStatsd/on-the-fly/gauge-8              29172970                40.7 ns/op             0 B/op          0 allocs/op
-//     BenchmarkStatsd/on-the-fly-with-labels/histogram-8               2687974               449 ns/op             191 B/op          4 allocs/op
-//     BenchmarkStatsd/on-the-fly-with-labels/counter-8                 2680872               448 ns/op             191 B/op          4 allocs/op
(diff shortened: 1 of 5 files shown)


fishy:
cc @isugimpy

metricsbp/statsd.go
```diff
@@ -135,12 +135,21 @@ func (st Statsd) Counter(name string) metrics.Counter {
...
 func (st Statsd) Histogram(name string) metrics.Histogram {
+	return st.Statsd.NewHistogram(name, st.sampleRate)
+}
+
+// Timing returns a histogram metrics to the name with milliseconds as the unit.
+//
+// It uses the DefaultSampleRate used to create Statsd object.
+// If you need a different sample rate,
+// you could use st.Statsd.NewTiming instead.
+func (st Statsd) Timing(name string) metrics.Histogram {
```
pacejackson: Should we use this rather than a Histogram in `Timer`?

fishy: They both return the same type, after returned there's no way we can tell which way it was created before.

pacejackson: Right, but it’s changing the suffix on the timer, doesn’t what it currently uses (ms) make more sense in the context of a timer?

fishy: Yes, but it requires a lot of extra efforts to block people from using a normal histogram in a timer, which is not worthing it.

fishy: Even if someone accidentally used Histogram in Timer, in the end they are still able to see all the p99/p90/p50 stuff, the only thing is that they need to manually set the unit on the graphs. That's not a big deal.

pacejackson: Oh right, I was thinking that we created the Histogram in “NewTimer”

metricsbp/doc.go
```diff
@@ -15,15 +15,18 @@
...
-//     BenchmarkStatsd/pre-create/gauge-8              100000000               10.7 ns/op             0 B/op          0 allocs/op
-//     BenchmarkStatsd/on-the-fly/histogram-8           4342429               247 ns/op              94 B/op          2 allocs/op
-//     BenchmarkStatsd/on-the-fly/counter-8             4850746               263 ns/op             125 B/op          2 allocs/op
-//     BenchmarkStatsd/on-the-fly/gauge-8              29172970                40.7 ns/op             0 B/op          0 allocs/op
-//     BenchmarkStatsd/on-the-fly-with-labels/histogram-8               2687974               449 ns/op             191 B/op          4 allocs/op
-//     BenchmarkStatsd/on-the-fly-with-labels/counter-8                 2680872               448 ns/op             191 B/op          4 allocs/op
-//     BenchmarkStatsd/on-the-fly-with-labels/gauge-8                   3645584               331 ns/op             112 B/op          3 allocs/op
+//     BenchmarkStatsd/pre-create/histogram-8         	 8583646	       124 ns/op	      44 B/op	       0 allocs/op
```
fishy: The slightly increased memory is caused by the fallback change not this one. THe increased time is around 2ns, or just statistic error.

PR #65, opened 2020-01-30
metricsbp: Add labels to RunSysStats

fishy:
Since RunSysStats sets gauge value on a per-machine basis, allowing
additional labels would make more sense, otherwise the gauge will just
keeps getting overwritten between the machines.

This is a breaking change.

### DIFF
--- a/metricsbp/statsd_test.go
+++ b/metricsbp/statsd_test.go
@@ -12,7 +12,7 @@ import (
 func TestGlobalStatsd(t *testing.T) {
 	// Make sure global statsd is safe to use and won't cause panics, no real
 	// tests here:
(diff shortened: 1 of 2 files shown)


metricsbp/statsd_test.go
```diff
@@ -12,7 +12,7 @@ import (
 func TestGlobalStatsd(t *testing.T) {
 	// Make sure global statsd is safe to use and won't cause panics, no real
 	// tests here:
-	metricsbp.M.RunSysStats()
+	metricsbp.M.RunSysStats(nil)
```
konradreiche: Nice, I didn't know that `range` over `nil` doesn't panic.

PR #66, opened 2020-01-30
Add http go-kit integrations

pacejackson:
- [x] Add `httpbp` package to provide the base helpers to get headers from the HTTP request and put them on the context object.
- [x] Add `httpbp.HMACTrustHandler` to allow us to trust HTTP headers based on an HMAC signature
- [x] Add tests for `httpbp` package
- [ ] Add function to build a new EdgeContext from HTTP headers
- [ ] Add tests for `EdgeContextFromHTTP`
- [ ] Add go-kit Middleware to inject an EdgeContext into the context
- [ ] Add tests for `InjectEdgeContext`
- [ ] Add function to build a new Span from HTTP headers
- [ ] Add tests for `SpanFromHTTP`
- [ ] Add go-kit Middleware to inject a server Span into the context
- [ ] Add tests for `InjectServerSpan`

### DIFF
--- a/httpbp/headers.go
+++ b/httpbp/headers.go
@@ -1,0 +1,209 @@
+package httpbp
+
+import (
+	"context"
+	"net/http"
+)
+
+const (
+	// EdgeContextHeader is the key use to get the raw edge context from
+	// the HTTP request headers.
+	EdgeContextHeader = "X-Edge-Request"
+
+	// ParentIDHeader is the key use to get the span parent ID from
+	// the HTTP request headers.
+	ParentIDHeader = "X-Parent"
+
+	// SpanIDHeader is the key use to get the span ID from the HTTP
+	// request headers.
+	SpanIDHeader = "X-Span"
+
+	// SpanFlagsHeader is the key use to get the span flags from the HTTP
+	// request headers.
+	SpanFlagsHeader = "X-Flags"
+
+	// SpanSampledHeader is the key use to get the sampled flag from the
+	// HTTP request headers.
+	SpanSampledHeader = "X-Sampled"
+
+	// TraceIDHeader is the key use to get the trace ID from the HTTP
+	// request headers.
+	TraceIDHeader = "X-Trace"
+)
+
+type headerContextKey int
+
+const (
+	edgeContextContextKey headerContextKey = iota
+	traceIDContextKey
+	parentIDContextKey
+	spanIDContextKey
+	spanFlagsContextKey
+	spanSampledContextKey
+)
+
+// HeaderContextKey is an Enum used to get HTTP headers from a context object.
+//
+// It is not used as the actual key in the context, these enums are mapped to
+// un-exported values that serve as the actual context keys.
+type HeaderContextKey int
+
+const (
+	// EdgeContextContextKey is the key for the raw edge request context
+	EdgeContextContextKey HeaderContextKey = iota
+
+	// TraceIDContextKey is the header for the trace ID passed by the caller
+	TraceIDContextKey
+
+	// ParentIDContextKey is the header for the parent ID passed by the caller
+	ParentIDContextKey
+
+	// SpanIDContextKey is the header for the span ID passed by the caller
+	SpanIDContextKey
+
+	// SpanFlagsContextKey is the header for the span flags passed by the caller
+	SpanFlagsContextKey
+
+	// SpanSampledContextKey is the header for the sampled flag passed by the caller
+	SpanSampledContextKey
+)
+
+var headerKeyToContextKey = map[HeaderContextKey]headerContextKey{
+	EdgeContextContextKey: edgeContextContextKey,
+	TraceIDContextKey:     traceIDContextKey,
+	ParentIDContextKey:    parentIDContextKey,
+	SpanIDContextKey:      spanIDContextKey,
+	SpanFlagsContextKey:   spanFlagsContextKey,
+	SpanSampledContextKey: spanSampledContextKey,
+}
+
+// GetHeader returns the HTTP header stored on the context at key.
+func GetHeader(ctx context.Context, key HeaderContextKey) (header string) {
+	if contextKey, ok := headerKeyToContextKey[key]; ok {
+		if h, ok := ctx.Value(contextKey).(string); ok {
+			header = h
+		}
+	}
+	return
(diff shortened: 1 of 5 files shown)


httpbp/headers.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// HeaderSpanFlags is the header for the span flags passed by the caller
+	HeaderSpanFlags
+
+	// HeaderSpanSampled is the header for the sampled flag passed by the caller
+	HeaderSpanSampled
+)
+
+// GetHeader returns the HTTP header stored on the context at key.
+func GetHeader(ctx context.Context, key headerContextKey) (string, bool) {
```
fishy: Using an unexported type (`headerContextKey`) in an exported function (`GetHeader`) won't work as the caller has no way to provide that arg.

We also should not export the context key type, as that could lead to context key collisions (someone use the exported type to set context value and bypass our logic, probably accidentally). See the doc on https://godoc.org/context#WithValue for more discussion.

So we probably should have a set of internal type/values for context key, and a set of exported enum for read-only access.

httpbp/headers.go
```diff
@@ -0,0 +1,117 @@
...
+//			DecodeIsHealthyRequest,
+//			EncodeResponse,
+//			httpgk.ServerBefore(
+//				httpbp.PopulateRequestContext(trustHandler),
+//			),
+//		)
+func PopulateBaseplateRequestContext(t HeaderTrustHandler) func(ctx context.Context, r *http.Request) context.Context {
+	return func(ctx context.Context, r *http.Request) context.Context {
+		if t.TrustEdgeContext(r) {
+			ctx = context.WithValue(ctx, HeaderEdgeContext, r.Header.Get("X-Edge-Request"))
```
fishy: We should constantize those headers.

httpbp/headers.go
```diff
@@ -0,0 +1,117 @@
...
+		return header, true
+	}
+	return "", false
+}
+
+// HeaderTrustHandler provides an interface PopulateBaseplateRequestContext to
+// verify that it should trust the HTTP headers it receives.
+type HeaderTrustHandler interface {
+	TrustEdgeContext(r *http.Request) bool
+	TrustSpan(r *http.Request) bool
```
fishy: Can you add more comments into the interface about what do these 2 functions do? I can figure that out by reading the code of `PopulateBaseplateRequestContext`, but that should be part of the documentation :)

httpbp/headers.go
pacejackson: Already happening in the next commit :)

httpbp/trust_handers.go
```diff
@@ -0,0 +1,220 @@
...
+// caller to provide the signature of a message in the following format:
+//
+// 		"{header0}:{value0}|{header1}|{value1}|...|{headerN}:{valueN}"
+//
+// where the headers are sorted lexicographically.  Additionally, the signature
+// should be generated using the baseplate provided `signing.Sign` function.
+//
+// HMACTrustHandler provides implementations for both signing and verifying
+// edge context and span headers.
+type HMACTrustHandler struct {
```
pacejackson: I wonder if we should rename this to something like `BPSignatureTrustHandler` since the handler itself doesn't know or care about the signing method used?

pacejackson: Or maybe `HeaderSignatureTrustHandler`

httpbp/headers.go
```diff
@@ -0,0 +1,188 @@
...
+// It is not used as the actual key in the context, these enums are mapped to
+// un-exported values that serve as the actual context keys.
+type HTTPHeaderKey int
+
+const (
+	// EdgeContextHTTPHeader is the key for the raw edge request context
+	EdgeContextHTTPHeader HTTPHeaderKey = iota
+
+	// TraceIDHTTPHeader is the header for the trace ID passed by the caller
+	TraceIDHTTPHeader
```
fishy: these two are with suffix `HTTPHeader` whole others are only with suffix `Header`. I think since this is `httpbp` package we could just use `Header` suffix consistently.

httpbp/headers.go
```diff
@@ -0,0 +1,188 @@
...
+// baseplate resource (spans and edge contexts) into a struct that provides an
+// easy way to convert them into HTTP headers.
+//
+// This interface exists so we can avoid having to do runtime checks on maps to
+// ensure that they have the right keys set when we are trying to sign or verify
+// a set of HTTP headers.
+type Headers interface {
+	// AsHTTPHeaders returns the Headers struct as HTTP headers, a map of header
+	// keys to header values.
+	AsHTTPHeaders() map[string]string
```
fishy: `http.Header` is actually `map[string][]string`, because you could have more than one of the same header: https://godoc.org/net/http#Header

I'm not sure if we should map it into that type, or call it `AsMap` instead to avoid confusion.

httpbp/headers.go
```diff
@@ -0,0 +1,188 @@
...
+//		trustHandler := httpbp.NeverTrustHeaders{}
+//		httpgk.NewServer(
+//			endpoints.IsHealthy,
+//			DecodeIsHealthyRequest,
+//			EncodeResponse,
+//			httpgk.ServerBefore(
+//				httpbp.PopulateRequestContext(trustHandler),
+//			),
+//		)
+func PopulateBaseplateRequestContext(t HeaderTrustHandler) func(ctx context.Context, r *http.Request) context.Context {
```
fishy: We probably should also provide the inner function in the API, like `InjectContextFromHTTPHeader` or something, because this one can only be used with go-kit http package, some microservices don't really want to be forced to use that :)

httpbp/trust_handers.go
```diff
@@ -0,0 +1,220 @@
...
+	secrets               *secrets.Store
+	edgeContextSecretPath string
+	spanSecretPath        string
+}
+
+// NewHMACTrustHandler returns a new HMACTrustHandler that uses the provided
+// secrets.Store and secret paths to validate the sigantures of HTTP headers.
+func NewHMACTrustHandler(store *secrets.Store, edgeContextSecretPath string, spanSecretPath string) HMACTrustHandler {
```
fishy: nit: turn the arg list into a struct, which makes adding more stuff in the future non-breaking

PR #67, opened 2020-01-31
Add 'httpbp' package

pacejackson:
This package provides Baseplate specific helpers and integrations for http services using go-kit.

### DIFF
--- a/httpbp/example_response_test.go
+++ b/httpbp/example_response_test.go
@@ -1,0 +1,144 @@
+package httpbp_test
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"log"
+	"net/http"
+
+	"github.com/go-kit/kit/endpoint"
+	httpgk "github.com/go-kit/kit/transport/http"
+	"github.com/reddit/baseplate.go/httpbp"
+)
+
+// ExampleResponse is an example response that implements the ErrorResponse
+// interface.
+type ExampleResponse struct {
+	Message string `json:"message,omitempty"`
+
+	err     httpbp.HTTPError
+	cookies []*http.Cookie
+	headers http.Header
+	code    int
+}
+
+func (r *ExampleResponse) AddCookie(name, value string) {
+	r.cookies = append(r.cookies, &http.Cookie{Name: name, Value: value})
+}
+
+func (r ExampleResponse) Cookies() []*http.Cookie {
+	return r.cookies
+}
+
+func (r ExampleResponse) Headers() http.Header {
+	return r.headers
+}
+
+func (r ExampleResponse) StatusCode() int {
+	return r.code
+}
+
+// Err returns the internal HTTPError set on r.
+func (r ExampleResponse) Err() error {
+	return r.err
+}
+
+var (
+	// Verify that both ExampleResponse and *ExampleResponse implement the
+	// ResponseCookies interface.
+	_ httpbp.ResponseCookies = ExampleResponse{}
+	_ httpbp.ResponseCookies = (*ExampleResponse)(nil)
+	// Verify that both ExampleResponse and *ExampleResponse implement the
+	// go-kit http.Headerer interface.
+	_ httpgk.Headerer = ExampleResponse{}
+	_ httpgk.Headerer = (*ExampleResponse)(nil)
+	// Verify that both ExampleResponse and *ExampleResponse implement the
+	// go-kit http.StatusCoder interface.
+	_ httpgk.StatusCoder = ExampleResponse{}
+	_ httpgk.StatusCoder = (*ExampleResponse)(nil)
+	// Verify that both ExampleResponse and *ExampleResponse implement the
+	// ErrorResponse interface.
+	_ httpbp.ErrorResponse = ExampleResponse{}
+	_ httpbp.ErrorResponse = (*ExampleResponse)(nil)
+)
+
+// ExampleRequest is the request struct for our Example endpoint.
+type ExampleRequest struct {
+	// Error signals to the example endpoint whether it should return an error
+	// or not.
+	Error bool `json:"error"`
+}
+
+// DecodeExampleRequest decodes the request body into an ExampleRequest.
+func DecodeExampleRequest(_ context.Context, r *http.Request) (interface{}, error) {
+	var req ExampleRequest
+	json.NewDecoder(r.Body).Decode(&req)
+	return req, nil
+}
+
+// MakeExampleEndpoint builds a go-kit endpoint.Endpoint function that simply
+// returns an ExampleResponse with an error set.
+func MakeExampleEndpoint() endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		var resp ExampleResponse
+		req := request.(ExampleRequest)
+		if req.Error {
+			// Return a response that returns a non-nil error when
+			// httpbp.EncodeJSONResponse checks response.Err() which will signal it
+			// to send an error response rather than a normal one.
+			resp = ExampleResponse{
+				Message: "you'll never see this",
+				err: httpbp.HTTPError{
+					// Code sets the status code to return, defaults to
+					// http.InternalServerError (500).
+					Code: http.StatusBadGateway,
+
+					// Message sets a custom message for the error response
+					// body, defaults to http.StatusText for the status code.
+					Message: "Disruption downstream",
+
+					// Cause holds the error that triggered an error response.
+					// This is not communicated to the client but inspected by
+					// your service.
+					Cause: errors.New("database offline"),
+				},
+			}
+		} else {
+			// Return a non-error response.
+			resp = ExampleResponse{Message: "hello world!"}
+		}
+		return resp, nil
+	}
+}
+
+// This example demonstrates how to use HTTPError along with ErrorResponse to
+// return errors to API clients.
+//
+// Example request and response:
+//
+// request: {"error": false}
+// response:
+//		code: 200
+//		content-type: "application/json; charset=utf-8"
+//		body: {"message": "hello world!"}
+//
+// request: {"error": true}
+// response:
+//		code: 502
+//		content-type: "text/plain; charset=utf-8"
+//		body: Disruption dowstream
+func ExampleErrorResponse() {
+	// Create server handler
+	handler := http.NewServeMux()
+
+	// Register our example endpoint
+	handler.Handle("/example", httpgk.NewServer(
+		MakeExampleEndpoint(),
+		DecodeExampleRequest,
+		httpbp.EncodeJSONResponse,
+	))
+
+	// Start the server
+	log.Fatal(http.ListenAndServe(":8080", handler))
+}
--- a/httpbp/headers.go
+++ b/httpbp/headers.go
@@ -1,0 +1,182 @@
+package httpbp
+
+import (
+	"context"
+	"net/http"
+
+	httpgk "github.com/go-kit/kit/transport/http"
+)
+
+const (
+	// EdgeContextHeader is the key use to get the raw edge context from
+	// the HTTP request headers.
+	EdgeContextHeader = "X-Edge-Request"
+
+	// ParentIDHeader is the key use to get the span parent ID from
+	// the HTTP request headers.
+	ParentIDHeader = "X-Parent"
+
+	// SpanIDHeader is the key use to get the span ID from the HTTP
+	// request headers.
+	SpanIDHeader = "X-Span"
+
+	// SpanFlagsHeader is the key use to get the span flags from the HTTP
+	// request headers.
+	SpanFlagsHeader = "X-Flags"
+
+	// SpanSampledHeader is the key use to get the sampled flag from the
+	// HTTP request headers.
+	SpanSampledHeader = "X-Sampled"
+
+	// TraceIDHeader is the key use to get the trace ID from the HTTP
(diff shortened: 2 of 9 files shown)


httpbp/trust_handers.go
```diff
@@ -0,0 +1,233 @@
...
+	//
+	// If it can trust those headers, then the headers will be copied into the
+	// context object to later be used to initialize the server span for the
+	// request.
+	TrustSpan(r *http.Request) bool
+}
+
+// NeverTrustHeaders implements the HeaderTrustHandler interface and always
+// returns false.
+type NeverTrustHeaders struct{}
```
konradreiche: Is there a use for those beyond tests? If not I would move them into the test.

pacejackson: Historically this is what most HTTP services used 😅 , if anything `AlwaysTrust` is the one to remove.

fishy: For users using different signatures than ours? I think these 2 makes sense to provide as part of your API as they provide useful and reasonable alternatives to the users.

httpbp/headers.go
```diff
@@ -0,0 +1,209 @@
...
+type headerContextKey int
+
+const (
+	edgeContextContextKey headerContextKey = iota
+	traceIDContextKey
+	parentIDContextKey
+	spanIDContextKey
+	spanFlagsContextKey
+	spanSampledContextKey
+)
```
fishy: we actually don't need define these consts. When using them to read/write from the context, we can just do:

```
type headerContextKey HeaderContextKey

ctx.Value(headerContextKey(key))
```

This can also avoid the trouble of using the map.

httpbp/headers.go
```diff
@@ -0,0 +1,209 @@
...
+// HTTP headers.
+func NewEdgeContextHeaders(h http.Header) *EdgeContextHeaders {
+	return &EdgeContextHeaders{
+		EdgeRequest: h.Get(EdgeContextHeader),
+	}
+}
+
+// AsMap returns the EdgeContextHeaders as a map of header keys to header
+// values.
+func (s *EdgeContextHeaders) AsMap() map[string]string {
```
fishy: nit: This does not need to be a pointer receiver. Same for `SpanHeaders`

pacejackson: Is this because strings aren't actually copied?  Since they're immutable you can just copy the pointer and metadata?

fishy: Yes.

httpbp/trust_handers.go
pacejackson: The way I would think about it is:

1. `NeverTrustHeaders` you are exposed to the public internet and also do not expect to receive these headers anyways.
2. `AlwaysTrustHeaders` you only accept calls from within a trusted network and you feel comfortable always trusting these headers.
3. `HeaderSignatureTrustHandler` you are somewhere in between.  Maybe you are exposed to the public internet but also receive calls from internal clients and you want to be able to trust these headers conditionally.

pacejackson: (I'm adding those blurbs to the documentation for the appropriate trust handlers.)

httpbp/response.go
```diff
@@ -0,0 +1,135 @@
...
+
+type responseEncoder func(w http.ResponseWriter, r interface{}) error
+
+type encodeArgs struct {
+	contentType string
+	encoder     responseEncoder
+	errorMapper *ErrorMapper
+}
+
+func encodeResponse(w http.ResponseWriter, response interface{}, args encodeArgs) error {
```
fishy: 🔕 for unexported functions using a struct as arg is unnecessary (but I'm not against it either)

httpbp/response.go
```diff
@@ -0,0 +1,135 @@
...
+func DefaultErrorMapper(err error) error {
+	return err
+}
+
+type responseEncoder func(w http.ResponseWriter, r interface{}) error
+
+type encodeArgs struct {
+	contentType string
+	encoder     responseEncoder
+	errorMapper *ErrorMapper
```
fishy: Don't use pointers to function types.

httpbp/response.go
```diff
@@ -0,0 +1,135 @@
...
+	if resp, ok := response.(ErrorResponse); ok {
+		var mapper ErrorMapper
+		if args.errorMapper == nil {
+			mapper = DefaultErrorMapper
+		} else {
+			mapper = *args.errorMapper
+		}
+		if resp.Err() != nil {
+			return mapper(resp.Err())
+		}
```
fishy: this can be simplified as (after removed the pointer to `errorMapper`):

```
if args.errorMapper == nil {
  args.errorMapper = DefaultErrorMapper
}
if resp.Err() != nil {
  return args.errorMapper(resp.Err())
}
```

httpbp/response.go
```diff
@@ -0,0 +1,135 @@
...
+// order to have the httpbp.Encode methods automatically return http errors.
+type ErrorResponse interface {
+	// Err returns the error set on the response.
+	//
+	// This method is named "Err" rather than "Error" to avoid confusion with
+	// the "Error" function that is specified in the error interface.
+	Err() error
+
+	// SetError sets an error to be returned in the response.
+	SetError(err error)
```
fishy: is this function ever used?

pacejackson: I'm not sure it's necessary... I was thinking it would be good to have to clarify "this is the thing you use to set the error returned by `Err`"

fishy: i think most implementations will just change the underlying field directly, without calling a setter.

httpbp/response.go
pacejackson: Ya, I did it _more_ to keep the argument list smaller

fishy: yea it's fine to keep it as is.

httpbp/response.go
fishy: Here we are defining an interface for callers to implement and for us to call, the principle to that is to only define the functions we need to call. `SetErr` is not something we'll call, even if it's useful to the user/implementation.

httpbp/response.go
```diff
@@ -0,0 +1,210 @@
...
+	// Code is the status code to set on the HTTP response.  Defaults to 500 if
+	// it is not set.
+	Code int
+
+	// ResponseMessage is an optional message that can be returned to the
+	// client.  Defaults to the native http.StatusText message for the
+	// StatusCode() of the HTTPError.
+	ResponseMessage string
+
+	// Cause is an optional error that can be used to
```
fishy: unfinished sentence :)

httpbp/response.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	return http.StatusText(e.StatusCode())
+}
+
+// As implements helper interface for errors.As.
+//
+// If v is pointer to either HTTPError or *HTTPError,
+// *v will be set into this error.
+// Otherwise, As will try errors.As against the Cause of the HTTPError
+func (e HTTPError) As(v interface{}) bool {
```
fishy: 🔕 `As` is optional, but I guess it's useful here as you don't know if they are gonna return `HTTPError{}` or `&HTTPError{}`.

httpbp/response.go
```diff
@@ -0,0 +1,210 @@
...
+		return true
+	}
+	return errors.As(e.Cause, v)
+}
+
+// Is implements helper interface for errors.Is.
+//
+// Returns true if target is equal to the HTTPError.
+// Otherwise, Is will try errors.Is against the Cause of the HTTPError
+func (e HTTPError) Is(target error) bool {
```
fishy: `Is` is certainly not useful here. It's only used to compare against simple errors (that comes from `errors.New`). For custom error types `Is` won't work.

httpbp/response.go
```diff
@@ -0,0 +1,210 @@
...
+	StatusCode() int
+}
+
+// ErrorResponse is an interface that your Response objects can implement in
+// order to have the httpbp.Encode methods automatically return http errors.
+type ErrorResponse interface {
+	// Err returns the HTTPError set on the response.
+	//
+	// This method is named "Err" rather than "Error" to avoid confusion with
+	// the "Error" function that is specified in the error interface.
```
fishy: The second paragraph is unnecessary, this is standard go convention :)

httpbp/response.go
```diff
@@ -0,0 +1,210 @@
...
+	Err() *HTTPError
```
fishy: I would prefer to not force the error type here. When using it we can just do a casting:

```go
if respErr := resp.Err(); respErr != nil {
  var he HTTPError
  if !errors.As(respErr, &he) {
    he.Code = 500
    he.Cause = respErr
  }
  http.Error(w, he.ResponseMessage(), he.StatusCode())
  return nil
}
```

httpbp/response.go
```diff
@@ -0,0 +1,210 @@
...
+	Cause error
+}
+
+// Error returns the error message to send to the client.
+func (e HTTPError) Error() string {
+	if e.ResponseMessage != "" {
+		return e.ResponseMessage
+	}
+	return http.StatusText(e.StatusCode())
+}
```
fishy: I suggest move this logic into a different function (maybe call it `ResponseMessage`), and for `Error` do something more standard:

```
func (e HTTPError) Error() string {
  return fmt.Sprintf("httpbp: http error with code %d, message %q and cause %v", e.Code, e.Message, e.Cause)
}
```

httpbp/response.go
pacejackson: 🔪

httpbp/headers.go
```diff
@@ -0,0 +1,182 @@
...
+// verifies that it should trust the headers using the provided
+// HeaderTrustHandler, and attaches the trusted headers to the context.
+//
+// These headers can be retrieved using httpbp.GetHeader.
+//
+// This method does not implement the go-kit http.RequestFunc interface, if you
+// want to use this with go-kit, use PopulateBaseplateRequestContext to
+// return a function that wraps InjectBaseplateContextFromHTTP with the
+// HeaderTrustHandler and does implement the http.RequestFunc interface.
+func InjectBaseplateContextFromHTTP(ctx context.Context, t HeaderTrustHandler, r *http.Request) context.Context {
```
fishy: 🔕 These function names are getting long, I'm wondering if we could get them shorter, for example `ContextFromHTTP` for this one, `PopulateRequestContext` for the next one.

httpbp/response.go
```diff
@@ -0,0 +1,218 @@
...
+// Otherwise, As will try errors.As against the Cause of the HTTPError
+func (e HTTPError) As(v interface{}) bool {
+	if target, ok := v.(*HTTPError); ok {
+		*target = e
+		return true
+	} else if target, ok := v.(**HTTPError); ok {
+		*target = &e
+		return true
+	}
+	return errors.As(e.Cause, v)
```
fishy: 🔕 This part is kinda unnecessary. The whole purpose of `As` is to be used by `errors.As`. If you look at [its implementation](https://github.com/golang/go/blob/7d2473dc81c659fba3f3b83bc6e93ca5fe37a898/src/errors/wrap.go#L80-L87), it basically does these 3 things, in order:

1. try cast err into the type of `*v` directly (which is half of what we did above, but we try both pointer and non-pointer cases)
2. call `err`'s `As` implementation, if it's implemented (this function)
3. unwrap `err` and go back to 1

So if we just return false here, step 3 will still take care of it (because in our `Unwrap` implementation we just return `e.Cause`), with a slightly shallower call stack :)

httpbp/response.go
```diff
@@ -0,0 +1,218 @@
...
+	// StatusCode returns the HTTP status code to set on the response.
+	StatusCode() int
+}
+
+// ErrorResponse is an interface that your Response objects can implement in
+// order to have the httpbp.Encode methods automatically return http errors.
+type ErrorResponse interface {
+	// Err returns the HTTPError set on the response.
+	Err() error
+}
```
fishy: We likely need to provide an example, either here or at `HTTPError`, to demonstrate how to use `HTTPError` to implement `ErrorResponse` (this does not necessarily need to be a full testable example, it can just be code snippet inside the doc comment).

httpbp/response.go
fishy: Note: if we don't implement `As`, we can achieve the same thing with trying to call `errors.As` against both `HTTPError` and `*HTTPError` (or only implement error interface in `*HTTPError` to eliminate the possibility of using `HTTPError`):

```
var he HTTPError
if errors.As(err, &he) {
  ...
}
var phe *HTTPError
if errors.As(err, &phe) {
  ...
}
```

httpbp/example_response_test.go
```diff
@@ -0,0 +1,144 @@
...
+//		code: 200
+//		content-type: "application/json; charset=utf-8"
+//		body: {"message": "hello world!"}
+//
+// request: {"error": true}
+// response:
+//		code: 502
+//		content-type: "text/plain; charset=utf-8"
+//		body: Disruption dowstream
+func ExampleHTTPErrorResponse() {
```
fishy: The name here caused `go vet` to fail. If you name it `ExampleHTTPError` this example will be attached to `HTTPError`, if you name it `ExampleErrorResponse` it will be attached to `ErrorResponse`. You can also just rename it to `Example_HTTPErrorResponse` which will make this a package example.

fishy: See the naming convention part of https://pkg.go.dev/testing?tab=doc#hdr-Examples

PR #70, opened 2020-02-02
secrets: Allow adding additional middlewares

fishy:
Also:

1. Remove the type check after filewatcher.Get, which would never fail.
2. Add Store.Stop()

### DIFF
--- a/edgecontext/validator.go
+++ b/edgecontext/validator.go
@@ -1,10 +1,16 @@
 package edgecontext
 
 import (
+	"crypto/rsa"
 	"errors"
+	"fmt"
 
 	jwt "gopkg.in/dgrijalva/jwt-go.v3"
+
+	"github.com/reddit/baseplate.go/secrets"
 )
+
+type keysType []*rsa.PublicKey
 
 const (
 	authenticationPubKeySecretPath = "secret/authentication/public-key"
@@ -43,22 +49,21 @@ func shouldShortCircutError(err error) bool {
 // ValidateToken parses and validates a jwt token, and return the decoded
 // AuthenticationToken.
 func ValidateToken(token string) (*AuthenticationToken, error) {
-	sec, err := store.GetVersionedSecret(authenticationPubKeySecretPath)
-	if err != nil {
-		return nil, err
+	keys, ok := keysValue.Load().(keysType)
+	if !ok {
+		// This would only happen when all previous middleware parsing failed.
+		return nil, errors.New("no public keys loaded")
 	}
 
-	// TODO 1: Patch upstream to support key rotation natively:
+	// TODO: Patch upstream to support key rotation natively:
 	// https://github.com/dgrijalva/jwt-go/pull/372
-	//
-	// TODO 2: Use secrets middleware to cache parsed pubkeys.
 	var lastErr error
-	for _, key := range sec.GetAll() {
+	for _, key := range keys {
 		token, err := jwt.ParseWithClaims(
 			token,
 			&AuthenticationToken{},
 			func(_ *jwt.Token) (interface{}, error) {
-				return jwt.ParseRSAPublicKeyFromPEM([]byte(key))
+				return key, nil
 			},
 		)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


fishy:
cc @andizzle

secrets/store.go
```diff
@@ -59,62 +58,63 @@ func (s *Store) parser(r io.Reader) (interface{}, error) {
...
-	}
+// AddMiddlewares registers new middlewares to the store.
+//
+// Every AddMiddlewares call will cause all already registered middlewares to be
+// called again with the latest data.
+//
+// AddMiddlewares call is not thread-safe, it should not be called concurrently.
+func (s *Store) AddMiddlewares(middlewares ...SecretMiddleware) {
+	s.secretHandler(middlewares...)
+	s.secretHandlerFunc(s.getSecrets())
```
andizzle: As commented this would also trigger existing middlewares. What do you think have this function accepts  `SecretHandlerFunc`, call them and wrap them as `SecretMiddleware`?

https://github.com/reddit/baseplate.go/pull/71/files#diff-1a3fa2e612f9f352390e03634cc256d0R82

fishy: triggering existing middlewares shouldn't be a problem, because we are triggering them with the same secret data.

If we want to accept `SecretHandlerFunc` instead of `SecretMiddleware` as the arg, we should make it consistent and also `SecretMiddleware` in `NewStore` as well (and not export `SecretMiddleware` at all).

secrets/store.go
```diff
@@ -59,62 +58,63 @@ func (s *Store) parser(r io.Reader) (interface{}, error) {
...
 
-	return secrets.GetVersionedSecret(path)
+// GetSimpleSecret loads secrets from watcher, and fetches a simple secret from secrets
+func (s Store) GetSimpleSecret(path string) (SimpleSecret, error) {
+	return s.getSecrets().GetSimpleSecret(path)
 }
 
-// GetCredentialSecret loads secrets from watcher, and fetches a credential secret from secrets
-func (s *Store) GetCredentialSecret(path string) (CredentialSecret, error) {
-	secrets, ok := s.watcher.Get().(*Secrets)
```
konradreiche: I don't think we should remove the type check. I don't like making assumptions based on upstream code that may or may not change in the future. If the type would be set in stone just a few lines above I'd be okay with it but this is a different package.

edgecontext/validator.go
```diff
@@ -43,22 +49,21 @@ func shouldShortCircutError(err error) bool {
...
-	if err != nil {
-		return nil, err
+	keys, ok := keysValue.Load().(keysType)
+	if !ok {
+		// This would only happen when all previous middleware parsing failed.
+		return nil, errors.New("no public keys loaded")
 	}
 
-	// TODO 1: Patch upstream to support key rotation natively:
+	// TODO: Patch upstream to support key rotation natively:
```
konradreiche: Nice, can we add an issue in our repository to keep track of this?

secrets/store.go
```diff
@@ -59,62 +58,63 @@ func (s *Store) parser(r io.Reader) (interface{}, error) {
...
+}
 
-	return secrets.GetSimpleSecret(path)
+// Stop stops the underlying filewatcher and release associated resources.
+//
+// After Stop is called, you won't get any updates to the secret file,
+// but can still access the secrets as they were before Stop is called.
+//
+// It's OK to call Stop multiple times. Calls after the first one are no-ops.
+func (s *Store) Stop() {
```
konradreiche: I wonder if we can find a method name that fits the semantic more. I was a bit confused about stopping the store. It doesn't really stop the store but just the updating of the underlying file I guess?

secrets/store.go
fishy: 1. This is the behavior of stdlib's `atomic.Value`, I'm pretty sure that's part of the the go 1 backward compatibility guarantee
2. We control the parser implementation, and there's no reason we would return a different type in the parser.

secrets/store.go
fishy: `StopUpdating`? `Release`? Do you have any suggestions?

pacejackson: `Release` sounds good or maybe `Close`?

secrets/store.go
konradreiche: But on these lines I don't see that it's using `atomic.Value`, it's referencing a different package. It's just defensive programming, doesn't really hurt to keep the practice of checking the type when you type assert. The opposite, does.

secrets/store.go
konradreiche: In which scenario would this usually be called?

secrets/store.go
fishy: Even though, the second point still stands, that we control what are the values stored in the same package (parser implementation).

secrets/store.go
fishy: defer this right after a store is initialized to have an clean exit.

konradreiche: I mean for instance in production code. I saw the usage in test, I was just wondering it's purpose on the public interface of store :)

fishy: You can still defer this in the main function in prod code.

konradreiche: Now where I understand this better I think both `Close()` or `Stop()` are fine.

secrets/store.go
fishy: See also: https://pkg.go.dev/github.com/reddit/baseplate.go/filewatcher?tab=doc#Parser & https://pkg.go.dev/github.com/reddit/baseplate.go/filewatcher?tab=doc#Result.Get

secrets/store.go
fishy: Renamed to `Close`.

edgecontext/validator.go
fishy: https://github.com/reddit/baseplate.go/issues/73

secrets/store.go
konradreiche: Okay, sounds good!

Issue #69, opened 2020-02-02
Support HTTP services


pacejackson:
#67 
- [x] Add `httpbp` package to provide the base helpers to get headers from the HTTP request and put them on the context object.

- [x] Add function to build a new EdgeContext from HTTP headers
- [x] Add go-kit Middleware to inject an EdgeContext into the context
- [x] Add function to build a new Span from HTTP headers
- [x] Add go-kit Middleware to inject a server Span into the context

pacejackson:
#67 
#72

PR #72, opened 2020-02-02
Finish HTTP integrations

pacejackson:
Finish up the remaining tasks in #69

### DIFF
--- a/edgecontext/edgecontext.go
+++ b/edgecontext/edgecontext.go
@@ -7,6 +7,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/reddit/baseplate.go/httpbp"
 	"github.com/reddit/baseplate.go/internal/gen-go/reddit/baseplate"
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/secrets"
@@ -54,6 +55,27 @@ var deserializerPool = thrift.NewTDeserializerPool(
 		}
 	},
 )
+
+type contextKey int
+
+const (
+	edgeContextKey contextKey = iota
+)
+
+// SetEdgeContext sets the given EdgeRequestContext on the context object.
+func SetEdgeContext(ctx context.Context, ec *EdgeRequestContext) context.Context {
+	return context.WithValue(ctx, edgeContextKey, ec)
+}
+
+// GetEdgeContext gets the current EdgeRequestContext from the context object,
+// if set.
+func GetEdgeContext(ctx context.Context) (ec *EdgeRequestContext, ok bool) {
+	if e, success := ctx.Value(edgeContextKey).(*EdgeRequestContext); success {
+		ec = e
+		ok = success
+	}
+	return
+}
 
 // Config for Init function.
 type Config struct {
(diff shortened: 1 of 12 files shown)


edgecontext/edgecontext.go
```diff
@@ -137,3 +153,33 @@ func FromThriftContext(ctx context.Context) (*EdgeRequestContext, error) {
 		raw:    raw,
 	}, nil
 }
+
+// ContextFactory builds an *EdgeRequestContext from a context object.
+type ContextFactory func(ctx context.Context) (*EdgeRequestContext, error)
```
fishy: 🔕 Naming a function type `*Factory` sounds a little bit weird, but I don't have a better suggestion either

httpbp/headers.go
```diff
@@ -58,14 +58,16 @@ const (
 
 type headerContextKey HeaderContextKey
 
-func setHeader(ctx context.Context, key HeaderContextKey, value string) context.Context {
+// SetHeader sets the value on the context at key.
+func SetHeader(ctx context.Context, key HeaderContextKey, value string) context.Context {
```
fishy: I don't think we want to export this function, this opens a can of worms we don't want to deal with.

I think the only reason you want to export this is to be used in the test of edgecontext? Can we move that test into this package instead? Or by constructing the actual http headers?

httpbp/headers.go
```diff
@@ -58,14 +58,16 @@ const (
...
 	return context.WithValue(ctx, headerContextKey(key), value)
 }
 
 // GetHeader returns the HTTP header stored on the context at key.
-func GetHeader(ctx context.Context, key HeaderContextKey) (header string) {
-	if h, ok := ctx.Value(headerContextKey(key)).(string); ok {
+func GetHeader(ctx context.Context, key HeaderContextKey) (header string, ok bool) {
+	if h, success := ctx.Value(headerContextKey(key)).(string); success {
 		header = h
+		ok = success
```
fishy: If you want to use named returns, you can do this instead:

```
header, ok = ctx.Value(headerContextKey(key)).(string)
return
```

tracing/http.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/reddit/baseplate.go/thriftbp"
+)
+
+// StartSpanFromHTTPContext creates a server span from http context object.
+//
+// This span would usually be used as the span of the whole http endpoint
+// handler, and the parent of the child-spans.
+//
+// Caller should pass in the context object they got from http library,
+// which would have all the required headers already injected.
```
fishy: I think we should mention here that registering `PopulateRequestContext` in `ServerBefore` is required.

httpbp/headers.go
fishy: On second thought, this is OK. We already do that in thrift library.

But in that case we no longer need the extra unexported type of `headerContextKey`, we should just use `HeaderContextKey` as the context key directly.

tracing/http.go
fishy: The last paragraph of the comment should be merged here. This paragraph works for the thrift case because the feature is built into thrift's `SimpleServer`, no extra steps needed. Here we do need the extra step of `PopulateRequestContext`. Saying "context object they got from http library ... have all the required headers already injected" is kinda untrue :)

Also I think you actually meant go-kit's http library here, which should also be called out (by default people would assume you are talking about stdlib http package, which does not come with contexts in handlers).

fishy: now we no longer need this paragraph.

edgecontext/middlewares_test.go
```diff
@@ -0,0 +1,247 @@
...
+		context.Background(),
+		httpbp.EdgeContextContextKey,
+		headerWithValidAuth,
+	)
+}
+
+func TestInitializeEdgeContext(t *testing.T) {
+	t.Parallel()
+
+	const expectedUserID = "t2_example"
```
konradreiche: This seems to be unused.

PR #75, opened 2020-02-05
metricsbp: Actually sample counters and histograms

fishy:
go-kit's influxstatsd implementation doesn't actually do sample. It
expects the user to do sampling manually, and report the sample rate to
them. Implement SampledCounter and SampledHistogram to actually do
sampling.

Also rename StatsdConfig.Default*SampleRate to *SampleRate, as the
semantic changed because of this change.

This is a breaking change.

### DIFF
--- a/metricsbp/statsd.go
+++ b/metricsbp/statsd.go
@@ -83,14 +83,14 @@ type StatsdConfig struct {
 	// If it's not ending with a period ("."), a period will be added.
 	Prefix string
 
-	// The default reporting sample rate used when creating counters and
-	// timings/histograms.
+	// The reporting sample rate used when creating counters and
+	// timings/histograms, respectively.
 	//
 	// For user convenience,
 	// we actually treat zero values (within 1e-9 since it's float) as 1 (100%),
 	// and <-1e-9 as 0 (0%).
-	DefaultCounterSampleRate   float64
-	DefaultHistogramSampleRate float64
+	CounterSampleRate   float64
+	HistogramSampleRate float64
 
 	// Address is the UDP address (in "host:port" format) of the statsd service.
 	//
@@ -143,8 +143,8 @@ func NewStatsd(ctx context.Context, cfg StatsdConfig) *Statsd {
 	st := &Statsd{
 		Statsd:              influxstatsd.New(prefix, log.KitLogger(cfg.LogLevel), labels...),
 		ctx:                 ctx,
-		counterSampleRate:   convertSampleRate(cfg.DefaultCounterSampleRate),
-		histogramSampleRate: convertSampleRate(cfg.DefaultHistogramSampleRate),
+		counterSampleRate:   convertSampleRate(cfg.CounterSampleRate),
+		histogramSampleRate: convertSampleRate(cfg.HistogramSampleRate),
 	}
 
 	if cfg.Address != "" {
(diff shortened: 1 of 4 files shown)


metricsbp/statsd.go
```diff
@@ -159,34 +159,46 @@ func NewStatsd(ctx context.Context, cfg StatsdConfig) *Statsd {
...
-// It uses the DefaultCounterSampleRate used to create Statsd object.
-// If you need a different sample rate,
-// you could use st.Statsd.NewCounter instead.
+// Counter returns a counter metrics to the name,
+// with sample rate inherited from StatsdConfig.
 func (st *Statsd) Counter(name string) metrics.Counter {
 	st = st.fallback()
-	return st.Statsd.NewCounter(name, st.counterSampleRate)
+	counter := st.Statsd.NewCounter(name, st.counterSampleRate)
+	if st.counterSampleRate >= 1 {
```
pacejackson: Should we also do this if `st.counterSampleRate == 0`?

metricsbp/statsd.go
```diff
@@ -159,34 +159,46 @@ func NewStatsd(ctx context.Context, cfg StatsdConfig) *Statsd {
...
-// It uses the DefaultHistogramSampleRate used to create Statsd object.
-// If you need a different sample rate,
-// you could use st.Statsd.NewHistogram instead.
+// Histogram returns a histogram metrics to the name with no specific unit,
+// with sample rate inherited from StatsdConfig.
 func (st *Statsd) Histogram(name string) metrics.Histogram {
 	st = st.fallback()
-	return st.Statsd.NewHistogram(name, st.histogramSampleRate)
+	histogram := st.Statsd.NewHistogram(name, st.histogramSampleRate)
+	if st.histogramSampleRate >= 1 {
```
pacejackson: Should we also do this if `st. histogramSampleRate == 0`?

metricsbp/statsd.go
```diff
@@ -159,34 +159,46 @@ func NewStatsd(ctx context.Context, cfg StatsdConfig) *Statsd {
...
-// It uses the DefaultHistogramSampleRate used to create Statsd object.
-// If you need a different sample rate,
-// you could use st.Statsd.NewTiming instead.
+// Timing returns a histogram metrics to the name with milliseconds as the unit,
+// with sample rate inherited from StatsdConfig.
 func (st *Statsd) Timing(name string) metrics.Histogram {
 	st = st.fallback()
-	return st.Statsd.NewTiming(name, st.histogramSampleRate)
+	histogram := st.Statsd.NewTiming(name, st.histogramSampleRate)
+	if st.histogramSampleRate >= 1 {
```
pacejackson: Same question.

metricsbp/statsd.go
fishy: That will make things slightly faster, but then again I don't think we actually use zero sample rate in real world so this sounds like a premature optimization. The current code still works when sample rate is 0, it's just slower.

pacejackson: I would only argue that `0` is the default value, so it might be worth adding the optimization just because of that?  That said:

1. I'm not sure how often that'll come up
2. I'm not sure how much of a gain it actually is

fishy: 1 is the actual default value now (since last change)

pacejackson: Welp, I think it’s more than safe to ignore then 😅

PR #76, opened 2020-02-06
Stop HTTP spans after endpoint finishes

pacejackson:
Use the same `endpoint.Middleware` that we are using to inject server spans into the context to also stop the server span after the endpoint has finished.

### DIFF
--- a/tracing/middlewares.go
+++ b/tracing/middlewares.go
@@ -19,21 +19,13 @@ import (
 // the context object.  This can be done by adding httpbp.PopulateRequestContext
 // as a ServerBefore option when setting up the request handler for an endpoint.
 func InjectHTTPServerSpan(name string) endpoint.Middleware {
-	return func(next endpoint.Endpoint) endpoint.Endpoint {
-		return func(ctx context.Context, request interface{}) (response interface{}, err error) {
-			ctx, span := StartSpanFromHTTPContext(ctx, name)
-			defer span.Stop(ctx, err)
-
-			response, err = next(ctx, request)
-			return
-		}
-	}
+	return InjectHTTPServerSpanWithTracer(name, nil)
 }
 
 // InjectHTTPServerSpanWithTracer is the same as InjectHTTPServerSpan except it
 // uses StartSpanFromHTTPContextWithTracer to initialize the server span rather
 // than StartSpanFromHTTPContext.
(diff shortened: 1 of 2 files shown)


tracing/middlewares.go
```diff
@@ -9,14 +9,20 @@ import (
...
-// Note, this depends on the edge context headers already being set on the
-// context object.  This can be done by adding httpbp.PopulateRequestContext as
-// a ServerBefore option when setting up the request handler for an endpoint.
+// Starts the server span before calling the `next` endpoint and stops the span
+// after the endpoint finishes.
+//
+// Note, this function depends on the edge context headers already being set on
+// the context object.  This can be done by adding httpbp.PopulateRequestContext
+// as a ServerBefore option when setting up the request handler for an endpoint.
 func InjectHTTPServerSpan(name string) endpoint.Middleware {
```
fishy: :no_bell: This function can be just a wrapper for the other one so you don't have to duplicate the logic.

pacejackson: Ya, previously it was just a basic wrapper for the `StartSpan` methods so I figured it was fine to just use the two separate functions but now that there's other logic, it's probably better to just have `InjectHTTPServerSpan` call `InjectHTTPServerSpanWithTracer` with a `nil` tracer.

tracing/middlewares.go
```diff
@@ -9,26 +9,32 @@ import (
...
-func InjectHTTPServerSpanWithTracer(tracer *Tracer, name string) endpoint.Middleware {
+func InjectHTTPServerSpanWithTracer(name string, tracer *Tracer) endpoint.Middleware {
 	return func(next endpoint.Endpoint) endpoint.Endpoint {
-		return func(ctx context.Context, request interface{}) (interface{}, error) {
-			ctx, _ = StartSpanFromHTTPContextWithTracer(ctx, name, tracer)
-			return next(ctx, request)
+		return func(ctx context.Context, request interface{}) (response interface{}, err error) {
+			ctx, span := StartSpanFromHTTPContextWithTracer(ctx, name, tracer)
+			defer func() {
+				span.Stop(ctx, err)
```
pacejackson: The "simplified" way didn't work because `ctx` and `err` were being evaluated when `defer` executed rather than when the `defer`-ed `span.Stop` is executed.

PR #78, opened 2020-02-07
metricsbp: Add Statsd.Ctx() function

fishy:


metricsbp/statsd.go
```diff
@@ -215,3 +215,30 @@ func (st *Statsd) fallback() *Statsd {
...
+//           select {
+//           case <- metricsbp.M.Ctx().Done():
+//             return
+//           case <- ticker.C:
+//             gauge.Set(getValue())
+//           }
+//         }
+//       }
+//     }
+func (st *Statsd) Ctx() context.Context {
```
pacejackson: Does this need to be a pointer?

fishy: No. `context.Context` is an interface.

PR #80, opened 2020-02-09
Add a global Store to the secrets package

pacejackson:


pacejackson:
> How will this be used?

It's really just there as a convenience for service owners.  Pretty much all services need access to some sort of secrets and secrets are something that seem fine to provide global access for.

_That said_ I'm going to ponder this a bit more.  One reason this is ok with the metrics client is we can initialize it to a default one that just pushes the metrics to nowhere so if you try to access it without updating it, it at least still works and does not panic.  I'm not sure that the same is true for this global `secrets.Store` so it might be worth considering if there's a way to initialize this so it's just empty, or make it a nil `Store` does not panic.

fishy:
> > How will this be used?
> 
> It's really just there as a convenience for service owners. Pretty much all services need access to some sort of secrets and secrets are something that seem fine to provide global access for.
> 
> _That said_ I'm going to ponder this a bit more. One reason this is ok with the metrics client is we can initialize it to a default one that just pushes the metrics to nowhere so if you try to access it without updating it, it at least still works and does not panic. I'm not sure that the same is true for this global `secrets.Store` so it might be worth considering if there's a way to initialize this so it's just empty, or make it a nil `Store` does not panic.

It should be trivial to make the global one initialized with an empty one (that won't panic, but trying to get anything from it will give you an error, but not panic)

konradreiche:
I see, I have personally never made access to those things globally but I can see a lot of convenience for that depending on the time of library, i.e. https://github.com/spf13/viper provides a global configuration accessible from everywhere. Only drawbacks I know of are that other packages can manipulate the state and it might make testing harder, but I don't think this applies here since it's just another way to use it. We only need to ask ourselves how we want this to be used by the implementing side.

pacejackson:
The more I think about this, the less sure I am that it makes sense to provide secrets as a global.  The reason being that, unlike metrics, you're going to need to initialize this to something both before your service can do something useful, and before you can test anything that actually uses it.

Making it a global obscures that and makes it harder to figure out if secrets are a dependency of what you are testing.  If you are passing it around either by initializing other structs with one or injecting it into your endpoint then you know that it is a dependency and you have to initialize it.

Given that, I'm _inclined_ to just close this but I can be persuaded either way (and I think services _can_ make this a global if they want to).

fishy:
from the microservices' point of view, they usually need to initialize a secrets store in main, then either 1) pass it to all the places (packages, objects, etc.) it's needed, or 2) set it to a global variable in a package that's imported by all the packages need to use secrets. option 1 doesn't sound too bad so I can live with that, but 2 is definitely much worse than defining the global variable here.

pacejackson:
Let's start with option 1 and see how that goes, I'm open to re-opening this PR (or more likely make a new one) down the line if we decide that we do want to support this.

PR #81, opened 2020-02-09
Add httpbp.BaseResponse

pacejackson:


httpbp/response.go
```diff
@@ -107,6 +107,84 @@ type ErrorResponse interface {
...
+	return r.code
+}
+
+// Headers returns the http.Header collection of headers to set on the response.
+func (r BaseResponse) Headers() http.Header {
+	return r.headers
+}
+
+// SetCookie adds a cookie to set on the response.
+func (r *BaseResponse) SetCookie(cookie *http.Cookie) {
```
fishy: We probably also need a `ClearCookies` api?

httpbp/response.go
```diff
@@ -107,6 +107,84 @@ type ErrorResponse interface {
...
+	r.cookies = append(r.cookies, cookie)
+}
+
+// Cookies returns the current list of cookies to set on the response.
+func (r BaseResponse) Cookies() []*http.Cookie {
+	return r.cookies
```
fishy: In go, returning a slice field in a function makes the ownership muddy. Say we have 3 items in the slice, the caller can actually change the 3 items inside the slice to other values, but they cannot add new items nor remove the size of the slice.

As a result, if the intention is to give read-only access, making a copy is recommended:

```go
cookies := make([]*http.Cookie, len(r.cookies))
copy(cookies, r.cookies)
return cookies
```

fishy: note that if after `Cookies()` is called we no longer care about the underlying cookies (e.g. the object should be gc'd after that) then this is fine.

pacejackson: In the way that `baseplate.go` uses `Cookies()`, we don't care about it _but_ I think it's still a good idea for cases where we want to use it outside of just the `EncodeResponse` methods in baseplate.

PR #82, opened 2020-02-11
Add redisbp.MonitoredCmdableFactory

pacejackson:
These Factory objects allow you to create new go-redis clients
that are monitored by RedisSpanHooks using the span for the
given context.

This also removes the 'integrations' package in favor of a
'redisbp' package, mostly to improve the readability of the
exported structs and functions.

### DIFF
--- a/redisbp/example_monitored_client_test.go
+++ b/redisbp/example_monitored_client_test.go
@@ -1,0 +1,97 @@
+package redisbp_test
+
+import (
+	"context"
+
+	"github.com/go-redis/redis/v7"
+
+	"github.com/reddit/baseplate.go/redisbp"
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+// This example demonstrates how to use a MonitoredCmdableFactory to create
+// monitored redis.Client objects.
+func ExampleMonitoredCmdableFactory_client() {
+	// variables should be properly initialized in production code
+	var tracer *tracing.Tracer
+	// Create a factory
+	factory := redisbp.NewMonitoredClientFactory(
+		"redis",
+		redis.NewClient(&redis.Options{Addr: ":6379"}),
+	)
+	// Get a context object and a server Span, with the server Span set on the
+	// context
+	ctx, _ := tracing.CreateServerSpanForContext(context.Background(), tracer, "test")
+	// Create a new client using the context for the Server Span
+	client := factory.BuildClient(ctx)
+	// Commands should now be wrapped using Client Spans
+	client.Ping()
+}
+
+// This example demonstrates how to use a MonitoredCmdableFactory to create
+// monitored redis.ClusterClient objects.
+func ExampleMonitoredCmdableFactory_cluster() {
+	// variables should be properly initialized in production code
+	var tracer *tracing.Tracer
+	// Create a factory
+	factory := redisbp.NewMonitoredClusterFactory(
+		"redis",
+		redis.NewClusterClient(&redis.ClusterOptions{
+			Addrs: []string{":7000", ":7001", ":7002"},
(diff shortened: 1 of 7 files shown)


redisbp/monitored_client.go
```diff
@@ -0,0 +1,99 @@
...
+
+// NewMonitoredRingFactory creates a MonitoredCmdableFactory for a redis.Ring
+// object.
+func NewMonitoredRingFactory(name string, client *redis.Ring) MonitoredCmdableFactory {
+	return newMonitoredCmdableFactory(name, &monitoredRing{Ring: client})
+}
+
+// BuildClient returns a new MonitoredCmdable with its context set to the
+// provided one.
+func (f MonitoredCmdableFactory) BuildClient(ctx context.Context) MonitoredCmdable {
```
fishy: This feels like an extra step that's unnecessary. What if we just put this into the 3 constructors (so it's for example `func NewMonitoredClient(ctx context.Context, name string, client *redis.Client) MonitoredCmdable` instead). This way we don't need `MonitoredCmdableFactory` type, and don't need `WithMonitoredContext` inside `MonitoredCmdable` interface.

pacejackson: The two things I like about the approach taken here are:

1. You only have to run the "setup" code to add the SpanHook once.
2. (This is the main one) You don't actually have access to the unmonitored, base client so it's harder to accidentally use it.

fishy: Do you mean `BuildClient` should be called every time a client is used (e.g. it should actually be `factory.BuildClient(ctx).Ping()`, rather than `client := factory.BuildClient(ctx); ... ; client.Ping()`?) If that's the case then yes this is better, but it should be reflected in the example :)

pacejackson: It's more targeted at service code where you would do something like:

```go
func (s service) MyEndpoint(ctx context.Context, req controllers.MyEndpointRequest, resp *controllers.MyEndpointResponse) error {
	redis := s.Redis.BuildClient(ctx)
	return controllers.MyEndpoint(ctx, req, resp, redis)
}
```

Where you use it on each new request to build a monitored client to inject into the actual request handler/controller.

fishy: Ah ok, probably worth explaining that in the doc comment and/or example.

pacejackson: Good call, I'll update the examples to reflect the intended usage.

redisbp/example_monitored_client_test.go
```diff
@@ -5,93 +5,147 @@ import (
...
 	"github.com/go-redis/redis/v7"
 
+	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/redisbp"
 	"github.com/reddit/baseplate.go/tracing"
 )
 
+// ExampleService is an example go-kit service to help demonstrate how to use
+// redisbp.type MonitoredCmdableFactory in a service.
+type ExampleService struct {
```
fishy: Try use `My*` instead of `Example*`, except that one function, so that the godoc server will render the example correctly: https://pkg.go.dev/testing?tab=doc#hdr-Examples

>The entire test file is presented as the example when it contains a single example function, at least one other function, type, variable, or constant declaration, and no test or benchmark functions.

pacejackson: Ya, `go vet` yelled at me so I changed it to just `Service`.

fishy: Oh right.

Now you have more than one example functions in that file, so it won't be rendered as whole file in the doc site. Maybe just use one and explain that you can also use other options in comments?

PR #83, opened 2020-02-12
Convert all span tags to strings

pacejackson:
Our current system for storing trace data only supports string values for binary annotations so we want to also encode boolean values as lower case strings.

### DIFF
--- a/tracing/trace.go
+++ b/tracing/trace.go
@@ -3,6 +3,7 @@ package tracing
 import (
 	"context"
 	"math/rand"
+	"strconv"
 	"time"
 
 	"github.com/reddit/baseplate.go/timebp"
@@ -64,6 +65,9 @@ func (t *trace) addCounter(key string, delta float64) {
 }
 
 func (t *trace) setTag(key string, value interface{}) {
+	if v, ok := value.(bool); ok {
+		value = interface{}(strconv.FormatBool(v))
+	}
 	t.tags[key] = value
 }
 
--- a/tracing/trace.go
+++ b/tracing/trace.go
@@ -65,13 +65,7 @@ func (t *trace) addCounter(key string, delta float64) {
 }
 
 func (t *trace) setTag(key string, value interface{}) {
-	var valStr string
-	if v, ok := value.(string); !ok {
-		valStr = fmt.Sprintf("%v", value)
-	} else {
-		valStr = v
-	}
-	t.tags[key] = valStr
+	t.tags[key] = fmt.Sprintf("%v", value)
 }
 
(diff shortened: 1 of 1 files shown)


tracing/trace.go
```diff
@@ -64,6 +65,9 @@ func (t *trace) addCounter(key string, delta float64) {
 }
 
 func (t *trace) setTag(key string, value interface{}) {
+	if v, ok := value.(bool); ok {
+		value = interface{}(strconv.FormatBool(v))
```
fishy: I don't think `interface{}()` is necessary? `value = strconv.FormatBool(v)` should just work?

pacejackson: You're right, I thought it would complain about setting an `interface{}` to another value.

tracing/trace.go
```diff
@@ -66,7 +66,7 @@ func (t *trace) addCounter(key string, delta float64) {
 
 func (t *trace) setTag(key string, value interface{}) {
 	if v, ok := value.(bool); ok {
-		value = interface{}(strconv.FormatBool(v))
+		value = strconv.FormatBool(v)
```
fishy: 🔕 Actually I wonder if we should just convert anything that's not string into string? e.g.

```
if _, ok := value.(string); !ok {
  value = fmt.Sprintf("%v", v)
}
```

Or maybe change `value` from `interface{}` into `string`?

pacejackson: 🤔 

I think converting everything to a string makes sense, since this is a detail of the current system we use for tracing rather than a part of the tracing spec, I don't think we want to encode that into the API.

We could make the type on `setTag` into string and do the conversion in `Span.SetTag` though?

tracing/trace.go
```diff
@@ -65,10 +65,13 @@ func (t *trace) addCounter(key string, delta float64) {
...
 
 func (t *trace) setTag(key string, value interface{}) {
-	if v, ok := value.(bool); ok {
-		value = strconv.FormatBool(v)
+	var valStr string
+	if v, ok := value.(string); !ok {
+		valStr = fmt.Sprintf("%v", value)
+	} else {
+		valStr = v
 	}
```
fishy: I'm not sure which is faster, always do `fmt.Sprintf("%v", value)` even if it's string, or do a typecasting test. 😅 Also `ok` being true is slightly different from "`value` is string", as it could be typedef'd string or some other weird cases.

konradreiche: I think using `fmt.Sprintf("%v", value)` makes the most sense here :)

PR #84, opened 2020-02-13
Add go-kit middleware for Thrift services

pacejackson:


edgecontext/middlewares.go
```diff
@@ -46,3 +40,24 @@ func InjectHTTPEdgeContext(logger log.Wrapper) endpoint.Middleware {
...
+	return InitializeEdgeContext(ctx, logger, FromThriftContext)
+}
+
+// InjectThriftEdgeContext returns a go-kit endpoint.Middleware that injects an
+// edge request context created from the Thrift headers set on the context into
+// the `next` endpoint.Endpoint.
+//
+// Note, this depends on the edge context headers already being set on the
+// context object.  You can configure the thrift server to do this by passing
+// thriftbp.HeadersToForward to the server's SetForwardHeaders method.
```
fishy: `SetForwardHeaders` only affects how this server will forward headers to upstreams. For this to work that's not required. thrift's TSimpleServer implementation already injected it for you.

tracing/middlewares.go
```diff
@@ -38,3 +38,34 @@ func InjectHTTPServerSpanWithTracer(name string, tracer *Tracer) endpoint.Middle
...
+
+// InjectThriftServerSpanWithTracer is the same as InjectThriftServerSpan except it
+// uses StartSpanFromThriftContextWithTracer to initialize the server span rather
+// than StartSpanFromThriftContext.
+func InjectThriftServerSpanWithTracer(name string, tracer *Tracer) endpoint.Middleware {
+	return func(next endpoint.Endpoint) endpoint.Endpoint {
+		return func(ctx context.Context, request interface{}) (response interface{}, err error) {
+			ctx, span := StartSpanFromThriftContextWithTracer(ctx, name, tracer)
+			defer func() {
+				span.Stop(ctx, err)
```
konradreiche: Should we check the error here?

fishy: good point. it might make sense to pass in a `log.Wrapper` and log the error here. but I won't worry about it too much (e.g. if it's too inconvenient to pass in `log.Wrapper` then I'm fine with ignoring the errors)

pacejackson: `tracer` has a Logger :)

PR #85, opened 2020-02-13
thriftpool: Change the benchmark test to parallel

fishy:
This is closer to how they are used in real world code.


fishy:
For comparison, this is the previous (non-parallel) benchmark result:

```
$ go test -bench . -benchmem
goos: darwin
goarch: amd64
pkg: github.com/reddit/baseplate.go/thriftpool
BenchmarkPoolGetRelease/channel-8         	12347830	        95.7 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/reddit/baseplate.go/thriftpool	1.404s
```

PR #86, opened 2020-02-14
thriftpool: Fix a bug on active connection counting

fishy:
Previously we only decrease number of active connections when we
successfully returned the client to the pool, which means when there are
errors (usually when Close reported error, or when we try to open
a new client in Release and the opener returned an error), we don't
decrease the number, even though semantically the client is returned to
the pool. That leads to the pool reporting exhausted when even we tried
to return every client we got from the pool.

### DIFF
--- a/thriftpool/channel.go
+++ b/thriftpool/channel.go
@@ -74,6 +74,10 @@ func (cp *channelPool) Release(c Client) error {
 		return nil
 	}
 
+	// As long as c is not nil, we always need to decrease numActive by 1,
+	// even if we encounter errors here, either due to close or opener.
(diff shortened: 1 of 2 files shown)


fishy:
> This seems right to me, are there any potential gotchas we should note (I wouldn't think so but 🤷‍♂ )?

one potential gotcha is that in normal cases we shouldn't see a lot of errors in `client.Close` or from client opener, so if it's causing us issues something is probably seriously wrong, and we _might_ prefer the pool to report exhausted so we can report the service as unhealthy? but that can be protected by a circuit breaker so we don't need to rely on the pool reporting unhealthy for it.

PR #88, opened 2020-02-17
Add thriftclient package

pacejackson:
This package adds a `MonitoredClient`, which is an implementation of `thrift.TClient` that wraps the `Call` to an internal client in a thrift client span.

This is in it's own package because it has to import from `tracing` which currently imports from `tracingbp`.

It also adds 2 other `thrift.TClient` implementations that can be used to mock out clients and help with unit testing.

### DIFF
--- a/thriftclient/client.go
+++ b/thriftclient/client.go
@@ -1,0 +1,58 @@
+package thriftclient
+
+import (
+	"context"
+
+	"github.com/apache/thrift/lib/go/thrift"
+
+	"github.com/reddit/baseplate.go/log"
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+// MonitoredClient implements the thrift.TClient interface and can be used to
+// wrap thrift calls in a client span.
+//
+// Calls will only be wrapped in client spans if the given context has a parent
+// span (either a server or "active" local/client span) set on it.  If no such
+// span is set, then the call will still proceed, it will just not be monitored
+// by a client span.
+type MonitoredClient struct {
+	// Client is the inner thrift.TClient implementation wrapped by the
+	// MonitoredClient.
+	Client thrift.TClient
+}
+
+// NewMonitoredClientFromFactory returns a pointer to a new MonitoredClient that
+// a new TStandardClient using the provided transport and protocol factory.
+func NewMonitoredClientFromFactory(transport thrift.TTransport, factory thrift.TProtocolFactory) *MonitoredClient {
+	return &MonitoredClient{thrift.NewTStandardClient(
+		factory.GetProtocol(transport),
+		factory.GetProtocol(transport),
+	)}
+}
+
+// Call wraps the underlying thrift.TClient.Call method with a client span.
+func (c MonitoredClient) Call(ctx context.Context, method string, args, result thrift.TStruct) (err error) {
+	var span *tracing.Span
+	if span = tracing.GetActiveSpan(ctx); span == nil {
+		span = tracing.GetServerSpan(ctx)
+	}
+	var child *tracing.Span
+	if span != nil {
+		ctx, child = span.ChildAndThriftContext(ctx, method)
+	}
+	defer func() {
+		if child != nil {
+			if stopErr := child.Stop(ctx, err); stopErr != nil {
+				log.Error("Error trying to stop span: " + stopErr.Error())
+			}
+		}
+	}()
+	return c.Client.Call(ctx, method, args, result)
+}
+
+// Validate that MonitoredClient implements thrift.TClient
+var (
+	_ thrift.TClient = MonitoredClient{}
+	_ thrift.TClient = (*MonitoredClient)(nil)
+)
--- a/thriftclient/client_test.go
+++ b/thriftclient/client_test.go
@@ -1,0 +1,178 @@
+package thriftclient_test
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/reddit/baseplate.go/mqsend"
+	"github.com/reddit/baseplate.go/runtimebp"
+	"github.com/reddit/baseplate.go/thriftclient"
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+const (
+	method = "foo"
+)
+
+func initClients() (*thriftclient.MockClient, *thriftclient.RecordedClient, *thriftclient.MonitoredClient) {
+	mock := &thriftclient.MockClient{}
+	recorder := thriftclient.NewRecordedClient(mock)
+	client := &thriftclient.MonitoredClient{Client: recorder}
+	return mock, recorder, client
+}
+
+func initServerSpan(ctx context.Context) (context.Context, *tracing.Tracer) {
+	ip, _ := runtimebp.GetFirstIPv4()
+	tracer := &tracing.Tracer{
+		SampleRate: 1.0,
+		Endpoint: tracing.ZipkinEndpointInfo{
+			ServiceName: "test-service",
+			IPv4:        ip,
+		},
+	}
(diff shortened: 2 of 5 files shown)


pacejackson:
I meant to open this as a draft PR since I have not added an example or any tests yet, sorry about that. 

😅

thriftclient/testing.go
```diff
@@ -0,0 +1,103 @@
+package thriftclient
```
pacejackson: Should this actually go in a `thrifttesting` subpackage or something like that, similar to the `httptest` subpackage in `net/http`?  Or is that overkill?

thriftclient/client.go
```diff
@@ -0,0 +1,67 @@
...
+	client thrift.TClient
+}
+
+// NewMonitoredClient returns a pointer to a new MonitoredClient that wraps the
+// given thrift.TClient calls in thrift client spans.
+//
+// If c is nil, will return a nil client and an error rather than panicking.
+func NewMonitoredClient(c thrift.TClient) (*MonitoredClient, error) {
+	if c == nil {
+		return nil, ErrNilTClient
```
pacejackson: Would it actually be better to just panic here rather than adding an error to the return type?  That doesn't seem like an uncommon approach to this sort of thing.

fishy: just let it panic should be fine.

thriftclient/client.go
```diff
@@ -29,13 +22,11 @@ type MonitoredClient struct {
 
 // NewMonitoredClient returns a pointer to a new MonitoredClient that wraps the
 // given thrift.TClient calls in thrift client spans.
-//
-// If c is nil, will return a nil client and an error rather than panicking.
-func NewMonitoredClient(c thrift.TClient) (*MonitoredClient, error) {
+func NewMonitoredClient(c thrift.TClient) *MonitoredClient {
 	if c == nil {
-		return nil, ErrNilTClient
+		panic("NewMonitoredClient received a nil client")
```
fishy: Oh I thought you meant let it panic naturally :badpokerface:

pacejackson: 😅 

I prefer to let it panic sooner if it's something that will panic if you ever actually use it, I'm fine to remove this too.

thriftclient/client.go
```diff
@@ -0,0 +1,67 @@
...
+		span = tracing.GetServerSpan(ctx)
+	}
+	var child *tracing.Span
+	var err error
+	if span != nil {
+		ctx, child = span.ChildAndThriftContext(ctx, method)
+	}
+	defer func() {
+		if child != nil {
+			if err = child.Stop(ctx, err); err != nil {
```
fishy: If you want to access the returning `err` from within the deferred function, the best practice is to use named return (you don't even need to do `err = c.client.Call(...); return err`, just `return c.client.Call(...)` would do). But in that case inside the deferred function you need to be careful not to overwrite it, which will also overwrite the returned err (unless you intend to do so), so here you need to use a different name for the error returned by `child.Stop`.

thriftclient/testing.go
fishy: Since this package is small enough this is fine.

thriftclient/client.go
fishy: I would lean against calling `panic` explicitly from the library code, but I don't feel strongly about it.

pacejackson: I also don't feel that strongly about it, I'll remove it.

thriftclient/client.go
```diff
@@ -0,0 +1,64 @@
...
+		span = tracing.GetServerSpan(ctx)
+	}
+	var child *tracing.Span
+	var err error
+	if span != nil {
+		ctx, child = span.ChildAndThriftContext(ctx, method)
+	}
+	defer func() {
+		if child != nil {
+			if stopErr := child.Stop(ctx, err); stopErr != nil {
```
fishy: What I mean is that at function signature you use named return (`(err error)` instead of just `error`), then you no longer need to define it in L45, and you can combine L56 and L57 into `return c.client.Call(ctx, method, args, result)`, and it still works in this defer'd block.

thriftclient/client.go
```diff
@@ -0,0 +1,62 @@
...
+		factory.GetProtocol(transport),
+		factory.GetProtocol(transport),
+	))
+}
+
+// NewMonitoredClient returns a pointer to a new MonitoredClient that wraps the
+// given thrift.TClient calls in thrift client spans.
+func NewMonitoredClient(c thrift.TClient) *MonitoredClient {
+	return &MonitoredClient{client: c}
+}
```
fishy: 🔕 Do you think it's better to just make `client` an exported field, so we no longer need this constructor, and we can just rename the `NewMonitoredClientFromFactory` constructor to this shorter name? I think in most of the cases we want people to use that constructor anyways.

pacejackson: That's probably fine?  I _like_ keeping thing un-exported but I probably over-favor that compared to how most go code seems to be.

thriftclient/client_test.go
```diff
@@ -0,0 +1,172 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
+	defer cancel()
+	return mmq.Receive(ctx)
+}
+
+func TestMonitoredClientNilSpan(t *testing.T) {
+	t.Parallel()
+
+	_, recorder, client := initClients()
+	client.Call(context.Background(), method, nil, nil)
```
konradreiche: Should we check the error here?

thriftclient/client_test.go
```diff
@@ -0,0 +1,172 @@
...
+		t.Run(
+			c.name,
+			func(t *testing.T) {
+				t.Parallel()
+
+				mock, recorder, client := initClients()
+				mock.AddMockCall(method, c.call)
+
+				ctx, tracer := c.initSpan(context.Background())
+				client.Call(ctx, method, nil, nil)
```
konradreiche: Should we check the error here?

thriftclient/example_client_test.go
```diff
@@ -0,0 +1,29 @@
...
+	)
+	// Create an actual service client
+	client := baseplate.NewBaseplateServiceClient(
+		// Use MonitoredClient to wrap a standard thrift client
+		thriftclient.NewMonitoredClientFromFactory(transport, factory),
+	)
+	// Create a context with a server span
+	ctx, _ := tracing.CreateServerSpanForContext(context.Background(), tracer, "test")
+	// Calls should be automatically wrapped using client spans
+	client.IsHealthy(ctx)
```
konradreiche: I've never written Example code, would it make sense to check the error here to display that the return value should be checked?

thriftclient/testing.go
```diff
@@ -0,0 +1,103 @@
...
+
+// NewRecordedClient returns a pointer to a new RecordedClient that wraps the
+// provided client.
+func NewRecordedClient(c thrift.TClient) *RecordedClient {
+	return &RecordedClient{client: c}
+}
+
+// Calls returns a copy of all of recorded Calls.
+//
+// Calls is not thread-safe and may panic if used accross threads if the number
```
konradreiche: s/accross/across

thriftclient/testing.go
```diff
@@ -0,0 +1,103 @@
...
+// that MockCall when that method is Call-ed.  If no MockCall is registered to a
+// method, Call will simply return nil.
+//
+// MockClient is provided to help with unit testing and should not be used in
+// production code.
+type MockClient struct {
+	methods map[string]MockCall
+}
+
+// AddMockCall registeres the given MockCall to the given method.
```
konradreiche: s/registeres/registers ?

thriftclient/client_test.go
pacejackson: It's not really necessary _as this code is written_ but it's probably a good idea since this is relying on the Mock behavior not changing so that would provide a nice guard to that.

PR #89, opened 2020-02-18
thriftpool: Add package example

fishy:


thriftpool/doc_test.go
```diff
@@ -0,0 +1,97 @@
...
+		TTLClient: thriftpool.NewTTLClient(trans, clientTTL),
+	}, nil
+}
+
+func callEndpoint(ctx context.Context, pool thriftpool.Pool) (*MyEndpointResponse, error) {
+	c, err := pool.Get()
+	if err != nil {
+		return nil, err
+	}
+	defer pool.Release(c)
```
konradreiche: Similar question to the previous pull request, should we check errors in example code for clarification?

thriftpool/doc_test.go
```diff
@@ -0,0 +1,97 @@
...
+		maxConnections,
+		func() (thriftpool.Client, error) {
+			return newClient(remoteAddr)
+		},
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	callEndpoint(context.Background(), pool)
```
konradreiche: Same here

thriftpool/doc_test.go
fishy: Added.

thriftpool/doc_test.go
fishy: This is just an example calling the endpoint. In real code we certainly would also handle the response and errors :)

thriftpool/doc_test.go
```diff
@@ -0,0 +1,125 @@
...
+	trans, err := thrift.NewTSocketTimeout(addr, socketTimeout)
+	if err != nil {
+		return nil, err
+	}
+	err = trans.Open()
+	if err != nil {
+		return nil, err
+	}
+	protoFactory := thrift.NewTHeaderProtocolFactory()
+	client := NewMyServiceClientFactory(trans, protoFactory)
```
pacejackson: Would you consider using:

```suggestion
	client := NewMyServiceClient(thriftclient.NewMonitoredClientFromFactory(trans, protoFactory))
```

so we set other devs up to use that by default?

fishy: That change was not merged in yet 😅

pacejackson: :badpokerface:

_yet_

https://github.com/reddit/baseplate.go/commit/be09f33afa2edea73bbec51bb096648796594450

fishy: :shakes-fist:

fishy: And the line is soooooo long!

PR #90, opened 2020-02-19
tracing: Implement opentracing interface

fishy:
Make our Tracer to implement opentracing.Tracer, and make our Span to
implement opentracing.Span and opentracing.SpanContext, so that our
tracing implementation can be considered as an opentracing
implementation and users can start migrating to opentracing APIs.

Currently some of the interfaces are implemented with no-ops, but all
functions we care about are implemented.

The only breaking change is Span.SetTag, which now returns self for
chaining (this is part of opentracing.Span interface).


tracing/start_options.go
```diff
@@ -0,0 +1,54 @@
...
+type StartSpanOptions struct {
+	Type          SpanType
+	ComponentName string
+}
+
+// StartSpanOption defines additional options for baseplate spans.
+type StartSpanOption interface {
+	opentracing.StartSpanOption
+
+	ApplyBP(*StartSpanOptions)
```
konradreiche: A bit confused about the semantic here. Why is it saying ApplyBP? BP stands for Baseplate here?

tracing/span.go
```diff
@@ -334,3 +343,104 @@ func (s *Span) preStop(err error) {
...
+// In this implementation it's a no-op.
+func (s *Span) LogFields(fields ...log.Field) {
+	// TODO
+}
+
+// LogKV implements opentracing.Span.
+//
+// In this implementation it's a no-op.
+func (s *Span) LogKV(alternatingKeyValues ...interface{}) {
+	// TODO
```
konradreiche: Should this log something? How do we keep track of those TODOs?

fishy:
> This seems reasonable modulo the two questions. Would we want to move the internal span code to use the `opentracing` interfaces or is it just presented as an option?

I think we do want to move all the internal code and examples to opentracing interfaces whenever possible (some of the context related ones might not be able to be migrated over), but I'll do that in a followup PR.

tracing/start_options.go
fishy: Yes BP stands for Baseplate. The [original API](https://pkg.go.dev/github.com/opentracing/opentracing-go?tab=doc#StartSpanOption) is limited to whatever [`opentracing.StartSpanOptions`](https://pkg.go.dev/github.com/opentracing/opentracing-go?tab=doc#StartSpanOptions) holds and that's not extensible from here, so we have to extend that in this way.

tracing/span.go
fishy: I'm not 100% sure how those APIs are used in real world yet. In the cases we care we likely would never need to use these 2 APIs, but leaving TODO here makes it more obvious that these are things that can be improved in the future when someone find some extra time :)

fishy: Removed the TODOs.

fishy:
:haircut:

konradreiche:
:fish:

PR #91, opened 2020-02-20
Rename 'thriftpool' package to 'clientpool'

pacejackson:
There's not really anything thrift specific in here, so we can pull it out into a generic package that can be used for thrift as well as other "client"s

### DIFF
--- a/clientpool/channel.go
+++ b/clientpool/channel.go
@@ -1,4 +1,4 @@
-package thriftpool
+package clientpool
 
 import (
 	"sync/atomic"
--- a/clientpool/channel.go
+++ b/clientpool/channel.go
@@ -15,7 +15,7 @@ type channelPool struct {
 // Make sure channelPool implements Pool interface.
 var _ Pool = (*channelPool)(nil)
 
-// NewChannelPool creates a new thrift client pool implemented via channel.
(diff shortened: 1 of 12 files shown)


clientpool/channel.go
```diff
@@ -1,31 +1,31 @@
...
+	opener     ClientOpener
+	numActive  int32
+	minClients int
+	maxClients int
 }
 
 // Make sure channelPool implements Pool interface.
 var _ Pool = (*channelPool)(nil)
 
 // NewChannelPool creates a new thrift client pool implemented via channel.
```
fishy: we need to remove the mention of thrift here :)

clientpool/errors.go
```diff
@@ -11,16 +11,16 @@ var ErrExhausted = errors.New("client pool exhausted")
 // ConfigError is the error type returned when trying to open a new thrift
```
fishy: here as well.

clientpool/interface.go
```diff
@@ -19,7 +20,7 @@ type Client interface {
 // ClientOpener defines a generator for clients.
 type ClientOpener func() (Client, error)
 
-// Pool defines the thrift client pool interface.
+// Pool defines thrift client pool interface.
```
fishy: here as well :)

PR #92, opened 2020-02-21
Add thriftclient.ClientPool interface

pacejackson:
This adds a thrift focused wrapper over a `clientpool.Pool` to the `thriftclient` package.

The clients that it creates implement both `clientpool.Client` and `thrift.TClient` so they can be used as the underlying `TClient` in a thrift service client and can be managed by a `clientpool.Pool`.


thriftclient/client_pool.go
```diff
@@ -0,0 +1,211 @@
...
+	// "${ServiceSlug}.pool-active-connections".
+	//
+	// The reporting goroutine is cancelled when the global metrics client
+	// context is Done.
+	ReportPoolStats bool
+
+	// PoolGaugeInterval indicates how often we should update the active
+	// connections gauge when collecting pool stats.
+	PoolGaugeInterval time.Duration
+}
```
fishy: I think we also need the ability to specify metrics tags for the metrics managed by the pool. This is useful when we have sharded requests, and have a pool for each shard. (Technically it's still possible to just have different slugs for each shard, but I think labels are preferred over that).

I think the easiest way to do that is to add a `MetricsLabels []string` field into config struct, and pass that into the metrics' `With` functions. nil/empty slice will still work so for most users who don't use labels the default value still works.

thriftclient/client_pool.go
```diff
@@ -0,0 +1,211 @@
...
+	// needed.
+	GetClient() (Client, error)
+
+	// ReleaseClient returns the given client to the pool.
+	ReleaseClient(Client)
+}
+
+// AddressGenerator defines a function that returns the address of a thrift
+// service.
+type AddressGenerator func() string
```
fishy: I think this generator needs to return `(string, error)` instead. For `SingleAddressGenerator` it can be that it always return nil error.

thriftclient/client_pool.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// NewCustomClientPool creates a ClientPool that uses a custom AddressGenerator
+// and ClientFactory.
+//
+// Most services will want to just use NewTTLClientPool, this has been provided
+// to support services that have non-standard and/or legacy needs.
+func NewCustomClientPool(cfg ClientPoolConfig, genAddr AddressGenerator, clientFact ClientFactory, protoFact thrift.TProtocolFactory) (ClientPool, error) {
+	if cfg.Addr != "" {
+		log.Warnf(
```
fishy: nit: When using zap logger interface, prefer *w version over *f version for performance reasons. when using *w version you just put a static string as the first arg, then key/value pairs follow for additional info.

thriftclient/client_pool.go
```diff
@@ -0,0 +1,211 @@
...
+	//		ImageUploadService -> image-upload
+	ServiceSlug string
+
+	// Addr is the address of a thrift service.  Addr must be in the format
+	// ${host}:${port}
+	Addr string
+
+	// MinConnections is the minimum number of thrift connections the client
+	// pool should maintain.
+	MinConnections int
```
fishy: Also the comment here is inaccurate because our implementation :badpokerface:

The only place min connection is used is on pool creation that we pre-populate that many clients and put into the pool. It's never used again. There's no trivial atomical way to guarantee that we always maintain the min connections in the pool and it's not really worth it to do it the non-trivial way.

pacejackson: Maybe I should call it `NumInitialConnection`?

fishy: works for me. we can also rename it in the `clientpool` interface :)

pacejackson: Will do

thriftclient/client_pool.go
```diff
@@ -40,6 +40,10 @@ type ClientPoolConfig struct {
 	// SocketTimeout is the timeout on the underling thrift.TSocket.
 	SocketTimeout time.Duration
 
+	// Any labels that should be applied to metrics logged by the ClientPool.
+	// This includes the optional pool stats.
+	MetricsLabels []string
```
fishy: Worth mention that it should be in even number in key value pairs, for example: `key1, value1, key2, value2`. The underlying implementation will actually panic if it's odd number.

fishy: Alternative is that we accept a map here and convert that into string slice in constructors. (nil map will still work)

pacejackson: I have a slightly different approach that I just pushed.

PR #93, opened 2020-02-21
metricsbp: add MetricsLabels type 

pacejackson:
This is a `map[string]string` typedef that allows you to define StatsD labels in a simple way and provides a function to convert itself to the format that the StatsD client expects.

### DIFF
--- a/thriftclient/client_pool.go
+++ b/thriftclient/client_pool.go
@@ -42,7 +42,7 @@ type ClientPoolConfig struct {
 
 	// Any labels that should be applied to metrics logged by the ClientPool.
 	// This includes the optional pool stats.
-	MetricsLabels map[string]string
+	MetricsLabels metricsbp.MetricsLabels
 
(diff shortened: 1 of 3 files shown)


thriftclient/client_pool.go
```diff
@@ -61,16 +61,6 @@ type ClientPoolConfig struct {
 	PoolGaugeInterval time.Duration
 }
 
-// getMetricLabels returns c.MetricsLabels as a slice of strings, appropriate
-// for passing to a metrics client.
-func (c ClientPoolConfig) getMetricLabels() []string {
-	labels := make([]string, len(c.MetricsLabels)*2)
```
pacejackson: We had a bug here, this was initializing a list of length `len(c.MetricsLabels)*2` full of empty strings and then appending the actual labels to that list.

fishy: 🤦‍♂

PR #94, opened 2020-02-21
tracing: Migrate to opentracing API

fishy:
This is a breaking change.

Remove some of the tracing API in favor of opentracing API, in details
(this is not an exhaustive list):

- GlobalTracer is no longer exported, use opentracing.GlobalTracer()
  instead.
- No longer provide *WithTracer APIs, always use the global tracer.
- Remove CreateServerSpan* APIs, use opentracing.StartSpan with
  SpanTypeServer as a SpanTypeOption instead.
- Remove Tracer.NewTrace API, use opentracing.StartSpan without parent
  instead.
- Remove Span.Create*Child API, use opentracing.StartSpan instead.

Added APIs:

- tracing.AsSpan() to convert opentracing.Span into *tracing.Span.
- tracing.TestWrapper() for special case of tracing related testing.
- tracing.CloseTracer() to stop the global tracer from reporting.

### DIFF
--- a/redisbp/hooks.go
+++ b/redisbp/hooks.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/go-redis/redis/v7"
+	opentracing "github.com/opentracing/opentracing-go"
 
 	"github.com/reddit/baseplate.go/batcherror"
 	"github.com/reddit/baseplate.go/tracing"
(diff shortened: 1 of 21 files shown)


redisbp/hooks.go
```diff
@@ -4,10 +4,11 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/go-redis/redis/v7"
-
 	"github.com/reddit/baseplate.go/batcherror"
 	"github.com/reddit/baseplate.go/tracing"
+
+	"github.com/go-redis/redis/v7"
+	opentracing "github.com/opentracing/opentracing-go"
```
pacejackson: Shouldn’t these be before the baseplate imports? Or am I mixing that up with the Python conventions ?

fishy: I think them as this way: stdlibs (1st party), same module (2nd party), and 3rd parties. that's why I _slightly_ prefer this order. but as long as the latter 2 are separated (except in examples, in examples we are in user's shoes so everything is 3rd party to them) I'm ok with it. if python programmers find it annoying I can switch it back :sweat_smile:

pacejackson: I think the other way is probably preferable? goimports even has a flag to support it (maybe we should use it?)

https://github.com/golang/tools/blob/e1da425f72fd3793b579f4e74d908ba96eb16c8a/cmd/goimports/goimports.go#L61

fishy: that option is likely from pre-modules days (as it accepts a list of prefixes), but fine, I'll switch back.

fishy: Done.

redisbp/hooks.go
```diff
@@ -49,27 +50,23 @@ func (h SpanHook) AfterProcessPipeline(ctx context.Context, cmds []redis.Cmder)
...
+	span := opentracing.SpanFromContext(ctx)
+	if span == nil {
+		return nil
 	}
-	return nil
+	defer span.FinishWithOptions(tracing.FinishOptions{
+		Ctx: ctx,
+		Err: err,
+	}.Convert())
+	return err
```
fishy: Note that I don't think returning the stop error to redis is the correct behavior. That wouldn't be some error redis care about. cc @pacejackson

pacejackson: Agree

fishy:
:haircut: That should be all the things I need to fix 😅

PR #96, opened 2020-02-22
timebp: Add TimestampSecondF type

fishy:
Also switch experiments.ExperimentConfig's timestamp json types to
TimestampSecondF.

### DIFF
--- a/experiments/experiments.go
+++ b/experiments/experiments.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/reddit/baseplate.go/filewatcher"
 	"github.com/reddit/baseplate.go/log"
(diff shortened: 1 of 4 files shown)


experiments/experiments.go
```diff
@@ -121,11 +122,11 @@ type ExperimentConfig struct {
 	// StartTimestamp is a float of seconds since the epoch of date and time
```
fishy: Note that the comment here mentioned that the number in json should actually be float instead of int. That's why I defined the json type as float type instead of int type.

konradreiche: Oh good catch :)

PR #97, opened 2020-02-23
thriftclient: ClientPool API changes

fishy:
- Add DefaultPoolGaugeInterval to avoid panic
- Add a few clientpool.Pool API into thriftclient.ClientPool as
  otherwise the callers don't have access to them

### DIFF
--- a/thriftclient/client_pool.go
+++ b/thriftclient/client_pool.go
@@ -3,6 +3,7 @@ package thriftclient
 import (
 	"context"
 	"errors"
+	"io"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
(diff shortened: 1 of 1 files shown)


thriftclient/client_pool.go
```diff
@@ -74,6 +82,12 @@ type Client interface {
 // ClientPool defines an object that can be used to manage a pool of
 // Client objects.
 type ClientPool interface {
+	// Passthrough APIs from clientpool.Pool:
+	io.Closer
+	NumActiveClients() int32
+	NumAllocated() int32
+	IsExhausted() bool
```
fishy: Two alternatives to this:

1. Just embed `clientpool.Pool`
2. Provide a `GetPool() clientpool.Pool` API

Both will expose the low level `Get`/`Release` API to the user, which is likely OK?

pacejackson: I think either 1 or the current approach, but I don’t think I feel super strong about any of the options.

fishy: I actually slightly like 2 more than 1. With 2 it's less likely for user to accidentally use the wrong pair (`GetClient` then `Release` while then are supposed to use `ReleaseClient`, as in 2 they need to use `GetPool().Release()` instead).

fishy: On second thought, only `Close` and `IsExhausted` are really useful to the users. Removed the two stats one but added them both to gauges.

PR #104, opened 2020-02-27
Add method to attach the Request Context header to HTTP headers

pacejackson:
This is so we can pass the RequestContext header between HTTP services.

### DIFF
--- a/edgecontext/req_context.go
+++ b/edgecontext/req_context.go
@@ -13,6 +13,8 @@ import (
 	"github.com/apache/thrift/lib/go/thrift"
 )
 
+type Signer func(s string) (string, error)
+
 // An EdgeRequestContext contains context info about an edge request.
 type EdgeRequestContext struct {
 	// header and raw should always be set during initialization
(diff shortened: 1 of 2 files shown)


pacejackson:
Hmm, that is a good question... Should there be 2 methods, one that just attaches it unsigned and one that adds a signature?  Or maybe we should just give read access to the raw header and let clients do what they want with it?

edgecontext/req_context.go
```diff
@@ -47,6 +52,23 @@ func (e *EdgeRequestContext) AttachToContext(ctx context.Context) context.Contex
...
 }
 
+// AttachHTTPHeader attaches the header to the http Headers
+func (e *EdgeRequestContext) AttachHTTPHeader(h http.Header) {
+	h.Add(httpbp.EdgeContextHeader, e.header)
+}
+
+// AttachSignedHTTPHeader attaches the header to the http Headers along with
+// a signature header.
+func (e *EdgeRequestContext) AttachSignedHTTPHeader(h http.Header, signer Signer) error {
```
fishy: I think instead of asking the users to implement `Signer`, we should actually take care of that? We already provide [`httpbp.TrustHeaderSignature`](https://pkg.go.dev/github.com/reddit/baseplate.go/httpbp?tab=doc#TrustHeaderSignature), and this really should just be the reverse of that (or at least, we should provide an implementation of `Signer` in `httpbp` package).

PR #105, opened 2020-03-02
Add DeviceID to edge context

pacejackson:


edgecontext/edgecontext_test.go
```diff
@@ -16,14 +16,15 @@ import (
 // copied from https://github.com/reddit/baseplate.py/blob/db9c1d7cddb1cb242546349e821cad0b0cbd6fce/tests/__init__.py#L56-L59
```
fishy: This line should also be updated?

pacejackson: Ya, I'll take a min to copy anything else over and update the tests if needed.

PR #110, opened 2020-03-06
experiments: Fix tests

fishy:
When checking expected error message, we should also fail the test when
err is nil.


fishy:
> How did you find that?

I was trying to deliberately fail some tests to make sure bazel test works, remembered these 2 tests needed to be changed with the error message prefix change, and noticed the bug 😅

PR #111, opened 2020-03-06
thriftclient: Add an error counter for monitored client

fishy:


thriftclient/monitored_tclient.go
```diff
@@ -46,6 +47,9 @@ func (c MonitoredClient) Call(ctx context.Context, method string, args, result t
 			Ctx: ctx,
 			Err: err,
 		}.Convert())
+		if err != nil {
+			metricsbp.M.Counter(method + ".error").Add(1)
```
pacejackson: I think the span hoof from metricsbp should already cover this?

fishy: Ah right.

PR #115, opened 2020-03-08
metricsbp: Use pointer to provide better default sample rates

fishy:
Instead of converting 0 to 1 for sample rate configs, use *float64
to provide default value. Also provide Float64Ptr, similar to the
function in thrift package, to help user set the config value.

This is a breaking change.


konradreiche:
What is the advantage of using a float pointer here?

fishy:
> What is the advantage of using a float pointer here?

We want the default to be 1 (100%), so that users don't need to specify that when they want 100% of metrics sampled. But without pointer the zero value of float64 is 0, which is a valid value here (means 0% sampling, e.g. no metrics will be sampled), so we used to do a special handling to convert 0 to 1 and ask users to use <0 if they really want 0.

Using pointer can avoid that situation.

PR #116, opened 2020-03-09
Add experiment logging by adding an Expose method

### LINKED ISSUE #3: Experiments
Details TBD. [baseplate.py reference implementation](https://github.com/reddit/baseplate.py/blob/master/baseplate/lib/experiments/__init__.py), [baseplate spec](https://pages.github.snooguts.net/reddit/baseplate.spec/component-apis/experiments)

konradreiche:
I have removed the unnecessary interface and added experiment logging. Experiment logging as part of bucketing seems deprecating so it is enough to include an `Expose` method that services can make use of if they want. The corresponding Python code can be found [here](https://github.com/reddit/baseplate.py/blob/master/baseplate/lib/experiments/__init__.py#L233-L266).

This resolves #3

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -35,7 +35,7 @@ http_archive(
     sha256 = "d8c45ee70ec39a57e7a05e5027c32b1576cc7f16d9dd37135b0eddde45cf1b10",
 )
 
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
 
 gazelle_dependencies()
 
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -43,10 +43,3 @@ load("//:external.bzl", "go_dependencies")
 
 # gazelle:repository_macro external.bzl%go_dependencies
 go_dependencies()
-
-go_repository(
-    name = "com_github_gofrs_uuid",
-    importpath = "github.com/gofrs/uuid",
-    sum = "h1:y12jRkkFxsd7GpqdSZ+/KCs/fJbqpEXSGd4+jfEaewE=",
-    version = "v3.2.0+incompatible",
-)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -35,7 +35,7 @@ http_archive(
     sha256 = "d8c45ee70ec39a57e7a05e5027c32b1576cc7f16d9dd37135b0eddde45cf1b10",
 )
 
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
 
 gazelle_dependencies()
 
@@ -43,3 +43,10 @@ load("//:external.bzl", "go_dependencies")
 
 # gazelle:repository_macro external.bzl%go_dependencies
 go_dependencies()
+
+go_repository(
+    name = "com_github_gofrs_uuid",
+    importpath = "github.com/gofrs/uuid",
+    sum = "h1:y12jRkkFxsd7GpqdSZ+/KCs/fJbqpEXSGd4+jfEaewE=",
+    version = "v3.2.0+incompatible",
+)
--- a/experiments/experiments.go
+++ b/experiments/experiments.go
@@ -78,7 +78,9 @@ func (e *Experiments) Expose(ctx context.Context, experimentName string, event E
 		return UnknownExperimentError(experimentName)
 	}
 	event.Experiment = experiment
-	event.EventType = "EXPOSE"
+	if event.EventType == "" {
+		event.EventType = "EXPOSE"
+	}
 	return e.eventLogger.Log(ctx, event)
 }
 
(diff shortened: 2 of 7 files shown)


WORKSPACE
```diff
@@ -30,11 +30,18 @@ http_archive(
...
 
 # gazelle:repository_macro external.bzl%go_dependencies
 go_dependencies()
+
+go_repository(
+    name = "com_github_gofrs_uuid",
+    importpath = "github.com/gofrs/uuid",
+    sum = "h1:y12jRkkFxsd7GpqdSZ+/KCs/fJbqpEXSGd4+jfEaewE=",
+    version = "v3.2.0+incompatible",
+)
```
fishy: This should be in `external.bzl` instead. If you remove this section and rerun `./scripts/bazel_cleanup.sh` it should auto do that for you (or you can just manually move this part into `external.bzl`).

experiments/experiments.go
```diff
@@ -334,3 +334,24 @@ func isSimpleExperiment(experimentType string) bool {
...
+type ExperimentEvent struct {
+	ID              uuid.UUID
+	CorrelationID   uuid.UUID
+	DeviceID        uuid.UUID
+	Experiment      *ExperimentConfig
+	VariantName     string
+	UserID          string
+	LoggedIn        bool
+	ClientTimestamp int64
+	CookieCreatedAt int64
```
fishy: For `ClientTimestamp` and `CookieCreatedAt`, can we use a more concrete type from `timebp` to indicate the unit? Or better yet, just use `time.Time` and convert that into required type of int in `Expose` function?

experiments/experiments.go
```diff
@@ -334,3 +334,24 @@ func isSimpleExperiment(experimentType string) bool {
...
+	AppName         string
+	EventType       string
+}
```
fishy: Also I assume on the API level, it's up to the caller to fill in `ExperimentEvent` struct and call `Expose` function, is that correct? Can we add some comments here to indicate which fields are required and which are optional?

experiments/experiments.go
konradreiche: Dang! I always miss those, thanks! :)

WORKSPACE
konradreiche: The script didn't fix it. I moved the `go_repository` part around.

experiments/experiments.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gofrs/uuid"
```
fishy: nit: Also please add a blank line after this line so it's in a different group with baseplate.go packages.

experiments/experiments.go
```diff
@@ -334,3 +336,41 @@ func isSimpleExperiment(experimentType string) bool {
...
+	// UserID identifies the user who is being exposed to the experimental
+	// treatment.
+	UserID string
+	// LoggedIn indiciates whether the user is authenticated.
+	LoggedIn bool
+	// ClientTimestamp is the time when the experiment has been applied. If
+	// this is not provided the logger should generate a timestamp (optional).
+	ClientTimestamp time.Time
+	// CookieCreatedAt is the timestamp when the cookie for the user has been
+	// generated.
```
fishy: should this one be optional as well?

experiments/experiments.go
```diff
@@ -334,3 +337,41 @@ func isSimpleExperiment(experimentType string) bool {
...
 }
+
+// ExperimentEvent is the playload used by Expose to log whether a user has
+// been exposed to an experimental treatment.
+type ExperimentEvent struct {
+	// ID uniquely identifies the experiment event. If you pass in uuid.Nil the
+	// logger handling this event should generate a UUID v4 (optional).
+	ID uuid.UUID
+	// CorrelationID are used to track events across different services.
+	CorrelationID uuid.UUID
```
fishy: 🔕 Is it that every field didn't mention "(optional)" is required? It doesn't sound right that ID is optional but CorrelationID is required.

experiments/experiments.go
```diff
@@ -334,3 +337,41 @@ func isSimpleExperiment(experimentType string) bool {
...
+	AppName string
+	// EventType is the type of the experiment event. Will be set to EXPOSE
+	// (optional).
+	EventType string
+}
+
+// Logger provides an interface for experiment events to be logged.
+type Logger interface {
+	Log(ctx context.Context, event ExperimentEvent) error
+}
```
fishy: 🔕 For an interface with only one function and there's no additional functions in the foreseeable future, I would slightly prefer to just define a function type instead of `interface`, as that way it's much easier for the caller to implement it inline. But I don't feel strongly about that. [`log.Wrapper`](https://pkg.go.dev/github.com/reddit/baseplate.go/log?tab=doc#Wrapper) is an example of that.

experiments/experiments.go
```diff
@@ -334,3 +337,41 @@ func isSimpleExperiment(experimentType string) bool {
...
+	CookieCreatedAt time.Time
+	// AppName if any specifies the application (optional).
+	AppName string
+	// EventType is the type of the experiment event. Will be set to EXPOSE
+	// (optional).
+	EventType string
+}
+
+// Logger provides an interface for experiment events to be logged.
+type Logger interface {
```
fishy, line 375: Can we rename this to `EventLogger`? I feel the name `Logger` is too generic and could cause confusions.

konradreiche:
Odd, I did a rebase

konradreiche:
Ah, there we go.

PR #123, opened 2020-03-26
edgecontext: Move away from singleton

fishy:
This fixes https://github.com/reddit/baseplate.go/issues/122.

Consider the common use case is initialize an implementation then
immediately pass that into the middleware and never worry about it
again, we no longer provide APIs for singleton usage.

This is a breaking change.

### DIFF
--- a/edgecontext/edgecontext.go
+++ b/edgecontext/edgecontext.go
@@ -17,20 +17,25 @@ import (
 	"github.com/apache/thrift/lib/go/thrift"
 )
 
+// LoIDPrefix is the prefix for all LoIDs.
+const LoIDPrefix = "t2_"
+
 // ErrLoIDWrongPrefix is an error could be returned by New() when passed in LoID
 // does not have the correct prefix.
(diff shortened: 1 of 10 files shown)


edgecontext/edgecontext.go
```diff
@@ -17,20 +17,25 @@ import (
...
 // ErrNoHeader is an error could be returned by FromThriftContext() when passed
 // in context does not have Edge-Request header set.
 var ErrNoHeader = errors.New("edgecontext: no Edge-Request header found")
 
-// global vars that will be initialized in Init function.
-var (
+// An Impl is an initialized edge context implementation.
+//
+// Please call Init function to initialize it.
+type Impl struct {
```
konradreiche: Could this maybe be just the `EdgeContext` struct?

fishy: We already have [`EdgeRequestContext`](https://pkg.go.dev/github.com/reddit/baseplate.go/edgecontext?tab=doc#EdgeRequestContext), so that name will be more confusing.

PR #126, opened 2020-04-01
Add Thrift middleware framework

pacejackson:


thriftbp/middlewares.go
```diff
@@ -0,0 +1,81 @@
...
+
+// BaseplateProcessor is a TProcessor that can be thriftbp.Wrap-ed and
+// thriftbp.Merge-d.
+//
+// The TProcessors generated by the Apache Thrift compiler fufill this
+// interface, but not all of them are a part of any interface within Apache
+// Thrift.
+type BaseplateProcessor interface {
+	// Process fufills the TProcessor interface.
+	Process(ctx context.Context, in, out thrift.TProtocol) (bool, thrift.TException)
```
fishy, line 19: 🔕 For this one you can just embed `thrift.TProcessor`. That also makes it more obvious that anything implementing this interface also implements `thrift.TProcessor`.

thriftbp/middlewares.go
```diff
@@ -0,0 +1,81 @@
...
+	// TProcessorFunction.
+	AddToProcessorMap(string, thrift.TProcessorFunction)
+}
+
+// Middleware is a function that can be passed to Wrap to wrap
+// the TProcessorFunctions for that TProcessor.
+//
+// Middlewares are passed in the name of the function as set in the processor
+// map of the TProcessor and a logger that can be used by the Middleware.
+type Middleware func(name string, logger log.Wrapper, next thrift.TProcessorFunction) thrift.TProcessorFunction
```
fishy, line 37: I don't feel that adding `logger` to `Middleware` API is a good idea. The one use case we have is that edgecontext needs it, but that can be passed in `edgecontext.InjectThriftEdgeContext` instead. For users implementing their own middleware on the service repo, they would most likely just use the zap logger directly in their middleware implementation. Putting `logger` into `Middleware` API and pass it through all the middlewares also means that all middlewares must share the same logging level, which is likely not true (in real world scenario some middleware will likely need more severity logging than others).

pacejackson: Great point that also conveniently makes the API neater 😅

fishy:
Oh and add doc comments to MockTProcessor: https://cloud.drone.io/reddit/baseplate.go/600/3/2

PR #128, opened 2020-04-02
Add function to take care of the boilerplate of creating a thrift server

pacejackson:


thriftbp/server.go
```diff
@@ -0,0 +1,41 @@
...
+// and protocol to serve the given BaseplateProcessor which is wrapped with the
+// given Middlewares.
+func NewThriftServer(cfg ServerConfig, processor BaseplateProcessor, middlewares ...Middleware) (*thrift.TSimpleServer, error) {
+	transport, err := thrift.NewTServerSocketTimeout(cfg.Addr, cfg.Timeout)
+	if err != nil {
+		return nil, err
+	}
+
+	server := thrift.NewTSimpleServer4(
+		Wrap(processor, cfg.Logger, middlewares...),
```
fishy, line 33: I didn't notice this in the other PR, but logger is unused in `Wrap` implementation so it should be removed from `Wrap`'s function signature: https://github.com/reddit/baseplate.go/blob/e21d93469801525d38a8848ee56d74ef83db3af5/thriftbp/middlewares.go#L71-L80

thriftbp/server.go
```diff
@@ -0,0 +1,41 @@
...
+
+// ServerConfig is the arg struct for NewThriftServer.
+type ServerConfig struct {
+	// The endpoint address of your thrift service
+	Addr string
+
+	// The timeout for the underlying thrift.TServerSocket transport.
+	Timeout time.Duration
+
+	// A log wrapper that is passed to the "Wrap" call and server.SetLogger
```
fishy, line 19: After you removed logger from Wrap function, this comment should be changed to clarify that it's only used by `thrift.SimpleServer`.

thriftbp/server.go
```diff
@@ -16,7 +16,7 @@ type ServerConfig struct {
 	// The timeout for the underlying thrift.TServerSocket transport.
 	Timeout time.Duration
 
-	// A log wrapper that is passed to the "Wrap" call and server.SetLogger
+	// A log wrapper that is used the TSimpleServer.
```
fishy, line 19: nit: "used **by** the thrift.TSimpleServer"

PR #130, opened 2020-04-04
thriftbp: Change the return type of Merge to BaseplateProcessor

fishy:


thriftbp/merger.go
```diff
@@ -1,17 +1,15 @@
...
-import (
-	"github.com/apache/thrift/lib/go/thrift"
-)
-
-// Merge merges together the specified processors.
+// Merge merges together multiple processors into the first one.
 //
-// It's useful if the server need to support two separated thrift files.
+// It's useful when the server need to support more than one separated thrift
+// files.
```
pacejackson, line 6: I think this should be “file”?

fishy: `more than one` so it should be plural?

pacejackson: What about “more than one thrift file”?

fishy: oh so you are thinking about "more than (one thrift file)" and I'm thinking about "(more than one) thrift files".

fishy: done.

PR #132, opened 2020-04-05
httpbp: refactor to remove go-kit

pacejackson:
The intention here is to replace go-kit as the HTTP/middleware
framework for HTTP services with a micro-framework that makes
it easier to work with the framework already provided by
"net/http" and add Middleware support.

Ultimately, this ends up being a similar amount and kind of code
to what we were previously using to smooth over using go-kit and
should simplify things for most services.


httpbp/edgecontext.go
```diff
@@ -0,0 +1,37 @@
...
+func HeaderTrustHandlerSigner(handler TrustHeaderSignature, duration time.Duration) Signer {
+	return func(s string) (string, error) {
+		h := EdgeContextHeaders{EdgeRequest: s}
+		return handler.SignEdgeContextHeader(h, duration)
+	}
+}
+
+// AttachHTTPHeader attaches the header to the http Headers
+func AttachHTTPHeader(h http.Header, e edgecontext.EdgeRequestContext) {
+	h.Add(EdgeContextHeader, e.Header())
```
fishy, line 24: This should be `Set` instead of `Add`. Same as the signature one.

httpbp/handler.go
```diff
@@ -0,0 +1,74 @@
+package httpbp
+
+import (
+	"context"
+	"html/template"
+	"net/http"
+)
+
+type HandlerFunc func(context.Context, *http.Request, Response) (interface{}, error)
```
fishy, line 9: A few design discussions regarding this type:

1. Currently it could return an error, is that supposed to be `HTTPError`, or is that not supposed to be `HTTPError` and the handler will wrap whatever error it returns into an `HTTPError`?
2. If it's `HTTPError`, since `Response` is a subset of `HTTPError`, should the implementation still try to write to Response? Or is it that in case of error, don't touch Response, return an `HTTPError` instead?

httpbp/handler.go
```diff
@@ -0,0 +1,74 @@
...
+
+type Middleware func(HandlerFunc) HandlerFunc
+
+func abort(w http.ResponseWriter) {
+	code := http.StatusInternalServerError
+	http.Error(w, http.StatusText(code), code)
+}
+
+func writeResponse(w http.ResponseWriter, resp Response, body interface{}) error {
+	if resp.StatusCode() != 0 && resp.StatusCode() != http.StatusOK {
```
fishy, line 19: nit: `&& resp.StatusCode() != http.StatusOK` part is unnecessary (it will be implicit, but doing that explicitly doesn't hurt anyways).

httpbp/response.go
```diff
@@ -1,313 +1,173 @@
...
-	)
+// ContentWriterFactory is the interface used by the handler to create new
+// ContentWriters when serving requests.
+type ContentWriterFactory func() ContentWriter
+
+// Response is an HTTP response that can be returned by a baseplate HTTP handler.
+//
+// Response is responsible for setting the values that are independant of the
+// body as well as the ContentWriter to write the response body.
+type Response interface {
```
fishy, line 45: There's no way to set `ContentWriter` in `HTTPHandler`, does that mean for a microservice all handlers must share the same `ContentWriter` coming from the factory? e.g. you can't mix json and html?

httpbp/response.go
```diff
@@ -1,313 +1,173 @@
...
-// e.Cause error.
-func (e HTTPError) Unwrap() error {
-	return e.Cause
+// TextContentWriter is a ContentWriterFactory that returns a ContentWriter
+// for writing plain text.
+//
+// When using a text content writer, your handler should return a string,
+// WriteResponse will return an error rather than writing the content if you
+// pass a value of a different type.
+func TextContentWriter() ContentWriter {
```
fishy, line 105: I think this should be `RawContentWriter` with the ability to specify content type instead, and the body should be `io.Reader` instead of `string`. Or maybe keep this one and add a `RawContentWriter`.

httpbp/handler.go
pacejackson: 1. It can be either, if you return an HTTPError, it uses that to formulate the error response, if it's another error, it just returns a generic `500` error (it doesn't actually use HTTPError for this, although it could, it just uses `http.Error`).
2. As it currently stands, HTTPError is separate from Response, so if you want to return an HTTP error you would return `nil, NewJSONError(...)` and wouldn't mess around with `Response`.  You _can_ do that but it won't do anything, the error will determine what is returned.

httpbp/handler.go
pacejackson: I was thinking that I would just skip the call since `http.StatusOK` is the default and let Go do it's thing.

httpbp/middlewares.go
```diff
@@ -0,0 +1,50 @@
...
+func InjectEdgeRequestContext(truster HeaderTrustHandler, impl *edgecontext.Impl, logger log.Wrapper) Middleware {
+	return func(next HandlerFunc) HandlerFunc {
+		return func(ctx context.Context, req *http.Request, resp Response) (interface{}, error) {
+			var header string
+			if truster.TrustEdgeContext(req) {
+				header = req.Header.Get(EdgeContextHeader)
+			}
+
+			factory := edgecontext.FromRawHeader(header)
+			ctx = edgecontext.InitializeEdgeContext(ctx, impl, logger, factory)
```
fishy, line 46: I think these 2 lines should be inside the if block above, otherwise we'll be spamming the log with errors... Or is the intention to use the error logging inside `InitializeEdgeContext` to log the situation that the http header doesn't work?

httpbp/error_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestHTTPError(t *testing.T) {
+	t.Parallel()
+
+	t.Run(
+		"default values",
+		func(t *testing.T) {
+			t.Parallel()
+
+			defaultCode := http.StatusInternalServerError
+			err := httpbp.HTTPError{}
```
fishy, line 21: `httpbp.HTTPError` is an interface so this won't compile I think?

Issue #102, opened 2020-02-25
Top level doc comment


fishy:
Add a top level `doc.go` and `doc_test.go` (at the root directory) to give users a high level overview of how to use the packages, how to manage their main function, etc.

pacejackson:
Can we also add a "tutorial" like what we have in baseplate.py? 

https://baseplate.readthedocs.io/en/stable/tutorial/index.html

The idea is giving devs something to step through to help them understand the parts that go into creating a basic service and let them peek under the hood a little.

PR #133, opened 2020-04-06
Pull out typical middleware into a package

fizx:


thriftsimple/thriftsimple.go
```diff
@@ -0,0 +1,160 @@
...
+	if err := raven.SetSampleRate(float32(cfg.Sentry.SampleRate)); err != nil {
+		return err
+	}
+	raven.SetEnvironment(cfg.Sentry.Environment)
+	return nil
+}
+
+// NewBaseplateThriftServer returns a server that includes the default middleware.
+func NewBaseplateThriftServer(cfg BaseplateServerConfig, processor thriftbp.BaseplateProcessor, additionalMiddlewares ...thriftbp.Middleware) (*thrift.TSimpleServer, error) {
+	ctx, _ := context.WithCancel(context.Background())
```
fishy, line 124: Don't discard cancel returned by `context.WithCancel`. If you don't want to cancel it you can just use `context.Background()` directly, but in this case ctx really should be just passed in.

thriftsimple/thriftsimple.go
```diff
@@ -0,0 +1,160 @@
...
+
+	log.Debugf("%#v", cfg)
+	return nil
+}
+func initLogger(debug bool) (log.Level, log.Wrapper) {
+	var logLevel log.Level
+	if debug {
+		logLevel = log.DebugLevel
+	} else {
+		logLevel = log.WarnLevel
```
fishy, line 76: There's also `InfoLevel` between `DebugLevel` and `WarnLevel`, so I'm not sure why when debug is false we skipped that one and go directly to `WarnLevel`.

thriftsimple/thriftsimple.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+type BaseplateServerConfig struct {
+	Addr string
+
+	Timeout time.Duration
+
+	Debug bool
+
+	Metrics struct {
```
pacejackson, line 27: Should these be separate `config` structs that we define in the packages and embed in this struct?

One potential benefit of that is we can change those `itit*` functions to exported `Init*` functions that use those specific configs rather than the top-level `BaseplateServerConfig`.

thriftsimple/thriftsimple.go
```diff
@@ -0,0 +1,160 @@
...
+
+	Tracing struct {
+		Namespace     string
+		Endpoint      string
+		RecordTimeout time.Duration `yaml:"recordTimeout"`
+		SampleRate    float64
+	}
+}
+
+func (c BaseplateServerConfig) AsConfig() BaseplateServerConfig {
```
pacejackson, line 50: What is this for?

thriftsimple/thriftsimple.go
```diff
@@ -0,0 +1,160 @@
...
+
+	metricsbp.M = metricsbp.NewStatsd(ctx, metricsbp.StatsdConfig{
+		Prefix:   cfg.Metrics.Namespace,
+		Address:  cfg.Metrics.Endpoint,
+		LogLevel: logLevel,
+	})
+
+	secretsStore, err := initSecrets(ctx, cfg, logger)
+	if err != nil {
+		log.Fatal(err)
```
fishy, line 136: This function has a return value of `error`, so I don't see why it would ever call `log.Fatal`. We should either remove that return value, or just return the errors.

thriftsimple/thriftsimple.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	defer secretsStore.Close()
```
pacejackson, line 138: This doesn't really work, it's going to `Close` as soon as you return this.

pacejackson: We also generally need to use the secrets store in other places.

thriftsimple/thriftsimple.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	innerCfg := thriftbp.ServerConfig{
+		Addr:    cfg.Addr,
+		Timeout: cfg.Timeout,
+		Logger:  logger,
+	}
+
+	middlewares := []thriftbp.Middleware{
+		edgecontext.InjectThriftEdgeContext(ecImpl, logger),
+		tracing.InjectThriftServerSpan,
```
pacejackson, line 155: I think we actually want to swap the order of these 🤔

fishy:
The `init*` functions really should be living in the top level `doc_test.go` as an example. That is what https://github.com/reddit/baseplate.go/issues/102 was for.

thriftsimple/thriftsimple.go
fizx: That sounds quite nice, and a wider-ranging refactor.

thriftsimple/thriftsimple.go
fizx: If you embed BaseplateServerConfig into a larger struct, then this let's you cast back into a BaseplateServerConfig

thriftsimple/thriftsimple.go
fizx: Yeah probably worth fixing!

thriftsimple/thriftsimple.go
fizx: Ah, bad copy-pasta.

thriftsimple/thriftsimple.go
fizx: I just copied from the template.

fizx:
So the feedback I'm getting is that the direction is fine, but please encapsulate and package things out more?

Mostly just trying to confirm the shape of things before investing too much time.

fishy:
I think we mostly only want a much simpler version, which is basically just [`thriftbp.NewThriftServer`](https://pkg.go.dev/github.com/reddit/baseplate.go/thriftbp?tab=doc#NewThriftServer) but with the two middlewares hardcoded (but still allow additional middlewares)?

thriftsimple/thriftsimple.go
pacejackson: Ya, I think the template is probably wrong too 😅

fizx:
> I think we mostly only want a much simpler version, which is basically just thriftbp.NewThriftServer but with the two middlewares hardcoded (but still allow additional middlewares)?

I think it's pretty hard to do this without doing the config extraction/setup for the middleware like @pacejackson suggests:

https://github.com/reddit/baseplate.go/pull/133/files#r404459394

fishy:
> > I think we mostly only want a much simpler version, which is basically just thriftbp.NewThriftServer but with the two middlewares hardcoded (but still allow additional middlewares)?
> 
> I think it's pretty hard to do this without doing the config extraction/setup for the middleware like @pacejackson suggests:
> 
> https://github.com/reddit/baseplate.go/pull/133/files#r404459394

You only need either an edge context implementation or the secret store passed in, plus a logger for edge context middleware, no?

konradreiche:
:running_man: will review after changes are applied

fishy:
Regarding package name, how about we put this under the root baseplate.go directory and call it `baseplate`?

thriftsimple/thriftsimple.go
fizx: Let's do this in a follow-up patch.  That should effectively be an internal refactor, as the inits are private now.

fizx:
Should have addressed the feedback.

> Some general feedback, I also don't think we really want to use NewBaseplateThriftServer as-is, most of this stuff is applicable to non-thrift servers as well or needs to be done in the context of the main method.

This could happen in a follow-up PR.  This PR introduces relatively little surface-area (one new public method and a simple interface) that could easily be deprecated/replaced.  I think there's a pretty big win here for defining and consolidating anything at all, and people with stronger opinions can use this as a stepping stone to the version they actually want.

baseplate/server.go
```diff
@@ -0,0 +1,70 @@
+package baseplate
+
+import (
+	"errors"
+	"io/ioutil"
+	"time"
+
+	"github.com/reddit/baseplate.go/log"
+	"gopkg.in/yaml.v2"
```
fishy, line 9: Style nit: Please group imports into 3 groups (separated by blank lines): stdlibs, third-party imports (anything that's not stdlib and don't have `github.com/reddit/baseplate.go` prefix), and packages from this module (anything with `github.com/reddit/baseplate.go` prefix).

baseplate/server.go
```diff
@@ -0,0 +1,70 @@
...
+		Endpoint      string
+		RecordTimeout time.Duration `yaml:"recordTimeout"`
+		SampleRate    float64
+	}
+}
+
+// Server is the primary interface for baseplate servers.
+type Server interface {
+	Config() ServerConfig
+	Impl() interface{}
```
fishy, line 48: I don't see how `Impl()` could be useful in real world scenarios. Do you have any example usage in mind?

baseplate/server.go
```diff
@@ -0,0 +1,70 @@
...
+	Sentry struct {
+		DSN         string
+		Environment string
+		SampleRate  float64
+	}
+
+	Tracing struct {
+		Namespace     string
+		Endpoint      string
+		RecordTimeout time.Duration `yaml:"recordTimeout"`
```
fishy, line 40: Why is this the only one in the whole `ServerConfig` with yaml annotation?

baseplate/server.go
```diff
@@ -0,0 +1,70 @@
...
+	Timeout time.Duration
+
+	Log struct {
+		Level log.Level
+	}
+
+	Metrics struct {
+		Namespace string
+		Endpoint  string
+	}
```
fishy, line 25: 🔕 From prod experience we likely would need histogram sample rate for metrics config.

baseplate/server.go
```diff
@@ -0,0 +1,70 @@
...
+// Server is the primary interface for baseplate servers.
+type Server interface {
+	Config() ServerConfig
+	Impl() interface{}
+	Serve() error
+	Stop() error
+}
+
+// ParseServerConfig will populate a ServerConfig from a YAML file.
+func ParseServerConfig(path string, cfg interface{}) error {
```
fishy, line 54: Why do we use `cfg interface{}` instead of just returning `ServerConfig`? Especially the doc comment already explicitly said that it "will populate a ServerConfig".

baseplate/thrift.go
```diff
@@ -0,0 +1,147 @@
...
+	"time"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/getsentry/raven-go"
+	"github.com/reddit/baseplate.go/edgecontext"
+	"github.com/reddit/baseplate.go/log"
+	"github.com/reddit/baseplate.go/metricsbp"
+	"github.com/reddit/baseplate.go/secrets"
+	"github.com/reddit/baseplate.go/thriftbp"
+	"github.com/reddit/baseplate.go/tracing"
```
fishy, line 15: Same import groups comment.

baseplate/thrift.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (bts *baseplateThriftServer) Serve() error {
+	return bts.thriftServer.Serve()
+}
+
+func (bts *baseplateThriftServer) Stop() error {
+	var err error = nil
+	for _, c := range bts.beforeStop {
+		c.Close()
```
fishy, line 41: We have [`batcherror` package](https://pkg.go.dev/github.com/reddit/baseplate.go/batcherror), please add the errors returned by closers to the batch.

baseplate/thrift.go
```diff
@@ -0,0 +1,147 @@
...
+func (bts *baseplateThriftServer) Impl() interface{} {
+	return bts.thriftServer
+}
+
+func (bts *baseplateThriftServer) Serve() error {
+	return bts.thriftServer.Serve()
+}
+
+func (bts *baseplateThriftServer) Stop() error {
+	var err error = nil
```
fishy, line 39: style nit: `var err error` is the same as and preferred over `var err error = nil`. Also if you use batcherror then this becomes irrelevant.

baseplate/thrift.go
```diff
@@ -0,0 +1,147 @@
...
+	if err := raven.SetSampleRate(float32(cfg.Sentry.SampleRate)); err != nil {
+		return err
+	}
+	raven.SetEnvironment(cfg.Sentry.Environment)
+	return nil
+}
+
+// NewBaseplateThriftServer returns a server that includes the default middleware.
+func NewBaseplateThriftServer(ctx context.Context, cfg ServerConfig, processor thriftbp.BaseplateProcessor, additionalMiddlewares ...thriftbp.Middleware) (Server, error) {
+	beforeStop := make([]io.Closer, 0)
```
fishy, line 100: No need to add `, 0`, and also I don't really see we would ever have any closer that's needed before stop. Even if that use case raises in the future, we can add it then instead of now.

baseplate/thrift.go
```diff
@@ -0,0 +1,147 @@
...
+		Prefix:   cfg.Metrics.Namespace,
+		Address:  cfg.Metrics.Endpoint,
+		LogLevel: cfg.Log.Level,
+	})
+
+	secretsStore, err := initSecrets(ctx, cfg, logger)
+	if err != nil {
+		return nil, err
+	}
+	afterStop = append(afterStop, secretsStore)
```
fishy, line 114: I think we need to add a defer that if returned err is non-nil, call all the already appended closers.

secrets/store.go
```diff
@@ -26,6 +26,7 @@ func nopSecretHandlerFunc(sec *Secrets) {}
 // memory so there's little performance impact to doing so and you will be sure
 // to always have the current version in the face of key rotation etc.
 type Store struct {
+	io.Closer
```
fishy, line 29: style nit: Please add a blank line after this line (always add blank lines after embedded structs/interfaces).

But then there's no reason to embed `io.Closer` to `Store` struct. As long as you implemented that interface it's satisfied, that's not controlled by embedding.

secrets/store.go
```diff
@@ -80,8 +81,9 @@ func (s *Store) getSecrets() *Secrets {
 // but can still access the secrets as they were before Close is called.
 //
 // It's OK to call Close multiple times. Calls after the first one are no-ops.
-func (s *Store) Close() {
+func (s *Store) Close() error {
```
fishy, line 84: Please add a line into the doc comment stating that this closer always returns nil error.

baseplate/server.go
fizx: lift-n-shift

baseplate/server.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Server is the primary interface for baseplate servers.
+type Server interface {
+	Config() ServerConfig
+	Serve() error
+	Stop() error
+}
+
+// ParseServerConfig will populate a ServerConfig from a YAML file.
+func ParseServerConfig(path string, cfg *ServerConfig) error {
```
fishy, line 54: Is there a reason we want `cfg` as an arg instead of return value?

baseplate/thrift.go
```diff
@@ -0,0 +1,153 @@
...
+
+	metricsbp.M = metricsbp.NewStatsd(ctx, metricsbp.StatsdConfig{
+		Prefix:   cfg.Metrics.Namespace,
+		Address:  cfg.Metrics.Endpoint,
+		LogLevel: cfg.Log.Level,
+	})
+
+	secretsStore, err := initSecrets(ctx, cfg, logger)
+	if err != nil {
+		cleanup(afterStop)
```
fishy, line 115: Use defer after initializing `afterStop` instead, if you make it named return you can access return value in deferred function:

```go
func NewBaseplateThriftServer(...) (_ Server, err error) {
  afterStop := make([]io.Closer)
  defer func() {
    if err != nil {
      cleanup(afterStop)
    }
  }()
  ...
}
```

baseplate/thrift.go
```diff
@@ -0,0 +1,153 @@
...
+func cleanup(closers []io.Closer) {
+	for _, c := range closers {
+		if err := c.Close(); err != nil {
+			log.Error("error cleaning up thrift server initialization:", err)
+		}
+	}
+}
+
+// NewBaseplateThriftServer returns a server that includes the default middleware.
+func NewBaseplateThriftServer(ctx context.Context, cfg ServerConfig, processor thriftbp.BaseplateProcessor, additionalMiddlewares ...thriftbp.Middleware) (Server, error) {
```
fishy, line 103: Since the package is already named `baseplate`, this could be just called `NewThriftServer`.

secrets/store.go
```diff
@@ -80,8 +80,10 @@ func (s *Store) getSecrets() *Secrets {
 // but can still access the secrets as they were before Close is called.
 //
 // It's OK to call Close multiple times. Calls after the first one are no-ops.
-func (s *Store) Close() {
+// Close doesn't return non-nil errors, but implements io.Closer
```
fishy, line 83: Add a blank line before this line (so it's a new paragraph), and end it with period (`.`).

baseplate/thrift.go
fizx: Yeah, I'm a little reluctant b/c there is already a NewThriftServer, and I'd not like them confused.  NewConfiguredThriftServer?

baseplate/server.go
fizx: Probably not anymore.  It made more sense when it was possible to embed the config in other larger configs in a previous iteration..

baseplate/thrift.go
fishy: From user's perspective, it's `baseplate.NewThriftServer` vs. `thriftbp.NewThriftServer`, so it's not very confusing? Also that reminds me the other one probably should be renamed into `thriftbp.NewServer` :)

baseplate/thrift.go
```diff
@@ -0,0 +1,153 @@
...
+
+func cleanup(closers []io.Closer) {
+	for _, c := range closers {
+		if err := c.Close(); err != nil {
+			log.Error("error cleaning up thrift server initialization:", err)
+		}
+	}
+}
+
+// NewBaseplateThriftServer returns a server that includes the default middleware.
```
fishy, line 102: Also we should give a high level walkthrough in this function's doc comment about that this function does, otherwise people would always need to check the source code to make sure.

Most things are directly from the config so it's probably obviously enough, the part that's implied are the middlewares added, so that should be documented properly.

baseplate/thrift.go
fizx: nifty

baseplate/thrift.go
```diff
@@ -0,0 +1,153 @@
...
+func NewBaseplateThriftServer(ctx context.Context, cfg ServerConfig, processor thriftbp.BaseplateProcessor, additionalMiddlewares ...thriftbp.Middleware) (Server, error) {
+	afterStop := make([]io.Closer)
```
fishy, line 107: Oh and for slices, prefer `var afterStop []io.Closer` over `afterStop := make([]io.Closer)`.

But here you actually know what's the expected size of the slice, so you could use `afterStop := make([]io.Closer, 0, N)` instead to make it faster.

But then again this is a run once function, so that optimization is not useful and more error-prone than just do `var afterStop []io.Closer` (because N might change in the future and people might forget to change N when that happens).

baseplate/thrift.go
fizx: 🔕 I do agree with the general principle, and I'd rather type an extra word in this particular case, because it's the user's *first* thing they're doing in the package, and I'd rather not have noobs get confused when the wrong one gets autocompleted.

baseplate/thrift.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/reddit/baseplate.go/metricsbp"
+	"github.com/reddit/baseplate.go/secrets"
+	"github.com/reddit/baseplate.go/thriftbp"
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+type baseplateThriftServer struct {
+	thriftServer *thrift.TSimpleServer
+	config       ServerConfig
+	afterStop    []io.Closer
```
fishy, line 21: 🔕 now without `beforeStop` maybe this should just be called `closers`? I don't feel strongly, though.

baseplate/thrift.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, err
+	}
+	srv = &baseplateThriftServer{
+		logger:       logger,
+		config:       cfg,
+		afterStop:    afterStop,
+		thriftServer: ts,
+	}
+
+	return srv, nil
```
fishy, line 157: optional: Since you used named return, this can be just `return` instead of `return srv, nil`.

baseplate/thrift.go
```diff
@@ -0,0 +1,158 @@
...
+	secretsStore, err := initSecrets(ctx, cfg, logger)
+	if err != nil {
+		return nil, err
+	}
+	afterStop = append(afterStop, secretsStore)
+
+	ecImpl := edgecontext.Init(edgecontext.Config{Store: secretsStore})
+	if err = initTracing(cfg, logger, metricsbp.M); err != nil {
+		return nil, err
+	}
```
fishy, line 130: tracing also have closer to be called upon cleanup. You can either try cast `opentracing.GlobalTracer()` into `*tracing.Tracer` and add that, or just write a wrapper locally.

baseplate/thrift.go
fishy: Actually looking at the `append` calls in constructor I do think the name `closers` makes more sense.

baseplate/thrift.go
```diff
@@ -0,0 +1,158 @@
...
+		if err := c.Close(); err != nil {
+			log.Error("error cleaning up thrift server initialization:", err)
+		}
+	}
+}
+
+// NewBaseplateThriftServer returns a server that includes the default middleware needed
+// in order to be a standard Reddit service.
+//
+// At the moment, this includes secrets management, metrics, edge contexts, and spans/tracing.
```
fishy, line 105: Semantically "this" should be referring to "middlewares" mentioned in the above paragraph, but then it's actually not listing the middlewares.

baseplate/thrift.go
```diff
@@ -0,0 +1,158 @@
...
+func cleanup(closers []io.Closer) {
+	for _, c := range closers {
+		if err := c.Close(); err != nil {
+			log.Error("error cleaning up thrift server initialization:", err)
+		}
+	}
+}
+
+// NewBaseplateThriftServer returns a server that includes the default middleware needed
+// in order to be a standard Reddit service.
```
fishy, line 103: Should we say "Baseplate service" instead of "Reddit service"?

baseplate/server.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// ServerConfig is a general purpose config for assembling a BaseplateServer
+type ServerConfig struct {
+	Addr string
+
+	Timeout time.Duration
+
+	Log struct {
+		Level log.Level
```
fishy, line 20: How does this type work in yaml parsing?

baseplate/thrift.go
fishy: Feel free to rename both. This is something `golint` would complain about (when you have `package.PackageSomething` defined), see also https://github.com/reddit/baseplate.go/pull/137.

baseplate/thrift.go
```diff
@@ -0,0 +1,158 @@
...
+	}()
+	logger := initLogger(cfg)
+
+	metricsbp.M = metricsbp.NewStatsd(ctx, metricsbp.StatsdConfig{
+		Prefix:   cfg.Metrics.Namespace,
+		Address:  cfg.Metrics.Endpoint,
+		LogLevel: cfg.Log.Level,
+	})
+
+	secretsStore, err := initSecrets(ctx, cfg, logger)
```
fishy, line 121: unlike other things initialized in this function, `secretsStore` is something the service code would also need access to, so we likely will need to provide something like `GetSecrets` in the `Server` interface.

baseplate/server.go
```diff
@@ -0,0 +1,71 @@
...
+		RecordTimeout time.Duration
+		SampleRate    float64
+	}
+}
+
+// Server is the primary interface for baseplate servers.
+type Server interface {
+	Config() ServerConfig
+	Serve() error
+	Stop() error
```
fishy, line 50: Also, why not make this `Close() error` so it's compatible with `io.Closer` as well? Since `Server` is an interface you can just embed `io.Closer` here.

baseplate/server.go
fizx: Level is a string

internal/closers.go
```diff
@@ -0,0 +1,17 @@
...
+import "io"
+
+type anonymousCloser struct {
+	io.Closer
+	closeFunc func() error
+}
+
+// NewAnonymousCloser is a convenience function to transform anonymous functions into
+// io.Closers
+func NewAnonymousCloser(f func() error) io.Closer {
```
fishy, line 12: putting this in an internal package is really an overkill. you can just put it under root with an unexported type.

also in the current tracing implementation even if user calls `InitGlobalTracer` multiple times with different configs, we always update the same one and it's always the same one registered inside opentracing (so `opentracing.GetGlobalTracer` would always return the same global tracer), so you can just cast that into `*tracing.Tracer` and add it into closers directly.

And that really should be called `closers` instead of `afterStop`.

fishy:
If you rebase on the latest master, `metricsbp.M` is an `io.Closer` now as well.

baseplate/server.go
```diff
@@ -0,0 +1,75 @@
...
+	if path == "" {
+		return nil, errors.New("no config path given")
+	}
+
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = yaml.Unmarshal(data, cfg); err != nil {
```
fishy, line 69: This is minor since this is once on start up thing instead of once per request thing, but yaml package does have [`NewDecoder`](https://pkg.go.dev/gopkg.in/yaml.v2?tab=doc#NewDecoder) to take an `io.Reader` directly, which means there's no need to read the file fully into memory before passing that into yaml for parsing.

baseplate/thrift.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (bts *baseplateThriftServer) Serve() error {
+	return bts.thriftServer.Serve()
+}
+
+func (bts *baseplateThriftServer) Close() error {
+	var err error
+	bts.thriftServer.Stop()
+	for _, c := range bts.closers {
+		c.Close()
```
fishy, line 42: Please don't throw away close errors.

baseplate/thrift.go
fishy: In case you forgot this one.

baseplate/thrift.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return err
+}
+func initLogger(cfg ServerConfig) log.Wrapper {
```
fishy, line 46: nit: Add a blank line between functions (I'm surprised that gofmt didn't complain about this).

baseplate/thrift.go
```diff
@@ -36,13 +37,14 @@ func (bts *baseplateThriftServer) Serve() error {
 }
 
 func (bts *baseplateThriftServer) Close() error {
-	var err error
-	bts.thriftServer.Stop()
+	errors := &batcherror.BatchError{}
```
fishy, line 40: nit: BatchError supports both reference and pointer, so `var errors batcherror.BatchError` is sufficient (but this still works)

baseplate/thrift.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+}
+
+// NewBaseplateThriftServer returns a server that initializes and includes the default
+// middleware and cross-cutting concerns needed in order to be a standard Baseplate service.
+//
+// At the moment, this includes secrets management, metrics, edge contexts
+// (edgecontext.InjectThriftEdgeContext), and spans/tracing (tracing.InjectThriftServerSpan).
+func NewBaseplateThriftServer(ctx context.Context, cfg ServerConfig, processor thriftbp.BaseplateProcessor, additionalMiddlewares ...thriftbp.Middleware) (srv Server, err error) {
```
fishy, line 110: The other rename is happening in https://github.com/reddit/baseplate.go/pull/144, so this could be renamed into `NewThriftServer`.

fizx:
💇‍♂️

baseplate/server.go
fishy: I still prefer this to be fixed. Although we care less about performance here, I consider baseplate.go codebase an example for engineers less familiar with go to follow, and this is a bad example.

baseplate/server.go
fishy: If some service want to set non-1 sample rate for histograms, do they have to make a baseplate.go change? If that's the case we need to support that.

baseplate/thrift.go
```diff
@@ -0,0 +1,164 @@
...
+
+import (
+	"context"
+	"io"
+	"time"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/getsentry/raven-go"
+	"github.com/opentracing/opentracing-go"
+	"github.com/reddit/baseplate.go/batcherror"
```
fishy, line 11: Add a blank line between Line 10 and 11.

baseplate/thrift.go
```diff
@@ -0,0 +1,164 @@
...
+		return err
+	}
+	raven.SetEnvironment(cfg.Sentry.Environment)
+	return nil
+}
+
+func cleanup(closers []io.Closer) {
+	for _, c := range closers {
+		if err := c.Close(); err != nil {
+			log.Error("error cleaning up thrift server initialization:", err)
```
fishy, line 100: Prefer `log.Errorw("msg", "err", err)` over this.

baseplate/thrift.go
```diff
@@ -0,0 +1,164 @@
...
+
+	secretsStore, err := initSecrets(ctx, cfg, logger)
+	if err != nil {
+		return nil, err
+	}
+	closers = append(closers, secretsStore)
+
+	ecImpl := edgecontext.Init(edgecontext.Config{Store: secretsStore})
+	if err = initTracing(cfg, logger, metricsbp.M); err != nil {
+		closers = append(closers, opentracing.GlobalTracer().(*tracing.Tracer))
```
fishy, line 134: why do we append this to closers when we failed to initialize it?

baseplate/thrift.go
```diff
@@ -0,0 +1,164 @@
...
+	return errors.Compile()
+}
+
+func initLogger(cfg ServerConfig) log.Wrapper {
+	if cfg.Log.Level == "" {
+		cfg.Log.Level = log.InfoLevel
+	}
+	level := cfg.Log.Level
+	log.InitLogger(level)
+	return log.ZapWrapper(level)
```
fishy, line 54: In most cases this shouldn't be the same level.

The level used in config is the minimal level to actually show in the logs, which in most cases should be info level. The log wrapper we used for other initialization are for logging errors happened inside those packages that cannot be communicated to the caller otherwise (for example, filewatcher has a background goroutine watching the file and try to load it and call parser when its content changed), so that in most cases should be error level.

fizx: It will be much less confusing if all components of the system log to the same level.

fishy: by that logic why do we even bother to provide different log levels?

fizx:
Thanks for the tips!

I don't think bufio.NewReader is more idiomatic than ioutil.ReadFile for reading config files.  Keeping it simple here usually wins, and I'm usually the Big-O advocate.

fishy:
> Thanks for the tips!
> 
> I don't think bufio.NewReader is more idiomatic than ioutil.ReadFile for reading config files. Keeping it simple here usually wins, and I'm usually the Big-O advocate.

I never said to use `bufio.NewReader`. `os.Open` will return you an `io.Reader`.

fizx:
> I never said to use bufio.NewReader. os.Open will return you an io.Reader.

I look at "os" package imports as a code smell.  If you import "os", your package is probably doing some more low-level things than it needs to.   Maybe that's Python bias though.  Don't really care enough to fight on this.

fishy:
You need to import os to open a file (unless you use that iouils.ReadFile).

The problem with ioutils.ReadFile is that it's a really bad habit. For example if someone accidentally put the wrong filename in the arg, pointing to some huge file, ioutil.ReadFile will try to read the huge file wholly into memory.

server.go
```diff
@@ -63,12 +63,13 @@ func ParseServerConfig(path string) (*ServerConfig, error) {
 		return nil, errors.New("no config path given")
 	}
 
-	data, err := ioutil.ReadFile(path)
+	reader, err := os.Open(path)
 	if err != nil {
 		return nil, err
 	}
```
fishy, line 69: you need to `defer data.Close()` after the error check.

server.go
```diff
@@ -64,6 +64,7 @@ func ParseServerConfig(path string) (*ServerConfig, error) {
...
 
 	reader, err := os.Open(path)
+	defer reader.Close()
```
fishy, line 67: It needs to be after error check, otherwise when error happens it will cause panic (because reader would be nil)

server.go
```diff
@@ -0,0 +1,79 @@
...
+
+	Log struct {
+		Level log.Level
+	}
+
+	Metrics struct {
+		Namespace           string
+		Endpoint            string
+		CounterSampleRate   float64
+		HistogramSampleRate float64
```
fishy, line 29: 🔕 how does yaml handle default values? if we change the 2 sample rates into `*float64`, will they be `nil` when that part is absent from the config file? if that's the case we would want to use pointers for the sample rates.

thrift.go
```diff
@@ -0,0 +1,168 @@
...
+	return errors.Compile()
+}
+
+func initLogger(cfg ServerConfig) log.Wrapper {
+	if cfg.Log.Level == "" {
+		cfg.Log.Level = log.InfoLevel
+	}
+	level := cfg.Log.Level
+	log.InitLogger(level)
+	return log.ZapWrapper(level)
```
fishy, line 55: this function should return nothing. The log level defined in the config file is supposed to mean the minimal level of log to be seen (so info level means that all debug level logs are suppressed, while error level means that they don't want to see info level log as well). for all the other initialization that needs an `log.Wrapper`, just use `log.ZapWrapper(log.ErrorLevel)`.

fishy: See also: https://github.com/reddit/baseplate.go/pull/153

fizx: As stated above, I strongly disagree with having the bootstrap process being at a different log level than what the user explicitly asked for.

fishy: The you need to provide two log level configs.

fishy: See also: https://pkg.go.dev/github.com/reddit/baseplate.go/log?tab=doc#ErrorWithSentryWrapper

fizx: This documentation is pretty unclear to me.  There's a little bit of the docs that is a direction of what to do, and not much how or why.

fizx: I guess let's leave it for now because it's already encoded in the public API, but if I understand correctly, I would strongly like ErrorWithSentryWrapper to be deprecated in favor of something like a simple WithSentryWrapper.

fizx: Forcing Error level logging isn't the right general purpose solution.  Bugs can happen in the bootstrap process and app authors should be able to adjust the log level.

server.go
```diff
@@ -0,0 +1,79 @@
+package baseplate
```
pacejackson, line 1: Having this be top-level rather than in an inner directory called `baseplate` seems like it'll probably get messy?  Also it means that the package name won't map up with what the importer presents which is kind of annoying (although not exactly _uncommon_).

thrift.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+func cleanup(closers []io.Closer) {
+	for _, c := range closers {
+		if err := c.Close(); err != nil {
+			log.Errorw("msg", "err", err)
+		}
+	}
+}
```
pacejackson, line 103: I feel lie _all_ of this should just be in the basic `baseplate` section, there's nothing `Thrift` about this and we'll want to use it on both HTTP and Thrift services.

thrift.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	closers = append(closers, opentracing.GlobalTracer().(*tracing.Tracer))
+	if err = initSentry(cfg); err != nil {
+		return nil, err
+	}
+	innerCfg := thriftbp.ServerConfig{
+		Addr:    cfg.Addr,
+		Timeout: cfg.Timeout,
+		Logger:  logger,
+	}
```
pacejackson, line 149: Same deal, I think all of this should be handled by a `Baseplate` struct possibly or some top level function, this feels like it's trying to do too much in one place.  I don't think we want to take the size of `main` methods down to _zero_.

thrift.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+	if err := raven.SetSampleRate(float32(cfg.Sentry.SampleRate)); err != nil {
+		return err
+	}
+	raven.SetEnvironment(cfg.Sentry.Environment)
+	return nil
+}
+
+func cleanup(closers []io.Closer) {
```
pacejackson, line 97: We should probably just expose this, maybe something like:

```
type Closers []io.Closer

func (closers Closers) Cleanup() {
	for _, c := range closers {
		if err := c.Close(); err != nil {
			log.Errorw("msg", "err", err)
		}
	}
}

File: thrift.go [hunk 52]
```diff
@@ -0,0 +1,169 @@
...
+	middlewares := []thriftbp.ProcessorMiddleware{
+		thriftbp.InjectServerSpan,
+		thriftbp.InjectEdgeContext(ecImpl),
+	}
+	middlewares = append(middlewares, additionalMiddlewares...)
+
+	ts, err := thriftbp.NewServer(innerCfg, processor, middlewares...)
+	if err != nil {
+		return nil, err
+	}
```
pacejackson, line 160: This should already be handled by `thritbp.NewBaseplateServer` or whatever we called it?

thrift.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/reddit/baseplate.go/batcherror"
+	"github.com/reddit/baseplate.go/edgecontext"
+	"github.com/reddit/baseplate.go/log"
+	"github.com/reddit/baseplate.go/metricsbp"
+	"github.com/reddit/baseplate.go/secrets"
+	"github.com/reddit/baseplate.go/thriftbp"
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+type baseplateThriftServer struct {
```
pacejackson, line 21: I don't think we need this, I think what we want is more like:

```
func main() {
    var closers baseplate.Closers
    defer closers.Cleanup()

    baseplate := SetupBaseplateStruct()
    closers = baseplate.start()
    server, err := thriftbp.NewBaseplateServer(baseplate, other_args...)
    if err {panic}
    return log.Fatal(server.serve())
}
```

where `main` is defined in each service.

server.go
fishy: by "messy" do you mean the dependencies it pulled in? that won't have a difference between being top level or subdirectory, because even at top level it won't be pulled into dependency tree unless the top level one is explicitly imported (e.g. `imporet "github.com/reddit/baseplate.go/log"` won't auto pull `github.com/reddit/baseplate.go`).

For the rename importing, yes but that's a minor issue. The alternative is to use `github.com/reddit/baseplate.go/baseplate`, which is arguably too redundant.

pacejackson: In this case, I mean `messy` in that now there's `go` code in the top level directory rather than having everything organized into sub-directories (with dedicated doc.go files).  For one file, maybe that's fine but if this top-level package grows, it's going to be messy to have code intermingled with the other non-code files in the top-level directory.

I think either side of the "import" argument is minor either way as you're likely only going to be importing this once in `main.go`, mostly just a point to call out.

fishy: Yea I understand your concern. I think to address that we just need to be more careful about what to pull into the top level package and what to not pull into it.

I think we should position the top-level package as a "all/most battery included starting point" for service implementations, the stuff they need to do in their main function. With that purpose in mind, do you think it's more acceptable?

Another reason I prefer top level vs. a "baseplate" subdirectory is for the doc rendering. Since this is a starting point, the doc rendered will be a good candidate to be put on to https://pkg.go.dev/mod/github.com/reddit/baseplate.go

pacejackson: I think that makes sense, especially from the documentation/onboarding perspective.

thrift.go
fizx: In the next PR!

thrift.go
fizx: Let's move it there in a followup PR.  The HTTP stuff has been fast moving, and all of this is private/moveable.

thrift.go
fizx: Generally, we do hope to take the main methods towards zero for the common cases and for the hello worlds.  Having a main that does a little too much, but that you can decompose 10% of the time for your advanced case, is ideal for gracefully onboarding new people to the system.

fizx:
Hello y'all!  We've now hit the dreaded centithread! (i.e. 100 comments)

Can we please agree that this is a step forward for the developer experience, but it's not perfect.  Ultimately, I've defined one function and one struct, which we can easily move away from, that wrapped code that everyone already thought was best practice. 

Rather than continuing to debate choices in this PR, I'm looking forward to many more PRs in the future that represent iterative improvements to this one.  This project will move faster and therefore get to higher quality sooner if we take a permissive and iterative approach.

thrift.go
fishy: I think we only have [`thriftbp.NewServer`](https://pkg.go.dev/github.com/reddit/baseplate.go/thriftbp?tab=doc#NewServer), which does not add default middlewares. We can certainly add a `thriftbp.NewBaseplateServer` to pull in them, though.

PR #134, opened 2020-04-07
Add middleware for thrift.TClient

pacejackson:
This is a breaking change that changes the API for how you
set up a Thrift client pool.  Specifically, we have removed
`MonitoredClient` entirely and replaced `NewMonitoredClientPool`
with `NewBaseplateClientPool` as the "golden path" option for
creating a Thrift connection pool in a Baseplate service.

The intention behind using Middleware rather than the previous 
approach is to trade off some having a potentially slightly higher
learning curve for having a system that is easier to extend within
individual services to make it easy to test new ideas and customize
your usage.


thriftclient/client_pool.go
```diff
@@ -99,92 +99,57 @@ type ClientPool interface {
...
 //
-// Most services will want to just use NewTTLClientPool, this has been provided
-// to support services that have non-standard and/or legacy needs.
-func NewCustomClientPool(cfg ClientPoolConfig, genAddr AddressGenerator, clientFact ClientFactory, protoFact thrift.TProtocolFactory) (ClientPool, error) {
+// Most services will want to just use NewBaseplateClientPool, this has been
+// provided to support services that have non-standard and/or legacy needs.
+func NewCustomClientPool(
+	cfg ClientPoolConfig,
+	genAddr AddressGenerator,
+	clientFact ClientFactory,
```
fishy, line 149: I kind of hate the abbreviation of factory -> fact but ¯\\\_(ツ)\_/¯

pacejackson: Would you say it's a... _fact_ that you hate it????

fishy: :get_out_frog:

pacejackson: For real though, I'll rename them :)

thriftclient/client_pool.go
```diff
@@ -217,9 +217,9 @@ func NewWrappedTClientFactory(base TClientFactory, middlewares ...Middleware) TC
 // convenience struct for passing around the different factories needed to
 // create a Client.
 type factories struct {
-	client  ClientFactory
-	tclient TClientFactory
-	proto   thrift.TProtocolFactory
+	Client   ClientFactory
+	TClient  TClientFactory
+	Protocol thrift.TProtocolFactory
```
fishy, line 222: This is an unexported struct so making the fields exported won't have any effect, but if your intention is to make it more obvious to tell fields and variables apart then this is fine :)

pacejackson: Yep! It’s just to improve readability

PR #139, opened 2020-04-10
metricsbp: Add Statsd.Close API

fishy:
Provide a flushing API for jobs that don't run indefinitely to make sure
that their metrics are emitted before exit.

### DIFF
--- a/metricsbp/statsd.go
+++ b/metricsbp/statsd.go
@@ -5,10 +5,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/reddit/baseplate.go/log"
-
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/influxstatsd"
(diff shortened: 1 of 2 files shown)


metricsbp/statsd.go
```diff
@@ -262,3 +266,37 @@ func (st *Statsd) fallback() *Statsd {
...
+// no more metrics will be send to the remote collector,
+// similar to the situation that this Statsd was initialized without Address
+// set,
+// but the difference is that calling Close() again will do the manual flush
+// again.
+//
+// After Close() is called,
+// Ctx() will always return an already canceled context.
+//
+// This function is useful for jobs that exits,
```
pacejackson, line 284:
```suggestion
// This function is useful for jobs that exit,
```

fishy: Done.

PR #140, opened 2020-04-11
Add code style guide

fishy:


Style.md
```diff
@@ -0,0 +1,144 @@
...
+As recommended by go modules version design,
+in the example of `"github.com/go-redis/redis/v7"`,
+`v7` is the major version number,
+so we consider `redis` as the actual last element of import path.
+As a result there's no need to rename import it.
+
+## One-per-line style
+
+When putting all args to a function signature/call to a single line makes the
+line too long, Use one-per-line style with closing parenthesis on the next line.
```
konradreiche, line 77: I think we need to clarify what too long here means.

fishy: Go's official stance is that there's no limit on the length of line. Personally I use tabsize of 2 and try not to exceed 80, but that's not a hard limit (e.g. if a line just barely cross 80 I might let it slip). I think the thing we want to enforce is that "if you are not gonna put everything in the same line, the do it one-per-line instead". I'm open to suggestions on how to word that better.

fishy:
>I assume this will evolve as we go forward.

Yes, by making PRs to this file :)

PR #141, opened 2020-04-13
thriftclient: Define BaseplateDefaultMiddlewares

fishy:
This makes service code using customized client pool easier to add all
the middlewares recommended by baseplate.

### DIFF
--- a/thriftclient/client_pool.go
+++ b/thriftclient/client_pool.go
@@ -125,10 +125,10 @@ type TClientFactory func(thrift.TTransport, thrift.TProtocolFactory) thrift.TCli
 //
 // A baseplate ClientPool:
 //		1. Uses a TTLClientPool with the given ttl.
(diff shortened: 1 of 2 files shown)


pacejackson:
We might want to make this a function rather than a mutable global?

fishy:
> We might want to make this a function rather than a mutable global?

I don't feel strongly. In general in go it's more like "if you try to mutate a global variable in another package, you need to know what you are doing and you are at your own risk".

fishy:
For example, if they _really wanted to_, they can mutate that list to add their own middlewares early in their main function, then use the `NewBaseplateClientPool` for several different client pools so they don't need to add those middlewares to those `NewBaseplateClientPool` calls one by one.

fishy:
💇 @pacejackson: I changed it to a function.

PR #142, opened 2020-04-14
runtimebp: Add HandleShutdown API

fishy:


fishy:
Note that I wrote a test but signal interception actually doesn't work in go tests so I removed it :(

runtimebp/signal.go
```diff
@@ -0,0 +1,41 @@
...
+	"os"
+	"os/signal"
+)
+
+// ShutdownHandler is the callback type used in HandleSignals.
+type ShutdownHandler func(signal os.Signal)
+
+var defaultSignals = []os.Signal{
+	os.Interrupt,
+	os.Kill,
```
konradreiche, line 14: I wonder if those two should be in the same default one. Usually interrupt waits for whatever is still being processed to finish and kill terminates them all immediately.

fishy: Good point. Updated the default list.

fishy:
Added an example.

runtimebp/signal.go
```diff
@@ -0,0 +1,44 @@
...
+	"os/signal"
+	"syscall"
+)
+
+// ShutdownHandler is the callback type used in HandleSignals.
+type ShutdownHandler func(signal os.Signal)
+
+var defaultSignals = []os.Signal{
+	// Ref: https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods
+	syscall.SIGTERM,
```
pacejackson, line 15: Should we also handle `SIGINT` by default? 

https://github.com/reddit/baseplate.py/blob/3638ffd8f561aa90f16e126acb167a41cc7b1727/baseplate/server/__init__.py#L210

fishy: Added.

fishy:
:haircut:

runtimebp/signal.go
konradreiche: Didn't we just remove `os.Interrupt` from the previous discussion? https://github.com/reddit/baseplate.go/pull/142#pullrequestreview-393311225

fishy: Yes. But we changed `os.Kill` into `syscall.SIGTERM` (`os.Kill` is 9 while `SIGTERM` is 15). Interrupt is still useful when developers run the service locally, in the foreground to do some local debugging (so they can use ^C to stop it).

konradreiche: Thanks for the clarification :)

PR #143, opened 2020-04-14
httpbp: refactor v1

pacejackson:
This PR trims down the `httpbp` package to be less framework-y and more just a small shim around `net/http` to handle errors in a simple way and one that lets us report them to spans/etc as well as some helpers to simplify writing HTTP responses.

### DIFF
--- a/httpbp/example_handler_test.go
+++ b/httpbp/example_handler_test.go
@@ -1,0 +1,75 @@
+package httpbp_test
+
+import (
+	"context"
+	"errors"
+	"net/http"
+
+	"github.com/reddit/baseplate.go/edgecontext"
+	"github.com/reddit/baseplate.go/httpbp"
+	"github.com/reddit/baseplate.go/log"
+)
+
+type body struct {
+	X int `json:"x"`
+	Y int `json:"y"`
+}
+
+type errorResponse struct {
+	Reason  string `json:"reason"`
+	Message string `json:"message,omitempty"`
+}
+
+func home(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+	response := httpbp.Response{
+		Body: body{
+			X: 1,
+			Y: 2,
+		},
+	}
+	return httpbp.WriteJSON(w, response)
+}
+
+func err(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+	body := errorResponse{
+		Reason:  "BAD_REQUEST",
+		Message: "this endpoint always returns an error",
+	}
+	return httpbp.NewJSONError(http.StatusBadRequest, body, errors.New("example"))
+}
+
+func loggingMiddleware(next httpbp.HandlerFunc) httpbp.HandlerFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+		log.Infof("Request: %#v", r)
+		return next(ctx, w, r)
+	}
+}
+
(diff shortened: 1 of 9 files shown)


httpbp/example_handler_test.go
```diff
@@ -0,0 +1,75 @@
...
+	_ httpbp.Middleware  = loggingMiddleware
+)
+
+func ExampleBaseplateHandlerFactory() {
+	var (
+		ecImpl       *edgecontext.Impl
+		logger       log.Wrapper
+		trustHandler httpbp.HeaderTrustHandler
+	)
+	handlerFactory := httpbp.BaseplateHandlerFactory{
```
fishy, line 59: 🔕 I think eventually this function should be moved to top level and becomes `baseplate.HTTPHandlerFactory` instead. But definitely not happening in this PR.

httpbp/handler.go
```diff
@@ -3,64 +3,47 @@ package httpbp
...
-
-	if resp.StatusCode() != 0 && resp.StatusCode() != http.StatusOK {
-		w.WriteHeader(resp.StatusCode())
-	}
-	return cw.WriteResponse(w, body)
-}
+// consistency across Baseplate.
+//
+// HandlerFuncs are free to write directly to the given ResponseWriter but
+// WriteResponse and it's helpers for common Content-Types have been provided to
```
fishy, line 18: nit: s/it's/its/

httpbp/handler_test.go
```diff
@@ -2,16 +2,23 @@ package httpbp_test
...
 	"net/http/httptest"
+	"os"
 	"reflect"
 	"sort"
 	"strings"
 	"testing"
 
+	"github.com/reddit/baseplate.go/edgecontext"
+	"github.com/reddit/baseplate.go/log"
+
```
fishy, line 18: nit: extra blank line.

httpbp/example_handler_test.go
pacejackson: I don't think I agree since that means that _every_ service has to include the HTTP stuff as well.

fishy: That's true, but isn't the direction https://github.com/reddit/baseplate.go/pull/133 is going means that eventually the top level will depend on both `thriftbp` and `httpbp` (so it provides battery-included helpers for both thrift and http implementations)?

pacejackson: I was under the impression that the `baseplate` package would _explicitly_ only be things that are agnostic towards the type of service, I haven't checked in on that PR recently so I could be wrong.

httpbp/handler_test.go
```diff
@@ -200,3 +178,84 @@ func TestHandler(t *testing.T) {
 		)
 	}
 }
+
+func TestBaseplateHandlerFactory(t *testing.T) {
+	defer func() {
+		tracing.CloseTracer()
```
konradreiche, line 184: Do we need to check the error here?

pacejackson: I don't think so? We don't in any other tests that do this.

fishy: That error comes from this function, which always return nil: https://github.com/reddit/baseplate.go/blob/cbf7b6770b15e2cfd948c0ac9dcdcf8262f09ac3/mqsend/mqsend_mock.go#L27

PR #147, opened 2020-04-15
httpbp: pass the endpoint name into the Middleware call

pacejackson:


httpbp/handler.go
```diff
@@ -59,8 +59,8 @@ var (
 // instead.
 // NewHander is provided for those who need to avoid the default Baseplate
 // middleware or for testing purposes.
-func NewHandler(handle HandlerFunc, middlewares ...Middleware) http.Handler {
-	return handler{handle: Wrap(handle, middlewares...)}
+func NewHandler(handle HandlerFunc, name string, middlewares ...Middleware) http.Handler {
```
pacejackson, line 62: style question, I feel ilke maybe `name` should come _first_ in these functions and in `Wrap`?  It lines up with the `Handle` functions on `http.ServeMux` that way too.

fishy: yea name being the first one does look better.

fishy: But I think we need to mention that this is just a name, it does not affect routing, as people might get confused.

httpbp/example_handler_test.go
```diff
@@ -38,9 +38,9 @@ func err(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	return httpbp.NewJSONError(http.StatusBadRequest, body, errors.New("example"))
 }
 
-func loggingMiddleware(next httpbp.HandlerFunc) httpbp.HandlerFunc {
+func loggingMiddleware(name string, next httpbp.HandlerFunc) httpbp.HandlerFunc {
 	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
-		log.Infof("Request: %#v", r)
+		log.Infof("Request <%s>: %#v", name, r)
```
fishy, line 43: nit: `<%s>` feels too python, maybe use `%q` instead?

PR #149, opened 2020-04-17
thriftbp: Remove header list dedup logic

fishy:
The THeaderTransport implementation will do the dedup for us before
sending to the wire, and the dedup logic we added requires a lot of
inefficient converting back and forth between slices and maps. So remove
them to rely on the thrift implementation to do the dedup'ing.


thriftbp/tracing.go
```diff
@@ -74,12 +73,12 @@ func CreateThriftContextFromSpan(ctx context.Context, span *tracing.Span) contex
 			HeaderTracingSampled,
 			HeaderTracingSampledTrue,
 		)
-		headers.Add(HeaderTracingSampled)
+		headers = append(headers, HeaderTracingSampled)
 	} else {
-		headers.Remove(HeaderTracingSampled)
+		ctx = thrift.UnsetHeader(ctx, HeaderTracingSampled)
```
pacejackson, line 78: I'm not sure that this actually works, if it was already `Set` then it would be in `headers` because we got that _before_ Unsetting it here, so it's just going to be re-set when we call `SetWriteHeaderList`

fishy: So this is how it works: https://github.com/apache/thrift/blob/9e864d57026b4905862108418ba9482892fb1f65/lib/go/thrift/client.go#L28-L35

We iterate through the write list, and only actually add the key-value to the wire when we can also get the key out of ctx.

UnsetHeader ensures that we can't get this key out of ctx, so it's actually removed even though it's still in the write list.

The write and read lists are designed to be just an index, not the source of truth. They are needed because:

1. There's no way to iterate through key-value pairs in ctx
2. This makes sure that we don't need to do extra sanitize when sharing a ctx object got from client to the upstream server (we only send headers defined in the write list over the wire, not all of them).

fishy: The commit message of the change added that feature also explained a bit more about the background. I probably should have added some of them into the doc comment of the code :badpokerface: https://github.com/apache/thrift/commit/26ef904600edc810f6514605c0611b7442a4c64e

pacejackson: ohohohoh I see, I think my brain got a little scrambled and was thinking that this updated the _write list_ but it's really just removing the value from the context if it was set, so it's still in the "write list" but there's nothing to write.

PR #150, opened 2020-04-20
edgecontext: add support for "origin" service

pacejackson:
This is intended to identify the 'origin' service of this request,
that is the service that is handling the request made from the client

Port from baseplate.py: https://github.com/reddit/baseplate.py/commit/3638ffd8f561aa90f16e126acb167a41cc7b1727


edgecontext/edgecontext_test.go
```diff
@@ -19,15 +19,16 @@ const (
 
 	// copied from https://github.com/reddit/baseplate.py/blob/865ce3e19c549983b383dd49f748599929aab2b5/tests/__init__.py#L55-L59
```
fishy, line 20: This link needs to be updated.

edgecontext/req_context.go
```diff
@@ -65,6 +65,13 @@ func (e *EdgeRequestContext) User() User {
 	}
 }
 
+// OriginService returns the info about the origin of this request.
+func (e *EdgeRequestContext) OriginService() OriginService {
```
fishy, line 69: 🔕 Should this just return a string instead, like [`SessionID`](https://pkg.go.dev/github.com/reddit/baseplate.go/edgecontext?tab=doc#EdgeRequestContext.SessionID)? Or do we expect that we would add more fields into origin service in the future?

edgecontext/req_context.go
```diff
@@ -123,3 +130,11 @@ func (e *EdgeRequestContext) UpdateExperimentEvent(ee *experiments.ExperimentEve
 		ee.DeviceID = uuid.Nil
 	}
 }
+
+type OriginService struct {
```
fishy, line 134: Missing doc comment here: https://cloud.drone.io/reddit/baseplate.go/790/3/2

edgecontext/req_context.go
```diff
@@ -123,3 +130,11 @@ func (e *EdgeRequestContext) UpdateExperimentEvent(ee *experiments.ExperimentEve
...
+	raw NewArgs
+}
+
+func (os OriginService) Name() string {
```
fishy, line 138: and here

edgecontext/req_context.go
pacejackson: Hmm, I wouldn't say we currently have anything _planned_ and there's a good chance we wouldn't add any more fields but it seems like it would be a _little_ more likely than DeviceID/SessionID.

PR #151, opened 2020-04-20
Merge thriftclient into thriftbp package

fishy:
Sometimes the line between thrift client and server is blurry. Also in
both upstream apache thrift package, and stdlib http package, they put
client and server code in the same package. So we follow this pattern.

Most changes in this commit are moving files around and renames.


thriftbp/doc_client_test.go
```diff
@@ -87,11 +87,11 @@ func LoggingMiddleware(next thrift.TClient) thrift.TClient {
 	}
 }
 
-// This example demonstrates a typical use case of thriftclient pool in
+// This example demonstrates a typical use case of thriftbp pool in
 // microservice code with custom middleware.
-func Example() {
-	pool, err := thriftclient.NewBaseplateClientPool(
-		thriftclient.ClientPoolConfig{
+func Example_clientPool() {
```
pacejackson, line 92: 🔕 should this be `Example_ClientPool`?

thriftbp/client_middlewares.go
```diff
@@ -7,31 +7,30 @@ import (
...
 
 	"github.com/reddit/baseplate.go/edgecontext"
-	"github.com/reddit/baseplate.go/thriftbp"
 	"github.com/reddit/baseplate.go/tracing"
 )
 
-// BaseplateDefaultMiddlewares returns the default middlewares should be used by
-// a baseplate service.
+// BaseplateDefaultClientMiddlewares returns the default client middlewares
+// should be used by a baseplate service.
```
pacejackson, line 14:
```suggestion
// that should be used by a baseplate service.
```

pacejackson: Or maybe:

```suggestion
// to be used by a baseplate service.
```

thriftbp/doc_client_test.go
fishy: golint will yell at you if you try to capitalize the `_Suffix`

pacejackson: Ok, I thought it might be something like that, thanks for confirming :)

thriftbp/client_middlewares.go
fishy: Done.

PR #152, opened 2020-04-21
filewatcher: Add size limit

fishy:
This is a basic sanity protection to avoid being bombed by bad
configuration or attacks.

This is a breaking change, but it should make future filewatcher changes
less breaking. :)

### DIFF
--- a/filewatcher/filewatcher.go
+++ b/filewatcher/filewatcher.go
@@ -20,6 +20,11 @@ import (
 // caller can tweak its value when needed.
 var InitialReadInterval = time.Second / 2
 
+// DefaultMaxFileSize is the default MaxFileSize used when it's <= 0.
+//
+// It's 1MiB, which is following the size limit of Apache ZooKeeper nodes.
+const DefaultMaxFileSize = 1024 * 1024
+
 // A Parser is a callback function to be called when a watched file has its
 // content changed, or is read for the first time.
 //
@@ -60,6 +65,10 @@ func (r *Result) watcherLoop(
 	parser Parser,
 	logger log.Wrapper,
 ) {
+	if logger == nil {
+		logger = log.NopWrapper
+	}
+
 	file := filepath.Base(path)
 	for {
 		select {
@@ -68,9 +77,7 @@ func (r *Result) watcherLoop(
 			return
 
 		case err := <-watcher.Errors:
-			if logger != nil {
-				logger("watcher error: " + err.Error())
-			}
+			logger("watcher error: " + err.Error())
 
 		case ev := <-watcher.Events:
 			if filepath.Base(ev.Name) != file {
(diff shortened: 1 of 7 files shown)


filewatcher/filewatcher.go
```diff
@@ -104,6 +105,35 @@ func (r *Result) watcherLoop(
...
+	Parser Parser
+
+	// Optional. When non-nil, it will be used to log errors,
+	// either returned by parser or by the underlying file system watcher.
+	// Please note that this does not include errors returned by the first parser
+	// call, which will be returned directly.
+	Logger log.Wrapper
+
+	// Optional. When <=0 DefaultMaxFileSize will be used instead.
+	// This limits the size of the file that will be read into memory and send to
```
pacejackson, line 123:
```suggestion
	// This limits the size of the file that will be read into memory and sent to
```

filewatcher/filewatcher.go
```diff
@@ -104,6 +105,35 @@ func (r *Result) watcherLoop(
...
+	// the parser, to limit memory usage.
+	// If the file content is larger than MaxFileSize,
+	// we don't treat that as an error,
+	// but the parser will only receive the first MaxFileSize bytes of content.
```
pacejackson, line 127: 🔕 we don't _treat_ this as an error, but generally won't the parser only receiving a portion of the file _result_ in an error?  Would it be better to actually treat this as an error rather than letting it return unpredictable errors later in the execution?

fishy: A few considerations here:

1. Some parser stops reading when they got everything they need (for example, [json](https://play.golang.org/p/QP9yudEszSd), but those parsers won't have a problem even we don't limit the data sent to them)
2. Yes for most parsers only receiving partial data will cause an error, but when a parser returns an error we won't update the parsed result, so that won't cause us problems.
3. Detect the fact that file is too large without actually reading it is tricky. You can use Lstats to get the file size, but that doesn't work for all cases (e.g. symlinks, special files like `/dev/urandom`, etc.). This is a more robust solution.

pacejackson: 👍 
It might be good to include those points in the comment for future reference/documentation?

filewatcher/filewatcher.go
fishy: Done.

filewatcher/filewatcher.go
```diff
@@ -20,6 +20,9 @@ import (
 // caller can tweak its value when needed.
 var InitialReadInterval = time.Second / 2
 
+// DefaultMaxFileSize is the default MaxFileSize used when it's <= 0.
+const DefaultMaxFileSize = 1024 * 1024 // 1MiB
```
konradreiche, line 24: Maybe a comment to reference why this is set to 1MiB:
https://lucene.apache.org/solr/guide/7_4/setting-up-an-external-zookeeper-ensemble.html
> ZooKeeper is designed to hold small files, on the order of kilobytes. By default, ZooKeeper’s file size limit is 1MB. Attempting to write or read files larger than this will cause errors.

filewatcher/filewatcher.go
konradreiche: Won't this hide the underlying cause of parse failures?

filewatcher/filewatcher_test.go
```diff
@@ -283,3 +291,87 @@ func TestParserFailure(t *testing.T) {
...
+func limitedParser(t *testing.T, expectedSize int) filewatcher.Parser {
+	return func(f io.Reader) (interface{}, error) {
+		buf, err := ioutil.ReadAll(f)
+		if err != nil {
+			t.Error(err)
+			return nil, err
+		}
+		if len(buf) != expectedSize {
+			t.Errorf(
+				"Expected size of %d, got %d, data %q",
```
konradreiche, line 304: :no_bell: nit: should we generally go for [lowercase error messages](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)?

filewatcher/limit_parser_test.go
```diff
@@ -0,0 +1,29 @@
...
+		if err != nil {
+			t.Error(err)
+			return nil, err
+		}
+		if string(buf) != expected {
+			t.Errorf("Data expected %q, got %q", expected, buf)
+		}
+		return nil, nil
+	}
+	limitParser(parser, limit)(strings.NewReader(origin))
```
konradreiche, line 28: Worth checking the error here?

filewatcher/filewatcher_test.go
fishy: The lower case requirement is only for the messages in constructed `error` values, not anything else.

filewatcher/limit_parser_test.go
fishy: `parser` implementation already checked that.

filewatcher/filewatcher.go
fishy: So filewatcher is general purpose and not limited to only watch files synced down from zookeeper, that's why it's a configuration, not just a hardcoded value :)

konradreiche: That's why I thought having the comment here and not in the `Config` struct. I didn't know that Zookeeper has a file limit.

fishy:
:haircut:

filewatcher/filewatcher.go
fishy: @konradreiche There's no reliable way to detect the underlying cause, though. Parser implementations can also count the bytes read to see if it's at the limit.

konradreiche: I think making some effort to at least inform the user that there are still more bytes available would greatly help in case there's some misconfiguration going on.

fishy: There's no reliable way to know that without actually reading the file either.

konradreiche: I think the fact that there are edge cases in detecting the file size can be discouraging but is it enough to justify not making an attempt to detect it while it will most likely work in many cases to prevent masked bugs sneaking in?

fishy: As long as their parser return an error, they can figure it out from the error logged/sent to raven (see also: https://github.com/reddit/baseplate.go/pull/153)

Have a unified way (that it's always a parser error) is better because if it usually is another error, but occasionally it's a parser error, then in the long run they might start to only handle it in the other error that doesn't work 100% of the time, and forgot that it could also be a parser error.

PR #153, opened 2020-04-21
log: Add ErrorWithRavenWrapper

fishy:


log/wrapper.go
```diff
@@ -60,3 +61,22 @@ func ZapWrapper(level Level) Wrapper {
...
+// ErrorWithRavenWrapper is a Wrapper implementation that both use Zap logger to
+// log at error level, and send the error to raven.
+//
+// In most cases this should be the one used to pass into Baseplate.go libraries
+// expecting a log.Wrapper.
+//
+// Note that this should not be used as the logger set into thrift.TSimpleServer,
+// as that would use the logger to log network I/O errors,
+// which would be too spammy to be sent to raven.
+func ErrorWithRavenWrapper(msg string) {
```
konradreiche, line 74: While we're at it, we should rename Raven to Sentry. In the Go library they also moved away from the word raven https://github.com/getsentry/sentry-go

log/wrapper.go
```diff
@@ -60,3 +61,22 @@ func ZapWrapper(level Level) Wrapper {
 		f(msg)
 	}
 }
+
+// ErrorWithRavenWrapper is a Wrapper implementation that both use Zap logger to
+// log at error level, and send the error to raven.
```
konradreiche, line 66: Does it make sense to handle the error here twice? I think this deviates a bit from the baseplate.py behavior. I am just wondering if it's the right course of action to surface an error in multiple log systems.

fishy: Send to raven/sentry will only happen when it's actually configured. For example people usually don't configure it in staging environment, and in those environments the send to sentry part will be a nop. This is why the logging is still useful.

log/wrapper.go
fishy: We are currently using `github.com/gensentry/raven-go`: https://github.com/reddit/baseplate.go/blob/e0f1201fa38d0132ba5acfab621435fef7b754dd/log/log.go#L4

I'm not against migrating to sentry-go instead, but sentry-go requires explicit initialization, while raven-go does not. That can happen in a future PR instead.

konradreiche: It's still Sentry we're logging though right? I think they figured Raven vs. Sentry is just confusing.

fishy: what I mean is it's probably better renamed when we migrated to the sentry-go library :)

konradreiche: Sounds good :+1:

PR #154, opened 2020-04-22
log: Migrate from raven-go to sentry-go

fishy:
This is a breaking change as with sentry-go we need to do initialization
explicitly.


log/sentry.go
```diff
@@ -0,0 +1,90 @@
...
+//
+// The io.Closer returned calls sentry.Flush with SentryFlushTimeout.
+// If it returns an error, that error is guaranteed to wrap
+// ErrSentryFlushFailed.
+//
+// You can also just call sentry.Init,
+// which provides even more customizations.
+// This function is provided to do the customizations we care most about,
+// and to provide a Closer to be more consistent with other baseplate packages.
+func InitSentry(cfg SentryConfig) (io.Closer, error) {
```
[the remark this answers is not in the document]
fishy: Mainly two reasons:

1. I'm skeptical on their handling of the default value of SampleRate
2. We do need an extra config that's not in `sentry.ClientOptions`, which is `FlushTimeout`.

Currently `SentryConfig` is basically an intersection between `sentry.ClientOptions` and the configurations we provided on baseplate.py, plus `FlushTimeout` :)

log/log.go
```diff
@@ -27,11 +27,33 @@ const (
 	ZapNopLevel zapcore.Level = zapcore.FatalLevel + 1
 )
 
-// InitLogger provides a quick way to start or replace a logger
-// Pass in debug as log level enables development mode (which makes DPanicLevel logs panic).
+// ToAtomicLevel converts Level to a zap acceptable atomic level struct
```
pacejackson, line 30: What does “atomic level” mean here? It might be good to clarify that in the comment for future devs.

fishy: Done.

metricsbp/baseplate_hooks.go
```diff
@@ -18,7 +21,12 @@ type CreateServerSpanHook struct {
 }
 
 // OnCreateServerSpan registers MetricSpanHooks on a server Span.
+//
+// It also configure the sentry scope with the current trace id.
 func (h CreateServerSpanHook) OnCreateServerSpan(span *tracing.Span) error {
+	sentry.ConfigureScope(func(scope *sentry.Scope) {
+		scope.SetTag("trace_id", strconv.FormatUint(span.TraceID(), 10))
```
fishy, line 28: on second thought, this won't work as this is too global and will fail miserably in parallel request handling :(

fishy: Done.

fishy:
:haircut: Fixed the trace id issue.

tracing/span.go
```diff
@@ -292,6 +297,31 @@ func (s *Span) preStop(err error) {
...
+// object as the current hub.
+//
+// It's called automatically by StartSpanFromHeaders and thriftbp/httpbp
+// middlewares,
+// so you don't need to call it for spans created automatically from requests.
+// But you should call it if you created a top level span manually.
+//
+// It's also not needed to be called for the child spans,
+// as the hub attached would be the same.
+func (s Span) InjectSentryHub(ctx context.Context) context.Context {
```
pacejackson, line 321: 🔕 Do we actually want to inject the hub on the context or just export the code to get it from the span?

fishy: This is needed by the new `log.ErrorWithSentry` API: https://github.com/reddit/baseplate.go/pull/154/files#diff-fbf850c1cb9fc6234051c4e9517a5274R240

If we don't inject it into the context, then we need to add span into that API instead (or try to extract span out of the context), that way this is less useful, as injecting hub into context is a pattern already provided by sentry-go API.

PR #155, opened 2020-04-22
thriftbp: Change the type of ServerConfig.Logger

fishy:
Make it more explicit and obvious that logger is only used by
thrift.TSimpleServer, and should not be set to
log.ErrorWithSentryWrapper.

### DIFF
--- a/log/wrapper.go
+++ b/log/wrapper.go
@@ -72,12 +72,15 @@ func ZapWrapper(level Level) Wrapper {
 // Note that this should not be used as the logger set into thrift.TSimpleServer,
 // as that would use the logger to log network I/O errors,
 // which would be too spammy to be sent to Sentry.
-func ErrorWithSentryWrapper(msg string) {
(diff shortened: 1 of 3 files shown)


fishy:
That extra typecasting is a good way to indicate that "this logger is different from all the other loggers I'm passing around in main function" and force them to pay more attention, no?

fishy:
You can't stop people from footgunning, but you can make it slightly harder :)

fishy:
> Why do we actually have this as a LogWrapper in the first place. Isn't everything covered by the `ErrorWithSentry` method? I thought the log wrapper is only for logging within the baseplate.go library.

I think by "this" you mean `thrift.Logger`? That's added by me following the same principals we used for `log.Wrapper` in baseplate.go. This was the upstream jira ticket discussion the change: https://issues.apache.org/jira/browse/THRIFT-4985.

This is also why it's compatible with `log.Wrapper` :)

PR #156, opened 2020-04-23
httpbp: refactor errors and support standard error format

pacejackson:


httpbp/errors.go
```diff
@@ -0,0 +1,221 @@
...
+	Unwrap() error
+}
+
+// NewHTTPError returns a new HTTPError object initialized with the given
+// values.
+//
+// NewHTTPError is provided for testing purposes and should not be used directly,
+// you should use one of the JSON/HTML/Raw error functions to return standard
+// HTTPErrors
+func NewHTTPError(code int, body interface{}, cause error, cw ContentWriter) HTTPError {
```
fishy, line 33: 🔕 Is it ever used outside of this package? Or do you expect service code might need to use it in their own unit tests?

If not we can probably unexport it.

httpbp/errors.go
```diff
@@ -0,0 +1,221 @@
...
+
+// ErrorResponse is the base struct used by all the of standard errors in httpbp.
+//
+// ErrorResponse can be used to construct JSON, HTML, and Raw HTTPError responses.
+//
+// You should not generally need to create ErrorResponses manually as standard
+// ones have been provided for each of the common 4xx and 5xx errors.
+type ErrorResponse struct {
+	// The HTTP status code of the response.
+	Code        int
```
fishy, line 77: nit1: I usually prefer to either add json annotation to all fields, or none at all (and rely on go's default behavior to convert field name into json keys). mixing them causes inconsistency.

nit2: since those are not consecutive lines, the align of type won't happen, so you don't need to add extra spaces :)

httpbp/errors.go
pacejackson: We _probably_ don't need to export it anymore.

httpbp/errors.go
```diff
@@ -0,0 +1,237 @@
...
+
+// RawError returns the given error as an HTTPError that will write a raw
+// response of the given content-type.
+func RawError(resp *ErrorResponse, cause error, contentType string) HTTPError {
+	return newHTTPError(resp.code, resp, cause, RawContentWriter(contentType))
+}
+
+// RetryableError can be used to wrap an ErrorResponse and communicate to the
+// caller that the request can be retried after the given duration.
+func RetryableError(resp ErrorResponse, w http.ResponseWriter, retryAfter time.Duration) ErrorResponse {
```
fishy, line 171: 🔕 Although semantically this API is related to an error, but technically it didn't touch the `resp` passed in and always return it as-is. This means this can be written completely as a helper around `http.ResponseWriter`, without involving `ErrorResponse`.

httpbp/headers.go
```diff
@@ -24,6 +24,9 @@ const (
 	// the HTTP request headers.
 	ParentIDHeader = "X-Parent"
 
+	// RetryAfterHeader is the standard "Retry-After" header key.
+	RetryAfterHeader = "Retry-After"
```
fishy, line 28: This is a weird place to put it. This is a standard HTTP header, not tracing related or something baseplate defined. We probably should put it out of this const block.

httpbp/errors.go
pacejackson: Ya, maybe this is _too_ cute?  My thinking was it makes for nice looking code to be able to do something like:

```go
return httpbp.JSONError(
    httpbp.RetryableError(httpbp.ServiceUnavailable(), w, time.Hour), errors.New("Downtime"),
)
```

fishy: hmm that's a reasonable use case, maybe just add to comment to explain it? also probably extract the version don't require `resp` as well :)

fishy: _technically_ you can still do:

```
httpbp.WriteRetryableError(e, time.Hour)
return httpbp.JSONError(httpbp.ServiceUnavailable(), errors.New("downtime"))
```

but I can see that language sugar still be useful :)

pacejackson: I'll leave it like this for now and we can mull over it while I grind out tests/errors for each response code 😅

pacejackson: I think I found a nicer way to do it actually:

```go
// Retryable communicates to the  caller that the request may be retried after
// the given duration.
func (r *ErrorResponse) Retryable(w http.ResponseWriter, retryAfter time.Duration) *ErrorResponse {
	after := strconv.FormatInt(int64(retryAfter/time.Second), 10)
	w.Header().Set(RetryAfterHeader, after)
	return r
}
```

So it's now:

```go
return httpbp.JSONError(
    httpbp.ServiceUnavailable().Retryable(w, time.Hour), errors.New("Downtime"),
)
```

fishy: Nice. The new approach makes a lot of sense :+1:

httpbp/errors.go
```diff
@@ -0,0 +1,293 @@
...
+import (
+	"fmt"
+	"html/template"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+const (
+	// RetryAfterHeader is the standard "Retry-After" header key.
```
fishy, line 12: nit: Probably worth mentioning RFC2616 here: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html. e.g.:

>RetryAfterHeader is the standard "Retry-After" header key defined in RFC2616.

pacejackson:
Screen Shot 2020-04-24 at 4 26 49 PM

The standard HTML error response looks like this (the table is not there if no `Details` are set).

httpbp/errors.go
```diff
@@ -0,0 +1,560 @@
...
+	return t.New(DefaultErrorTemplateName).Parse(text)
+}
+
+// BadRequest is for 400 responses.
+//
+// This is appropriate for when the client sends a malformed or invalid request.
+//
+// If the client sent an invalid request, you can send the details using the
+// Details map in the ErrorResponse.
+func BadRequest() *ErrorResponse {
```
fishy, line 325: 🔕 Do you think it makes sense to construct a map of http code -> *ErrorResponse? We can also provide a function instead, fallback to a generic one (maybe the 500 one) if it's not in the map.

pacejackson: We can't just _directly_ map code to *ErrorResponse because some of the methods mutate the error response, so we either need to update the existing code to clone the ErrorResponse in those methods and return a new ErrorResponse, or have a function that takes in the http code and returns a new *ErrorResponse.

Also, to clarify, are you suggesting that rather than having a dedicated function for each error code, we have a single one that takes in an http error code?  I think that the ergonomics of that are probably _worse_ than the current approach.  Or maybe this map is hidden and all the these functions do is return the appropriate error from that map?

fishy: It doesn't have to be a map of code -> response, it could be from code to a function to give you a response. (or a function that takes in code as arg and return a function)

pacejackson: Hmm, maybe have both options? You the give me a code, you get back an error can be used in cases where someone has multiple potential errors and wants to easily pick one while the majority of cases use the specific functions directly?

It also might help to give these functions a standard prefix, say `Err` so it's `httpbp.ErrForbidden`, etc?

fishy: yea give both options makes sense. as for common prefix, I don't think `Err` is a good one (it's more commonly used by pre defined error variables), and I can't think of a good one so maybe not?

httpbp/errors.go
```diff
@@ -0,0 +1,621 @@
...
+		http.StatusUnsupportedMediaType,
+		"UNSUPPORTED_MEDIA_TYPE",
+		"The given media type is not supported.",
+	)
+}
+
+// Teapot is for 418 responses.
+//
+// This is appropriate for when the server is a teapot rather than a coffee maker.
+func Teapot() *ErrorResponse {
```
konradreiche, line 439: I thought this has been deprecated but there has been the [Save 418 Movement](https://en.wikipedia.org/wiki/Hyper_Text_Coffee_Pot_Control_Protocol#Save_418_movement)

Glad we're taking a stance here.

PR #157, opened 2020-04-23
log: More detailed doc around Wrapper interface

fishy:
Also add FallbackWrapper for nil-safe usage of it.

### DIFF
--- a/log/wrapper.go
+++ b/log/wrapper.go
@@ -7,19 +7,71 @@ import (
 	sentry "github.com/getsentry/sentry-go"
 )
 
-// Wrapper is a simple wrapper of a logging function.
+// Wrapper defines a simple interface to wrap logging functions.
 //
+// As principles, library code should
+//
+// 1. Not do any logging.
+//    The library code should communicate errors back to the caller,
+//    and let the caller decide how to deal with them
+//    (log them, ignore them, panic, etc.)
+//
+// 2. In some rare cases, 1 is not possible,
+//    for example the error might happen in a background goroutine.
+//    In those cases some logging is necessary,
+//    but those should be kept at minimal,
+//    and the library code should provide control to the caller on how to do
+//    those logging.
+//
+// This interface is meant to solve Principle 2 above.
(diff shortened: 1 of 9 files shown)


log/wrapper.go
```diff
@@ -7,20 +7,53 @@ import (
...
+//
+// This interface is meant to solve Principle 2 above.
 // In reality we might actually use different logging libraries in different
 // services, and they are not always compatible with each other.
 // Wrapper is a simple common ground that it's easy to wrap whatever logging
 // library we use into.
 //
-// This is also the same type as thrift.Logger and can be used interchangeably
-// (sometimes with a typecast).
+// For production code,
```
pacejackson, line 29: Should we also call out that this is really only meant for passing into libraries, production code can just use the functions provided by the `log` module?

fishy: Done. PTAL?

fishy:
:haircut:

I added even more clarification about when to use log.Wrapper.

I also did an audit on all the current use of it, and found out that we are fine :) But as a result I unexported Span.LogError to avoid it being abused by service code.

log/wrapper.go
```diff
@@ -7,20 +7,72 @@ import (
...
 // services, and they are not always compatible with each other.
 // Wrapper is a simple common ground that it's easy to wrap whatever logging
 // library we use into.
 //
-// This is also the same type as thrift.Logger and can be used interchangeably
-// (sometimes with a typecast).
+// With that in mind, this interface should only be used by library code,
+// when the case satisfies all of the following 3 criteria:
+//
+// 1. This is a bad thing happened.
```
pacejackson, line 35:
```suggestion
1. A bad thing happened.
```

log/wrapper.go
```diff
@@ -7,20 +7,72 @@ import (
...
+//
+// 2. This is an unexpected thing happened.
```
pacejackson, line 37:
```suggestion
// 2. This is unexpected.
```

PR #158, opened 2020-04-23
[httpbp]base64 encoding decoding edgecontext

suzie-su:
## Description
Since our current edgecontext header transfer is using base64encoded so we need to keep it consistent to attach to the header with base64 format
👓 @pacejackson @fishy

### DIFF
--- a/httpbp/headers.go
+++ b/httpbp/headers.go
@@ -1,12 +1,14 @@
 package httpbp
 
 import (
+	"encoding/base64"
 	"fmt"
 	"net/http"
 	"sort"
 	"strings"
 	"time"
 
+	"github.com/reddit/baseplate.go/edgecontext"
 	"github.com/reddit/baseplate.go/secrets"
 	"github.com/reddit/baseplate.go/signing"
 )
@@ -66,10 +68,24 @@ type EdgeContextHeaders struct {
 
 // NewEdgeContextHeaders returns a new EdgeContextHeaders object from the given
 // HTTP headers.
+//
+// The edge context header using base64 encoding for http transport, so it needs decode first
 func NewEdgeContextHeaders(h http.Header) EdgeContextHeaders {
-	return EdgeContextHeaders{
-		EdgeRequest: h.Get(EdgeContextHeader),
+	ec, err := base64.StdEncoding.DecodeString(h.Get(EdgeContextHeader))
+	if err != nil {
+		ec = []byte("")
 	}
+	return EdgeContextHeaders{
+		EdgeRequest: string(ec),
+	}
+}
+
+// AttachEdgeContextHeaders attach EdgeRequestContext into request/response header
+//
+// The base64 encoding is only for http transport
+func AttachEdgeContextHeaders(ec *edgecontext.EdgeRequestContext, h http.Header) {
+	ecencoded := base64.StdEncoding.EncodeToString([]byte(ec.Header()))
+	h.Set(EdgeContextHeader, ecencoded)
 }
 
 // AsMap returns the EdgeContextHeaders as a map of header keys to header
--- a/httpbp/headers.go
+++ b/httpbp/headers.go
@@ -70,22 +70,17 @@ type EdgeContextHeaders struct {
 // HTTP headers.
 //
 // The edge context header using base64 encoding for http transport, so it needs decode first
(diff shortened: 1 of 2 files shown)


httpbp/headers.go
```diff
@@ -67,11 +69,21 @@ type EdgeContextHeaders struct {
...
+	}
 	return EdgeContextHeaders{
-		EdgeRequest: h.Get(EdgeContextHeader),
+		EdgeRequest: string(ec),
 	}
 }
 
+// AttachEdgeContextHeaders attach EdgeRequestContext into request/response header
+func AttachEdgeContextHeaders(ec *edgecontext.EdgeRequestContext, h http.Header) {
+	ecencoded := base64.StdEncoding.EncodeToString([]byte(ec.Header()))
```
pacejackson, line 83: typo:

```suggestion
	encoded := base64.StdEncoding.EncodeToString([]byte(ec.Header()))
```

httpbp/headers.go
```diff
@@ -67,11 +69,21 @@ type EdgeContextHeaders struct {
...
+		ec = []byte("")
+	}
 	return EdgeContextHeaders{
-		EdgeRequest: h.Get(EdgeContextHeader),
+		EdgeRequest: string(ec),
 	}
 }
 
+// AttachEdgeContextHeaders attach EdgeRequestContext into request/response header
+func AttachEdgeContextHeaders(ec *edgecontext.EdgeRequestContext, h http.Header) {
```
pacejackson, line 82: Maybe call this `SetEdgeContextHeader`?

httpbp/headers.go
```diff
@@ -67,11 +69,21 @@ type EdgeContextHeaders struct {
 // NewEdgeContextHeaders returns a new EdgeContextHeaders object from the given
 // HTTP headers.
 func NewEdgeContextHeaders(h http.Header) EdgeContextHeaders {
+	ec, err := base64.StdEncoding.DecodeString(h.Get(EdgeContextHeader))
+	if err != nil {
```
fishy, line 73: I'm not so sure that we want to throw the error away.

`NewEdgeContextHeaders` is called by `TrustEdgeContext`, and if the decoding failed, `TrustEdgeContext` should also report failure.

Although in most cases returning an empty edge request here will also cause `TrustEdgeContext` to fail, I still prefer this to be done explicitly.

suzie-su:
change notes:
1. renaming
2. explicit the error and and return trust edge context false when error happens
3. use w http.ResponseWriter as argument instead of http.header for SetEdgeContextHeader

httpbp/headers.go
```diff
@@ -66,10 +68,19 @@ type EdgeContextHeaders struct {
...
-		EdgeRequest: h.Get(EdgeContextHeader),
-	}
+//
+// The edge context header using base64 encoding for http transport, so it needs decode first
+func NewEdgeContextHeaders(h http.Header) (EdgeContextHeaders, error) {
+	ec, err := base64.StdEncoding.DecodeString(h.Get(EdgeContextHeader))
+	return EdgeContextHeaders{EdgeRequest: string(ec)}, err
+}
+
+// SetEdgeContextHeader attach EdgeRequestContext into request/response header
```
pacejackson, line 78: I think this is just `response header` now.

httpbp/headers_test.go
```diff
@@ -278,7 +289,29 @@ func TestTrustHeaderSignatureSignAndVerify(t *testing.T) {
 		},
 	)
 }
+func TestInvalidEdgeContextHeader(t *testing.T) {
```
fishy, line 292: nit: Add a blank line between functions.

httpbp/headers_test.go
```diff
@@ -278,7 +289,29 @@ func TestTrustHeaderSignatureSignAndVerify(t *testing.T) {
...
+	ecSignature, err := truster.SignEdgeContextHeader(
+		httpbp.EdgeContextHeaders{EdgeRequest: "=="},
+		time.Minute,
+	)
+	invalidEdgeContextHeader.Header.Set(httpbp.EdgeContextSignatureHeader, ecSignature)
+
+	if truster.TrustEdgeContext(invalidEdgeContextHeader) {
+		t.Error("trust mismatch, expected False, got True")
+	}
+}
```
fishy, line 314: Here, too.

PR #160, opened 2020-04-25
top: Fix sentry initialization and add test for config parsing

fishy:


server.go
```diff
@@ -57,23 +58,26 @@ type Server interface {
 }
 
 // ParseServerConfig will populate a ServerConfig from a YAML file.
-func ParseServerConfig(path string) (*ServerConfig, error) {
-	cfg := &ServerConfig{}
+func ParseServerConfig(path string) (ServerConfig, error) {
 	if path == "" {
-		return nil, errors.New("no config path given")
+		return ServerConfig{}, errors.New("baseplate.ParseServerConfig: no config path given")
```
pacejackson, line 63: Should we export this as an error rather than just returning `errors.New` here?

fishy: Consider this is usually called by main function and the most likely way to handle this error is to just panic, probably no?

Also most of this package would get rewritten later on. This change is to fix some immediate issue to make it more usable 😅

PR #164, opened 2020-04-28
signing: align interface with baseplate.py

pacejackson:
* Require that the caller pass in a VersionedSecret for both Sign and Verify operations.
* Add a helper method to SimpleSecret to easily get a VersionedSecret.
* Update baseplate to account for this breaking change.

### DIFF
--- a/httpbp/headers.go
+++ b/httpbp/headers.go
@@ -241,7 +241,7 @@ func (h TrustHeaderSignature) signHeaders(headers Headers, secretPath string, ex
 	}
 	return signing.Sign(signing.SignArgs{
 		Message:   headerMessage(headers),
-		Key:       secret.Current,
+		Secret:    secret,
 		ExpiresIn: expiresIn,
 	})
 }
@@ -256,7 +256,7 @@ func (h TrustHeaderSignature) verifyHeaders(headers Headers, signature string, s
 		return false, err
 	}
 
-	if err = signing.Verify(headerMessage(headers), signature, secret.GetAll()...); err != nil {
+	if err = signing.Verify(headerMessage(headers), signature, secret); err != nil {
 		return false, err
 	}
 	return true, nil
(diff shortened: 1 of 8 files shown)


pacejackson:
I re-ran the benchmark too:

```
goos: darwin
goarch: amd64
pkg: github.com/reddit/baseplate.go/signing
BenchmarkV1/sign-12         	  749191	      1592 ns/op	     736 B/op	      15 allocs/op
BenchmarkV1/verify-12       	  716444	      1601 ns/op	     696 B/op	      12 allocs/op
BenchmarkV1/verify-3keys-12 	  312458	      3892 ns/op	    1864 B/op	      26 allocs/op
PASS
ok  	github.com/reddit/baseplate.go/signing	4.597s
Success: Benchmarks passed.
```

I'm not sure how much of a change that is? It wouldn't be an _unexpected_ change (also I would argue the old benchmark was just hiding this by forcing the caller to do it themself).

pacejackson:
Benchmark on master is pretty similar, probably very slightly faster because it's not calling `GetAll` multiple times:

```
Running tool: /usr/local/bin/go test -benchmem -run=^$ github.com/reddit/baseplate.go/signing -bench ^(BenchmarkV1)$

goos: darwin
goarch: amd64
pkg: github.com/reddit/baseplate.go/signing
BenchmarkV1/sign-12         	  731336	      1622 ns/op	     736 B/op	      15 allocs/op
BenchmarkV1/verify-12       	  769293	      1690 ns/op	     728 B/op	      13 allocs/op
BenchmarkV1/verify-3keys-12 	  326895	      3773 ns/op	    1720 B/op	      24 allocs/op
PASS
ok  	github.com/reddit/baseplate.go/signing	5.801s
Success: Benchmarks passed.
```

We could _technically_ speed that up by caching `GetAll` after the first call but that seems like a benchmark-only optimization since you aren't likely to call `GetAll` on a single secret many times in a single request.

fishy:
nah that's premature optimization for very marginal performance gain, don't worry about it. :)

PR #168, opened 2020-05-01
baseplate package 1.0

pacejackson:
Overhaul the top level `baseplate` package and disperse some of its functionality into different packages.  

- The `Config` formats for a particular package along with the `Init` methods that go along with those `Config`s are now in each package rather than being done manually in the `baseplate` package.
- Add a `baseplate.Baseplate` object that in responsible for initializing the monitoring/logging frameworks and forms the base of a `baseplate.Server`.
- Add a `baseplate.Server` implementation to `httpbp`.
- Add a `baseplate.Server` implementation to `thriftbp`.
- Add `baseplate.Serve` to manage starting and shutting down a `baseplate.Server` using `runtimebp.HandleShutdown`.  Accepts a `StopTimeout` configuration value to control the amount of time you wait for a `Server` to stop before exiting.


thriftbp/server_middlewares.go
```diff
@@ -17,6 +17,24 @@ var (
...
+// BaseplateDefaultProcessorMiddlewares returns the default processor
+//  middlewares that should be used by a baseplate Thrift service.
+//
+// Currently they are (in order):
+//
+// 1. InjectServerSpan
+//
+// 2. InjectEdgeContext
+//
+// 3. ExtractDeadlineBudget
```
fishy, line 29: I do think DeadlineBudget should be the first middleware on server and last on client (but on client all the additional ones will be after the default ones, so that's not possible, but it should at least be the last of the default ones). This is the relative time, so every extra in between will make it less accurate (e.g. if there's a middleware takes more than 1ms before it on the server or after it on the client, that is not take into consideration of the budget).

This won't be a problem for phase 2 as that's absolute time, but it is a problem for phase 1.

metricsbp/config.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/reddit/baseplate.go/log"
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+type Config struct {
+	Namespace           string
+	Endpoint            string
+	CounterSampleRate   *float64
+	HistogramSampleRate *float64
```
fishy, line 15: I'd suggest that for those more complex config field names, add the yaml annotation here.

Otherwise you can only use the yaml's default conversion from field name to key name in yaml file, which is not great: https://github.com/reddit/baseplate.go/blob/63c4dda088d9f8f53af9080618bdf60135972872/server_test.go#L27

thriftbp/server.go
```diff
@@ -4,8 +4,30 @@ import (
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/reddit/baseplate.go"
```
fishy, line 7: nit: Add a blank line to separate the import groups :)

thriftbp/server_middlewares.go
pacejackson: Makes sense with one caveat, I think it needs to come _second_ on the server side after the Span step, reason being that Spans are how we monitor these requests, if we don’t set those up first, then requests could fail silently (from a server monitoring perspective)

fishy: This one is resilient (won't fail or something), but as long as span creation one is fast enough I'm fine with that.

baseplate.go
```diff
@@ -0,0 +1,140 @@
...
+	Config() Config
+	EdgeContextImpl() *edgecontext.Impl
+	Secrets() *secrets.Store
+}
+
+// Server is the primary interface for baseplate servers.
+type Server interface {
+	Baseplate() Baseplate
+	Serve() error
+}
```
fishy, line 46: `Server` should have a `Stop` or `StopServe` or whatever it's called in the interface that's called from shutdown handler for graceful shutdown.

thriftbp/server.go
```diff
@@ -3,6 +3,9 @@ package thriftbp
 import (
 	"time"
 
+	"github.com/reddit/baseplate.go"
+	"github.com/reddit/baseplate.go/log"
```
fishy, line 7: This group should be after the third party imports (thrift) :)

httpbp/server.go
```diff
@@ -0,0 +1,143 @@
+package httpbp
+
+import (
+	"context"
+	"errors"
+	"net/http"
+
+	"github.com/reddit/baseplate.go"
```
fishy, line 8: nit: use rename import (`baseplate "github.com/reddit/baseplate.go"`)

metricsbp/config.go
```diff
@@ -0,0 +1,50 @@
...
+// Can be deserialized from YAML.
+type Config struct {
+	// Namespace is the standard prefix applied to all of your metrics, it should
+	// include the name of your service.
+	Namespace string
+
+	// Endpoint is the endpoint you send metrics to.
+	Endpoint string
+
+	// CounterSampleRate is the % of counters that you want to send to your metrics
```
fishy, line 22: nit: say "%" in the comment will mislead user from putting `50` when they want to mean 50% sample rate, but that will give them 100% sampling (because 50>1), and they actually need to use `0.5` for actually 50% sample rate.

baseplate.go
```diff
@@ -0,0 +1,245 @@
...
+				ch <- server.Stop()
+			}()
+
+			// Wait for either the context to timeout or Stop() to finish.
+			select {
+			case <-ctx.Done():
+				err = ctx.Err()
+			case e := <-ch:
+				err = e
+			}
```
pacejackson, line 110: Originally this was handled by `Server.Stop` (largely because `http.Server` handles this), but it seemed like it made more sense to just do this here rather than in each implementation.

pacejackson:
I think we still want to add some more tests (specifically around the top-level package) but I think the overall approach is ready for review.

baseplate.go
```diff
@@ -0,0 +1,245 @@
...
+	Addr string
+
+	// Timeout is the socket connection timeout for Servers that
+	// support that.
+	Timeout time.Duration
+
+	// StopTimeout is the timeout for the Stop command for the service.
+	//
+	// If this is not set, then no timeout will be set on the Stop command.
+	StopTimeout time.Duration `yaml:"stopTimeout"`
```
konradreiche, line 37: I think we should make the other YAML also explicit even though the mapping is correct implicitly. It communicates right away that this whole config is mapped to a file where just seeing this one `yaml` tag I'd thought only that one is mapped.

baseplate.go
```diff
@@ -0,0 +1,245 @@
...
+	Log     log.Config
+	Metrics metricsbp.Config
+	Secrets secrets.Config
+	Sentry  log.SentryConfig
+	Tracing tracing.Config
+}
+
+// Baseplate is the general purpose object that you build a Server on.
+type Baseplate interface {
+	io.Closer
```
konradreiche, line 48: Nice

baseplate.go
```diff
@@ -0,0 +1,245 @@
...
+
+	Config() Config
+	EdgeContextImpl() *edgecontext.Impl
+	Secrets() *secrets.Store
+}
+
+// Server is the primary interface for baseplate servers.
+type Server interface {
```
konradreiche, line 56: Why is this an interface?

baseplate.go
```diff
@@ -0,0 +1,245 @@
...
+	// shutting it down gracefully.
+	Serve() error
+
+	// Stop should stop the server gracefully and only return after the server has
+	// finished shutting down.
+	//
+	// It is recommended that you use baseplate.Serve() rather than calling Stop
+	// directly as baseplate.Serve will manage starting your service as well as
+	// shutting it down gracefully in response to a shutdown signal.
+	Stop() error
```
konradreiche, line 74: Shouldn't `Close()` take care of what `Stop` does or are these separate steps we want to differentiate?

baseplate.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (c cancelCloser) Close() error {
+	c.cancel()
+	return nil
+}
+
+// New parses the config file at the given path, initializes the monitoring and
+// logging frameworks, and returns the "serve" context and a new Baseplate to
+// run your service on.
+func New(path string) (context.Context, Baseplate, error) {
```
konradreiche, line 154: I think this should still accept a `context.Context` rather than producing its own, for instance for testing. I think `context.Context` is so idiomatic in Go we don't need to necessarily lift that part of the shoulders of baseplate users?

httpbp/server.go
```diff
@@ -0,0 +1,271 @@
...
+func (f httpHandlerFactory) NewHandler(name string, handle HandlerFunc, middlewares ...Middleware) http.Handler {
+	wrappers := make([]Middleware, 0, len(f.middlewares)+len(middlewares))
+	wrappers = append(wrappers, f.middlewares...)
+	wrappers = append(wrappers, middlewares...)
+	return NewHandler(name, handle, wrappers...)
+}
+
+// Pattern is the pattern passed to a EndpointRegistry when registering an
+// Endpoint.
+type Pattern string
```
konradreiche, line 42: I wanted to say something like `Path` or `EndpointPath` would be better, but `ServerMux` also calls it pattern :man_shrugging:

secrets/config.go
```diff
@@ -0,0 +1,29 @@
...
+//
+// Can be deserialized from YAML.
+type Config struct {
+	// Path is the path to the secrets.json file file to load your service's
+	// secrets from.
+	Path string
+}
+
+// InitFromConfig returns a new *secrets.Store using the given context and config.
+func InitFromConfig(ctx context.Context, cfg Config) (*Store, error) {
```
konradreiche, line 20: :no_bell: `NewFromConfig`?

httpbp/server.go
```diff
@@ -0,0 +1,271 @@
...
+	return s.bp
+}
+
+func (s *testServer) Serve() error {
+	s.wg.Wait()
+	return nil
+}
+
+func (s *testServer) Stop() error {
+	s.srv.Close()
```
konradreiche, line 261: return the error of `Close`?

baseplate.go
pacejackson: Because we have methods in baseplate that operate on a `Server` (specifically `baseplate.Serve`) and have multiple implementations.  I think also _generally_ for something like this an interface is good for just communicating what is the expected API.

baseplate.go
pacejackson: I didn't really want to tie the `Baseplate` to the `Server` in that way, `Close` closes out all of the non-server stuff that `Baseplate` starts up while this is for stopping the `Server`.

httpbp/server.go
pacejackson: It doesn't return one:

https://golang.org/pkg/net/http/httptest/#Server.Close

secrets/config.go
pacejackson: I can see either approach 😅 

1. The nice thing about this is it follows the same convention as the other packages so it's clear what it's for.
2. The part that is a little weird is it's the only one that isn't setting up some global state and just returning an `io.Closer` so I can see why renaming it might make sense.

What does everyone else think?

httpbp/server.go
```diff
@@ -0,0 +1,271 @@
...
+	// Baseplate is a required argument to NewBaseplateServer and must
+	// be non-nil.
+	Baseplate baseplate.Baseplate
+
+	// Endpoints is the mapping of endpoint patterns to Endpoint objects that
+	// the Server will handle.
+	//
+	// While endpoints is not technically required, if none are provided, your
+	// server will not handle any Endpoints.
+	Endpoints map[Pattern]Endpoint
```
fishy, line 84: 🔕 Looking at [the doc of `http.ServeMux`](https://pkg.go.dev/net/http?tab=doc#ServeMux), it handles pattern conflicting in a deterministic way, not depending on the order of them being registered, so using a map is fine (as the iteration of a map is non-deterministic in go). But if the user uses a different mux implementation that depending on the order of endpoints are registered, using a map will cause non-deterministic result. Maybe add `Pattern` as a field of `Endpoint` and make this a slice of `Endpoint`s is better. But I don't feel strongly (using map can make sure that we don't have 2 exactly same patterns, which is a case that would panic `http.ServeMux`, so that's a pro for using a map.)

secrets/config.go
fishy: I slightly prefer `InitFromConfig` to keep it consistent.

httpbp/server.go
pacejackson: I think that's a _reasonable_ approach too (we could always update `ValidateAndSetDefaults` to check for duplicate patterns as well).

baseplate.go
konradreiche: Fair enough, I thought the only implementation is the test server.

baseplate.go
konradreiche: Why not? :sweat_smile:

httpbp/server.go
pacejackson: @konradreiche do you have any feelings on the matter?  The _entire_ reason I made an interface to plug this in is so we could potentially support things other than `http.ServeMux` so it might make sense.

log/config.go
```diff
@@ -0,0 +1,18 @@
...
+	Level Level
+}
+
+// InitFromConfig initializes the log package using the given Config.
+func InitFromConfig(cfg Config) {
+	if cfg.Level == "" {
+		cfg.Level = InfoLevel
+	}
+	level := cfg.Level
+	InitLogger(level)
```
fishy, line 17: I just merged the other PR so you need to rebase on master and use `InitLoggerJSON` here instead 😅

baseplate.go
pacejackson: Part of it is around separation of concerns, the `Baseplate` is what you built a `Server` on, not the manager of the `Server`.  

It also means that you either need to build the Server and Baseplate together so you can register the Server with the Baseplate closers (which means you have to have a whole `ServerFactory` system that you pass in to New and write/test more code for each Server type) _or_ each server type has to know about this dependency and register their closer with the Baseplate they get passed in.

Finally though, I just don't know that it really makes sense?  The current flow seems the most intuitive to me where you `Close` the Baseplate _after_ the `Server` is shut down, but they aren't actually directly tied together in any way, it just happens when you `defer bp.Close()` because your `main` method won't return until the server is stopped.

httpbp/server.go
fishy: I feel that the way `http.ServeMux` handles conflicts, as described in its doc, makes a lot of sense so I would imagine most of the actual good mux implementations should follow it to some level, as a result map is likely to work fine (if it depends on the order, we might not want to use that mux implementation in the first place?)

pacejackson: WFM

PR #169, opened 2020-05-02
log: Add InitLoggerJSON for full json logging

fishy:
Also update zap to latest version (v1.15.0).


log/log.go
```diff
@@ -71,6 +71,34 @@ func InitLogger(logLevel Level) {
...
 }
 
+// InitLoggerJSON initializes global logger with full json format.
+//
+// The JSON format is also compatible with logdna's ingestion format:
+// https://docs.logdna.com/docs/ingestion
+//
+// Pass in debug as log level enables development mode,
+// which makes DPanicLevel logs panic.
+func InitLoggerJSON(logLevel Level) {
```
pacejackson, line 81: 🔕 Should we use this as the default `Init` rather than `InitLogger`?

fishy: The default one will be controlled by `InitFromConfig` from https://github.com/reddit/baseplate.go/pull/168, eventually :)

log/log.go
```diff
@@ -71,6 +71,34 @@ func InitLogger(logLevel Level) {
...
+	config.Encoding = "json"
+	config.EncoderConfig.EncodeCaller = zapcore.ShortCallerEncoder
+	config.EncoderConfig.EncodeTime = JSONTimeEncoder
+	config.EncoderConfig.EncodeLevel = zapcore.CapitalLevelEncoder
+	// json keys expected by logdna:
+	config.EncoderConfig.MessageKey = "message"
+	config.EncoderConfig.TimeKey = "timestamp"
+	config.Level = zap.NewAtomicLevelAt(lvl)
+
+	if lvl == zap.DebugLevel {
```
konradreiche, line 93: I am not sure we should bind the log level to behavior as drastic as panicing. I'd imagine putting a production instance into debug logging for debugging but without wanting the instance to actually panic or are there use cases where that's the wanted behavior on production.

I'd imagine passing in a development bool or having a `InitLoggerJSONDevelopment`.

Should we make use of `NewDevelopmentConfig()` too?

log/log.go
```diff
@@ -71,6 +71,34 @@ func InitLogger(logLevel Level) {
...
+//
+// The JSON format is also compatible with logdna's ingestion format:
+// https://docs.logdna.com/docs/ingestion
+//
+// Pass in debug as log level enables development mode,
+// which makes DPanicLevel logs panic.
+func InitLoggerJSON(logLevel Level) {
+	config := zap.NewProductionConfig()
+	lvl := logLevel.ToZapLevel()
+	config.Encoding = "json"
```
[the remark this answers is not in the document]
fishy: Yes that's implicit from `zap.NewProductionConfig()`, but we want to make it explicit here.

log/log.go
fishy: The difference `config.Development = true` makes is mainly just how `DPanic` behaves (whether it actually panics). But I think your point makes sense and I'll make the change.

fishy: So I removed this `if` block. If user really want to setup development logging, they can use `InitLoggerWithConfig` instead.

PR #170, opened 2020-05-05
Baseplate package 1.0.1

pacejackson:


baseplate.go
```diff
@@ -38,11 +39,23 @@ type Config struct {
...
 	Log     log.Config       `yaml:"log"`
 	Metrics metricsbp.Config `yaml:"metrics"`
+	Runtime runtimebp.Config `yaml:"runtime"`
 	Secrets secrets.Config   `yaml:"secrets"`
 	Sentry  log.SentryConfig `yaml:"setry"`
 	Tracing tracing.Config   `yaml:"tracing"`
 }
 
+// Args is the argument struct for baseplate.New
+type Args struct {
```
konradreiche, line 49: Maybe just have `New` accept three arguments, context, path, and the config. I feel like this wrapping makes it harder to follow, lookup what needs to be done, etc. Unless we already know that there will be more things to add.

fishy: Since go doesn't allow default values to args, using struct is a way to support that. Using 3 args directly means that even if the user don't have any additional service confg, they always need to explicitly pass `nil` into `New()`, while using `Args` struct means that it can be omitted completely.

pacejackson: I'm not opposed to that, I mostly did this as a way to make `ServiceCfg` optional but that's not a huge benefit for this.

runtimebp/config.go
```diff
@@ -0,0 +1,38 @@
...
+		// Max controls the minimum number of Go processes.
+		//
+		// Defaults to 1 if not set.
+		Min int `yaml:"min"`
+	} `yaml:"numProcesses"`
+}
+
+// InitFromConfig sets GOMAXPROCS using the given config.
+func InitFromConfig(cfg Config) {
+	max := 50
```
pacejackson, line 28: I'm not sure if this _actually_ makes sense as a default 😅

fishy: I think Amazon EC2 allows maximum 64 cpu, so this might be low for some cases but should be fine for a default value :)

pacejackson:
I have a commit now without the `Args` struct, I don't feel _super_ strongly about leaving it in since I think _most_ services will have some sort of service-specific configuration to parse along with the baseplate level config but I also don't feel all that strongly about the reverse.

runtimebp/config.go
pacejackson: 64 also seams reasonable

fishy: We are currently using cpu*2-1 as the formula, so if they configure a very heavy service using all 64 cpu that would be 127 😅

pacejackson: This is a ceiling though IIRC? the actual value is based on the CPUs available right?

fishy: Yea this is the ceiling. The main purpose is to prevent some bug gives you some crazily high value.

PR #173, opened 2020-05-05
Add more comments to 'baseplate.Serve'

pacejackson:
The intention here is to make it clear what this method does, it isn't doing anything particularly tricky but it can be a little confusing, especially if you are new to Go.

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -102,8 +102,10 @@ func Serve(ctx context.Context, server Server) error {
 			// for the server to close.
 			ctx := context.Background()
 
-			// Check if the server has a StopTimeout configured, if it does, use that.
+			// Check if the server has a StopTimeout configured.
 			//
+			// If one is set, we will only wait for that duration for the server to
+			// stop gracefully and will exit after the deadline is exceeded.
 			// If one is not set, we will wait indefinetly for the server to stop.
 			timeout := server.Baseplate().Config().StopTimeout
 			if timeout > 0 {
--- a/baseplate.go
+++ b/baseplate.go
@@ -86,12 +86,15 @@ type Server interface {
 // This is the recommended way to run a Baseplate Server rather than calling
 // server.Start/Stop directly.
 func Serve(ctx context.Context, server Server) error {
-	// Initialize a channel to pass the response from server.Close() back to Serve
-	// in order to return it to the caller of Serve.
+	// Initialize a channel to return the response from server.Close() as our
+	// return value.
 	shutdownChannel := make(chan error)
 
-	// Listen for a shutdown command.  This is a blocking call so it is in a
-	// separate goroutine.
+	// Listen for a shutdown command.
+	//
+	// This is a blocking call so it is in a separate goroutine.  It will exit
+	// either when it is triggered via a shutdown command or if the context passed
+	// in is cancelled.
 	go runtimebp.HandleShutdown(
 		ctx,
 		func(signal os.Signal) {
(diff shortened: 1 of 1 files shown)


baseplate.go
```diff
@@ -86,29 +86,57 @@ type Server interface {
...
 		ctx,
 		func(signal os.Signal) {
-			timeout := server.Baseplate().Config().StopTimeout
+			// Initialize a context to potentially control the length of time we wait
+			// for the server to close.
 			ctx := context.Background()
+
+			// Check if the server has a StopTimeout configured, if it does, use that.
+			//
+			// If one is not set, we will wait indefinetly for the server to stop.
```
suzie-su, line 107: maybe we can say set the context with timeout which indicate after this period of time then we close the server

baseplate.go
```diff
@@ -86,29 +86,57 @@ type Server interface {
...
+			timeout := server.Baseplate().Config().StopTimeout
 			if timeout > 0 {
+				// Declare cancel in advance so we can just use `=` when calling
+				// context.WithTimeout.  If we used `:=` it would bind `ctx` to the scope
+				// of this `if` statement rather than updating the value declared before.
```
suzie-su, line 112: it might be better to say that after the timeout reached what happens in here 
like the cancel will execute and close the done channel 
oh no  this `cancel()` in here just close out everything if `<-ctx.Done()` did not gets called .
if timeout, the `cancal` will called internal and unblock  execute here `<-ctx.Done()`
is it correct?

baseplate.go
```diff
@@ -86,29 +86,57 @@ type Server interface {
...
+			// the result of the switch statement below.
+			var err error
+
+			// Wait for either the context to timeout or server.Close() to return.
 			select {
 			case <-ctx.Done():
+				// The context timed-out or was cancelled so use that error.
 				err = ctx.Err()
 			case e := <-closeChannel:
+				// server.Close() completed and passed it's result to closeChannel, so
```
suzie-su, line 138: does the io.closer indicate the server has been finished processing?

pacejackson: Yes, `Close()` should only shut down the server "gracefully", so it should allow any processing requests to finish returning, both the http server and thrift servers do this.

baseplate.go
pacejackson: Close, the results of you are right that both `cancel` and timing out send a value through the `ctx.Done()` channel, but the value is different for when `cancel()` is called than if it's deadline is exceeded:

https://golang.org/pkg/context/#pkg-variables

Note that when the timeout is reached `cancel()` is not called, `defer cancel()` is so the context is cancelled after the function returns which would signal to anything that was using that context but was still hanging around that it doesn't need to do any more work.

pacejackson: > unblock execute here <-ctx.Done() is it correct?

Ya, essentially the `select` statement is waiting for _one_ of the channels to return a value and whichever one returns first, that is the `case` that gets executed.  If it's the channel from `ctx.Done()` that means the deadline we set was exceeded, if it's `closeChannel`, that means that `Close()` finished.

fishy: the implementation of context is actually that nothing gets send to the `ctx.Done()` channel, it's just closed (which also unblocks any read from that channel).

`defer cancel()` is more to prevent memory leak, so that even if you exit this goroutine via other means (like your other channel gets unblocked first), you are still making sure that the resources associated with this context object is released. If you don't do that the linter will yell at you :)

pacejackson: Oh right, that's why you use `ctx.Err()` to get the value :)

PR #175, opened 2020-05-06
redisbp: add config.go

pacejackson:
This allows us to parse redis client configurations from a YAML file and get the `Options`/`ClusterOptions` objects that you use to create the actual redis clients.

### DIFF
--- a/httpbp/BUILD.bazel
+++ b/httpbp/BUILD.bazel
@@ -43,6 +43,7 @@ go_test(
         "//edgecontext:go_default_library",
         "//log:go_default_library",
         "//mqsend:go_default_library",
+        "//redisbp:go_default_library",
         "//secrets:go_default_library",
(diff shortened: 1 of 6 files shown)


pacejackson:
Sorry for the re-review, I made some final tweaks:

1. Remove the `Backoff` sub-field from `RetryOptions` and instead just name the 2 fields `MinBackoff` and `MaxBackoff`.  The sub-field was cute, but I think it's probably better to minimize YAML testing.
2. Added some examples, both a code example and example configs.
3. Added `OptionsMust`, similar to `templates.Must` which lets you just panic if `ClientConfig.Options` returns an error.  This is just a little bit of sugar that you can use when you're setting this up at server startup so it just fails if your config is bad.  This only applies to `ClientConfig.Options` because `ClusterConfig.Options` doesn't return an error.

PR #178, opened 2020-05-06
metricsbp: Add total counter to span hooks

fishy:
We only need 2 of fail, success, and total. Before this change we use
fail and success, and eventually I think we would want to use fail and
total instead, but keep success for now for backward compatibility.

The reason success and total won't work is that there's no way to
guarantee that the success and total counter gets reported atomically,
so it would lead to situations that success reported a slightly higher
rate than total, which screws up the rate calculating. Although
theoretically this is also an issue with fail+total (or fail+success),
they are much more tolerable than success+total.

### DIFF
--- a/metricsbp/baseplate_hooks.go
+++ b/metricsbp/baseplate_hooks.go
@@ -9,6 +9,7 @@ import (
 const (
 	success = "success"
 	fail    = "fail"
+	total   = "total"
 )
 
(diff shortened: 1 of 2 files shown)


fishy:
cc @spladug

fishy:
Yea the main concern is around sampling. The experience is that we usually do need to sample on timers/histograms, and do not need to sample on counters.

PR #180, opened 2020-05-07
baseplate: default StopTimeout to 30 seconds

pacejackson:
You can still use negative numbers to indicate that you should wait indefinitely, but we don't want that to be the "default".

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -33,7 +33,7 @@ type Config struct {
 
 	// StopTimeout is the timeout for the Stop command for the service.
 	//
-	// If this is not set, then a default value of 30 seconds will be used.
+	// If this is not set, then a default value of 10 seconds will be used.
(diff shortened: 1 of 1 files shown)


pacejackson:
It's more that we want the "default" behavior to be to timeout after a little bit, since it should really take only a few seconds at most to shut-down, so rather than forcing every service to set this value, we pick a sane default and use that if you don't set it.

This is what we already do in baseplate.py:

https://github.com/reddit/baseplate.py/blob/887f9be99f367be27d0e58fecdb274cb090a1585/baseplate/server/wsgi.py#L29-L31

I waffled between picking the same time as baseplate.py (10 seconds) and a longer one (30 seconds).  I _think_ 30 might make a little more sense in Go since it's a little more likely that people will just spawn off async processes in a Go service than in a Python one.

PR #183, opened 2020-05-13
Implement TClient with ClientPool

pacejackson:
The big change in this PR is updating `thriftbp.ClientPool` to implement `TClient`.  It pulls a client from it's pool to make the actual `Call` and releases is after the call finishes.  This allows you to pass a `ClientPool` in to a thrift client as the `TClient` and avoid the boilerplate of getting a client from a pool and releasing it before each call.

This PR also allows us to trim down the exported interface in `thriftbp`.

1. Don't export `ttlClient`.
2. Update `ttlClient` to allow you to keep a client in a pool indefinetly and configure it using the `ClientPoolConfig`.
3. Remove all the Factory methods that were used to create a ClientPool.
4. Don't export `getClient` or `releaseClient`.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -3,6 +3,7 @@ package thriftbp
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"time"
 
@@ -17,6 +18,26 @@ import (
 // DefaultPoolGaugeInterval is the fallback value to be used when
 // ClientPoolConfig.PoolGaugeInterval <= 0.
 const DefaultPoolGaugeInterval = time.Second * 10
+
+// PoolError is returned by ClientPool.Call when it fails to get a client from
+// its pool.
+type PoolError struct {
+	// Cause is the inner error wrapped by PoolError.
+	Cause error
+}
+
+func (err PoolError) Error() string {
+	return "thriftpb: error getting a client from the pool. " + err.Cause.Error()
+}
+
+func (err PoolError) Unwrap() error {
+	return err.Cause
+}
+
+var (
+	_ error = PoolError{}
+	_ error = (*PoolError)(nil)
+)
 
 // ClientPoolConfig is the configuration struct for creating a new ClientPool.
 type ClientPoolConfig struct {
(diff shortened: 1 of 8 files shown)


thriftbp/client_pool.go
```diff
@@ -106,139 +113,69 @@ type ClientPool interface {
...
+				cfg.MaxConnectionAge,
+				genAddr,
+				proto,
+				middlewares...,
+			)
 		},
 	)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("thriftbp: error initializing thrift clientpool. %w", err)
```
fishy, line 178: 🔕 I don't feel strongly about it but I think this error (and the ones in `newClient`) makes less sense to wrap. The context ones are generic that can happen everywhere so wrapping makes it easier to figure out where do they come from, but here the underlying error comes directly from `clientpool` (either `ErrExhausted`, `ConfigError` or network error from the underlying openers). But I guess it won't hurt to wrap them.

thriftbp/client_pool.go
```diff
@@ -303,7 +255,17 @@ type clientPool struct {
 	releaseErrorCounter  metrics.Counter
 }
 
-func (p *clientPool) GetClient() (Client, error) {
+func (p *clientPool) Call(ctx context.Context, method string, args, result thrift.TStruct) error {
+	client, err := p.getClient()
+	if err != nil {
+		return fmt.Errorf("thriftbp: error retrieving thrift client from pool. %w", err)
```
fishy, line 261: I think any error returned by `Call` will be wrapped by thrift library code again, so here we might want:

1. Define this as a type to be inspectable (but I also need to make an upstream change to make sure the thrift wrapped errors has `Unwrap` implemented, currently they don't)
2. Since we also do error logging in `releaseClient`, we probably should do error logging here as well.

pacejackson: `%w` makes it unwrappable :)

fishy: yes but it won't work if it's wrapped again (by `TApplicationException` or `TTransportException`) and that wrapping is not unwrappable.

pacejackson: Fair, it's also _probably_ fine to just not do this.

thriftbp/client_pool.go
fishy: To clarify you can totally ignore that comment, I was just thinking out loud. I don't think there's any real downside of wrapping them.

thriftbp/client_pool.go
fishy: Also what I meant is we want to make it easy for the service code to distinguish between errors caused by unable to get a client from the pool, and errors caused by the actual call itself.

fishy: Just checked thrift code and error returned by `Call` is returned to the caller directly without further wrapping, so defining `PoolError` is sufficient.

PR #184, opened 2020-05-14
thriftbp: move mocks/test helpers into 'thriftbp/thrifttest' package

pacejackson:


thriftbp/thrifttest/mocks.go
```diff
@@ -272,7 +273,7 @@ func (m MockClientPool) getClient() (Client, error) {
 //     myMockClient.AddMockCall(
 //       "myEndpoint",
 //       func(ctx context.Context, args, result thrift.TStruct) error {
-//         return thriftbp.CopyTStruct(
+//         return thrifttest.CopyTStruct(
```
fishy, line 276: I just realized that the go convention is actually in the order of `to, from` instead of `from, to` we used here, see: https://golang.org/pkg/builtin/#copy, https://pkg.go.dev/io?tab=doc#Copy, etc.

Can you flip the order of `from, to` in the function signature (and also fix all the usages in this repo) in this PR? It's already a breaking change anyways 😅

fishy: And maybe also rename them to `dst, src` from `to, from` 😅

PR #185, opened 2020-05-14
thriftbp: Add more doc regarding ClientPool.Call

fishy:


thriftbp/client_pool.go
```diff
@@ -118,6 +118,22 @@ type Client interface {
...
+	//       // It's unable to get a client from the pool
+	//     } else {
+	//       // It's error from the actual thrift call
+	//     }
+	//
+	// If the error is not of type PoolError that means it's returned by the
+	// Call from the actual client.
+	//
+	// If Call fails to release the client back to the pool,
+	// it will log the error on error level but not return it to the caller.
```
pacejackson, line 136: I think it also increments a metric counter, want to include that as well?

fishy: Added.

PR #186, opened 2020-05-14
log: Allow attaching trace id into loggers

fishy:
This is a breaking change, both on how service code should be using the
logger, and on Span.InjectSentryHub has been renamed into
Span.InjectTraceContext.

### DIFF
--- a/log/context.go
+++ b/log/context.go
@@ -1,0 +1,62 @@
+package log
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+)
+
+type contextKeyType struct{}
+
+var contextKey contextKeyType
+
+// logger keys for attached data.
+const (
+	traceIDKey = "trace_id"
+)
+
+// AttachArgs are used to create loggers to be attached to context object with
+// pre-filled key-value pairs.
+//
+// All zero value fields will be ignored and only non-zero values will be
+// attached.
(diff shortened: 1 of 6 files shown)


log/context.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+)
+
+type contextKeyType struct{}
+
+var contextKey contextKeyType
```
pacejackson, line 11: OOC, why use a struct rather than just a typed int?

fishy: an empty `struct{}` consumes 0 bytes of memory (plus type info), an `int` is 8 bytes (plus type info) :)

log/context.go
```diff
@@ -0,0 +1,62 @@
...
+// the global logger if none is found.
+//
+// When you have a context object and want to do logging,
+// you should always use this one instead of the global one.
+// For example:
+//
+//     log.C(ctx).Errorw("Something went wrong!", "err", err)
+//
+// The return value is guaranteed to be non-nil.
+func C(ctx context.Context) *zap.SugaredLogger {
```
konradreiche, line 57: What is the meaning of `C` here, `Context`?

fishy: Yes 😅

PR #187, opened 2020-05-15
Initialize a new, top-level server span if no headers are set

pacejackson:


tracing/span.go
```diff
@@ -443,8 +457,109 @@ type Headers struct {
...
+}
+
+// TraceIDUint attempts to convert h.TraceID into a uint64, if it succeeds it
+// returns the value and 'true'.  If it fails, either because h.TraceID is not
+// set or it is malformed, ok will be 'false' and you should not rely on the ID
+// returned.
+//
+// If h.TraceID was malformed, an error will be logged using the global tracer's
+// logger but no error will be returned.
+func (h Headers) TraceIDUint() (id uint64, ok bool) {
```
fishy, line 475: nit: Adding the type suffix to the function name feels unnecessary. I guess you want to avoid naming conflicting from the exported fields, in that case those functions can be called `Get*` instead?

tracing/span.go
```diff
@@ -443,8 +457,109 @@ type Headers struct {
...
+		return false, false
+	}
+	return *h.Sampled, true
+}
+
 // StartSpanFromHeaders creates a server span from the passed in Headers.
 //
+// This should be used when you are initializing a server span *from headers*.
+// If you do not have headers to initialize your span from, you should use
+// StartTopLevelServerSpan instead.
```
fishy, line 561: nit: Since we already handled that in this function (the added first `if`), this should be reworded?

tracing/span.go
pacejackson: Fair, maybe `Parse*` instead?

fishy: yea that works

PR #189, opened 2020-05-15
thriftbp: Avoid reusing bad client in ClientPool.Call

fishy:
When releasing the client back to the pool, check the error returned by
Call, and close the client to avoid reusing if the error is caused by
network error.


fishy:
:haircut: Forgot that we need https://github.com/apache/thrift/commit/be3f7321cf0b7cff5d178ac09da02fc68dac6bd5 to do it, otherwise it will be wrapped by TTransportException and that's not unwrappable.

PR #190, opened 2020-05-16
Ignore redis nil errors

vkuptcov:
Baseplate adds a pipeline hook to batch Redis pipeline errors.
The problem is that Redis returns `redis.Nil` error in some cases, e.g. key absence.
These errors should be ignored.
Also, the fix doesn't return errors from the hook in order not to change the default go-redis client error.

### DIFF
--- a/redisbp/hooks.go
+++ b/redisbp/hooks.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
 	"github.com/go-redis/redis/v7"
 	opentracing "github.com/opentracing/opentracing-go"
 
@@ -69,7 +70,7 @@ func (h SpanHook) startChildSpan(ctx context.Context, cmdName string) context.Co
 	return ctx
 }
 
-func (h SpanHook) endChildSpan(ctx context.Context, err error)  {
+func (h SpanHook) endChildSpan(ctx context.Context, err error) {
 	if span := opentracing.SpanFromContext(ctx); span != nil {
 		span.FinishWithOptions(tracing.FinishOptions{
 			Ctx: ctx,
--- a/redisbp/hooks.go
+++ b/redisbp/hooks.go
@@ -2,6 +2,7 @@ package redisbp
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


redisbp/hooks.go
```diff
@@ -44,7 +44,9 @@ func (h SpanHook) BeforeProcessPipeline(ctx context.Context, cmds []redis.Cmder)
 func (h SpanHook) AfterProcessPipeline(ctx context.Context, cmds []redis.Cmder) error {
 	var errs batcherror.BatchError
 	for _, cmd := range cmds {
-		errs.Add(cmd.Err())
+		if cmd.Err() != redis.Nil {
```
pacejackson, line 47: Use `errors.Is` rather than a direct comparison:

```suggestion
		if !errors.Is(cmd.Err(), redis.Nil) {
```

vkuptcov: done

redisbp/hooks.go
```diff
@@ -49,7 +48,10 @@ func (h SpanHook) AfterProcessPipeline(ctx context.Context, cmds []redis.Cmder)
 			errs.Add(cmd.Err())
 		}
 	}
-	return h.endChildSpan(ctx, errs.Compile())
+	_ = h.endChildSpan(ctx, errs.Compile())
```
fishy, line 51: nit: you can remove the `_ =` part.

redisbp/hooks.go
```diff
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
```
fishy, line 6: And don't remove this blank line. See: https://github.com/reddit/baseplate.go/blob/master/Style.md#import-groups

redisbp/hooks.go
fishy: Also I think at this point there's no reason for `endChildSpan` to return an error, so that function can be changed to return nothing.

pacejackson: Ya, I think `endChildSpan` should return nothing and we just return `nil` for the error in each `After` function?

vkuptcov: Yeah, I also changed another part where it was used.

PR #191, opened 2020-05-16
thrifttest: add 'NewBaseplateServer'

pacejackson:
This is inspired by httptest.NewServer and allows you to create a new Thrift server listening using the local loopback interface as well as a ClientPool for creating a client to talk to that server.
    
This can be used to test your thrift service in automated testing.
    
It _can_ also be used to create fake services for testing your integration, although you might also be served by using the Mocks available in 'thrifttest' for that.

### DIFF
--- a/thriftbp/thrifttest/server.go
+++ b/thriftbp/thrifttest/server.go
@@ -1,0 +1,141 @@
+package thrifttest
+
+import (
+	"context"
+	"time"
+
+	"github.com/apache/thrift/lib/go/thrift"
+
+	baseplate "github.com/reddit/baseplate.go"
+	"github.com/reddit/baseplate.go/batcherror"
+	"github.com/reddit/baseplate.go/secrets"
+	"github.com/reddit/baseplate.go/thriftbp"
+)
+
+const loopbackAddr = "127.0.0.1:0"
+
+// ServerConfig can be used to pass in custom configuration options for the
+// server and/or client created by NewBaseplateServer.
+type ServerConfig struct {
+	// ServerConfig is an optional value, sane defaults will be chosen where
+	// appropriate.
+	//
+	// ServerConfig.Socket will always be replaced with one created in
+	// NewBaseplateServer using the local loopback address.
+	ServerConfig baseplate.Config
+
+	// ClientConfig is an optional value, sane defaults will be chosen where
+	// appropriate.
+	//
+	// Addr will always be set to the address of the test server.
+	// ReportPoolStats will always be set to false.
+	// InitialConnections will always be set to 0 since the test server will not
+	// be started yet.
+	// MaxConnections will be 10 if it is not set.
+	ClientConfig thriftbp.ClientPoolConfig
+
+	// Optional, additional ClientMiddleware to wrap the client with.
+	ClientMiddlewares []thrift.ClientMiddleware
+
+	// Optional, additional ProcessorMiddleware to wrap the server with.
+	ProcessorMiddlewares []thrift.ProcessorMiddleware
+}
+
+// Server is a test server returned by NewBaseplateServer.  It contains both
+// the baseplate.Server and a ClientPool to use to interact with the server.
+//
+// Server implements baseplate.Server.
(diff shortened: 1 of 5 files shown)


thriftbp/thrifttest/server.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// ClientPool provides a thriftbp.ClientPool that connects to this Server and
+	// can be used for making Thrift client objects to interact with this Server.
+	ClientPool thriftbp.ClientPool
+}
+
+// Start starts the server using baseplate.Serve in a background goroutine and
+// waits for a short period of time to give the server time to start up.
+//
+// The server can be closed manually, with shutdown commands, or by cancelling
```
fishy, line 59: s/shutdown/close/?

thriftbp/thrifttest/server.go
```diff
@@ -0,0 +1,141 @@
...
+// local loopback interface and a Baseplate ClientPool for use with that server.
+//
+// This is inspired by httptest.NewServer from the go standard library and can
+// be used to test a thrift service.
+//
+// "cfg" may be nil, if it is, sane defaults will be chosen.
+// The server and pool that are returned should be closed when done, but the
+// Baseplate used by the server does not need to be.
+func NewBaseplateServer(
+	store *secrets.Store,
```
fishy, line 90: 🔕 it's a bit annoying that we need to mock a secret store in order to use the mock server, but I don't see a better way :(

thriftbp/thrifttest/server.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return nil, err
+	}
+	server := &Server{Server: thriftbp.ApplyBaseplate(bp, srv)}
+
+	cfg.ClientConfig.Addr = server.Baseplate().Config().Addr
+	cfg.ClientConfig.ReportPoolStats = false
+	cfg.ClientConfig.InitialConnections = 0
+	if cfg.ClientConfig.SocketTimeout == 0 {
+		cfg.ClientConfig.SocketTimeout = 10 * time.Millisecond
```
fishy, line 126: nit: constantize those magic numbers? (this one and `10` for `MaxConnections` below)

thriftbp/thrifttest/server.go
```diff
@@ -0,0 +1,211 @@
...
+//		}
+//
+//		if !success {
+// 			t.Errorf("result mismatch, expected %v, got %v", c.expected.result, result)
+//		}
+//	}
+func NewBaseplateServer(
+	store *secrets.Store,
+	processor thrift.TProcessor,
+	cfg *ServerConfig,
```
fishy, line 161: I don't feel strongly about it, but from the code we can actually handle empty `ServerConfig{}` just fine without using pointer, and that's more consistent with the other config structs we used across baseplate.go.

thriftbp/thrifttest/server.go
```diff
@@ -0,0 +1,211 @@
...
+	socket, err := thrift.NewTServerSocket(loopbackAddr)
+	if err != nil {
+		return nil, err
+	}
+	// Call listen to reserve a port and check for any issues early
+	if err := socket.Listen(); err != nil {
+		return nil, err
+	}
+
+	cfg.ServerConfig.Addr = socket.Addr().String()
```
fishy, line 176: one side effect of using pointer for `cfg` is that if the passed in cfg is non-nil, then these kind of changes will be reflected back to the caller (e.g. the values in `cfg` will change after calling `NewBaseplateServer`), which in most cases won't be an issue but it's something to be mindful about.

pacejackson:
I updated this to use `batchcloser` and `NewTestSecrets`.

pacejackson:
These two commits:

https://github.com/reddit/baseplate.go/pull/191/commits/a77cbe7b7140daeb22003c09b6973db8f13e78ee

https://github.com/reddit/baseplate.go/pull/191/commits/1e272f68ac3a403171668d9519aff31556e3a131

PR #197, opened 2020-05-16
Upgrade build/CI system

fishy:
Upgrade:
- go to 1.14.3
- bazel to 3.1.0
- rules_go to 0.23.1
- gazelle to 0.21.0

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,1 +1,1 @@
-test --features=race --test_output=errors
+test --@io_bazel_rules_go//go/config:race --test_output=errors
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -1 +1 @@
-test --features=race --test_output=errors
+test --@io_bazel_rules_go//go/config:race --test_output=errors
```
fishy, line 1: NOTE: this change is optional but it's introduced in [rules_go v0.23.0](https://github.com/bazelbuild/rules_go/releases/tag/v0.23.0) and likely will be the future.

PR #198, opened 2020-05-17
secrets: add 'NewTestSecrets' to simplify tests

pacejackson:


secrets/testing.go
```diff
@@ -0,0 +1,94 @@
...
+
+	testVaultURL   = "vault.reddit.ue1.snooguts.net"
+	testVaultToken = "17213328-36d4-11e7-8459-525400f56d04"
+)
+
+// JWTPubKeySecret is the default EdgeRequestContext public key secret.
+//
+// pubkey copied from:
+// https://github.com/reddit/baseplate.py/blob/db9c1d7cddb1cb242546349e821cad0b0cbd6fce/tests/__init__.py#L12
+var JWTPubKeySecret = GenericSecret{
```
fishy, line 24: Since this is still an exported variable in `secrets` package, the name will likely cause confusion and we should make it more obvious that this is a test pub key, not a real thing. Maybe rename it `TestJWTPubKeySecret` or something like that?

secrets/testing.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+const (
+	// JWTPubKeyPath is the expected key for the EdgeRequestContext public
+	// key.
+	JWTPubKeyPath = "secret/authentication/public-key"
+
+	testVaultURL   = "vault.reddit.ue1.snooguts.net"
+	testVaultToken = "17213328-36d4-11e7-8459-525400f56d04"
+)
```
fishy, line 18: nit: I would prefer to separate exported and unexported consts into different const blocks.

secrets/testing.go
```diff
@@ -0,0 +1,94 @@
...
+	store.watcher = watcher
+	return store, watcher, nil
+}
+
+// UpdateTestSecrets replaces the secrets returned by the MockFileWatcher with the
+// the given raw secrets.
+//
+// Like NewTestSecrets, if you do not provide a value for the key defined by
+// JWTPubKeyPath, then we will add a default secret for you.
+func UpdateTestSecrets(fw *filewatcher.MockFileWatcher, raw map[string]GenericSecret) error {
```
fishy, line 83: 🔕 So this function is to replace the whole secret store, not only a single key, right? I think maybe provide an API to only replace a single key might be more useful? But that can be a future change instead.

pacejackson: Hmm, just replacing a single key might be more useful than this?  I can just change it?

filewatcher/filewatcher.go
```diff
@@ -210,3 +230,45 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
 
 	return res, nil
 }
+
+// NewMockFilewatcher returns a pointer to a new MockFileWatcher object
+// initialized with the given io.Reader and Parser.
+func NewMockFilewatcher(r io.Reader, parser Parser) (*MockFileWatcher, error) {
```
konradreiche, line 236: Any reason why this should be in `filewatcher` rather than `filewatcher_test`?

fishy: if it's in `filewatcher_test` it can only be used by tests under `filewatcher`, the intention here is to make it usable by other packages using filewatcher.

secrets/testing.go
fishy: yea but it involves slightly more work (you need to store the whole map in the struct in order to do single key update), so I'm fine if you punt it for future work :)

filewatcher/filewatcher.go
```diff
@@ -255,17 +262,31 @@ type MockFileWatcher struct {
...
 
+// Raw returns the raw, un-parsed data.
+func (fw *MockFileWatcher) Raw() []byte {
+	return fw.data.Load().(mockData).raw
+}
+
 // Get returns the parsed data.
 func (fw *MockFileWatcher) Get() interface{} {
-	return fw.data.Load()
+	return fw.data.Load().(mockData).parsed
```
fishy, line 289: this will panic if `Get` is called before all `Update` calls.

Is the reason to store the struct to support single key replacement in secrets mock? that can be done by storing the whole map in mock secret struct, instead of using the filewatcher.

pacejackson: > this will panic if Get is called before all Update calls.

`¯\_(ツ)_/¯ ` This is for test code and should be created using `NewMockFileWatcher`, if you don't do that and it panics that's on you IMO.

> Is the reason to store the struct to support single key replacement in secrets mock?

That's what _prompted_ it but it's to support incremental updates to the data stored in the mock file watcher in general.

pacejackson: > that can be done by storing the whole map in mock secret struct, instead of using the filewatcher.

Ya, although we'd need some other changes, (like make `SecretsStore` an interface and using that, not a crazy thing to do though).

pacejackson: I'll just take this out for now, we can do this in a separate PR :)

filewatcher/filewatcher.go
konradreiche: That makes sense. Which package would also use it?

fishy: currently `experiments`, but any package using filewatcher could benefit from it.

konradreiche: `experiments` is switching to use the interface, is this for testing purposes only? It seems the interface is added only to allow for supplying a mock implementation in tests but not another implementation for non-test code.

fishy: Oh sorry, I meant `secrets`, not `experiments`.

konradreiche: I can't find it anywhere besides tests, speaking of `NewMockFilewatcher` here.

fishy: it's in `secrets/testing.go`.

pacejackson: It's not going to be used anywhere besides tests likely, but we can't put it in `filewatcher_tests` because `_test` packages aren't usable externally.  The reason for including these things is to make it easier to write tests for code involving file watchers.

Right now, only `secrets` uses it, but `experiments` could too as well as any custom code using file watchers in services.

PR #203, opened 2020-05-21
thriftbp: wrap the ClientPool Call method

pacejackson:
This is mainly so we apply middleware before getting a client from the pool, this lets us do things like use a new client if we trigger a retry due to a bad client in middleware.  It also has the side effect of only calling Wrap once rather than every time a new Client object is built.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -295,9 +295,18 @@ type clientPool struct {
 }
 
 func (p *clientPool) Call(ctx context.Context, method string, args, result thrift.TStruct) (err error) {
+	// A clientPool needs to be set up properly before it can be used,
+	// specifically, ouse p.wrapCalls to set up p.wrappedCall before using it.
+	//
+	// newClientPool already takes care of this for you.
 	return p.wrappedCall.Call(ctx, method, args, result)
 }
 
+// wrapCalls wraps p.pooledCall in the given middleware and sets p.wrappedCall
+// to the resulting thrift.TClient.
+//
+// This must be called before the clientPool can be used, but newClientPool
+// already takes care of this for you.
 func (p *clientPool) wrapCalls(middlewares ...thrift.ClientMiddleware) {
 	p.wrappedCall = thrift.WrapClient(thrift.WrappedTClient{
 		Wrapped: func(ctx context.Context, method string, args, result thrift.TStruct) error {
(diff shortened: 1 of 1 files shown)


thriftbp/client_pool.go
```diff
@@ -295,9 +295,18 @@ type clientPool struct {
 }
 
 func (p *clientPool) Call(ctx context.Context, method string, args, result thrift.TStruct) (err error) {
+	// A clientPool needs to be set up properly before it can be used,
+	// specifically, ouse p.wrapCalls to set up p.wrappedCall before using it.
```
fishy, line 299: typo: "ouse"

thriftbp/client_pool.go
```diff
@@ -291,9 +290,37 @@ type clientPool struct {
 
 	poolExhaustedCounter metrics.Counter
 	releaseErrorCounter  metrics.Counter
+
+	wrappedCall thrift.TClient
```
fishy, line 294: maybe rename to `wrappedClient`?

thriftbp/client_pool.go
```diff
@@ -291,9 +290,37 @@ type clientPool struct {
...
+
+	wrappedClient thrift.TClient
 }
 
 func (p *clientPool) Call(ctx context.Context, method string, args, result thrift.TStruct) (err error) {
+	// A clientPool needs to be set up properly before it can be used,
+	// specifically use p.wrapCalls to set up p.wrappedClient before using it.
+	//
+	// newClientPool already takes care of this for you.
+	return p.wrappedClient.Call(ctx, method, args, result)
```
fishy, line 308: It _might_ make sense to be more defensive here and do a nil check on wrappedClient, fallback to use pooledCall directly if that's nil, but since this is an unexported type I don't feel strongly that's needed.

pacejackson: Could also use a mutex or something to set it up lazily by just keeping track of the middleware?

pacejackson: I'm ok with the fallback though, that seems reasonable.

fishy: That's definitely overkill. the appropriate thing for lazy init is `sync.Once`, but if a few calls happen at the same time before it's initialized, they all need to wait for the initialize to finish. consider we usually initialize pools in main function shift the initialization delay there sounds more efficient.

The only thing defensive coding here prevents is someone not initialize the pool properly. Since it's unexported, that could only happen inside this package, which is fine.

pacejackson: Ok, I think it's probably fine to leave this as-is?

fishy: yes :)

PR #210, opened 2020-05-26
thriftbp: include the client service name in the span name

pacejackson:
Right now we are not including this so the metrics just look like `client.${method}` when they should be `client.${service}.${method}`


pacejackson:
Note: this is technically a breaking change although services using `NewBaseplateClient` shouldn't actually require any changes.

PR #211, opened 2020-05-28
Add retrybp package

pacejackson:


retrybp/retry.go
```diff
@@ -0,0 +1,60 @@
...
+	retry "github.com/avast/retry-go"
+
+	"github.com/reddit/baseplate.go/batcherror"
+)
+
+func init() {
+	retry.DefaultAttempts = uint(1)
+	retry.DefaultDelay = 5 * time.Millisecond
+	retry.DefaultMaxJitter = 25 * time.Millisecond
+}
```
fishy, line 17: 🔕 I have a few concerns with this approach, but I don't know if there's a better way to do it:

1. Those values might still be too high for some latency sensitive services, so I could see some devs still want to tweak them in their code, but as long as they don't use `init` (the order of `init` functions from different packages are not guaranteed) to do that that shouldn't be a big problem.
2. These values here are more specific for this version of upstream we are using here, but service could upgrade them to a newer version, which might have changes to the default values, and _might_ change the meaning of some of those default values, and cause problems. With go's no breaking change policy those shouldn't happen, but there's no guarantee of it.

One alternative is to have an exported function, something like `OverrideDefaultValues`, so service code could do the overriding explicitly. But I do see one disadvantage of that approach, is they might forget to do that and end up with the higher defaults (which using `init` avoids).

retrybp/retry.go
```diff
@@ -0,0 +1,60 @@
...
+func Do(ctx context.Context, fn func() error, defaults ...retry.Option) error {
+	options, _ := GetOptions(ctx)
+	options = append(defaults, options...)
+	err := retry.Do(fn, options...)
+
+	var retryErr retry.Error
+	if errors.As(err, &retryErr) {
+		var batchErr batcherror.BatchError
+		for _, err := range retryErr.WrappedErrors() {
+			batchErr.Add(err)
```
fishy, line 54: 🔕 Should we make the change in `batcherror` to make that a vararg in this PR? :)

thriftbp/client_middlewares.go
```diff
@@ -108,6 +110,24 @@ func SetDeadlineBudget(next thrift.TClient) thrift.TClient {
 	}
 }
 
+// Retry returns a thrift.ClientMiddleware that can be used to automatically
+// retry thrift requests.
```
fishy, line 114: nit: Probably worth mentioning in the doc comment that this is not enabled by default in baseplate and service code need to add it explicitly.

retrybp/retry.go
konradreiche: I wonder if it's worth attaching a state to those retry functions to ensure the values set are the ones that the applier wants to have.

thriftbp/client_middlewares.go
pacejackson: I was actually considering making it enabled by default, just with only 1 attempt >.>

Then you can still set per-call retry behavior if you want with the context function.

fishy: Does "1 attempt" mean no retry or one retry?

pacejackson: It means no retries: https://github.com/avast/retry-go/blob/a8f6dc7e8f46a5d11c01d9da1e887bb5a51fd107/retry.go#L113

fishy: and just to double check, if the one and only attempt failed, it won't do any delay and will return the error directly, right? in that case make retry middleware default with one attempt is fine.

pacejackson: Hmm, it looks like it doesn't do that, it will at least go _through_ the logic to see if it should "retry" and might do the delay if it does.

fishy: https://github.com/avast/retry-go/blob/a8f6dc7e8f46a5d11c01d9da1e887bb5a51fd107/retry.go#L125-L128

Looks like we should be good?

fishy: This should also be verifiable via a unit test. You just create a retry policy, then pointing the client to a non-existing host:port, make sure that error is retryable, set a large delay, and check whether the time of the call included the delay or not.

pacejackson: Oh great! My eyes totally glazed over that part, I'll update this and add that.

pacejackson:
I made the change to `BatchError` and made `Retry` (with no actual retries) a default middleware.

pacejackson:
Note, I don't feel strongly about making `Retry` an actual default option if others prefer to leave it out but I think it'll be useful.

fishy:
> Note, I don't feel strongly about making `Retry` an actual default option if others prefer to leave it out but I think it'll be useful.

As long as the overhead is acceptable, I prefer making it default. That means when we need to add retry, we can just do that by adding a context value.

pacejackson:
Ok, I also put up a commit of it _not_ being a default so we can look at that as well, I can just revert it if we decide to make it a default.

pacejackson:
(note I also pushed the revert since it _seems_ like having it be a default is the approach we want)

batcherror/batch.go
```diff
@@ -89,22 +89,28 @@ func (be *BatchError) addBatch(batch BatchError) {
...
 // its underlying error(s) will be added instead of the BatchError itself.
 //
-// Nil error will be skipped.
-func (be *BatchError) Add(err error) {
-	if err == nil {
+// Nil errors will be skipped.
+func (be *BatchError) Add(errs ...error) {
+	if errs == nil {
 		return
 	}
```
fishy, line 101: nit: This if is no longer needed (in that case the for loop will just be executed 0 times).

retrybp/filters.go
```diff
@@ -0,0 +1,160 @@
...
+	)
+}
+
+// FilterDecision is returned by a RetryFilter to communicate whether it has
+// decided whether or not a request should be retried.
+//
+// Generally, the default decision for a RetryFilter should be "NoDecision".  A
+// RetryFilter should check a single case to determine one of "DoSend" or
+// "DoNotSend" and should return "NoDecision" if that check fails.
+type FilterDecision struct {
```
fizx, line 47: Any particular reason for this Decision struct over the approach I've seen more typically, which is to have a FilterChain and each filter has the choice on whether to call the next one?

I guess I'm saying that a filter "linked-list" is more common choice than a filter "array".  The "linked-list" might provide a little more flexibility and not require secondary structs.

Don't hate me for quoting the Servlet API :) https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/Filter.html

retrybp/retry.go
```diff
@@ -0,0 +1,60 @@
...
+	return
+}
+
+// Do retries the given function using retry.Do with the default retry.Options
+// provided and overriding them with any options set on the context via
+// WithOptions.
+//
+// The changes this has compared to retry.Do are:
+//
+// 1. Pulling options from the context.  This allows it to be used in middleware
```
fizx, line 42: "Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions."

🔕 We might be stretching this limit.

pacejackson: > We might be stretching this limit.

Fair, although this is a fairly common way to stretch the limit, this is _really_ only for when you are using retry middleware though where you don't really have another option.

retrybp/filters.go
pacejackson: This is just an approach I'm familiar with, but the function chain makes sense as well.

I'm open to changing it unless others feel strongly about one way or the other.

@fishy / @konradreiche ?

pacejackson: I think both have their pros and cons, the chain approach seems a little more... elegant?

The nice thing about the current approach though is the return struct forces you to say "Done" vs "Not Done" while the chain approach means that any custom filters have to remember to call `chain.Next()`.

fishy: I like the chain design more than a struct with two bools :)

pacejackson: That works for me :)

OOC, I'm going to benchmark both and see what it looks like (probably pretty much the same).

pacejackson:
If anyone is curious about the chain approach vs the current:

```
Running tool: /usr/local/bin/go test -benchmem -run=^$ github.com/reddit/baseplate.go/retrybp -bench ^(BenchmarkChain)$

goos: darwin
goarch: amd64
pkg: github.com/reddit/baseplate.go/retrybp
BenchmarkChain/Filters/1-12         	1000000000	         0.000012 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/Filters/10-12        	1000000000	         0.000016 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/Filters/100-12       	1000000000	         0.000022 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/Filters/1000-12      	1000000000	         0.000035 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/Filters/10000-12     	1000000000	         0.000229 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/ChainedFilters/1-12  	1000000000	         0.000020 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/ChainedFilters/10-12 	1000000000	         0.000024 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/ChainedFilters/100-12         	1000000000	         0.000126 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/ChainedFilters/1000-12        	1000000000	         0.000713 ns/op	       0 B/op	       0 allocs/op
BenchmarkChain/ChainedFilters/10000-12       	1000000000	         0.00567 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/reddit/baseplate.go/retrybp	5.624s
```

... I think we'll be fine with the chain approach

pacejackson:
💇 

Using the "chain" approach proposed

retrybp/filters.go
```diff
@@ -0,0 +1,134 @@
...
+	retry "github.com/avast/retry-go"
+
+	"github.com/reddit/baseplate.go/clientpool"
+)
+
+var (
+	// DefaultFilterDecision is the default decision returned by Filters at the end
+	// of the FilterChain.
+	DefaultFilterDecision = false
+)
```
fishy, line 17: nit: Now this is a primitive type, it can be `const` instead of `var` now.

retrybp/filters.go
```diff
@@ -0,0 +1,134 @@
...
+// if the request should be retried given the error.
+//
+// RetryFilters should only implement a single check that will generally
+// determine whether the request should be retried or not.  If it cannot make
+// that decision on it's own, it should call the next filter in the given
+// FilterChain using chain.Next.
+//
+// If a filter is doing more than that, there's a good chance that it is doing
+// too much.
+type RetryFilter func(err error, chain *FilterChain) bool
```
fishy, line 70: Is there a reason we need different types between `RetryFilter` and `FilterChain`? why can't it just be:

```
type RetryFilter func(err error, next RetryFilter) bool
```

We just need a termination implementation that ignore next, like:

```
func finalFilter(_ error, _ RetryFilter) bool {
  return DefaultFilterDecision
}
```

right?

retrybp/filters.go
pacejackson: The point of making it `var` was so you _could_ change it if you wanted to.

retrybp/filters.go
pacejackson: That approach doesn't quite work because you need to pass the next next into next, the Chain exists to hold that so you don't need it in `RetryFilter`.

pacejackson: i.e.:

```
func noDecision(err error, next RetryFilter) bool {
  return next(err, ?????)
```

retrybp/filters.go
fishy: That can be achieved by just add another one to the chain, but fine :)

pacejackson: although, you can just terminate the filters with one that always returns true if you really want that... I'll change this.

retrybp/filters.go
pacejackson: nifty side benefit, `FilterChain.Next` is the same signature that `retry.RetryIf` wants so we can pass it right on into that.

retrybp/filters.go
```diff
@@ -0,0 +1,132 @@
...
+// FilterChain is an ordered chain of RetryFilters that is used by Filter to
+// determine if a function should be retried based on the error it returns.
+//
+// A FilterChain is passed to RetryFilter functions and if the filter cannot make
+// a decison about whether to retry or not, it should instead return chain.Next
+// so the next filter in the chain can check the error.
+type FilterChain struct {
+	filter RetryFilter
+	next   *FilterChain
+}
```
fishy, line 48: nit: We usually define middleware interfaces by using two function types instead of one function type and one struct, might be better to keep the style more consistent? (the chain is essentially a middleware 😅)

pacejackson: done

thriftbp/example_client_test.go
```diff
@@ -4,9 +4,12 @@ import (
 	"context"
 
 	"github.com/apache/thrift/lib/go/thrift"
+	retry "github.com/avast/retry-go"
 	opentracing "github.com/opentracing/opentracing-go"
+
```
fishy, line 9: nit: this is an exception in the style guide: https://github.com/reddit/baseplate.go/blob/master/Style.md#exceptions-1. we don't separate third party imports with baseplate.go imports in full file examples.

thriftbp/client_middlewares.go
```diff
@@ -14,6 +14,21 @@ import (
 	"github.com/reddit/baseplate.go/tracing"
 )
 
+// WithDefaultRetryFilters returns a list of retrybp.Filters by appending the
+// given filters to the "default" retry filters:
+//
+// 1. ContextErrorFilter - do not retry on context cancellation/timeout.
+// 2. UnrecoverableErrorFilter - do not retry errors marked as unrecoverable with
+//    retry.Unrecoverable.
+// 3. PoolExhaustedFilter - do retry on clientpool.PoolExhausted errors.
```
fishy, line 23: nit: Add blank lines between ordered list, or they will be rendered into the same line in godoc :(

You should be able to check locally via command:

```
go doc github.com/reddit/baseplate.go/thriftbp.WithDefaultRetryFilters
```

retrybp/filters.go
konradreiche: Same

retrybp/retry.go
konradreiche: I am not sure if this is actually a common way, other libraries might misbehave in that case too. Usually this calls for a separate type.

pacejackson: The main one I'm thinking of is Thrift where you have to use the context object to pass through headers.

I'm not sure there's a good way around using it this way if we want to have retry middleware (so you can have default retries plus customization).  One option would be to just _remove_ the retry middleware and explicitly use `retrybp.Do` each time?

PR #212, opened 2020-05-28
mqsend: Don't fail send if ctx is canceled

fishy:


mqsend/mqsend_test.go
```diff
@@ -91,3 +91,27 @@ func TestLinuxMessageQueue(t *testing.T) {
 		},
 	)
 }
+
+func TestContextBehavior(t *testing.T) {
+	// We rely on the behavior that we can sill get the deadline out of a context
```
pacejackson, line 96:
```suggestion
	// We rely on the behavior that we can still get the deadline out of a context
```

fishy:
Closed in favor of #213

PR #213, opened 2020-05-29
Avoid using already canceled context for mqsend

fishy:


events/events.go
```diff
@@ -55,6 +55,8 @@ type Config struct {
 	//
 	// If the passed in context object already has an earlier deadline set,
 	// that deadline will be respected instead.
+	// But if the passed in context is already canceld,
```
pacejackson, line 58: Small typo, should be “ cancelled”

events/events.go
```diff
@@ -103,6 +105,11 @@ func (q *Queue) Close() error {
 
 // Put serializes and puts an event into the event queue.
 func (q *Queue) Put(ctx context.Context, event thrift.TStruct) error {
+	if ctx.Err() != nil {
+		// The request context is already canceled,
```
pacejackson, line 109: Small typo, should be “ cancelled”

events/events.go
fishy: The variable is called `context.Canceled`: https://pkg.go.dev/context?tab=doc#pkg-variables

fishy: technically both `cancelled` and `canceled` are correct in English, I think.

pacejackson: Ya, I guess they are...

:til:

(I’ve probably written it both ways...)

pacejackson: Well, _this_ one still has a small type 😅

“canceld”

fishy: :shakes-fist: fixed.

PR #218, opened 2020-06-02
secrets: clone the map passed into NewTestSecrets

pacejackson:
If you use the same underlying raw map to initialize test secrets in parallel, it will trigger race condition checks.  To avoid this, clone the map

### DIFF
--- a/secrets/testing_test.go
+++ b/secrets/testing_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"fmt"
 	"strings"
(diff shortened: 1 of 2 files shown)


secrets/testing_test.go
```diff
@@ -205,6 +206,39 @@ func TestNewTestSecrets(t *testing.T) {
 	}
 }
 
+func TestRaceConditionFlag(t *testing.T) {
```
pacejackson, line 209: This test fails when running `make test` without the accompanying changes because it triggers the race condition check.

PR #221, opened 2020-06-03
mqsend: Set ulimit automatically during Open

fishy:
When using mqsend in prod we need to set ulimit, mqsend.Open is the most
nature place to set that.


fishy:
> This doesn’t feel like the right place to do this, _maybe_ add a helper to runtimebp that a service can call manually? I don’t think we want to add this into the core path though.

I would argue that this is the perfect place to do this.

Any separated helper will create a disjoint between setting ulimit and actually using the message queue. In reality the same service binary would only actually enable the message queues on prod environment. They won't enable it if it's running locally on developer's laptop to test (the real message queue doesn't even work on mac), by just set the tracer's queue name to empty string in the local config file, or by disabling event publishing. If the helper is separated from `OpenMessageQueue`, then they would either need another config/flag to control whether to call the helper (because calling the helper won't work on macs either), or do some complex logic to check whether the tracer's configured queue name is empty and event publisher is also disabled.

With this happens automatically with OpenMessageQueue, the "helper" will only be called when we are really opening the system message queue on linux. Forgot to set the ulimit will also likely to causing opening the message queue to fail. So bundle them together is the most intuitive and hide away way to do it.

PR #222, opened 2020-06-04
thriftbp: Double MonitorClient middleware by default

fishy:
Add another MonitorClient before retry client middleware, with
".wrapped" suffix added to the service slug, so we have a view of the
total requests/failures/latency for all retries grouped together, from
the client's point of view.

### DIFF
--- a/thriftbp/client_middlewares.go
+++ b/thriftbp/client_middlewares.go
@@ -14,6 +14,21 @@ import (
 	"github.com/reddit/baseplate.go/tracing"
 )
 
+// MonitorClientWrappedSlugSuffix is a suffix to be added to the service slug
+// arg of MonitorClient function, in order to distinguish from the spans that
+// are the raw client calls..
+//
+// The MonitorClient with this suffix will have span operation names like:
+//
+//     service-with-retry.endpointName
+//
+// Which groups all retries of the same client call together,
+// while the MonitorClient without this suffix will have span operation names
+// like:
+//
+//     service.endpointName
+const MonitorClientWrappedSlugSuffix = "-with-retry"
+
 // WithDefaultRetryFilters returns a list of retrybp.Filters by appending the
 // given filters to the "default" retry filters:
 //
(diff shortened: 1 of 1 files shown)


thriftbp/client_middlewares.go
```diff
@@ -14,6 +14,21 @@ import (
...
+// The MonitorClient with this suffix will have span operation names like:
+//
+//     service.wrapped.endpointName
+//
+// Which groups all retries of the same client call together,
+// while the MonitorClient without this suffix will have span operation names
+// like:
+//
+//     service.endpointName
+const MonitorClientWrappedSlugSuffix = ".wrapped"
```
fishy, line 30: Other options might be ".with-retry"?

fishy: Or maybe "wrapped."/"with-retry." prefixes are better? Since on metrics we already have "clients." prefix before them.

pacejackson: I think _maybe_ use `-wrapped` or `-with-retry`?  Then it's separated as:

```
client.foo-wrapped.fizz
client.foo.fizz
```

pacejackson: That way it doesn't possibly conflict with either having a client named `wrapped` or a method named `wrapped`?

fishy: What if we have another service called foo-wrapped :trollface: 

I think that makes sense. @konradreiche what do you think?

pacejackson: :shakes-fist: that's true!!!

konradreiche: I'd love to see this being explicit about the retry middleware, i.e. `wrapped-with-retry` if we're not too bothered by a longer name.

pacejackson: `-with-retry` sounds great

fishy: Changed to "-with-retry"

PR #223, opened 2020-06-04
experiments: Return special error for empty/nil bucket key

fishy:
This could be caused by empty loids. We consider them expected and they
usually need to be handled differently from other errors.

### DIFF
--- a/experiments/experiments.go
+++ b/experiments/experiments.go
@@ -19,6 +19,25 @@ import (
 
 const targetAllOverride = `{"OVERRIDE": true}`
 
+// MissingBucketKeyError is a special error returned by Variant functions,
+// to indicate that the bucket key from the args map is missing.
+//
+// This error is usually considered "normal",
+// the caller might still want to log it,
+// but usually don't need to send it to sentry.
+type MissingBucketKeyError struct {
+	ExperimentName string
+	ArgsKey        string
+}
+
+func (e MissingBucketKeyError) Error() string {
+	return fmt.Sprintf(
+		"experiment: must specify %q in call to variant for experiment %s",
+		e.ArgsKey,
+		e.ExperimentName,
+	)
+}
+
 // Experiments offers access to the experiment framework with automatic refresh
 // when there are change.
 //
@@ -69,6 +88,10 @@ func NewExperiments(ctx context.Context, path string, eventLogger EventLogger, l
 //
 // Returns the name of the enabled variant as a string if any variant is
 // enabled. If no variant is enabled returns an empty string.
+//
+// This function might return MissingBucketKeyError as the error.
+// Caller usually want to check for that and handle it differently from other
+// errors. See its documentation for more details.
 func (e *Experiments) Variant(name string, args map[string]interface{}, bucketingEventOverride bool) (string, error) {
 	experiment, err := e.experiment(name)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


fishy:
> Yes, I think we can just allow `nil` same as in Python. We're on very untyped territory here so it makes sense to be consequent about that unless we decide to type it at some point.

@konradreiche OK I implemented that with [second commit](https://github.com/reddit/baseplate.go/pull/223/commits/133ad1b15b55060e6932b6cab8f5fa78799 3f9e4). If we are happy with that I'll drop the first one (that returns empty string instead of nil from edgecontext).

fishy:
> If bucket val is nil, we I think we should just return an error:
> 
> https://github.com/reddit/baseplate.py/blob/e557576f5b0c4e618069efe5b11c038b181c3a39/baseplate/lib/experiments/providers/simple_experiment.py#L276-L289

hmm but in .py we just print a log, instead of raising an exception. So that's actually different from "returning an error"

This kinda means this kind of situation (loid is empty) is expected and we just need to carry on?

pacejackson:
Ya, you're right and I changed my response, it _seems_ like what we want to do is just return the default variant (which I guess is `""`?).

fishy:
:haircut:

Changed to return a special error for that case so caller can handle it appropriately.

experiments/experiments.go
```diff
@@ -270,14 +297,21 @@ func (e *SimpleExperiment) Variant(args map[string]interface{}) (string, error)
...
 		)
 	}
 
+	if bucketVal == "" {
+		return "", MissingBucketKeyError{
+			ExperimentName: e.name,
+			ArgsKey:        e.bucketVal,
+		}
+	}
+
```
pacejackson, line 314: I think all this should happen _before_ the overrides:

https://github.com/reddit/baseplate.py/blob/e557576f5b0c4e618069efe5b11c038b181c3a39/baseplate/lib/experiments/providers/simple_experiment.py#L271-L293

fishy:
> This is a slightly weird one since it doesn't actually come up in python... we _might_ also want to do this before the `overrides` section as well?
> 
> https://github.com/reddit/baseplate.go/pull/223/files#diff-79b2aa0ff58b64f5fa5236a79671fda3R300-R306

I don't think that link is working... can you try to link the lines from here instead?

https://github.com/reddit/baseplate.go/blob/60051385043002918bde6cac6f93811dd9d49c44/experiments/experiments.go

pacejackson:
Yep:

https://github.com/reddit/baseplate.go/blob/60051385043002918bde6cac6f93811dd9d49c44/experiments/experiments.go#L300-L306

fishy:
That one is needed because the map itself is untyped, but when we pass it to the next function (`e.calculateBucket`) it has to be a string. So we need to convert it to string, and handle the case when it's not actually string somehow. So we either need to return an error, or silently ignore that and use some fallback (like empty string).

fishy:
So basically if we check it before override, it could be a string before override, but then overridden as something else.

pacejackson:
> So basically if we check it before override, it could be a string before override, but then overridden as something else.

Overrides doesn't effect the bucketing key though, it's a way to short-circuit to a particular variant.

fishy:
So the difference is that if the bucket key is something not string but also not empty/nil, do we always return error or do we allow some overridden rule to still give them a bucket. Do that check after override means we can still short circuit them, while doing it before override means we always return error in those cases.

Compare to .py implementation, py only checks None and not in map before override, and if it's actually not string and will cause errors, it happens after override. So our current order seems to be inline with that?

pacejackson:
I think that sounds fair, just wanted to feel out the ambiguity between this implementation and the Python one.

PR #224, opened 2020-06-05
Update metrics to follow reddit's spec

pacejackson:


metricsbp/baseplate_hooks.go
```diff
@@ -10,7 +12,7 @@ import (
 
 const (
 	success = "success"
-	fail    = "fail"
+	failure = "failure"
```
fishy, line 15: For easier transition, can we keep both `fail` and `failure` for at a while (e.g. two weeks), then remove the wrong one (`fail`)?

pacejackson: good call

metricsbp/baseplate_hooks.go
```diff
@@ -101,3 +106,76 @@ var (
...
+func (c *atomicCounter) incr() {
+	atomic.AddInt64(c.count, 1)
+}
+
+func (c *atomicCounter) decr() {
+	atomic.AddInt64(c.count, -1)
+}
+
+func (c *atomicCounter) current() int64 {
+	return *c.count
```
fishy, line 152: this needs to be `atomic.LoadInt64(c.count)` instead.

metricsbp/baseplate_hooks.go
```diff
@@ -101,3 +106,76 @@ var (
...
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "UNKOWN-HOSTNAME"
+	}
+	gauge := metrics.Gauge(
+		fmt.Sprintf("runtime.%s.PID%d.active_requests", hostname, os.Getpid()),
+	)
+	go func() {
+		for {
+			time.Sleep(interval)
```
fishy, line 129: don't use `time.Sleep` in a for loop, use `time.Tick`/`time.Ticker` instead.

but I think this should be handled by [`RunSysStats`](https://pkg.go.dev/github.com/reddit/baseplate.go/metricsbp?tab=doc#Statsd.RunSysStats) function, using same prefixes and stuff (we are actually using labels for that one, and I think we should still use labels for this one as well).

metricsbp/baseplate_hooks.go
```diff
@@ -101,3 +106,76 @@ var (
...
+// The gauge will be named "runtime.{hostname}.PID{process_id}.active_requests".
+func NewConcurrencyCreateSpanHook(metrics *Statsd, interval time.Duration) tracing.CreateServerSpanHook {
+	var activeRequests int64
+	counter := &atomicCounter{count: &activeRequests}
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "UNKOWN-HOSTNAME"
+	}
+	gauge := metrics.Gauge(
+		fmt.Sprintf("runtime.%s.PID%d.active_requests", hostname, os.Getpid()),
```
fishy, line 125: shouldn't the name be `concurrency` instead of `active_requests`?

metricsbp/baseplate_hooks.go
```diff
@@ -101,3 +106,76 @@ var (
...
+			}
+			gauge.Set(float64(counter.current()))
+		}
+	}()
+	return &concurrencyCreateServerSpanHook{counter: counter}
+}
+
+type atomicCounter struct {
+	count *int64
+}
```
fishy, line 141: I also feel this is overkill/overcomplex.

That gauge, by its definition, is not per-request but only one per process. So it perfectly fine to just declare a global variable of int64, and just call `atomic.AddInt64` 1/-1 in the hook, and call `atomic.LoadInt64` in `RunSysStats` to report it.

metricsbp/baseplate_hooks.go
pacejackson: No it's `active_requests`, `concurrency` was a mistake.

metricsbp/baseplate_hooks.go
pacejackson: I dunno, to me directly editing a global value using the `atomic` functions seems like a bad example to set?  I don't think this is complex at all and sets a better example for other developers reading the code.

I can change it if that's not the case.

pacejackson:
Should we also add the `fmt.Sprintf("runtime.%s.PID%d.", hostname, os.Getpid())` prefix to the other `RunSysStats` metrics?

metricsbp/sys_stats.go
```diff
@@ -24,45 +28,49 @@ func pullRuntimeStats() (cpu cpuStats, mem runtime.MemStats) {
 // RunSysStats starts a goroutine to periodically pull and report sys stats.
 //
 // Canceling the context passed into NewStatsd will stop this goroutine.
-func (st *Statsd) RunSysStats(labels Labels) {
+func (st *Statsd) RunSysStats(_ Tags) {
```
fishy, line 31: Oh I thought it was part of a config struct so I said leave it but ignore it to keep backward compatibility. Since this is a function arg, doing that is kinda defeating the point and we can just remove it. Also this is already a breaking change :)

PR #226, opened 2020-06-06
Baseplate error struct

pacejackson:


retrybp/filters.go
```diff
@@ -110,6 +118,40 @@ func UnrecoverableErrorFilter(err error, next retry.RetryIfFunc) bool {
...
+// BaseplateErrorFilter returns true if the given error is a BaseplateError and
+// returns one of the given codes, other wise it calls the next filter in the
+// chain.
+func BaseplateErrorFilter(codes ...int32) Filter {
+	codeMap := map[int32]bool{}
+	for _, code := range codes {
+		codeMap[code] = true
+	}
+	return func(err error, next retry.RetryIfFunc) bool {
+		bpErr := new(BaseplateError)
```
fishy, line 146: this should be `var bpErr BaseplateError` since it's an interface.

retrybp/filters.go
```diff
@@ -110,6 +118,40 @@ func UnrecoverableErrorFilter(err error, next retry.RetryIfFunc) bool {
...
+func WithDefaultRetryableCodes(codes ...int32) []int32 {
+	return append([]int32{
+		baseplate.ErrorCode_TOO_EARLY,
+		baseplate.ErrorCode_SERVICE_UNAVAILABLE,
+		baseplate.ErrorCode_TIMEOUT,
+	}, codes...)
+}
+
+// BaseplateErrorFilter returns true if the given error is a BaseplateError and
+// returns one of the given codes, other wise it calls the next filter in the
```
fishy, line 138: typo: `other wise` -> `otherwise`

retrybp/filters.go
```diff
@@ -110,6 +118,40 @@ func UnrecoverableErrorFilter(err error, next retry.RetryIfFunc) bool {
...
+// chain.
+func BaseplateErrorFilter(codes ...int32) Filter {
+	codeMap := map[int32]bool{}
+	for _, code := range codes {
+		codeMap[code] = true
+	}
+	return func(err error, next retry.RetryIfFunc) bool {
+		bpErr := new(BaseplateError)
+		if errors.As(err, &bpErr) {
+			_, ok := codeMap[bpErr.GetCode()]
```
fishy, line 148: Since the map has bool instead of struct{} as the value type, you can just do `return codeMap[bpErr.GetCode()]` instead.

retrybp/filters.go
```diff
@@ -110,6 +118,40 @@ func UnrecoverableErrorFilter(err error, next retry.RetryIfFunc) bool {
...
+		baseplate.ErrorCode_SERVICE_UNAVAILABLE,
+		baseplate.ErrorCode_TIMEOUT,
+	}, codes...)
+}
+
+// BaseplateErrorFilter returns true if the given error is a BaseplateError and
+// returns one of the given codes, other wise it calls the next filter in the
+// chain.
+func BaseplateErrorFilter(codes ...int32) Filter {
+	codeMap := map[int32]bool{}
```
fishy, line 141: This works but it's kinda unorthodox . I'd prefer `codeMap := make(map[int32]bool)` or even better, `codeMap := make(map[int32]bool, len(codes))`.

retrybp/filters.go
```diff
@@ -8,8 +8,16 @@ import (
 	retry "github.com/avast/retry-go"
 
 	"github.com/reddit/baseplate.go/clientpool"
+	"github.com/reddit/baseplate.go/internal/gen-go/reddit/baseplate"
 )
 
+// BaseplateError is implemented by the Error struct in baseplate.thrift.
+type BaseplateError interface {
```
fishy, line 15: consider we only used `GetCode`, it's probably worth unexport this type and only have `GetCode` defined.

edgecontext/edgecontext.go
```diff
@@ -109,6 +109,8 @@ type NewArgs struct {
 	AuthToken string
 
 	OriginServiceName string
+
```
fishy, line 112: nit: unnecessary black line?

thriftbp/errors.go
```diff
@@ -0,0 +1,89 @@
...
+	retry "github.com/avast/retry-go"
+	baseplatethrift "github.com/reddit/baseplate.go/internal/gen-go/reddit/baseplate"
+	"github.com/reddit/baseplate.go/retrybp"
+)
+
+// BaseplateErrorCode defines the minimum interface for a Baseplate Error that is
+// used by logic within baseplate.go.
+//
+// Additional fields are defined in BaseplateErrorDetails.
+type BaseplateErrorCode interface {
```
fishy, line 15: :no_bell: I still don't see the value of exporting those interfaces. service code (users of this library) don't really need them because they will have the thrift compiled go code, which has the  concrete types in them.

pacejackson: I can be useful for services that are interacting with multiple thrift services and what to be able to write helper methods for dealing with Baseplate errors. Otherwise they have a separate “Error” type for each service

fishy: even they have multiple thrift files, those should include the same `baseplate.thrift`, so there should be only one compiled error strict ?

PR #227, opened 2020-06-06
httpbp: add middleware for specifying supported HTTP methods

pacejackson:


httpbp/middlewares.go
```diff
@@ -158,3 +160,31 @@ func InjectEdgeRequestContext(truster HeaderTrustHandler, impl *edgecontext.Impl
 		}
 	}
 }
+
+// SupportedMethods returns a middleware that checks if the request is made
+// using one of the given HTTP methods.
+//
+// Returns a raw, plain text 405 error response if the method is not supported.
+func SupportedMethods(method string, additional ...string) Middleware {
```
fishy, line 168: 🔕 Is the separation of `method` and `additional` for forcing at least one arg passed in?

httpbp/middlewares.go
```diff
@@ -158,3 +160,31 @@ func InjectEdgeRequestContext(truster HeaderTrustHandler, impl *edgecontext.Impl
...
+	supported := make(map[string]bool, len(additional)+1)
+	supported[strings.ToUpper(method)] = true
+	for _, m := range additional {
```
fishy, line 171: 🔕 Should we implicitly add HEAD as long as GET is allowed? Go's stdlib http package will handle HEAD requests automatically (drop the body and just return the headers you returned in your handler implementation), and developers are likely to forget add HEAD to the allowed list, and I think in most cases we don't really want to disallow HEAD requests.

httpbp/middlewares.go
```diff
@@ -158,3 +160,31 @@ func InjectEdgeRequestContext(truster HeaderTrustHandler, impl *edgecontext.Impl
...
+		supported[strings.ToUpper(m)] = true
+	}
+	return func(name string, next HandlerFunc) HandlerFunc {
+		return func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+			method := r.Method
+			if method == "" {
+				method = http.MethodGet
+			}
+			if ok := supported[method]; !ok {
```
fishy, line 180: you don't even need `ok` here, just `if supported[method] {` is sufficient :)

httpbp/middlewares.go
pacejackson: Yes

httpbp/middlewares.go
```diff
@@ -158,3 +160,31 @@ func InjectEdgeRequestContext(truster HeaderTrustHandler, impl *edgecontext.Impl
...
+func SupportedMethods(method string, additional ...string) Middleware {
+	supported := make(map[string]bool, len(additional)+1)
+	supported[strings.ToUpper(method)] = true
+	for _, m := range additional {
+		supported[strings.ToUpper(m)] = true
+	}
+	return func(name string, next HandlerFunc) HandlerFunc {
+		return func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+			method := r.Method
+			if method == "" {
```
fishy, line 177: 🔕 In what scenario would this happen?

httpbp/middlewares.go
pacejackson: That makes sense

httpbp/middlewares.go
```diff
@@ -158,3 +160,40 @@ func InjectEdgeRequestContext(truster HeaderTrustHandler, impl *edgecontext.Impl
...
+	supported := make(map[string]bool, len(additional)+1)
+	supported[strings.ToUpper(method)] = true
+	for _, m := range additional {
+		supported[strings.ToUpper(m)] = true
+	}
+
+	hasGet := false
+	for m := range supported {
+		hasGet = hasGet || strings.Compare(m, http.MethodGet) == 0
+	}
```
fishy, line 179: you don't need this for loop. you already built the map, so you can just do a map lookup :)

but implicitly adding HEAD means that if some service really want to allow GET and disallow HEAD, there's no way they can achieve that with this middleware. I don't really know if that's a legit case in real life, though.

httpbp/middlewares.go
pacejackson: I realized that I just misinterpreted the documentation for this:

```go
    // Method specifies the HTTP method (GET, POST, PUT, etc.).
    // For client requests, an empty string means GET.
    //
    // Go's HTTP client does not support sending a request with
    // the CONNECT method. See the documentation on Transport for
    // details.
    Method string
```

I don't think this should happen on the server side so I can probably remove this.

httpbp/middlewares.go
pacejackson: 🤦 

As far as not allowing HEAD... I'm not sure there's _really_ a use case for that and if there is, that endpoint can just not use this middleware and make their own.

pacejackson:
You're also supposed to return an `Allow` header with a list of the supported HTTP methods so I added that as well.

httpbp/middlewares.go
```diff
@@ -158,3 +165,43 @@ func InjectEdgeRequestContext(truster HeaderTrustHandler, impl *edgecontext.Impl
...
+	for m := range supported {
+		allowed[i] = m
+		i++
+	}
+	sort.Strings(allowed)
+	allowedHeader := strings.Join(allowed, ",")
+
+	return func(name string, next HandlerFunc) HandlerFunc {
+		return func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+			if ok := supported[r.Method]; !ok {
```
fishy, line 196: nit: Here it can be just `if !supported[r.Method] {`

pacejackson: ahh, meant to fix that 😬

PR #229, opened 2020-06-08
httpbp: force http servers to define the methods their endpoints support

pacejackson:


httpbp/server.go
```diff
@@ -47,6 +61,15 @@ type Endpoint struct {
 	// to any Middleware wrapping the HandlerFunc.
 	Name string
 
+	// Methods is the list of HTTP methods that the endpoint supports.  Methods
+	// must have at least one entry and all entries must be valid HTTP methods.
+	//
+	// Method names should be in all upper case.
+	// Use the http.Method* constants from "net/http" for the values in this slice
+	// to ensure that you are using methods that are supported and in the format
+	// we expect.
```
fishy, line 70: should we also mention that HEAD will be implicitly added if GET is added here?

httpbp/server.go
```diff
@@ -30,11 +43,12 @@ type httpHandlerFactory struct {
 	middlewares []Middleware
 }
 
-func (f httpHandlerFactory) NewHandler(name string, handle HandlerFunc, middlewares ...Middleware) http.Handler {
-	wrappers := make([]Middleware, 0, len(f.middlewares)+len(middlewares))
+func (f httpHandlerFactory) NewHandler(endpoint Endpoint) http.Handler {
+	wrappers := make([]Middleware, 0, len(f.middlewares)+len(endpoint.Middlewares)+1)
+	wrappers = append(wrappers, SupportedMethods(endpoint.Methods[0], endpoint.Methods[1:]...))
```
pacejackson, line 48: note, this will panic if you pass in an empty string... let it!  We already check this earlier by calling `endpoint.Validate`

fishy, line 48: 🔕 til this won't cause panic when the slice contains one item: https://play.golang.org/p/HtkVu8laHF_n

PR #230, opened 2020-06-09
draft: http client with concurrency limits, circuit breaking...

fizx:


httpbp/client.go
```diff
@@ -0,0 +1,164 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"sync/atomic"
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/reddit/baseplate.go/tracing"
```
fishy, line 11: style nit: put tracing into its own import group.

httpbp/client.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/sony/gobreaker"
+)
+
+// LowLevelClientMiddleware is suitable for transport-level functions.  Retries
+// and request munging should be done by wrapping the client.
```
fishy, line 16: typo? s/munging/mangling/

httpbp/client.go
```diff
@@ -0,0 +1,164 @@
...
+	next           http.RoundTripper
+	activeRequests int64
+	maxConcurrency int64
+}
+
+func (m *maxConcurrency) RoundTrip(req *http.Request) (*http.Response, error) {
+	attemptedRequests := atomic.AddInt64(&m.activeRequests, 1)
+	defer atomic.AddInt64(&m.activeRequests, -1)
+	if m.maxConcurrency > 0 && attemptedRequests > m.maxConcurrency {
+		return nil, errors.New("hit concurrency limit on http round-tripper")
```
fishy, line 29: nit: Please define this error as a global var, or at least prefix the error message with `httpbp: `.

httpbp/client.go
```diff
@@ -0,0 +1,164 @@
...
+		r, e := c.next.RoundTrip(req)
+		if e != nil {
+			return nil, e
+		}
+		if r.StatusCode > 499 {
+			e = r.Body.Close()
+			if e != nil {
+				return nil, e
+			}
+			return nil, fmt.Errorf("received http %d", r.StatusCode)
```
fishy, line 61: Converting 5xx status code into errors seems against how `RoundTripper` is defined: https://pkg.go.dev/net/http?tab=doc#RoundTripper

```
	// RoundTrip should not attempt to interpret the response. In
	// particular, RoundTrip must return err == nil if it obtained
	// a response, regardless of the response's HTTP status code.
	// A non-nil err should be reserved for failure to obtain a
	// response.
```

httpbp/client.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return rsp.(*http.Response), nil
+}
+
+// CircuitBreaker enables circuit breaking for your client.
+func CircuitBreaker(minRequests uint32, minFailureRatio float64) LowLevelClientMiddleware {
+	settings := gobreaker.Settings{
+		Interval: 60 * time.Second, // Reset the counts every 60 seconds
```
fishy, line 74: Should this be configurable?

httpbp/client.go
```diff
@@ -0,0 +1,164 @@
...
+// transport middleware.  Note that retries are not suitable for low-level
+// middleware, and you should consider bringing your own retries, wrapping
+// this client.
+func NewClient(config *ClientConfig) *http.Client {
+	var tripper http.RoundTripper = &http.Transport{
+		MaxConnsPerHost: config.MaxConnsPerHost,
+	}
+	for i := len(config.Middlewares) - 1; i >= 0; i-- {
+		mw := config.Middlewares[i]
+		tripper = mw(fmt.Sprintf("%v", mw), tripper)
```
fishy, line 157: What is the intention of using `fmt.Sprintf("%v", mw)`? I don't see you define stringer in any of the middleware implementations, so this would look like some `"{foo <nil>}"` (see https://play.golang.org/p/EkRg7IkWS4b), and that's not suitable to be used in a span name.

fizx:
> Can you actually do this in your service code first, run it in prod, then upstream it to baseplate.go after that run was successful?

That's the plan :) Throwing a draft here so that people can comment if they like!

httpbp/client.go
fizx: Yeah, at this point, have we run into enough issues with trying to build at this layer that we'd prefer to just wrap a full http client, rather than try to mess with RoundTripper?

fishy: If converting 5xx codes into errors is a hard requirement, then `RoundTripper` is probably the wrong solution for this.

PR #233, opened 2020-06-11
Add active_requests runtime gauge

fishy:


metricsbp/statsd.go
```diff
@@ -313,3 +316,25 @@ func (st *Statsd) Close() error {
...
 	return st.fallback().statsd.WriteTo(w)
 }
+
+// IncActiveRequests increases the internal active requests counter by 1.
+//
+// The internal active requests counter is used by RunSysStats to report the
+// active_requests gauge.
+//
+// It returns the value after increasing.
+func (st *Statsd) IncActiveRequests() int64 {
```
pacejackson, line 326: On the statsd object seems like the wrong place to put this, or at least exporting it seems like not what we want to do?

fishy:
The problem with span hook is that not all server spans are active requests. There are queue consumers, background jobs, etc.

metricsbp/statsd.go
fishy: statsd object is what reports that gauge, so this sounds like the right place to do it. no?

pacejackson: I'm just thinking that I don't think we really want to use this outside of baseplate, adding it onto the Statsd object means it's being put on something that is used throughout services and we don't _really_ want people to use this in their services directly.

fishy: I can add a comment to say it should only be used by middleware implementations :)

fishy:
:haircut: Added comment and also added httpbp.CountActiveRequests middleware.

metricsbp/sys_stats.go
```diff
@@ -115,6 +117,7 @@ func (st *Statsd) RunSysStats() {
 				mcacheSys.Set(float64(mem.MCacheSys))
 				// other
 				memOther.Set(float64(mem.OtherSys))
+				activeRequests.Set(float64(atomic.LoadInt64(&st.activeRequests)))
```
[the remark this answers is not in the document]
fishy: We just don't have the need of that yet. We can add that in the future if there's a legit use case of it, but prefer not to export APIs not needed to give us the flexibility to alter them without doing breaking changes :)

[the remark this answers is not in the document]
pacejackson: that's fair, I don't think we _need_ it but it can be un-exported and keep references to `st.activeRequests` more contained which I think is nice.

fishy:
:haircut: Changed to use server span hook.

metricsbp/baseplate_hooks.go
```diff
@@ -98,9 +105,24 @@ func (h *spanHook) OnAddCounter(span *tracing.Span, key string, delta float64) e
 	return nil
 }
 
+type countActiveRequestsHook struct {
+	metrics *Statsd
+}
+
+func (h countActiveRequestsHook) OnPostStart(_ *tracing.Span) error {
+	atomic.AddInt64(&h.metrics.activeRequests, 1)
```
pacejackson, line 113: I still think it's _cleaner_ to have these as methods on the actual object that owns `activeRequests` rather than mutating them separately, this is probably a confusing thing to see from an outside perspective.  They don't have to be exported, but I do think that `Statsd` should _own_ the code that mutates it's internal state.

fishy: I just feel if we do that, that's would be just a very thin wrap around `atomic.AddInt64` and `atomic.LoadInt64`, which feels really unnecessary.

all the stuff touching it are unexported now (except `RunSysStats`) so there's not really any "outside perspective" 😅

pacejackson: Right... what I'm saying is it's icky for one thing to reach into another struct to mutate it's internal state, that seems like a _major_ code smell to me.

fishy: Fine. Functions added.

PR #236, opened 2020-06-16
Fix httpbp.InjectEdgeRequestContext middleware

fishy:
We forgot to do base64 decoding when handling the edge request header.

### DIFF
--- a/httpbp/fixtures_test.go
+++ b/httpbp/fixtures_test.go
@@ -2,6 +2,7 @@ package httpbp_test
 
 import (
 	"context"
+	"encoding/base64"
 	"net/http"
 	"strings"
 	"testing"
@@ -116,7 +117,7 @@ func newRequest(t testing.TB) *http.Request {
 	if err != nil {
 		t.Fatal(err)
 	}
-	req.Header.Add(httpbp.EdgeContextHeader, headerWithValidAuth)
+	req.Header.Add(httpbp.EdgeContextHeader, base64.StdEncoding.EncodeToString([]byte(headerWithValidAuth)))
 	req.Header.Add(httpbp.SpanSampledHeader, "1")
 	return req
 }
--- a/httpbp/middlewares.go
+++ b/httpbp/middlewares.go
@@ -44,16 +44,30 @@ func Wrap(name string, handle HandlerFunc, middlewares ...Middleware) HandlerFun
 // DefaultMiddlewareArgs provides the arguments for the default, Baseplate
 // Middlewares
 type DefaultMiddlewareArgs struct {
-	TrustHandler    HeaderTrustHandler
+	// The edgecontext implementation to use. Required.
 	EdgeContextImpl *edgecontext.Impl
+
+	// The HeaderTrustHandler to use.
+	// If empty, NeverTrustHeaders will be used instead.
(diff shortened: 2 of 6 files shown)


httpbp/middlewares.go
```diff
@@ -44,16 +44,30 @@ func Wrap(name string, handle HandlerFunc, middlewares ...Middleware) HandlerFun
...
+	// The logger to be called when edgecontext parsing failed.
+	Logger log.Wrapper
 }
 
 // DefaultMiddleware returns a slice of all of the default Middleware for a
 // Baseplate HTTP server.
 func DefaultMiddleware(args DefaultMiddlewareArgs) []Middleware {
+	if args.TrustHandler == nil {
+		args.TrustHandler = NeverTrustHeaders{}
+	}
```
pacejackson, line 63: Should we actually bother with this in here since we're already doing it in `InjectEdgeRequestContext`?

httpbp/middlewares.go
```diff
@@ -131,35 +145,56 @@ func InjectServerSpan(truster HeaderTrustHandler) Middleware {
...
 	ctx context.Context,
-	truster HeaderTrustHandler,
-	impl *edgecontext.Impl,
 	r *http.Request,
+	args InjectEdgeRequestContextArgs,
 ) context.Context {
-	if !truster.TrustEdgeContext(r) {
+	if args.TrustHandler == nil {
+		args.TrustHandler = NeverTrustHeaders{}
+	}
```
pacejackson, line 153: 👍

httpbp/fixtures_test.go
```diff
@@ -116,7 +117,7 @@ func newRequest(t testing.TB) *http.Request {
 	if err != nil {
 		t.Fatal(err)
 	}
-	req.Header.Add(httpbp.EdgeContextHeader, headerWithValidAuth)
+	req.Header.Add(httpbp.EdgeContextHeader, base64.StdEncoding.EncodeToString([]byte(headerWithValidAuth)))
```
pacejackson, line 120: Should we just have `EncodeEdgeContextHeader` and `DecodeEdgeContextHeader` methods?

fishy: maybe, but that's more breaking :)

fishy: I'm fine if you decided to make a future breaking change to clean that up. Right now I just want to fix that bug that's spamming the error log like crazy :)

httpbp/middlewares.go
fishy: This is also used by `InjectServerSpan`

httpbp/fixtures_test.go
pacejackson: I'm not really sure how that's a breaking change? It's just a way to wrap the `base64` encoding.

fishy: I assume `EncodeEdgeContextHeader` will replace `NewEdgeContextHeaders` (which does the same thing), and `NewEdgeContextHeaders` is exported so if it goes away that's a breaking change.

pacejackson: No, I was just thinking we would have methods to do the b64 encoding/decoding not replace anything 😅

PR #239, opened 2020-06-17
errorsbp package

fishy:
With `batcherror.BatchError` renamed into `errorsbp.Batch`, and `errorsbp.Suppressor` defined.

### DIFF
--- a/errorsbp/doc.go
+++ b/errorsbp/doc.go
@@ -1,5 +1,8 @@
-// Package batcherror provides BatchError, which can be used to compile multiple
-// errors into a single one.
+// Package errorsbp provides some error utilities for Baseplate.go project.
+//
+// Batch
+//
+// errorsbp.Batch can be used to compile multiple errors into a single one.
 //
 // An example of how to use it in your functions:
 //
@@ -18,7 +21,7 @@
 //         }
 //
 //         wg.Wait()
-//         var batch batcherror.BatchError
+//         var batch errorsbp.Batch
 //         for err := range errChan {
 //             // nil errors will be auto skipped
 //             batch.Add(err)
@@ -29,6 +32,6 @@
 //         return batch.Compile()
 //     }
 //
-// This package is not thread-safe.
+// errorsbp.Batch is not thread-safe.
 // The same batch should not be operated on different goroutines concurrently.
-package batcherror
+package errorsbp
--- a/errorsbp/doc.go
+++ b/errorsbp/doc.go
@@ -34,4 +34,22 @@
 //
 // errorsbp.Batch is not thread-safe.
 // The same batch should not be operated on different goroutines concurrently.
+//
+// Suppressor
+//
+// Suppressor is a type defined to provide an unified way to allow certain
+// functions/features to ignore certain errors.
+//
+// It's currently used by thriftbp package in both server and client
+// middlewares, to not treat certain errors defined in thrift IDL as span
+// errors. Because of go's type system, we cannot provide a Suppressor
+// implementation to suppress all errors defined in all thrift IDLs, as a result
+// we rely on service/client implementations to implement it for the
+// middlewares.
+//
+// An example of how to implement it for your thrift defined errors:
+//
+//     func MyThriftSuppressor(err error) bool {
+//         return errors.As(err, new(*mythrift.MyThriftErrorType))
+//     }
 package errorsbp
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -6,21 +6,30 @@ import (
 	"github.com/apache/thrift/lib/go/thrift"
 
 	baseplate "github.com/reddit/baseplate.go"
+	"github.com/reddit/baseplate.go/errorsbp"
 	"github.com/reddit/baseplate.go/log"
 )
 
-// ServerConfig is the arg struct for NewServer.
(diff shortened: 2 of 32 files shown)


fishy:
This replaces #225

thriftbp/thrifttest/server.go
```diff
@@ -157,6 +157,7 @@ func (s *Server) Close() error {
 func NewBaseplateServer(
 	store *secrets.Store,
 	processor thrift.TProcessor,
+	suppressor errorsbp.Suppressor,
```
fishy, line 160: I'm not super happy with this function signature change, but this function also makes less sense to convert to a struct arg.

errorsbp/doc.go
```diff
@@ -29,6 +32,6 @@
 //         return batch.Compile()
 //     }
 //
-// This package is not thread-safe.
+// errorsbp.Batch is not thread-safe.
 // The same batch should not be operated on different goroutines concurrently.
```
pacejackson, line 36: Should we add `Suppressor` to this too?

thriftbp/thrifttest/server.go
pacejackson: I think this probably makes sense to put in `ServerConfig`?

fishy: That's the tricky part. `ServerConfig` is used by `NewServer`, which is the more customized version of `NewBaseplateServer`. `NewBaseplateServer` actually constructs `ServerConfig` and calls `NewServer` underneath.

In `NewServer`, this is not needed as `NewServer` does not create any middlewares by itself, it just applies all the middlewares passed in.

This is also the reason I said this function makes less sense to have a struct arg, because that new struct will be very confusing alongside `ServerConfig`.

errorsbp/doc.go
fishy: Yes I planned to but forgot. Added.

thriftbp/thrifttest/server.go
fishy: OK I changed `ServerConfig` to be used by both `NewServer` and `NewBaseplateServer`. This makes the function signatures much better.

fishy:
:haircut:

thriftbp/thrifttest/server.go
pacejackson: Ya, I agree that neither option is ideal.  Looking forward though, I could see a world where we deprecate and remove `NewServer`, and `NewBaseplateServer` is the real focus anyways.

thriftbp/server.go
```diff
@@ -69,19 +100,27 @@ func NewServer(
...
-	srv, err := NewServer(cfg, processor, wrapped...)
+	middlewares := BaseplateDefaultProcessorMiddlewares(
+		DefaultProcessorMiddlewaresArgs{
+			EdgeContextImpl:     bp.EdgeContextImpl(),
+			ErrorSpanSuppressor: cfg.ErrorSpanSuppressor,
+		},
+	)
+	middlewares = append(middlewares, cfg.Middlewares...)
+	cfg.Middlewares = middlewares
+	cfg.Logger = thrift.Logger(log.ZapWrapper(bp.Config().Log.Level))
```
pacejackson, line 119: 🔕  Maybe `Baseplate` should have a log wrapper?

thriftbp/server.go
```diff
@@ -29,19 +38,41 @@ type ServerConfig struct {
...
-	// Optional TServerSocket you can use instead of setting one up using Addr
-	// plus timeout.  If provided, this will be used rather than Addr and Timeout.
+	// Optional, used only by NewBaseplateServer.
+	//
+	// Please refer to the documentation of
+	// DefaultProcessorMiddlewaresArgs.ErrorSpanSuppressor for more details
+	// regarding how it is used.
+	ErrorSpanSuppressor errorsbp.Suppressor
+
+	// Optional, used only by NewServer.
```
pacejackson, line 48: I don't think we really want to say `used only by NewServer`? I think that's confusing since it's still used by `NewBaseplateServer` because that calls `NewServer`.

thriftbp/server.go
```diff
@@ -69,19 +100,27 @@ func NewServer(
 //
 // The TProcessor underlying the server will be wrapped in the default
 // Baseplate Middleware and any additional middleware passed in.
+//
+// The suppressor arg will be used as
+// DefaultProcessorMiddlewaresArgs.ErrorSpanSuppressor,
+// please refer to its documentation for more details.
```
pacejackson, line 106: It might be good to add that the `Addr` and `Timeout` will be pulled from `bp.Config()`

thriftbp/server.go
pacejackson: I agree that it is _technically_ correct but I think it's misleading and can cause confusion.

fishy: How about now?

pacejackson: 1. `NewBaseplateServer` uses `NewServer` so the field is _still_ used by `NewBaseplateServer`.
2. I don't think it's important to people trying to read the documentation that `NewBaseplateServer` calls `NewServer` and if that ever changes, we'll likely have to use it directly in `NewBaseplateServer`.  From the perspective of someone trying to integrate with this, it _is_ used by `NewBaseplateServer` because that's all they interact with.

pacejackson: Ok, I think this works well 🐟

PR #242, opened 2020-06-19
Upgrade thrift to have better client socket timeout

fishy:
This is a big breaking change. Starting from this commit, Baseplate.go
would require all its users using HEAD version of thrift compiler
(minimal required version is bcae3bb, but 4db7a0a is recommended),
until thrift 0.14 is released.

This change allows client calls to have a controlled, low overhead over
the deadline they set in the context object, in the cases of slow
servers.


thriftbp/client_pool.go
```diff
@@ -82,9 +82,23 @@ type ClientPoolConfig struct {
...
+	// then SocketTimeout needs to be at least your upstream service's p99 latency
+	// SLA. If it's shorter than that you are gonna close connections and fail
+	// requests prematurely.
+	//
+	// It's recommended to make sure all your client call context objects have a
+	// deadline set, and set SocketTimeout to a short value. For example:
+	//
+	//     clientCtx, cancel := context.WithTimeout(ctx, myCallTimeout)
+	//     defer cancel()
+	//     resp, err := client.MyCall(clientCtx, args)
```
fishy, line 101: This is the only part really changed in baseplate.go. Everything else in this pr are just passing ctx around in accompany to the breaking change. The real change is in thrift.

fishy:
:haircut: Added in another thrift compiler change.

PR #246, opened 2020-06-25
Add missing test sizes for bazel build files

fishy:


fishy:
> Is this an autogenerated thing? Or are you adding them manually?

manually. the autogenerate part (gazelle) will keep it, though.

PR #248, opened 2020-07-01
Add Thrift Middleware for circuit breakers

PetGorignak:
Adding a Thrift Middleware for CircuitBreakers, based on existing internal usage.

This setup uses a breaker function based on error rate with a minimum requests during the period to trip. However it wouldn't be that hard to add other trip criterias (or even allow passing a trip function as an input arg), I just don't have a compelling use for that today and therefore this follows YAGNI.


thriftbp/circuit_breaker.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+	}
+}
+
+// ShouldTripCircuitBreaker checks if the circuit breaker should be tripped, based on the provided breaker counts.
+func (cb FailureRatioBreaker) ShouldTripCircuitBreaker(counts gobreaker.Counts) bool {
+	if counts.Requests > 0 && counts.Requests >= uint32(cb.minRequestsToTrip) {
+		failureRatio := float64(counts.TotalFailures) / float64(counts.Requests)
+		if failureRatio >= cb.failureThreshold {
+			log.Warnw(
```
fishy, line 85: As library code, we should refrain from using `log.Warnw` directly. Please use `log.Wrapper` to allow callers to control logging (we'll lose the json formatting of the logs).

Since this and the log in `stateChanged` are on different log levels, we probably want to add two `log.Wrapper`s into the config.

thriftbp/circuit_breaker.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	config.Settings.OnStateChange = stateChanged
+	config.Settings.ReadyToTrip = failureBreaker.ShouldTripCircuitBreaker
+	failureBreaker.CircuitBreaker = gobreaker.NewCircuitBreaker(config.Settings)
+	go failureBreaker.runStatsProducer()
+	return failureBreaker
+}
+
+func (cb FailureRatioBreaker) runStatsProducer() {
+	circuitBreakerGauge := metricsbp.M.RuntimeGauge(cb.metricPrefix + ".circuit.breaker.closed")
```
fishy, line 50: nit: slightly prefer `".circuit-breaker-closed"` over `".circuit.breaker.closed"`. We are not gaining anything from adding 2 extra layers of the metrics path.

thriftbp/circuit_breaker.go
```diff
@@ -0,0 +1,102 @@
...
+	// Minimum requests that need to be sent during a time period before the breaker is eligible to transition from closed to open.
+	MinRequestsToTrip int
+	// Percentage of requests that need to fail during a time period for the breaker to transition from closed to open.
+	// Represented as a float in {0,1}, where .05 means >5% failures will trip the breaker.
+	FailureThreshold float64
+	// Gobreaker settings used to configure the underlying circuit breaker.
+	Settings gobreaker.Settings
+}
+
+// NewFailureRatioBreaker creates a new FailureRatioBreaker with the provided configuration.
```
fishy, line 34: Please document that this function also creates a new goroutine.

We probably should also make that configurable (by adding a bool to the config struct).

thriftbp/circuit_breaker.go
```diff
@@ -0,0 +1,102 @@
...
+	minRequestsToTrip int
+	failureThreshold  float64
+}
+
+// CircuitBreakerConfig represents the configuration for a FailureRatioBreaker
+type CircuitBreakerConfig struct {
+	// Minimum requests that need to be sent during a time period before the breaker is eligible to transition from closed to open.
+	MinRequestsToTrip int
+	// Percentage of requests that need to fail during a time period for the breaker to transition from closed to open.
+	// Represented as a float in {0,1}, where .05 means >5% failures will trip the breaker.
```
fishy, line 28: The proper math-y way to write it is `[0, 1]` or maybe `[0,1)`, depending on whether 1 actually makes sense here (it's documented as `>` here but in `ShouldTripCircuitBreaker` it's actually `>=`).

thriftbp/circuit_breaker.go
```diff
@@ -0,0 +1,102 @@
...
+				circuitBreakerGauge.Set(0)
+			} else {
+				circuitBreakerGauge.Set(1)
+			}
+		}
+	}
+}
+
+// ThriftMiddleware is a thrift.ClientMiddleware that handles circuit breaking.
+func (cb FailureRatioBreaker) ThriftMiddleware(next thrift.TClient) thrift.TClient {
```
fishy, line 69: This is inside `thriftbp` package so the name `ThriftMiddleware` is kinda redundant. Make rename it into `ClientMiddleware` or just `Middleware` (as it doesn't make sense as a server middleware).

pacejackson: Maybe this should instead be in its own package? Then `ThriftClientMiddleware` makes sense (and also is cleaner to add new middleware's over time).

fishy: I'm kind of on the fence on this one. This only makes sense as a client middleware, and currently we don't have any http client middleware framework, so thrift client is the only one, thus it makes sense to live under `thriftbp` for now, and move to its own package in the future when we really need that (as compare to retry this is a thinner wrapper around a third party library). But then if we really move it to its own package in the future that would be a breaking change.

pacejackson: Ya, I kinda prefer having it be in its own package to not set up a potential breaking change but I don’t feel super strongly.

thriftbp/circuit_breaker_test.go
```diff
@@ -0,0 +1,106 @@
...
+	cb := newTestCircuitBreaker()
+	counts := gobreaker.Counts{
+		Requests:      uint32(testMinRequests + 1),
+		TotalFailures: 0,
+	}
+	tripped := cb.ShouldTripCircuitBreaker(counts)
+	assert.False(t, tripped)
+}
+
+func TestShouldTrip_zeroRequests(t *testing.T) {
```
pacejackson, line 31: :no_bell: I’m not sure if anyone cares too much but we usually do table driven tests in baseplate.go rather than multiple discrete test functions.

thriftbp/circuit_breaker_test.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/reddit/baseplate.go/thriftbp"
+	"github.com/reddit/baseplate.go/thriftbp/thrifttest"
+	"github.com/sony/gobreaker"
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
PetGorignak: ok I removed all the asserts.

thriftbp/circuit_breaker.go
```diff
@@ -0,0 +1,102 @@
...
+	config.Settings.ReadyToTrip = failureBreaker.ShouldTripCircuitBreaker
+	failureBreaker.CircuitBreaker = gobreaker.NewCircuitBreaker(config.Settings)
+	go failureBreaker.runStatsProducer()
+	return failureBreaker
+}
+
+func (cb FailureRatioBreaker) runStatsProducer() {
+	circuitBreakerGauge := metricsbp.M.RuntimeGauge(cb.metricPrefix + ".circuit.breaker.closed")
+
+	tick := time.NewTicker(10 * time.Second)
```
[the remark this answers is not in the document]
fishy: Leaving the interval non-configurable is fine to me. I don't feel strongly about it.

thriftbp/circuit_breaker.go
fishy: just `breakerbp`?

thriftbp/circuit_breaker.go
```diff
@@ -0,0 +1,123 @@
...
+	// Minimum requests that need to be sent during a time period before the breaker is eligible to transition from closed to open.
+	MinRequestsToTrip int
+	// Percentage of requests that need to fail during a time period for the breaker to transition from closed to open.
+	// Represented as a float in [0,1], where .05 means >=5% failures will trip the breaker.
+	FailureThreshold float64
+	// Name for this circuit breaker, mostly used as a prefix to disambiguate logs when multiple cb are used.
+	Name string
+	// EmitStatusMetrics sets whether the failure breaker will regularly update a gauge on the breakers state (closed or open/halfopen)
+	EmitStatusMetrics bool
+	// Logger is the logger to be called when edgecontext parsing failed.
```
fishy, line 37: copy-pasted comment :)

Also, please add a blank line between each comment-field groups (here you didn't group any two fields together so that's between each field).

thriftbp/circuit_breaker.go
```diff
@@ -0,0 +1,123 @@
...
+			_, err := cb.goBreaker.Execute(func() (interface{}, error) {
+				return nil, next.Call(ctx, method, args, result)
+			})
+			return err
+		},
+	}
+}
+
+// ShouldTripCircuitBreaker checks if the circuit breaker should be tripped, based on the provided breaker counts.
+func (cb FailureRatioBreaker) ShouldTripCircuitBreaker(counts gobreaker.Counts) bool {
```
fishy, line 107: Do we really want this function to be exported? It's not required to be used by go-breaker library, so exporting it only makes sense if we intend it to be used by some other ways, and I'm not sure we have any other usage for it?

thriftbp/circuit_breaker_test.go
fishy: Table test makes perfect sense here for combining `TestShouldTrip_zeroRequests`, `TestShouldTrip_failures`, `TestShouldTrip_tooFewRequests`, & `TestShouldTrip_lowFaiureRate`. I'd prefer it, even if just for having more consistent coding style within this repo.

But table tests are also good and you should embrace it :)

thriftbp/circuit_breaker_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+func TestShouldTrip_zeroRequests(t *testing.T) {
+	cb := newTestCircuitBreaker()
+	counts := gobreaker.Counts{
+		Requests:      0,
+		TotalFailures: 0,
+	}
+	tripped := cb.ShouldTripCircuitBreaker(counts)
+	if !tripped {
+		t.Errorf("expected circuit breaker not to trip but it did")
```
fishy, line 41: There's not any formatting variables in the string so `t.Error` is sufficient, but you will probably need `Errorf` once you converted it into table test.

thriftbp/circuit_breaker.go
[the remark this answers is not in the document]
fishy: Or maybe just reuse `metricsbp.SysStatsTickerInterval`, so if library user changed that one it also applies here?

thriftbp/circuit_breaker.go
PetGorignak: Changed it all to mocks, wasnt as bad as I thought.

thriftbp/circuit_breaker_test.go
PetGorignak: Changed to table. I think what I dislike about it is how cluttered they usually look -- especially the deep nesting with embedding the t.run() portion. But I guess that is fixable, I like how the code I wrote cleaned up.

thriftbp/circuit_breaker_test.go
PetGorignak: refactored this whole part

breakerbp/failure_ratio.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// EmitStatusMetrics sets whether the failure breaker will regularly update a gauge on the breakers state (closed or open/halfopen)
+	// When enabled, it emits metrics using the interval defined by metricsbp.SysStatsTickerInterval
+	EmitStatusMetrics bool
+
+	// Logger is the logger to be called when the breaker changes states.
+	Logger log.Wrapper
+
+	// MaxRequestsHalfOpen represents he Maximum amount of requests that will be allowed through while the breaker
+	// is in half-open state. Unset or 0 is treated as 1
```
fishy, line 46: unset and 0 is the same thing. there's no way to tell them apart (unless you use `*uint32` here) :)

what does 1 mean here? does that mean to completely disable half-open states?

also, please finish the sentences in doc comments with periods. :)

breakerbp/failure_ratio_test.go
```diff
@@ -0,0 +1,105 @@
...
+		shouldFail:   false,
+		numFailures:  testMinRequests - 1,
+		numSuccesses: 0,
+	},
+	{
+		name:         "low failure rate",
+		shouldFail:   false,
+		numFailures:  499,
+		numSuccesses: 501, // 50.1% just above threshold.
+	}}
```
fishy, line 57: style nit: `},\n}`.

breakerbp/failure_ratio_test.go
```diff
@@ -0,0 +1,105 @@
...
+		t.Errorf("test case {%v} expected to succeed, but call returned error: %v", config.name, err)
+	}
+}
+
+func mockRequests(config testConfig, mock *thrifttest.MockClient, client thrift.TClient) {
+	mock.AddNopMockCalls(testMethod)
+	for i := 0; i < config.numSuccesses; i++ {
+		client.Call(context.Background(), testMethod, nil, nil)
+	}
+	mock.AddMockCall(testMethod,
```
fishy, line 90: style nit: https://github.com/reddit/baseplate.go/blob/master/Style.md#one-per-line-style

breakerbp/failure_ratio_test.go
PetGorignak: done

breakerbp/failure_ratio_test.go
PetGorignak: fixed I think

breakerbp/failure_ratio.go
PetGorignak: I'll update the comment. I was trying to paraphrase the sony/gobreaker code (which has some meh comments itself) 

0, unset, and 1 all allow exactly 1 request through while half-open

PetGorignak:
💇‍♂️ 

Drone still broken, but the tests run just fine locally and i did run the bazel_cleanup script, so hopefully it will succeed when things are fixed.

breakerbp/failure_ratio_test.go
```diff
@@ -0,0 +1,107 @@
...
+func mockRequests(config testConfig, mock *thrifttest.MockClient, client thrift.TClient) {
+	mock.AddNopMockCalls(testMethod)
+	for i := 0; i < config.numSuccesses; i++ {
+		client.Call(context.Background(), testMethod, nil, nil)
+	}
+	mock.AddMockCall(
+		testMethod,
+		func(_ context.Context, args, result thrift.TStruct) error {
+			return errors.New("backend down")
+		})
```
fishy, line 95: here as well: `}\n)`

breakerbp/failure_ratio_test.go
```diff
@@ -0,0 +1,107 @@
...
+	client := thrift.WrapClient(
+		mock,
+		cb.ThriftMiddleware,
+	)
+	mockRequests(config, mock, client)
+	// reset client to succeed
+	mock.AddNopMockCalls(testMethod)
+	err := client.Call(context.Background(), testMethod, nil, nil)
+	if err == nil && config.shouldFail {
+		t.Errorf("test case {%v} expected to fail, but call returned without error", config.name)
```
fishy, line 80: nit: `test case {name}` part is redundant because in table test that will already in test's name (that's why table tests are good)

breakerbp/failure_ratio_test.go
PetGorignak: done

breakerbp/failure_ratio_test.go
PetGorignak: ah ok, removed that and made this one just .Error since it no longer has formatting needs.

fishy:
almost forgot: since it's in its own package now, please add a `doc.go` file and write doc comments for the package.

breakerbp/failure_ratio.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// EmitStatusMetrics sets whether the failure breaker will regularly update a gauge on the breakers state (closed or open/halfopen)
+	// When enabled, it emits metrics using the interval defined by metricsbp.SysStatsTickerInterval.
+	EmitStatusMetrics bool
+
+	// Logger is the logger to be called when the breaker changes states.
+	Logger log.Wrapper
+
+	// MaxRequestsHalfOpen represents he Maximum amount of requests that will be allowed through while the breaker
+	// is in half-open state. If left unset (or set to 0), exactly 1 request will be allwed through while half-open.
```
pacejackson, line 46: typo: `allwed` -> `allowed`

pacejackson:
```suggestion
	// is in half-open state. If left unset (or set to 0), exactly 1 request will be allowed through while half-open.
```

breakerbp/failure_ratio.go
```diff
@@ -0,0 +1,131 @@
...
+	MinRequestsToTrip int
+
+	// Percentage of requests that need to fail during a time period for the breaker to transition from closed to open.
+	// Represented as a float in [0,1], where .05 means >=5% failures will trip the breaker.
+	FailureThreshold float64
+
+	// Name for this circuit breaker, mostly used as a prefix to disambiguate logs when multiple cb are used.
+	Name string
+
+	// EmitStatusMetrics sets whether the failure breaker will regularly update a gauge on the breakers state (closed or open/halfopen)
```
pacejackson, line 38: nit: add a period to the end of this:

```suggestion
	// EmitStatusMetrics sets whether the failure breaker will regularly update a gauge on the breakers state (closed or open/halfopen).
```

breakerbp/failure_ratio.go
```diff
@@ -0,0 +1,131 @@
...
+	logger            log.Wrapper
+}
+
+// Config represents the configuration for a FailureRatioBreaker.
+type Config struct {
+	// Minimum requests that need to be sent during a time period before the breaker is eligible to transition from closed to open.
+	MinRequestsToTrip int
+
+	// Percentage of requests that need to fail during a time period for the breaker to transition from closed to open.
+	// Represented as a float in [0,1], where .05 means >=5% failures will trip the breaker.
```
pacejackson, line 32: 🔕  Style nit: I _thiiiiink_ there should be a blank line between the first "sentence" in a docstring and the more detailed parts that come after?

breakerbp/failure_ratio.go
```diff
@@ -0,0 +1,131 @@
...
+	// Interval represents the cyclical period of the 'Closed' state.
+	// If 0, internal counts do not get reset while the breaker remains in the Closed state.
+	Interval time.Duration
+
+	// Timeout is the duration of the 'Open' state. After an 'Open' timeout duration has passed, the breaker enters 'half-open' state.
+	Timeout time.Duration
+}
+
+// NewFailureRatioBreaker creates a new FailureRatioBreaker with the provided configuration. Creates a new goroutine to emit
+// breaker state metrics if EmitStatusMetrics is set to true.
```
pacejackson, line 58: It would be good to note in the comment when the goroutine is stopped (when `metricsbp.M.Ctx()` is `Done()`).

breakerbp/failure_ratio.go
PetGorignak: fixed this and below

breakerbp/failure_ratio.go
PetGorignak: it feels weird to me to have the extra space in all of these so I guess I would prefer to leave it as-is? IDK I will admit I end up writing these in javadoc style often.

pacejackson: I think it's _probably_ fine either way?  I probably inherited the extra blank line from Python docstrings so the lack of one looks weird to _me_ 😅

fishy: yea that's not required by go's doc string, and I don't think the first paragraph summary is used anywhere (unlike git's commit message).

PR #251, opened 2020-07-06
Baseplate health check v2

fishy:
Update thrift compiled files to latest baseplate.thrift, update thriftbp
package accordingly, and add a helper function to httpbp package to help
extracting health check probe out of health check requests.


httpbp/healthcheck.go
```diff
@@ -0,0 +1,62 @@
...
+//
+// - STARTUP
+//
+// If the probe specified in the request is the int value,
+// this function just blindly return the parsed int value,
+// even if it's not one of the defined enum values known to this version of
+// Baseplate.go.
+//
+// Note that because of go's type system,
+// to make this function more useful the returned probe value is casted back to
```
pacejackson, line 41: Does that _actually_ make it more useful? It means that if you want to compare it to the actual enum values you have to convert it _back_?

httpbp/healthcheck_test.go
```diff
@@ -0,0 +1,81 @@
...
+					}
+				}
+				probe, err := httpbp.GetHealthCheckProbe(query)
+				if probe != c.expected {
+					t.Errorf("Expected probe value %d, got %d", c.expected, probe)
+				}
+				if err != nil && !c.shouldErr {
+					t.Errorf("Did not expect error, got %v", err)
+				}
+				if err == nil && c.shouldErr {
```
pacejackson, line 75: nit: make this an `else if` since it should be one or the other?

httpbp/healthcheck.go
fishy: So if we just return the thrift type, it would be actually `github.com/reddit/baseplate.go/internal/gen-go/reddit/baseplate.IsHealthyProbe` type. Services don't really have access to that type (it's under `internal/`), and their own thrift compiled type would have a different import path and they need a typecasting anyways.

httpbp/healthcheck_test.go
fishy: I don't think this actually matters because these 2 `if`s cannot be satisfied at the same time? Also logically we don't really use `else if` for other test failures, we fail every possible test failures anyways.

httpbp/healthcheck.go
pacejackson: Right, I keep forgetting that 🤦

httpbp/healthcheck_test.go
pacejackson: I think _I_ usually write these sort of cases as an `if/else if` but I agree that it doesn't matter too much, up to you :) .

PR #253, opened 2020-07-08
Fix infinite recursion in unknown experiment error

konradreiche:
The existing error string construction results in a recursive invocation to itself. This fixes it.

### DIFF
--- a/experiments/experiments.go
+++ b/experiments/experiments.go
@@ -357,7 +357,7 @@ type Variant struct {
 type UnknownExperimentError string
 
 func (name UnknownExperimentError) Error() string {
-	return "experiments: experiment with name" + string(name) + " unknown"
+	return "experiments: experiment with name " + string(name) + " unknown"
 }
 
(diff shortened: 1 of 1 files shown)


experiments/experiments.go
```diff
@@ -357,7 +357,7 @@ type Variant struct {
 type UnknownExperimentError string
 
 func (name UnknownExperimentError) Error() string {
-	return fmt.Sprintf("experiments: experiment with name %s unknown", name)
+	return "experiments: experiment with name" + string(name) + " unknown"
```
pacejackson, line 360: You want a space after “name”

konradreiche: > `%v` would cause infinite recursion but `%s` _should_ be fine?

I would have thought so too but I was running into this bug on missing experiments.

konradreiche:
> `%v` would cause infinite recursion but `%s` _should_ be fine?

I would have thought so too but I was running into this bug on missing experiments.

PR #254, opened 2020-07-08
thriftb: Add breaker to BaseplateDefaultClientMiddlewares

fishy:
This the only way we can apply the breaker middleware before retry
middleware while using BaseplateDefaultClientMiddlewares. Without doing
it the breaker middleware can only be added as additional middlewares,
and all additional middlewares are after all the default middlewares.

### DIFF
--- a/thriftbp/BUILD.bazel
+++ b/thriftbp/BUILD.bazel
@@ -18,6 +18,7 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//:go_default_library",
+        "//breakerbp:go_default_library",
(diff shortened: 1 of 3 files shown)


fishy:
> This seems reasonable, looking forward, I wonder if we want to give finer grain control over where custom middlewares land in the chain?

In worst case scenario, services can always get default middlewares, and then add their own middlewares inbetween, then use the custom client pool constructor.

Giving it too much fine-tuning ability has the risk of making the API too complex, which I'm not a fan of.

I think it might make more sense to always try to keep the default ones sensible and leave it there.

PR #255, opened 2020-07-09
redisbp: Add function to publish connection pool stats

pacejackson:


redisbp/monitored_client.go
```diff
@@ -94,6 +100,40 @@ func (f MonitoredCmdableFactory) Close() error {
...
+//
+// The goroutine will be closed when metricsbp.M.Ctx() is Done().
+func (f MonitoredCmdableFactory) MonitorPoolStats(tags metricsbp.Tags) {
+	t := tags.AsStatsdTags()
+	hitsGauge := metricsbp.M.RuntimeGauge(f.name + "-hits").With(t...)
+	missesGauge := metricsbp.M.RuntimeGauge(f.name + "-misses").With(t...)
+	timeoutsGauge := metricsbp.M.RuntimeGauge(f.name + "-timeouts").With(t...)
+	totalConnectionsGauge := metricsbp.M.RuntimeGauge(f.name + "-connections.total").With(t...)
+	idleConnectionsGauge := metricsbp.M.RuntimeGauge(f.name + "-connections.idle").With(t...)
+	staleConnectionsGauge := metricsbp.M.RuntimeGauge(f.name + "-connections.stale").With(t...)
```
fishy, line 114: 🔕 switch `-` and `.` might make more sense?

redisbp/monitored_client.go
```diff
@@ -94,6 +100,40 @@ func (f MonitoredCmdableFactory) Close() error {
...
+	client := f.BuildClient(context.TODO())
+	interval := 10 * time.Second
```
fishy, line 116: this should also use `metricsbp.SysStatsTickerInterval` instead of hardcoded 10s?

redisbp/monitored_client.go
```diff
@@ -94,6 +100,40 @@ func (f MonitoredCmdableFactory) Close() error {
...
+	t := tags.AsStatsdTags()
+	hitsGauge := metricsbp.M.RuntimeGauge(f.name + ".hits").With(t...)
+	missesGauge := metricsbp.M.RuntimeGauge(f.name + ".misses").With(t...)
+	timeoutsGauge := metricsbp.M.RuntimeGauge(f.name + ".timeouts").With(t...)
+	totalConnectionsGauge := metricsbp.M.RuntimeGauge(f.name + ".connections.total").With(t...)
+	idleConnectionsGauge := metricsbp.M.RuntimeGauge(f.name + ".connections.idle").With(t...)
+	staleConnectionsGauge := metricsbp.M.RuntimeGauge(f.name + ".connections.stale").With(t...)
+	client := f.BuildClient(context.TODO())
+
+	go func() {
```
fishy, line 117: 🔕 also, since this is an exported function, I would slightly prefer we make this function an infinite loop, so people using this function need to use goroutine explicitly. and this function should also take in ctx arg for cancellation, but caller can just choose to use `metricsbp.M.Ctx()`.

So caller should do `go factory.MonitorPoolStats(metricsbp.M.Ctx(), tags)` instead.

But I don't feel strongly about that.

pacejackson: I think if we want to do something like that, it _might_ be better to have it pass in `metricsbp.Statsd` and use `metrics.Ctx().Done()`?  Then you call it as `go factory.MonitorPoolStats(metricsbp.M, tags)`

fishy: pass in a `context.Context` means they can choose to have a different way to control when to stop this reporting goroutine. You can't achieve the same with passing in a `metricsbp.Statsd` (well you can, but you need to construct the whole `Statsd` object just for that).

fishy: passing in a ctx to a function doing infinite loop to control when to stop the loop is normal, passing in `Statsd` for that is just... weird and inefficient.

pacejackson: Ok fixed

PR #269, opened 2020-08-05
errorsbp: Add Batch.AddPrefix

fishy:
Also deprecate batchcloser.CloseError, and change
batchcloser.BatchCloser to use AddPrefix instead of CloseError.

### DIFF
--- a/batchcloser/closers.go
+++ b/batchcloser/closers.go
@@ -14,6 +14,8 @@ import (
 //
 // It can be used to inspect both the error that was returned and the io.Closer
 // that caused it.
+//
+// DEPRECATED: This is no longer used and will be removed in a future version.
 type CloseError struct {
 	Cause  error
 	Closer io.Closer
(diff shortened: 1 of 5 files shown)


batchcloser/closers.go
```diff
@@ -84,12 +88,7 @@ type BatchCloser struct {
 func (bc *BatchCloser) Close() error {
 	var errs errorsbp.Batch
 	for _, closer := range bc.closers {
-		if err := closer.Close(); err != nil {
-			errs.Add(CloseError{
-				Cause:  err,
-				Closer: closer,
-			})
-		}
+		errs.AddPrefix(fmt.Sprintf("%#v", closer), closer.Close())
```
pacejackson, line 91: 🔕  I kinda feel like this is a little too magic? It still makes sense to me to check:

```go
if err := closer.Close(); err != nil {
}
```

I know this way works, but it's not clear that the errors are only added if `err` is non-nil from this part of the code.  I don't feel super strongly though if you prefer this way.

fishy: I think we should discourage users from doing the extra nil check and leave that to `Batch`. This leads to simpler code, and Baseplate.go code should serve as an example :)

Also that behavior is documented in `Batch`, so I disagree that it's "too magic" :)

fishy: Oh wait I forgot to document that part. Let me add it real quick 😬

PR #270, opened 2020-08-05
log: Send extra key values to sentry in ErrorWithSentry

fishy:
Also move ErrorWithSentry from log.go to sentry.go.

### DIFF
--- a/log/sentry.go
+++ b/log/sentry.go
@@ -1,12 +1,14 @@
 package log
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
 	"time"
 
 	sentry "github.com/getsentry/sentry-go"
+	"go.uber.org/zap/zapcore"
 )
 
 // DefaultSentryFlushTimeout is the timeout used to call sentry.Flush().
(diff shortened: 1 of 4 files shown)


log/sentry.go
```diff
@@ -88,3 +89,45 @@ func (c closer) Close() error {
...
+			// There's a dangling key, log and append an empty value to avoid panic.
+			Errorw(
+				"Dangling key in ErrorWithSentry",
+				"keysAndValues", keysAndValues,
+			)
+			keysAndValues = append(keysAndValues, "")
+		}
+		hub = hub.Clone()
+		hub.ConfigureScope(func(scope *sentry.Scope) {
+			for i := 0; i < len(keysAndValues); i++ {
```
pacejackson, line 120: Shouldn't this be `i += 2`?

```suggestion
			for i := 0; i < len(keysAndValues); i+=2 {
```

fishy: Yes 🤦

pacejackson: Easy mistake to make 😅

fishy:
:haircut: Changed to ignore zap.Field in ErrorWithSentry instead of break it

fishy:
💇 Added tests because this is a place that's really easy to make mistakes 😬

PR #271, opened 2020-08-07
log: Auto report hostname tag in sentry

fishy:
In k8s world hostname is by default the pod name, so this could be
useful to see if an issue is widespread or isolated on a single pod.

### DIFF
--- a/log/sentry.go
+++ b/log/sentry.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"time"
 
 	sentry "github.com/getsentry/sentry-go"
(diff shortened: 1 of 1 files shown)


log/sentry.go
```diff
@@ -108,17 +124,15 @@ func ErrorWithSentry(ctx context.Context, msg string, err error, keysAndValues .
...
-	}
+	hub = hub.Clone()
+	hub.ConfigureScope(func(scope *sentry.Scope) {
+		if extractKeyValuePairs(keysAndValues, scope.SetTag) {
+			Errorw(
+				"Dangling key in ErrorWithSentry",
+				"keysAndValues", keysAndValues,
+			)
+		}
+	})
```
fishy, line 135: this change was unintentional. reverting.

PR #274, opened 2020-08-18
randbp: Add min length arg to GenerateRandomString

fishy:
Currently GenerateRandomString could generate empty strings. In some
cases we want to avoid empty strings, so add min length arg to control
that.

Also convert its args into a struct. As a result this is a breaking
change.

### DIFF
--- a/randbp/string.go
+++ b/randbp/string.go
@@ -11,12 +11,42 @@ import (
 // GenerateRandomString.
 const Base64Runes = `ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_+/=`
 
-// GenerateRandomString generates a random string with length [0, maxLength),
-// and all characters limited to runes.
+// RandomStringArgs defines the args used by GenerateRandomString.
+type RandomStringArgs struct {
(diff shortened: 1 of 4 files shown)


randbp/string.go
```diff
@@ -11,12 +11,42 @@ import (
 // GenerateRandomString.
 const Base64Runes = `ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_+/=`
 
-// GenerateRandomString generates a random string with length [0, maxLength),
-// and all characters limited to runes.
+// RandomStringArgs defines the args used by GenerateRandomString.
+type RandomStringArgs struct {
+	// Required. If MaxLength <= MinLength it will cause panic.
```
pacejackson, line 16: Shouldn't `MaxLen == MinLen` be fine? I feel like it should only be a problem if `min > max`.

fishy: The length will be in range of `[MinLength, MaxLength)`, the max end is non-inclusive.

fishy:
:haircut: Added a full file example for `GenerateRandomString`.

PR #275, opened 2020-08-31
randbp: Fix typed nil error

fishy:
Also add an unit test to make sure of it.


fishy:
:haircut: `args.R != nil` is sufficient.

PR #277, opened 2020-09-02
retrybp: Add CappedExponentialBackoff option

fishy:
Although by default upstream retry package already use exponential
backoff with ability to cap max delay, the capping only happens after
the exponential backoff value is calculated. When n (number of retries)
is large enough the calculation of the delay could overflow and go back
to 0, resulting in incorrect delay values.

CappedExponentialBackoff supports the extra capping of n used in the
calculation, making sure that the result delay never overflows.

Also adding FixedDelay to avoid confusion.

### DIFF
--- a/retrybp/delay.go
+++ b/retrybp/delay.go
@@ -1,0 +1,117 @@
+package retrybp
+
+import (
+	"math"
+	"time"
+
+	retry "github.com/avast/retry-go"
+
+	"github.com/reddit/baseplate.go/randbp"
+)
+
+// CappedExponentialBackoffArgs defines the args used in
+// CappedExponentialBackoff retry option.
+//
+// All args are optional.
+type CappedExponentialBackoffArgs struct {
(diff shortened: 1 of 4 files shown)


retrybp/delay.go
```diff
@@ -0,0 +1,115 @@
...
+	"math"
+	"time"
+
+	retry "github.com/avast/retry-go"
+
+	"github.com/reddit/baseplate.go/randbp"
+)
+
+// CappedExponentialBackoffArgs defines the args used in
+// CappedExponentialBackoffArgs retry option.
```
pacejackson: Typo here?

```suggestion
// CappedExponentialBackoff retry option.
```

pacejackson (approved):

retrybp/delay.go
fishy: Already fixed that :)

konradreiche (approved):

PR #278, opened 2020-09-08
retrybp: Rename MaxN to MaxExponent

fishy:
Based on the user feedback, the name "MaxN" would confuse user to think
that it works as "max attempts". Rename it to MaxExponent to avoid this
confusion.

### DIFF
--- a/retrybp/delay_test.go
+++ b/retrybp/delay_test.go
@@ -31,7 +31,7 @@ func TestActualMaxNQuick(t *testing.T) {
 		if actualBase <= 0 {
 			actualBase = 1
 		}
-		n := actualMaxN(time.Duration(base))
+		n := actualMaxExponent(time.Duration(base))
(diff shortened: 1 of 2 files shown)


retrybp/delay_test.go
```diff
@@ -114,19 +114,19 @@ func TestCappedExponentialBackoff(t *testing.T) {
 			max:     time.Duration(math.MaxInt64),
 		},
 		{
-			label:   "max-n-too-high",
-			n:       9999,
-			initial: time.Millisecond,
-			maxN:    9998,
-			max:     time.Duration(math.MaxInt64),
+			label:       "max-n-too-high",
```
pacejackson, line 117: should we change these labels to `max-exponent-...`?

pacejackson (approved):

fishy:
cc @andizzle

konradreiche (approved):

PR #281, opened 2020-09-16
Add struct arg with optional pre- and postShutdown closers to Serve()

suversky:
Add an argument struct to Serve(), allowing for optional slices of `io.Closers` that can shut down gracefully before and after the server.

-----

This is a precursor for adding queue consumers, so that they can be cleanly shutdown alongside the server.

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -77,17 +77,42 @@ type Server interface {
 	Serve() error
 }
 
-// Serve runs the given Server until it is given an external shutdown signal
-// using runtimebp.HandleShutdown to handle the signal and shut down the server
-// and any provided closers gracefully.  Returns the (possibly nil) error
-// returned by "Close", or context.DeadlineExceeded if it times out.
+// ServeArgs provides a list of arguments to Serve.
+//    server: Mandatory. Indicates the Server that should be run until
+//            receiving a shutdown signal.
+//    preShutdown: Optional. Indicates a slice of io.Closers that should
+//						be gracefully shut down before the server upon receipt of a
+//            shutdown signal.
+//    preShutdown: Optional. Indicates a slice of io.Closers that should
+//						be gracefully shut down after the server upon receipt of a
+//            shutdown signal.
+type ServeArgs struct {
+	server       Server
+	preShutdown  []io.Closer
+	postShutdown []io.Closer
+}
+
+// Serve runs the given Server until it is given an external shutdown signal.
+//
+// It uses runtimebp.HandleShutdown to handle the signal and gracefully shut
+// down, in order:
+//     * any provided preShutdown closers,
+//     * the server, and
+//     * any provided postShutdown closers.
+//
+// Returns the (possibly nil) error returned by "Close", or
+// context.DeadlineExceeded if it times out.
 //
 // If a StopTimeout is configure, Serve will wait for that duration for the
 // server to stop before timing out and returning to force a shutdown.
 //
 // This is the recommended way to run a Baseplate Server rather than calling
 // server.Start/Stop directly.
-func Serve(ctx context.Context, server Server, closers ...io.Closer) error {
+func Serve(ctx context.Context, serveArgs ServeArgs) error {
+	server := serveArgs.server
+	preShutdown := serveArgs.preShutdown
+	postShutdown := serveArgs.postShutdown
+
 	// Initialize a channel to return the response from server.Close() as our
 	// return value.
 	shutdownChannel := make(chan error)
@@ -124,16 +149,14 @@ func Serve(ctx context.Context, server Server, closers ...io.Closer) error {
 			// Initialize a channel to pass the result of server.Close().
 			closeChannel := make(chan error)
 
-			// Tell the server to close.
+			// Tell the server and any provided closers to close.
 			//
 			// This is a blocking call, so it is called in a separate goroutine.
 			go func() {
-				bc := batchcloser.New(closers...)
-				be := errorsbp.Batch{}
-
-				be.AddPrefix("closers", bc.Close())
-				be.AddPrefix("server", server.Close())
-				closeChannel <- be.Compile()
+				bc := batchcloser.New(preShutdown...)
+				bc.Add(server)
+				bc.Add(postShutdown...)
+				closeChannel <- bc.Close()
 			}()
 
 			// Declare the error variable we will use later here so we can set it to
--- a/baseplate.go
+++ b/baseplate.go
@@ -78,16 +78,16 @@ type Server interface {
 }
 
 // Serve runs the given Server until it is given an external shutdown signal
-// using runtimebp.HandleShutdown to handle the signal and shut down the
-// server gracefully.  Returns the (possibly nil) error returned by "Close" or
-// context.DeadlineExceeded if it times out.
+// using runtimebp.HandleShutdown to handle the signal and shut down the server
+// and any provided closers gracefully.  Returns the (possibly nil) error
+// returned by "Close", or context.DeadlineExceeded if it times out.
 //
 // If a StopTimeout is configure, Serve will wait for that duration for the
 // server to stop before timing out and returning to force a shutdown.
 //
 // This is the recommended way to run a Baseplate Server rather than calling
 // server.Start/Stop directly.
-func Serve(ctx context.Context, server Server) error {
+func Serve(ctx context.Context, server Server, closers ...io.Closer) error {
 	// Initialize a channel to return the response from server.Close() as our
 	// return value.
 	shutdownChannel := make(chan error)
@@ -128,7 +128,12 @@ func Serve(ctx context.Context, server Server) error {
 			//
 			// This is a blocking call, so it is called in a separate goroutine.
 			go func() {
-				closeChannel <- server.Close()
+				bc := batchcloser.New(closers...)
+				be := errorsbp.Batch{}
+
+				be.AddPrefix("closers", bc.Close())
+				be.AddPrefix("server", server.Close())
+				closeChannel <- be.Compile()
 			}()
 
 			// Declare the error variable we will use later here so we can set it to
--- a/baseplate.go
+++ b/baseplate.go
@@ -96,9 +96,12 @@ type ServeArgs struct {
 //
 // It uses runtimebp.HandleShutdown to handle the signal and gracefully shut
 // down, in order:
-//     * any provided PreShutdown closers,
-//     * the Server, and
-//     * any provided PostShutdown closers.
+//
+// * any provided PreShutdown closers,
+//
+// * the Server, and
+//
+// * any provided PostShutdown closers.
 //
 // Returns the (possibly nil) error returned by "Close", or
 // context.DeadlineExceeded if it times out.
--- a/baseplate.go
+++ b/baseplate.go
@@ -79,16 +79,16 @@ type Server interface {
 
 // ServeArgs provides a list of arguments to Serve.
 type ServeArgs struct {
-	// Server: Mandatory. Indicates the Server that should be run until receiving
-	// a shutdown signal.
+	// Server is the Server that should be run until receiving a shutdown signal.
+	// This is a required argument and baseplate.Serve will panic if this is nil.
 	Server Server
 
-	// PreShutdown: Optional. Indicates a slice of io.Closers that should be
-	// gracefully shut down before the server upon receipt of a shutdown signal.
+	// PreShutdown is an optional slice of io.Closers that should be gracefully
+	// shut down before the server upon receipt of a shutdown signal.
 	PreShutdown []io.Closer
 
-	// PostShutdown: Optional. Indicates a slice of io.Closers that should be
-	// gracefully shut down after the server upon receipt of a shutdown signal.
+	// PostShutdown is an optional slice of io.Closers that should be gracefully
+	// shut down after the server upon receipt of a shutdown signal.
 	PostShutdown []io.Closer
 }
 
--- a/baseplate.go
+++ b/baseplate.go
@@ -78,27 +78,27 @@ type Server interface {
 }
 
 // ServeArgs provides a list of arguments to Serve.
-//    server: Mandatory. Indicates the Server that should be run until
-//            receiving a shutdown signal.
-//    preShutdown: Optional. Indicates a slice of io.Closers that should
-//						be gracefully shut down before the server upon receipt of a
-//            shutdown signal.
-//    preShutdown: Optional. Indicates a slice of io.Closers that should
-//						be gracefully shut down after the server upon receipt of a
-//            shutdown signal.
 type ServeArgs struct {
-	server       Server
-	preShutdown  []io.Closer
-	postShutdown []io.Closer
+	// Server: Mandatory. Indicates the Server that should be run until receiving
+	// a shutdown signal.
+	Server Server
+
+	// PreShutdown: Optional. Indicates a slice of io.Closers that should be
+	// gracefully shut down before the server upon receipt of a shutdown signal.
+	PreShutdown []io.Closer
+
+	// PostShutdown: Optional. Indicates a slice of io.Closers that should be
+	// gracefully shut down after the server upon receipt of a shutdown signal.
+	PostShutdown []io.Closer
 }
 
 // Serve runs the given Server until it is given an external shutdown signal.
 //
 // It uses runtimebp.HandleShutdown to handle the signal and gracefully shut
 // down, in order:
-//     * any provided preShutdown closers,
-//     * the server, and
-//     * any provided postShutdown closers.
+//     * any provided PreShutdown closers,
+//     * the Server, and
+//     * any provided PostShutdown closers.
 //
 // Returns the (possibly nil) error returned by "Close", or
 // context.DeadlineExceeded if it times out.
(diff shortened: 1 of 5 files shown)


baseplate.go
```diff
@@ -128,7 +128,12 @@ func Serve(ctx context.Context, server Server) error {
 			//
 			// This is a blocking call, so it is called in a separate goroutine.
 			go func() {
-				closeChannel <- server.Close()
+				bc := batchcloser.New(closers...)
+				be := errorsbp.Batch{}
```
fishy, line 132: style nit: prefer `var be errorsbp.Batch`.

but `server` itself is also a closer, so it's actually easier to just add server into the batch closer, e.g.:

```
bc := batchcloser.New(closers...)
bc.Add(server)
closeChannel <- bc.Close()
```

fishy:
Oh also I just noticed this part:

>This is a precursor for adding queue consumers, so that they can be cleanly shutdown alongside the server.

In that case, the order of close does matter. I think for the default implementation we should close the server before the additional closers, because the server is usually the main drive of all the sub goroutines/tasks. And that behavior should be clearly documented in the function comment doc, so if for some service it makes more sense to close the server the last, they know that they need to handle that manually.

pacejackson:
>  I think for the default implementation we should close the server before the additional closers, because the server is usually the main drive of all the sub goroutines/tasks.

Interesting, I was thinking the _other_ way, but mostly just because the "server" is what drives the healthcheck but I can see it this way too...

fishy:
> > I think for the default implementation we should close the server before the additional closers, because the server is usually the main drive of all the sub goroutines/tasks.
> 
> Interesting, I was thinking the _other_ way, but mostly just because the "server" is what drives the healthcheck but I can see it this way too...

so the case you are talking about is that the queue consumer is the main "server" driving tasks, and the thrift "server" is only for health check? in that case I think we should actually make the queue consumer as the server and thrift server as the additional closer.

here the api doesn't specify what are those closers, so this thing gets muddy fast :)  in our use case we often have thrift server also running kafka producer, so if we close the kafka producer before closing the thrift server, thrift requests can come in after the kafka producer is already closed, and we might try to produce after the producer is closed, and that will cause troubles.

pacejackson:
Part of the point was to avoid writing a specific "queue consumer server", you just use a basic HTTP or Thrift (we use HTTP in python) "server" that only serves healthchecks and include the queue consumer as an extra closer.

fishy:
> Part of the point was to avoid writing a specific "queue consumer server", you just use a basic HTTP or Thrift (we use HTTP in python) "server" that only serves healthchecks and include the queue consumer as an extra closer, but maybe we can't get away from that?

Ok in that case the order is fine,but it does need to be documented

baseplate.go
```diff
@@ -78,16 +78,16 @@ type Server interface {
...
+// and any provided closers gracefully.  Returns the (possibly nil) error
+// returned by "Close", or context.DeadlineExceeded if it times out.
 //
 // If a StopTimeout is configure, Serve will wait for that duration for the
 // server to stop before timing out and returning to force a shutdown.
 //
 // This is the recommended way to run a Baseplate Server rather than calling
 // server.Start/Stop directly.
-func Serve(ctx context.Context, server Server) error {
+func Serve(ctx context.Context, server Server, closers ...io.Closer) error {
```
pacejackson, line 90: In addition to some extra documentation, maybe we should name these closers something more descriptive?  Like "onShutdown" or maybe even "preShutdown"?

pacejackson (changes requested):

fishy:
Actually looking at it more, when people uses `baseplate.Serve` this is the shutdown handler, so if they do have anything that need to be closed after `server.Close` they can't do that. So I can see we likely will need to add something like `PostClosers` in the future.

I'd suggest since we already have breaking changes lined up, it's probably time to change that function's args into a struct, and the closers we add now are `PreClosers`. We don't need to add `PostClosers` now but when we add them in the future that won't be a breaking change.

suversky:
Sounds good to me! I figured the variadic approach would be good for a first stab at it and I'd use the struct approach if you more-knowledgeable folks gave me a reason to :slightly_smiling_face:

baseplate.go
```diff
@@ -77,17 +77,42 @@ type Server interface {
...
-// server gracefully.  Returns the (possibly nil) error returned by "Close" or
+// ServeArgs provides a list of arguments to Serve.
+//    server: Mandatory. Indicates the Server that should be run until
+//            receiving a shutdown signal.
+//    preShutdown: Optional. Indicates a slice of io.Closers that should
+//						be gracefully shut down before the server upon receipt of a
+//            shutdown signal.
+//    preShutdown: Optional. Indicates a slice of io.Closers that should
+//						be gracefully shut down after the server upon receipt of a
+//            shutdown signal.
```
suversky, line 88: I'm not actually sure how formatting (and tabs vs spaces) works in whatever utility is generating API documentation from this, so I'd love feedback on that!

baseplate.go
```diff
@@ -124,11 +149,14 @@ func Serve(ctx context.Context, server Server) error {
...
 
-			// Tell the server to close.
+			// Tell the server and any provided closers to close.
 			//
 			// This is a blocking call, so it is called in a separate goroutine.
 			go func() {
-				closeChannel <- server.Close()
+				bc := batchcloser.New(preShutdown...)
+				bc.Add(server)
+				bc.Add(postShutdown...)
```
suversky, line 158: I'm perhaps naively relying on the assumption that `preShutdown` and `postShutdown` will work appropriately if unset (presumably either `nil` or an empty slice, and in either case not problematic for `batchcloser.New` or `batchcloser.Add`.

baseplate.go
suversky: Interestingly it shows up properly in the `/files` endpoint but improperly in this snippet.

I used spaces when laying out the comment but I guess `go fmt` might have tabified them.

pacejackson: I believe tabs make things look like code and yiou should use spaces.  Also, you don't want to do this sort of spacing, i.e.:

```go
// ServeArgs provides a list of arguments to Serve.
// server: Mandatory. Indicates the Server that should be run until
// receiving a shutdown signal.
```

baseplate.go
```diff
@@ -77,17 +77,42 @@ type Server interface {
...
+//    preShutdown: Optional. Indicates a slice of io.Closers that should
+//						be gracefully shut down before the server upon receipt of a
+//            shutdown signal.
+//    preShutdown: Optional. Indicates a slice of io.Closers that should
+//						be gracefully shut down after the server upon receipt of a
+//            shutdown signal.
+type ServeArgs struct {
+	server       Server
+	preShutdown  []io.Closer
+	postShutdown []io.Closer
```
pacejackson, line 92: 2 things:

1. You actually want to document the fields just above the field itself, not when documenting the struct itself.
2. These need to be capitalized so they are exported, as it is, only things in this package actually have access to these fields (so they could never actually be set by a service).

baseplate.go
pacejackson: I think this should be fine.

baseplate.go
```diff
@@ -124,11 +149,14 @@ func Serve(ctx context.Context, server Server) error {
 			// Initialize a channel to pass the result of server.Close().
 			closeChannel := make(chan error)
 
-			// Tell the server to close.
+			// Tell the server and any provided closers to close.
 			//
 			// This is a blocking call, so it is called in a separate goroutine.
 			go func() {
-				closeChannel <- server.Close()
+				bc := batchcloser.New(preShutdown...)
```
pacejackson, line 156: style: I would just use `args.preShutdown...` directly here rather than assigning it to a variable that is only used once.

```go
				bc.Add(args.postShutdown...)
```

baseplate.go
pacejackson: Same style comment as above for this line:

```go
				bc.Add(args.postShutdown...)
```

baseplate.go
```diff
@@ -77,17 +77,42 @@ type Server interface {
...
+// Returns the (possibly nil) error returned by "Close", or
 // context.DeadlineExceeded if it times out.
 //
 // If a StopTimeout is configure, Serve will wait for that duration for the
 // server to stop before timing out and returning to force a shutdown.
 //
 // This is the recommended way to run a Baseplate Server rather than calling
 // server.Start/Stop directly.
-func Serve(ctx context.Context, server Server) error {
+func Serve(ctx context.Context, serveArgs ServeArgs) error {
```
pacejackson, line 111: style nit, calling this "serveArgs" is redundant, I think just "args" is simpler.

pacejackson (changes requested):
Can you also add some tests to https://github.com/reddit/baseplate.go/blob/fc4a6864ce01ce5322b65dab8552558dea7278fc/baseplate_test.go#L89 ?

baseplate.go
suversky: Ooh, I didn't realize that capitalization exporting applied to fields of structs as well as the struct itself. Clever!

baseplate.go
```diff
@@ -77,17 +77,40 @@ type Server interface {
...
+	PostShutdown []io.Closer
+}
+
+// Serve runs the given Server until it is given an external shutdown signal.
+//
+// It uses runtimebp.HandleShutdown to handle the signal and gracefully shut
+// down, in order:
+//     * any provided PreShutdown closers,
+//     * the Server, and
+//     * any provided PostShutdown closers.
```
fishy, line 101: super nit: this way the 3 bullet points will be rendered as code block in the godoc site, which works but might look weird. it's slightly better to not put 4 extra spaces before them, and add blank lines between them, e.g.:

```
// other comments
//
// * bullet 1
//
// * bullet 2
```

fishy (changes requested):
You need to fix the test, but otherwise lgtm.

baseplate.go
suversky: Great nit, and thanks for pointing me to the name `godoc` so I know what to google for now :P

httpbp/example_server_test.go
```diff
@@ -107,8 +107,9 @@ func ExampleNewBaseplateServer() {
 		Endpoints:   handlers.Endpoints(),
 		Middlewares: []httpbp.Middleware{loggingMiddleware},
 	})
+	args := baseplate.ServeArgs{Server: server}
```
fishy, line 110: It's a bit weird to define `args` before the `err` check. In general I'd prefer to just inline it in the line it's used, especially in examples. Same for the thrift example.

suversky:
The last thing I've got to do is add some tests for the `PreShutdown` and `PostShutdown` closers, which I'm still trying to reason about.

fishy:
> The last thing I've got to do is add some tests for the `PreShutdown` and `PostShutdown` closers, which I'm still trying to reason about.

We should add that test to make sure that we are closing them in the correct order.

baseplate.go
```diff
@@ -77,17 +77,43 @@ type Server interface {
 	Serve() error
 }
 
-// Serve runs the given Server until it is given an external shutdown signal
-// using runtimebp.HandleShutdown to handle the signal and shut down the
-// server gracefully.  Returns the (possibly nil) error returned by "Close" or
+// ServeArgs provides a list of arguments to Serve.
+type ServeArgs struct {
+	// Server: Mandatory. Indicates the Server that should be run until receiving
+	// a shutdown signal.
```
pacejackson, line 83: You should write comments like a normal sentence so this should be more like:

```suggestion
	// Server is the Server that should be run until receiving a shutdown signal.
	// This is a required argument and baseplate.Serve will panic if this is nil.
```

baseplate.go
```diff
@@ -77,17 +77,43 @@ type Server interface {
...
+	Server Server
+
+	// PreShutdown: Optional. Indicates a slice of io.Closers that should be
+	// gracefully shut down before the server upon receipt of a shutdown signal.
```
pacejackson, line 87: Similar to above:

```suggestion
	// PreShutdown is an optional slice of io.Closers that should be gracefully shut down
	// before the server upon receipt of a shutdown signal.
```

pacejackson (changes requested):

baseplate_test.go
```diff
@@ -149,6 +150,72 @@ func TestServe(t *testing.T) {
...
+			}()
+
+			time.Sleep(time.Millisecond)
+			p, err := os.FindProcess(syscall.Getpid())
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			p.Signal(os.Interrupt)
+			err = <-ch
```
suversky, line 199: I cribbed this from `TestServe`, but am not actually too sure if this block (and the provided `args.Server`) are correct for testing a happy shutdown path.

baseplate_test.go
```diff
@@ -145,6 +150,72 @@ func TestServe(t *testing.T) {
...
+		"serve closer order",
+		func(t *testing.T) {
+			ch := make(chan error)
+
+			go func() {
+				// Run Serve in a goroutine since it is blocking
+				ch <- baseplate.Serve(
+					context.Background(),
+					args,
+				)
```
pacejackson, line 189: 2 style nits:

1. This should be fine to fit on one line...
2. Modulo that you should just define `args` inline here

```suggestion
				ch <- baseplate.Serve(context.Background(), baseplate.ServeArgs{
					Server:       newWaitServer(t, bp, time.Millisecond),
					PreShutdown:  []io.Closer{pre},
					PostShutdown: []io.Closer{post},
				})
```

baseplate_test.go
```diff
@@ -145,6 +150,72 @@ func TestServe(t *testing.T) {
...
+
+	args := baseplate.ServeArgs{
+		Server:       newWaitServer(t, bp, time.Millisecond),
+		PreShutdown:  []io.Closer{pre},
+		PostShutdown: []io.Closer{post},
+	}
+
+	t.Run(
+		"serve closer order",
+		func(t *testing.T) {
```
pacejackson, line 181: This probably doesn't need to be it's own subtest, it's just this one test and I don't think you really need to add more.

baseplate_test.go
pacejackson: Seems fine to me, do you have any specific concerns?

pacejackson (approved):
LGTM modulo a couple style updates in the new test.

baseplate_test.go
fishy: `err` is assigned but never used in its current state? Do you mean to check it? if not you can just do `<-ch` without assigning it to anything.

baseplate_test.go
```diff
@@ -145,6 +150,72 @@ func TestServe(t *testing.T) {
...
+			p, err := os.FindProcess(syscall.Getpid())
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			p.Signal(os.Interrupt)
+			err = <-ch
+
+			if len(pre.ts) > 1 {
+				t.Fatalf("PreShutdown called too many times: expected 1, got %#v", len(pre.ts))
```
fishy, line 202: nit: for ints (`len()`) you don't really need `%#v`. just `%v` or `%d` will do.

baseplate_test.go
```diff
@@ -145,6 +150,72 @@ func TestServe(t *testing.T) {
...
+			}
+			if len(post.ts) > 1 {
+				t.Fatalf("PostShutdown called too many times: expected 1, got %#v", len(post.ts))
+			}
+
+			if !pre.ts[0].Before(post.ts[0]) {
+				t.Fatalf(
+					"PreShutdown finished after PostShutdown: pre: %#v, post: %#v",
```
fishy, line 210: for `time.Time` types, `%v` will work better than `%#v`.

baseplate_test.go
```diff
@@ -145,6 +150,72 @@ func TestServe(t *testing.T) {
...
+			time.Sleep(time.Millisecond)
+			p, err := os.FindProcess(syscall.Getpid())
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			p.Signal(os.Interrupt)
+			err = <-ch
+
+			if len(pre.ts) > 1 {
```
fishy, line 201: shouldn't this be `!=` instead of `>`? you want to fail it if it's not called as well.

suversky:
Thanks so much for all the feedback!

baseplate_test.go
```diff
@@ -145,6 +150,64 @@ func TestServe(t *testing.T) {
...
+
+	go func() {
+		// Run Serve in a goroutine since it is blocking
+		ch <- baseplate.Serve(context.Background(), args)
+	}()
+
+	time.Sleep(time.Millisecond)
+	p, err := os.FindProcess(syscall.Getpid())
+	if err != nil {
+		t.Fatal(err)
```
fishy, line 189: If we failed to get the pid, the already started goroutine will just be leaked (until all tests are done). I'd prefer to get pid first, fatal if failed, then start goroutine, sleep, and signal.

baseplate_test.go
```diff
@@ -145,6 +150,64 @@ func TestServe(t *testing.T) {
...
+
+	if len(pre.ts) != 1 {
+		t.Fatalf("Unexpected number of PreShutdown calls: expected 1, got %v", len(pre.ts))
+	}
+	if len(post.ts) != 1 {
+		t.Fatalf("Unexpected number of PostShutdown calls: expected 1, got %v", len(post.ts))
+	}
+
+	if !pre.ts[0].Before(post.ts[0]) {
+		t.Fatalf(
```
fishy, line 203: I'd prefer not to overly use Fatal over Error in tests. The previous ones are fine: We can't continue the test if we can't trigger graceful shutdown, getting `pref.ts[0]` could panic if `len(pre.ts) < 1`, but here this is just an error, not some thing fatal.

fishy (approved):
The two remaining comments are mostly nitpicks and not blocking.

baseplate_test.go
fishy: 🔕 We probably should check `err != nil` though?

fishy (approved):

PR #282, opened 2020-09-22
log: Log trace_id as string instead of int

fishy:
Some log ingesters have trouble with uint64 that's larger than max
signed int64. Log them as string to avoid troubles.

### DIFF
--- a/log/context.go
+++ b/log/context.go
@@ -2,6 +2,7 @@ package log
 
 import (
 	"context"
+	"strconv"
 
 	"go.uber.org/zap"
 )
@@ -30,7 +31,7 @@ func Attach(ctx context.Context, args AttachArgs) context.Context {
 	var kv []interface{}
 
 	if args.TraceID != 0 {
-		kv = append(kv, traceIDKey, args.TraceID)
+		kv = append(kv, traceIDKey, strconv.FormatUint(args.TraceID, 10))
 	}
 
 	logger := C(ctx)


log/context.go
```diff
@@ -21,16 +22,28 @@ const (
...
+	// numbers larger than max signed int64.
+	TraceID       uint64
+	StringTraceID bool
 }
 
 // Attach attaches a logger with data extracted from args into the context
 // object.
 func Attach(ctx context.Context, args AttachArgs) context.Context {
-	var kv []interface{}
+	kv := make([]interface{}, 0, 2)
```
pacejackson, line 39: what's the reasoning behind this? Just that we're almost always going to have a TraceID so it's probably better to just pre-allocate this?

log/context.go
```diff
@@ -21,16 +22,28 @@ const (
...
-	TraceID uint64
+	// The trace id of the span.
+	//
+	// If StringTraceID is set to true,
+	// it will be logged as string instead of int.
+	// The conversion to string will make it slightly slower,
+	// but avoids troubles when some of the log ingesters cannot deal with uint64
+	// numbers larger than max signed int64.
+	TraceID       uint64
+	StringTraceID bool
```
pacejackson, line 33: Maybe we should just string-ify it always?

pacejackson (commented):
I left some questions but no blockers.

log/context.go
fishy: I reverted this part back. In the current situation we at most do one allocation of 2 (because it's a single `append`, even we appended two items it won't be two allocations), but when there's no trace id set we can avoid this allocation. With `make` we always do the same allocation.

log/context.go
fishy: That was my first version. But in case in the future our log ingester fixed the issue and can handle uint64 fine, we would want to avoid the extra string conversion to save some performance, so leave it flexible seems slightly more valuable?

pacejackson: I imagine the performance hit isn't that much?  I also tend to err on the side of making things simpler when we don't need the flexibility and adding flexibility later if we find we need it.

pacejackson (approved):
I have a mild preference for removing the option and just always string-ifying but if you feel otherwise this is fine.

log/context.go
fishy: It's ~55ns per request (since this is usually done once per request):

```
$ go test -bench . -benchmem
goos: linux
goarch: amd64
BenchmarkFormatUint/baseline-12                 256251235                4.50 ns/op            0 B/op          0 allocs/op
BenchmarkFormatUint/format-uint-12              19179976                59.2 ns/op            31 B/op          1 allocs/op
PASS
$ cat bench_test.go 
package main

import (
        "math/rand"
        "strconv"
        "testing"
)

func BenchmarkFormatUint(b *testing.B) {
        r := rand.New(rand.NewSource(0))

        r.Seed(0)
        b.Run(
                "baseline",
                func(b *testing.B) {
                        for i := 0; i < b.N; i++ {
                                r.Uint64()
                        }
                },
        )

        r.Seed(0)
        b.Run(
                "format-uint",
                func(b *testing.B) {
                        for i := 0; i < b.N; i++ {
                                strconv.FormatUint(r.Uint64(), 10)
                        }
                },
        )
}
```

is it "much" is up to your definition :) I don't feel strongly either way either. will remove it.

konradreiche (approved):

PR #284, opened 2020-09-23
Add path to file watcher error & disallow empty paths

konradreiche:
I ran into this issue myself or have seen other run into it. Adding a path to the file watcher error message and disallowing empty paths to begin with could help give more directions when setting these things up or debugging them.

### DIFF
--- a/filewatcher/filewatcher.go
+++ b/filewatcher/filewatcher.go
@@ -186,7 +186,7 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
 		select {
 		default:
 		case <-ctx.Done():
-			return nil, fmt.Errorf("filewatcher: context cancelled while waiting for file to load. %w", ctx.Err())
+			return nil, fmt.Errorf("filewatcher: context cancelled while waiting for file under %q to load. %w", cfg.Path, ctx.Err())
 		}
 
 		var err error
--- a/secrets/errors.go
+++ b/secrets/errors.go
@@ -9,6 +9,9 @@ import (
 // encoding in the secrets.json file.
 var ErrInvalidEncoding = errors.New("secrets: invalid encoding, expected identity, base64 or empty")
 
+// ErrEmptySecretKey is returned when the path for a secret is empty.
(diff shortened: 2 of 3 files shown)


fishy (approved):
This probably should be split into 2 commits? But I don't feel super strongly about that so :shrug:

filewatcher/filewatcher.go
```diff
@@ -186,7 +186,7 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
 		select {
 		default:
 		case <-ctx.Done():
-			return nil, fmt.Errorf("filewatcher: context cancelled while waiting for file to load. %w", ctx.Err())
+			return nil, fmt.Errorf("filewatcher: context cancelled while waiting for file under %s to load. %w", cfg.Path, ctx.Err())
```
pacejackson, line 189: nit, `%q` rather than `%s` will probably give a clearer message.

secrets/errors.go
```diff
@@ -9,6 +9,9 @@ import (
 // encoding in the secrets.json file.
 var ErrInvalidEncoding = errors.New("secrets: invalid encoding, expected identity, base64 or empty")
 
+// ErrEmptySecretPath is returned when the key for a secret is empty.
+var ErrEmptySecretPath = errors.New("secrets: secret path cannot be empty")
```
pacejackson, line 13: Maybe this should be "key" rather than "path"?  Path sounds like the path to the secrets file IMO.

konradreiche: Agreed, as I was re-reading it I was thinking of the file path.

fishy: we do call them "secret path" in some context. our terminology is confusing 😞

fishy (approved):

secrets/errors.go
konradreiche: I did rename the parameter of those methods too.

pacejackson (approved):

fishy (approved):

PR #286, opened 2020-09-30
Kafkabp: Add consumer implementation using sarama

suversky:
Cribbed (very) heavily from the existing ads-go implementation. Expanded the tests a bit, cleaned up potential data races in the tests, and cleaned up the config somewhat, but the primary functionality should be about the same.


kafkabp/consumer.go
```diff
@@ -0,0 +1,262 @@
...
+	return p
+}
+
+func (kc *consumer) getPartitionConsumers() []sarama.PartitionConsumer {
+	pc, _ := kc.partitionConsumers.Load().([]sarama.PartitionConsumer)
+	return pc
+}
+
+// Reset recreates the consumer and assigns partitions.
+func (kc *consumer) Reset() error {
```
pacejackson, line 76: Reset probably shouldn't be exported since it doesn't seem like something you would want people to just be able to call? (granted, it's effectively not exported because `consumer` isn't but it is still probably better from a hygiene/setting a good example perspective)

kafkabp/config.go
```diff
@@ -0,0 +1,23 @@
...
+	Topic   string
+
+	// Optional. Defaults to DefaultSaramaConfig.
+	SaramaConfig *sarama.Config
+
+	// Optional. If set, overrides the client ID set in SaramaConfig.
+	ClientID string
+
+	// Optional. Defaults to OffsetOldest.
+	Offset KafkaOffset
```
pacejackson, line 18: For getting this from YAML, we might want to make it a `string` and have it support 2 values, "oldest" and "newest" and then convert that to the appropriate sarama value?

kafkabp/sarama_wrapper.go
```diff
@@ -0,0 +1,35 @@
+package kafkabp
+
+import (
+	"errors"
+
+	"github.com/Shopify/sarama"
+)
+
+// KafkaOffset returns supported Kafka offsets.
+type KafkaOffset int64
```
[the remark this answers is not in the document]
fishy: The main reason we did this in the internal code is to make this kind of like an enum.

sarama actually allows you to use some arbitrary offset value that's neither oldest nor newest, but in our constructor we convert everything that's not newest to oldest (essentially disallow all the arbitrary values), so give it a typedef and make it more like an enum helps (but mostly in a documentation way).

kafkabp/config.go
suversky: Ooh, that's a better idea than the typedef!

fishy: note that you don't have to use string type to do that. as long as the typedef implemented [`yaml.Unmarshaler`](https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshaler) it would work.

but yaml v2's `Unmarshaler` interface is defined so confusing I have no idea how it's supposed to work even after reading their code. [v3's `Unmarshaler`](https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshaler) is much better.

kafkabp/consumer.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+// ConsumerMessage represents a kafka consumer message.
+type ConsumerMessage struct {
+	Key, Value []byte
+	Topic      string
+	Partition  int32
+	Offset     int64
+	Timestamp  time.Time
+}
```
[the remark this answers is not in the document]
suversky: I suppose there's no harm in that, if we don't mind baking in the dependency on Sarama a bit more.

kafkabp/consumer.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+		kc.partitionConsumers.Store(partitionConsumers)
+
+		wg.Wait()
+
+		if atomic.LoadInt64(&kc.closed) != 0 {
+			return nil
+		}
+
+		if err := kc.reset(); err != nil {
```
pacejackson, line 220: maybe add a comment here like "Sarama closed the message channel due to a partition rebalance, reset the consumer and restart the goroutines."  @fishy does that sound right?

kafkabp/consumer.go
```diff
@@ -0,0 +1,229 @@
...
+				for err := range pc.Errors() {
+					errorsFunc(err)
+				}
+			}(partitionConsumer)
+		}
+		kc.partitionConsumers.Store(partitionConsumers)
+
+		wg.Wait()
+
+		if atomic.LoadInt64(&kc.closed) != 0 {
```
pacejackson, line 216: maybe add a comment to the effect of "close was called, shutting down the consumer"

kafkabp/consumer.go
```diff
@@ -0,0 +1,229 @@
...
+			return err
+		}
+
+		partitions, err := c.Partitions(kc.cfg.Topic)
+		if err != nil {
+			return err
+		}
+
+		kc.consumer.Store(c)
+		kc.partitions.Store(partitions)
```
pacejackson, line 119: @fishy This seems kind of weird to me. Using atomic values doesn't seem like the right way to protect these, if reset() got called by 2 separate goroutines at the same time, this would guard concurrent access to each one individually, but could they potentially be out of sync?  Like where goroutine A sets the consumer but goroutine B sets partitions?

This feels like something that should guarded by a `RWMutex` or rather than using `atomic.Value` or is there something I'm missing.

kafkabp/consumer.go
```diff
@@ -0,0 +1,229 @@
...
+// ConsumeErrorFunc is a function type for consuming consumer errors.
+type ConsumeErrorFunc func(err error)
+
+// consumer is an instance of a Kafka consumer.
+type consumer struct {
+	cfg ConsumerConfig
+
+	consumer           atomic.Value // sarama.Consumer
+	partitions         atomic.Value // []int32
+	partitionConsumers atomic.Value // []sarama.PartitionConsumer
```
pacejackson, line 28: @fishy do you know why we are using an atomic value for this?  It seems like it might be because it's read by `Close` in addition to being read/set in `Consume`?

🔕  Is this how we should even be handling this? This seems like it might be an over-use of the `atomic` package which has some pretty strong opinions about when to use it.

kafkabp/consumer.go
fishy: that comment is actually already on Lines 190-193 :)

kafkabp/consumer.go
fishy: so we use individual atomic values mainly for guarding against read and write at the same time. two writes at the same time could happen in theory, but it won't in current implementation (we have only two places calling `reset`, one is in initialization and one is in the infinite loop), so we didn't guard against that.

you can have a struct to store both in the same atomic value. I don't feel strongly either way.

kafkabp/config_test.go
```diff
@@ -0,0 +1,17 @@
+package kafkabp
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
pacejackson, line 6: We don't use testify in baseplate.go, we just check things directly.

kafkabp/config_test.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func TestConfig(t *testing.T) {
+	var cfg ConsumerConfig
+	// Config with invalid Offset should not create a new consumer and throw
+	// ErrOffsetInvalid
+	cfg.Offset = "fanciest"
+	sc, err := cfg.NewSaramaConfig()
+	assert.Nil(t, sc)
```
pacejackson, line 15: so this would be:

```suggestion
	if sc != nil {
		t.Errorf("expected config to be nil, got %v", sc)
	}
```

kafkabp/config_test.go
```diff
@@ -0,0 +1,17 @@
...
+	assert.Equal(t, ErrOffsetInvalid, err)
```
pacejackson, line 16: and this would be:

```suggestion
	if !errors.Is(err, ErrOffsetInvalid) {
		t.Errorf("expected error %v, got %v", ErrOffsetInvalid, err)
	}
```

kafkabp/consumer_test.go
```diff
@@ -0,0 +1,187 @@
...
+import (
+	"context"
+	"errors"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/Shopify/sarama"
+	"github.com/Shopify/sarama/mocks"
+	"github.com/stretchr/testify/assert"
```
pacejackson, line 12: Similar to the above comment, we don't use testify in baseplate.go.

pacejackson (changes requested):

kafkabp/consumer.go
pacejackson: Ya, I think it helps to have it here though so it's closer to the actual code that is running based on that? Especially for the partition rebalance case, the code here doesn't really tell you much about why we would hit this case.

kafkabp/consumer_test.go
```diff
@@ -0,0 +1,206 @@
...
+				consumedErrs = append(consumedErrs, err)
+				wg.Done()
+			},
+		)
+	}()
+
+	wg.Wait() // wait for all kafka messages and errors to be consumed
+
+	if len(consumedMsgs) != 2 {
+		t.Errorf("expected len(consumedMsgs) == 2, got %v", len(consumedMsgs))
```
pacejackson, line 79: for numbers, use `%d`

```suggestion
		t.Errorf("expected len(consumedMsgs) == 2, got %d", len(consumedMsgs))

```

kafkabp/consumer_test.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	if len(consumedErrs) != 2 {
+		t.Errorf("expected len(consumedErrs) == 2, got %v", len(consumedErrs))
```
pacejackson, line 82:
```suggestion
		t.Errorf("expected len(consumedErrs) == 2, got %d", len(consumedErrs))
```

kafkabp/consumer_test.go
```diff
@@ -0,0 +1,206 @@
...
+		pc1.YieldMessage(kMsg1)
+		pc1.YieldError(kErr1)
+	}
+
+	// close kafkaConsumer and assert all
+	// messages and error channels are drained
+	kc.Close()
+
+	if len(consumedMsgs) != 10 {
+		t.Errorf("expected len(consumedMsgs) == 10, got %v", len(consumedMsgs))
```
pacejackson, line 144:
```suggestion
		t.Errorf("expected len(consumedMsgs) == 10, got %d", len(consumedMsgs))

File: kafkabp/consumer_test.go [hunk 16]
```diff
@@ -0,0 +1,206 @@
...
+	}
+	if len(consumedErrs) != 10 {
+		t.Errorf("expected len(consumedErrs) == 10, got %v", len(consumedErrs))
```
pacejackson, line 147:
```suggestion
		t.Errorf("expected len(consumedErrs) == 10, got %d", len(consumedErrs))

File: kafkabp/config.go [hunk 17]
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	switch cfg.Offset {
+	case "oldest":
+		c.Consumer.Offsets.Initial = sarama.OffsetOldest
+	case "newest":
+		c.Consumer.Offsets.Initial = sarama.OffsetNewest
+	case "":
+		// This is the "true" default case (in that it will be reached if an offset isn't specified).
+		c.Consumer.Offsets.Initial = sarama.OffsetOldest
```
fishy, line 54: nit: I'd slight prefer to move the `""` case to be before `"oldest"` case, and use `fallthrough` instead of assigning here, e.g.:

```
case "":
  // This is the true default...
  fallthrough
case "oldest":
  c.Consumer.Offsets.Initial = sarama.OffsetOldest
case ...
```

This way the intention is more obvious that empty falls back to the oldest case, and we eliminated one line of duplicated code :)

suversky: Ooh, thanks! Always appreciate feedback on having more idiomatic and readable code :)

pacejackson (approved):

external.bzl
```diff
@@ -121,6 +121,13 @@ def go_dependencies():
         sum = "h1:BSKMNlYxDvnunlTymqtgONjNnaRV1sTpcovwwjF22jk=",
         version = "v1.0.10",
     )
+    go_repository(
+        name = "com_github_creack_pty",
+        importpath = "github.com/creack/pty",
+        sum = "h1:uDmaGzcdjhF4i/plgjmEsriH11Y0o7RKapEf/LDaM3w=",
+        version = "v1.1.9",
+    )
+
```
fishy, line 130: nit: Can you manually remove those blank lines? It's kinda a bug of gazelle, that when it create this file for the first time there's no blank lines, but when it add some new dependencies it will add blank lines after those new dependencies, which is inconsistent.

kafkabp/config.go
```diff
@@ -0,0 +1,61 @@
...
+//   clientID: myclient
+//   offset: oldest
+type ConsumerConfig struct {
+	// Required. Brokers specifies a slice of broker addresses.
+	Brokers []string `yaml:"brokers"`
+
+	// Required. Topic is used to specify the topic to consume.
+	Topic string `yaml:"topic"`
+
+	// Optional. A user-provided string sent with every request to the brokers
```
fishy, line 27: Saying this is optional is a little bit misleading.

In ads-go we have two different consumer implementations. The `GroupConsumer` is for the more traditional, delivery every message once queue consumer (when every consumer in the same group have the same client id). This one doesn't do that. This consumer is for the use cases of using kafka to send some config/data to all the running instances use-case, so we actually want every consumer to have the whole view of the kafka topic. As a result, we always need to make sure that every consumer has an unique client id (if you give a group of pods the same client id with this consumer, they won't actually make sure that every kafka message is consumed by exactly one pod, and also every pod won't have the whole view of the kafka topic either).

I'd say this is required, and the comment should mention that every consumer should have an unique `ClientID`. for k8s pods, the k8s pod id is usually a good candidate to be used as the `ClientID`.

fishy: If we are going to share the same `ConsumerConfig` between `NewConsumer` and `NewGroupConsumer`, then I guess this is actually optional for `NewGroupConsumer`, but still using the default value for `NewGroupConsumer` is not really recommended.

kafkabp/consumer.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	// Validate input parameters.
+	if len(cfg.Brokers) == 0 {
+		return nil, ErrBrokersEmpty
+	}
+
+	if cfg.Topic == "" {
+		return nil, ErrTopicEmpty
+	}
```
fishy, line 62: shouldn't we move these 2 checks into `cfg.NewSaramaConfig()` as well?

kafkabp/consumer.go
```diff
@@ -0,0 +1,228 @@
...
+func (kc *consumer) getPartitionConsumers() []sarama.PartitionConsumer {
+	pc, _ := kc.partitionConsumers.Load().([]sarama.PartitionConsumer)
+	return pc
+}
+
+// reset recreates the consumer and assigns partitions.
+func (kc *consumer) reset() error {
+	if c := kc.getConsumer(); c != nil {
+		if err := c.Close(); err != nil {
+			log.Warnw("Error closing the consumer", "err", err)
```
fishy, line 97: For a truly library code calling `log.Warnw` directly is not recommended. Looks like this is the only instance direct log call is used in the whole PR, so I'd recommend make it a `log.Wrapper` field of the config struct.

kafkabp/consumer.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+
+	rebalance := func() (err error) {
```
fishy, line 101: super nit: `err` doesn't need to be a named return (we don't use any feature provided by named return).

kafkabp/consumer.go
```diff
@@ -0,0 +1,228 @@
...
+type Consumer interface {
+	io.Closer
+
+	Consume(ConsumeMessageFunc, ConsumeErrorFunc) error
+
+	// IsHealthy returns false after Consume returns.
+	IsHealthy() bool
+}
+
+// NewConsumer creates a new Kafka consumer.
```
fishy, line 48: We should also explain the differences between the two consumers here. I know in this PR we only added one of them so there's no "differences" yet, but we still should add some explanation :)

fishy (changes requested):
Mostly lgtm, but I have some minor-ish comments.

kafkabp/consumer.go
suversky:
```suggestion
			return err
```

Any qualms to just returning the error? It seems like this is more in line with the philosophy outlined in `../log/wrapper.go`.

kafkabp/consumer.go
```diff
@@ -0,0 +1,222 @@
...
+func (kc *consumer) getPartitionConsumers() []sarama.PartitionConsumer {
+	pc, _ := kc.partitionConsumers.Load().([]sarama.PartitionConsumer)
+	return pc
+}
+
+// reset recreates the consumer and assigns partitions.
+func (kc *consumer) reset() error {
+	if c := kc.getConsumer(); c != nil {
+		if err := c.Close(); err != nil {
+			return err
```
fishy, line 91: I don't think this is the right way to do it. Close errors are usually considered much less severe compare to other errors (especially close errors from readers), and they can usually just be logged and ignored. If we just return the error here that means close errors will cause the whole resetting and consuming loop to break and fail.

But in prod I think this rarely happens so maybe this is ok? I still prefer this to be just logged and ignored instead.

fishy (approved):
I don't think any remaining comments are blocking.

kafkabp/config.go
```diff
@@ -0,0 +1,74 @@
...
+	//
+	// The Kubernetes pod ID is usually a good candidate for this unique ID.
+	ClientID string `yaml:"clientID"`
+
+	// Optional. Defaults to "oldest". Valid values are "oldest" and "newest".
+	Offset string `yaml:"offset"`
+}
+
+// NewSaramaConfig instantiates a sarama.Config with sane defaults from
+// sarama.NewConfig(), overwritten by values parsed from cfg.Overrides.
```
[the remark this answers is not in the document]
suversky: Yep, I actually addressed this in my initial producer commit that I haven't been able to PR yet here (until this is merged).

kafkabp/config.go
```diff
@@ -0,0 +1,74 @@
...
+func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
+	c := sarama.NewConfig()
+
+	// Return any errors that occurred while consuming on the Errors channel.
+	c.Consumer.Return.Errors = true
```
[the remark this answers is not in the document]
suversky: Good call!

kafkabp/consumer.go
suversky: I figure adding a `log.Wrapper` to the Config and letting the user decide if they want to set it is the right approach. Thanks for the guidance!

kafkabp/config.go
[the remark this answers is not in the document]
suversky: I added it in here as well.

suversky:
I imagine getting rid of atomics in favor of a `RWMutex` in relevant methods would be considerably simpler, but marginally slower for two reasons:
  1. I think atomics are implemented at the CPU instruction level these days, while I'm not sure if Golang mutexes are implemented in userspace or if they require a system call.
  2. The amount of non-parallelizable mutex-protected instructions would be by nature larger than the atomic setup (which again I think can load, store and compare-and-swap in a single CPU cycle).

konradreiche:
> I imagine getting rid of atomics in favor of a `RWMutex` in relevant methods would be considerably simpler, but marginally slower for two reasons:
> 
> 1. I think atomics are implemented at the CPU instruction level these days, while I'm not sure if Golang mutexes are implemented in userspace or if they require a system call.
> 2. The amount of non-parallelizable mutex-protected instructions would be by nature larger than the atomic setup (which again I think can use a single CPU cycle to either load, store, or compare-and-swap).

Mutexes are also just bootstrapped on atomics https://github.com/golang/go/blob/master/src/sync/rwmutex.go#L48

I'd be curious if there are performance differences at all. Personally, I favor readable/maintainable code over negligible performance differences.

suversky:
> Personally, I favor readable/maintainable code over negligible performance differences.

I agree entirely! I'd be happy to tackle the RWMutex approach (and even happier if I could do it in a separate PR so I can PR the producer code using this base.)

kafkabp/config.go
```diff
@@ -0,0 +1,81 @@
...
+	// Consumer implementation in this library expects every Consumer to have a
+	// unique ClientID.
+	//
+	// The Kubernetes pod ID is usually a good candidate for this unique ID.
+	ClientID string `yaml:"clientID"`
+
+	// Optional. Defaults to "oldest". Valid values are "oldest" and "newest".
+	Offset string `yaml:"offset"`
+
+	// Optional. If non-nil, will be used to log errors.
```
fishy, line 39: Maybe we can be more specific on what errors will be logged via this logger?

fishy (approved):
Depending on how RWMutex is used the performance hit could be significant. I do think atomic values are much better fit over RWMutex in the current implementation.

konradreiche:
Also an amendment to my comment earlier, I oversaw 

```
runtime_SemacquireMutex
```

https://github.com/golang/go/blob/master/src/sync/rwmutex.go#L50

Which is where the real performance dip would happen sending threads to sleep rather than trying to do a non-blocking operation.

fishy (approved):

pacejackson (approved):

kafkabp/config.go
```diff
@@ -0,0 +1,83 @@
...
+	if cfg.ClientID == "" {
+		return nil, ErrClientIDEmpty
+	}
+
+	var offset int64
+	switch cfg.Offset {
+	case "":
+		// OffsetOldest is the "true" default case (in that it will be reached if
+		// an offset isn't specified).
+		fallthrough
```
konradreiche, line 66: :no_bell: Not sure if this would be equivalent or more readable
```
case "", "oldest":
```

kafkabp/consumer.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+
+		kc.consumer.Store(c)
+		kc.partitions.Store(partitions)
+		return nil
+	}
+
+	err := rebalance()
+	if err != nil {
+		metricsbp.M.Counter("kafka.consumer.rebalance.failure").Add(1)
```
konradreiche, line 113: :no_bell: not sure if pointed out elsewhere but I assume we want metric on/off configurable in the future?

kafkabp/consumer.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+			partitionConsumers = append(partitionConsumers, partitionConsumer) // for closing individual partitions when Close() is called
+
+			// consume partition consumer messages
+			wg.Add(1)
+			go func(pc sarama.PartitionConsumer) {
+				defer wg.Done()
+				for m := range pc.Messages() {
+					// Wrap in anonymous function for easier defer.
+					func() {
```
konradreiche, line 173: :no_bell: At this point we're fairly deeply intended, I wonder if moving some of this into methods will improve the cognitive complexity

konradreiche (approved):
Thanks for taking this on, cannot wait to bring Kafka processing to Go :partying_face:

fishy (approved):

kafkabp/config.go
```diff
@@ -0,0 +1,81 @@
...
+	// Required. ClientID is a user-provided string sent with every request to
+	// the brokers for logging, debugging, and auditing purposes. The default
+	// Consumer implementation in this library expects every Consumer to have a
+	// unique ClientID.
+	//
+	// The Kubernetes pod ID is usually a good candidate for this unique ID.
+	ClientID string `yaml:"clientID"`
+
+	// Optional. Defaults to "oldest". Valid values are "oldest" and "newest".
+	Offset string `yaml:"offset"`
```
fishy, line 37: nit: I almost forgot but I think we should define `"oldest"` and `"newest"` as exported string constants, so that users not using yaml can make sure they assign the correct value here. e.g.:

```go
// Allowed Offset values
const (
  OffsetOldest = "oldest"
  OffsetNewest = "newest"
)
```

kafkabp/config.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, ErrClientIDEmpty
+	}
+
+	var offset int64
+	switch cfg.Offset {
+	case "", "oldest":
+		// OffsetOldest is the "true" default case (in that it will be reached if
+		// an offset isn't specified).
+		offset = sarama.OffsetOldest
+	case "newest":
```
fishy, line 67: and we should use the defined constants here :)

suversky: facepalm

fishy (approved):

PR #287, opened 2020-09-30
metricsbp: Make Timer functions return self for chaining

fishy:
This makes it possible to write something like:

    metricsbp.NewTimer(...).OverrideStartTime(start).ObserveDuration()

### DIFF
--- a/metricsbp/timer.go
+++ b/metricsbp/timer.go
@@ -36,17 +36,22 @@ func NewTimer(h metrics.Histogram) *Timer {
 //     t.OverrideStartTime(time.Now())
 //
 // If t is nil, it will be no-op.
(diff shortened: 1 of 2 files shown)


pacejackson (commented):
I'm not 100% opposed to this, but it also doesn't really seem like something that is necessary. Does this really come up outside of the one test that is being updated to use this?

fishy:
> I'm not 100% opposed to this, but it also doesn't really seem like something that is necessary. Does this really come up outside of the one test that is being updated to use this?

The example in the PR description :)

pacejackson (approved):

konradreiche (approved):

PR #292, opened 2020-10-14
Add a validate method to thriftbp.ClientPoolConfig

konradreiche:
It's often useful to have a way to check configuration files for completeness. I usually create a validate command that will check the files for correctness. I believe it makes sense to have configurations themselves provide a validate method too which at the minimum check for required values or returns an error if a value is out of range. This way the validation chain can be partially delegated.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -171,6 +171,9 @@ type ClientPoolConfig struct {
 
 // Validate checks the ClientPoolConfig for any missing or erroneous values. It
 // returns an errorsbp.Batch to be combineable with other error batching.
+//
+// This method is designated to be used when passing a configuration to
+// NewBaseplateClientPool, for NewCustomClientPool other constraints apply.
 func (c ClientPoolConfig) Validate() error {
 	var batch errorsbp.Batch
 	if c.ServiceSlug == "" {
@@ -247,6 +250,10 @@ func SingleAddressGenerator(addr string) AddressGenerator {
 // BaseplateDefaultClientMiddlewares plus any additional client middlewares
 // passed into this function.
 func NewBaseplateClientPool(cfg ClientPoolConfig, middlewares ...thrift.ClientMiddleware) (ClientPool, error) {
+	err := cfg.Validate()
+	if err != nil {
+		return nil, err
+	}
 	defaults := BaseplateDefaultClientMiddlewares(
 		DefaultClientMiddlewareArgs{
 			ServiceSlug:         cfg.ServiceSlug,
@@ -275,10 +282,6 @@ func NewCustomClientPool(
 	protoFactory thrift.TProtocolFactory,
 	middlewares ...thrift.ClientMiddleware,
 ) (ClientPool, error) {
-	err := cfg.Validate()
-	if err != nil {
-		return nil, err
-	}
 	return newClientPool(cfg, genAddr, protoFactory, middlewares...)
 }
 
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -174,15 +174,15 @@ type ClientPoolConfig struct {
 func (c ClientPoolConfig) Validate() error {
 	var batch errorsbp.Batch
 	if c.ServiceSlug == "" {
-		batch.Add(errors.New("ServiceSlug cannot be empty"))
+		batch.Add(ErrConfigMissingServiceSlug)
 	}
 	if c.Addr == "" {
-		batch.Add(errors.New("Addr cannot be empty"))
+		batch.Add(ErrConfigMissingAddr)
 	}
 	if c.InitialConnections > c.MaxConnections {
-		batch.Add(errors.New("InitialConnections cannot be bigger than MaxConnections"))
+		batch.Add(ErrConfigInvalidConnections)
 	}
-	return batch
+	return batch.Compile()
 }
 
 // Client is a client object that implements both the clientpool.Client and
@@ -275,6 +275,10 @@ func NewCustomClientPool(
 	protoFactory thrift.TProtocolFactory,
 	middlewares ...thrift.ClientMiddleware,
 ) (ClientPool, error) {
+	err := cfg.Validate()
+	if err != nil {
+		return nil, err
+	}
 	return newClientPool(cfg, genAddr, protoFactory, middlewares...)
 }
 
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -169,6 +169,22 @@ type ClientPoolConfig struct {
 	BreakerConfig *breakerbp.Config
 }
 
+// Validate checks the ClientPoolConfig for any missing or erroneous values. It
+// returns an errorsbp.Batch to be combineable with other error batching.
+func (c ClientPoolConfig) Validate() error {
+	var batch errorsbp.Batch
+	if c.ServiceSlug == "" {
+		batch.Add(errors.New("ServiceSlug cannot be empty"))
+	}
+	if c.Addr == "" {
+		batch.Add(errors.New("Addr cannot be empty"))
+	}
+	if c.InitialConnections > c.MaxConnections {
+		batch.Add(errors.New("InitialConnections cannot be bigger than MaxConnections"))
+	}
+	return batch
+}
+
 // Client is a client object that implements both the clientpool.Client and
 // thrift.TCLient interfaces.
 //
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -169,12 +169,27 @@ type ClientPoolConfig struct {
 	BreakerConfig *breakerbp.Config
 }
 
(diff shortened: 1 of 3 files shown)


thriftbp/client_pool.go
```diff
@@ -169,6 +169,22 @@ type ClientPoolConfig struct {
...
+	if c.ServiceSlug == "" {
+		batch.Add(errors.New("ServiceSlug cannot be empty"))
+	}
+	if c.Addr == "" {
+		batch.Add(errors.New("Addr cannot be empty"))
+	}
+	if c.InitialConnections > c.MaxConnections {
+		batch.Add(errors.New("InitialConnections cannot be bigger than MaxConnections"))
+	}
+	return batch
```
pacejackson, line 185: nit, use batch.Compile for this

```suggestion
	return batch.Compile()
```

That takes care of returning nil, one error, or the batch depending on the number of errors in the batch.

pacejackson (changes requested):
I support this sort of thing, can we also use this in `NewBaseplateClientPool`/`NewCustomClientPool`?

thriftbp/client_pool.go
```diff
@@ -169,6 +169,22 @@ type ClientPoolConfig struct {
 	BreakerConfig *breakerbp.Config
 }
 
+// Validate checks the ClientPoolConfig for any missing or erroneous values. It
+// returns an errorsbp.Batch to be combineable with other error batching.
+func (c ClientPoolConfig) Validate() error {
+	var batch errorsbp.Batch
+	if c.ServiceSlug == "" {
+		batch.Add(errors.New("ServiceSlug cannot be empty"))
```
pacejackson, line 177: Can you also use named errors for these rather than using errors.New each time?

fishy: +1, and also linter will yell at you when you try to capitalize error messages (but that seems to only check `fmt.Errorf` but not `errors.New`).

konradreiche: For some reason I thought `batch.Compile()` produces a new error

konradreiche:
I only added the call to `Validate()` in `NewCustomClientPool` since `NewBaseplateClientPool` jsut delegates to that method too :thinking:

thriftbp/client_pool.go
```diff
@@ -259,6 +275,10 @@ func NewCustomClientPool(
 	protoFactory thrift.TProtocolFactory,
 	middlewares ...thrift.ClientMiddleware,
 ) (ClientPool, error) {
+	err := cfg.Validate()
+	if err != nil {
+		return nil, err
+	}
```
fishy, line 281: 🔕 I do think this approach could lead to diverging behavior in the long term, though. e.g. there are some new restrictions added in the future, that then `newClientPool` would return an error because the config is bad, but we didn't add the same check into `Validate` so it still returns `nil`. I'm not sure what's the best solution for that, though.

thriftbp/errors.go
```diff
@@ -16,7 +16,10 @@ type BaseplateErrorCode interface {
 }
 
 var (
-	_ BaseplateErrorCode = (*baseplatethrift.Error)(nil)
+	_                           BaseplateErrorCode = (*baseplatethrift.Error)(nil)
```
fishy, line 19: I don't think this line does anything here. If your intention is to make the `Err*` vars to be of type `BaseplateErrorCode`, this doesn't do that.

konradreiche: My formatter just produced that intention after I added lines below.

thriftbp/client_pool.go
konradreiche: True, I've been trying to dig a bit deeper into this but in any case it's hard to keep track of this so it stays consistent.

thriftbp/errors.go
fishy: oh I see, that's an already existing line.

I'd suggest add the `Err*` into its own `var` group.

thriftbp/client_pool.go
konradreiche: Does that mean we want to lowercase those strings, i.e.
```
serviceSlug cannot be empty
```

fishy: nevermind I just realized that it's capitalized because it's referring to the field name that's already capitalized. I think just leave them as they are now is fine.

thriftbp/errors.go
```diff
@@ -19,6 +19,16 @@ var (
 	_ BaseplateErrorCode = (*baseplatethrift.Error)(nil)
 )
 
+var (
```
fishy, line 22: pro tip: golint also allows you to just have a single doc comment before the `var (` line for the whole var block, so you could just say something like `// ClientPoolConfig errors.` as the names are mostly self-explanatory.

fishy (approved):

thriftbp/client_pool_test.go
```diff
@@ -60,6 +60,7 @@ func TestCustomClientPool(t *testing.T) {
 
 	if _, err := thriftbp.NewCustomClientPool(
 		thriftbp.ClientPoolConfig{
+			Addr:               ":9090",
```
fishy, line 63: Actually this is a problem.

In `NewCustomClientPool` `Addr` is allowed to (and should) be empty, because in that function it uses `genAddr` arg instead of the `Addr` field from the config. So a single validation function for different usage is kinda weird.

We should only call `Validate` from `NewBaseplateClientPool`, and also mention it in `Validate`'s doc comment that it's only meant for `NewBaseplateClientPool`.

fishy (changes requested):

thriftbp/client_pool_test.go
konradreiche: Good catch,  that explains some of the behavior I've seen. My configuration didn't parse but the thrift client initialized just fine.

konradreiche: Makes me thing whether that should be codified somehow through sub-types but maybe it's making it unnecessarily complex.

fishy: I think we can revert this change now as you changed to only check `Validate` from `NewBaseplateClientPool`? we can probably add a new test to make sure that `NewBaseplateClientPool` checks `Validate` and returns config errors.

thriftbp/client_pool.go
```diff
@@ -169,6 +169,25 @@ type ClientPoolConfig struct {
 	BreakerConfig *breakerbp.Config
 }
 
+// Validate checks the ClientPoolConfig for any missing or erroneous values. It
+// returns an errorsbp.Batch to be combineable with other error batching.
```
pacejackson, line 173: The comment about what it returns is not always going to be true and should probably be left off. The way `errorsbp.Batch` works whatever is returned can be `Add`-ed into another `Batch` if needed and it will work the way you expect (nil is not added, a single error just adds that error, a batch adds each error in the batch rather than the batch itself).

thriftbp/client_pool.go
```diff
@@ -169,6 +169,25 @@ type ClientPoolConfig struct {
...
+//
+// This method is designated to be used when passing a configuration to
+// NewBaseplateClientPool, for NewCustomClientPool other constraints apply.
```
pacejackson, line 176: I wonder if it makes more sense to just have these validations split between `NewBaseplateClientPool` and `NewCustomClientPool` as appropriate rather than adding a function to the Config itself?

konradreiche: I think so, because from an API point of view, having the method defined on the struct gives you the impression, it's always okay to call it, which it's not.

konradreiche: What about having a 
```
ValidateBaseplateClientPoolConfig
```
method? :thinking: 

Alternatively
```
type BaseplateClientPoolConfig {
  ClientPoolConfig
}

func (c *BaseplateClientPoolConfig) Validate() error {
  // ...
}
```

pacejackson: `ValidateBaseplateClientPoolConfig` seems reasonable.  I actually _kind of_ like the second option but it might be a little too cute 😅 

@fishy do you have any opinions?

fishy: I can see the value of adding a function to the config struct itself, so that users can do the validation without actually trying to use it (e.g. they can write static checking tools, kinda). With that spirit it might be better to split it into two functions, `ValidateCustom` which doesn't check `Addr`, and then `Validate` which calls `ValidateCustom` and also checks `Addr`.

fishy: If we go the embedding route, it might be easier to just do typedef instead? e.g.

```
type BaseplateClientPoolConfig Config
```

Then we can override its `Validate` implementation.

fishy: but that gets tricky as in the original `ClientPoolConfig` some of the fields are actually never used (they are only used by `BaseplateClientPoolConfig`). so maybe just split into two functions?

konradreiche:
I went with the `typedef` route, what do you think?

fishy (approved):

pacejackson (approved):

PR #295, opened 2020-10-30
metricsbp: Buffer udp messages together

fishy:
When using telegraf as the statsd collector, telegraf has a limit of how
many UDP messages can be buffered, and the default go-kit influxstatsd
implementation sends one UDP message per metric on the report interval,
so when we have a lot of metrics, telegraf could be dropping those UDP
messages, thus dropping metrics.

Add a configurable buffer size can help reduce the number of UDP
messages.

Eventually we probably want to upstream this to go-kit.

### DIFF
--- a/log/kit_wrapper.go
+++ b/log/kit_wrapper.go
@@ -9,7 +9,7 @@ type KitWrapper zapcore.Level
 
 // Log implements go-kit/log.Logger interface.
 func (w KitWrapper) Log(keyvals ...interface{}) error {
-	const msg = ""
+	const msg = "log.KitWrapper"
 	switch zapcore.Level(w) {
 	default:
 		// for unknown values, fallback to info level.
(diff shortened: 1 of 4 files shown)


pacejackson (approved):
LGTM, I'm assuming that `st.statsd.WriteTo` just `Writes` each metric to whatever writer you pass it one at a time rather than doing any sort of batching on it's end?

fishy:
> LGTM, I'm assuming that `st.statsd.WriteTo` just `Writes` each metric to whatever writer you pass it one at a time rather than doing any sort of batching on it's end?

Yes exactly. This is the code: https://github.com/go-kit/kit/blob/v0.9.0/metrics/influxstatsd/influxstatsd.go#L142

konradreiche (approved):
:clap:

pacejackson (approved):
:doh:

PR #300, opened 2020-11-18
filewatcher: Short circuit when os.Open returned error

fishy:
The current code could cause panic when os.Open returned error, as f
would be nil.

While I'm here, also make the log messages slightly more descriptive.

### DIFF
--- a/filewatcher/filewatcher.go
+++ b/filewatcher/filewatcher.go
@@ -90,7 +90,7 @@ func (r *Result) watcherLoop(
 			return
 
 		case err := <-watcher.Errors:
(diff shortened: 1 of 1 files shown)


konradreiche (commented):
Is another file missing in this PR?

fishy:
No. What other file?

konradreiche:
> No. What other file?

The part where you short-circuit, I only see the log message update

fishy:
> > No. What other file?
> 
> The part where you short-circuit, I only see the log message update

The added `return` on Line 109.

konradreiche (approved):
:coffee:

pacejackson (approved):

PR #303, opened 2020-11-21
kafkabp: Add GroupID to ConsumerConfig

fishy:
This adds the support for group consumers, for exactly-once message
queue consumer use cases.

Also some tweaks on the existing shared consumer interfaces. Most of the
tweaks are on the wording of the doc comment to add reference to the
newly added GroupID, but one notable change is that the error
return from ConsumeMessageFunc is removed. The reason is with added
group consumers, the error return could make users to think "If I
return an error in this function maybe the message won't be committed",
which is not true and also impossible to do, as in Apache Kafka message
committing is a watermark, not individually per message, so there's no
way to skip committing a single failed message.

Also there's no tests for group consumer, as sarama doesn't provide a
good mock implementation for it.


fishy:
cc @suversky

kafkabp/group_consumer.go
```diff
@@ -0,0 +1,140 @@
...
+			defer func() {
+				span.FinishWithOptions(tracing.FinishOptions{
+					Ctx: ctx,
+				}.Convert())
+			}()
+
+			h.callback(ctx, m)
+			session.MarkMessage(
+				m,
+				"", // metadata
```
konradreiche, line 135: I wonder if this is ever useful. I did not find anything about this being a general concept in Kafka but it's documented in sarama as:

```
	// MarkOffset marks the provided offset, alongside a metadata string
	// that represents the state of the partition consumer at that point in time. The
	// metadata string can be used by another consumer to restore that state, so it
	// can resume consumption.
```
I don't think we have any real world use for this at the moment but when I looked at the implementation I tried to figure out if it makes sense to integrate it into the API :man_shrugging:

kafkabp/consumer_interface.go
```diff
@@ -25,20 +29,25 @@ type ConsumerConfig struct {
...
-	// unique ClientID.
+	// Required. ClientID is used by Kafka broker to track clients' consuming
+	// progresses on the topics.
 	//
+	// For NewConsumer, every instance is expected to have a unique ClientID.
 	// The Kubernetes pod ID is usually a good candidate for this unique ID.
+	//
+	// For NewGroupConsumer, instances from the same group must share the same
+	// ClientID.
 	ClientID string `yaml:"clientID"`
```
konradreiche, line 40: Should this become `GroupID` whereas I believe `ClientID` is used to distinguish single requests? The client ID here is at least passed into the sarama implementation which declares it as group ID too.

konradreiche (changes requested):
:clap: Mostly questions.

Agreeing on removing the error return. Would we generally assume that an implementer could delegate any error that happens inside the consume method to delegate to `ConsumeErrorFunc` or is its purpose solely to handle errors that happen in the underlying implementation?

kafkabp/consumer_interface.go
fishy: Good point. Added `GroupID` to `ConsumerConfig` instead, this way we don't even need to export `NewGroupConsumer`, and `NewConsumer` can be smart enough to choose the implementation automatically.

konradreiche (approved):

kafkabp/group_consumer.go
```diff
@@ -0,0 +1,132 @@
...
+// once all ConsumeClaim goroutines have exited.
+func (h GroupConsumerHandler) Cleanup(sarama.ConsumerGroupSession) error {
+	// Do nothing.
+	return nil
+}
+
+// ConsumeClaim starts a consumer loop of ConsumerGroupClaim's Messages() chan.
+func (h GroupConsumerHandler) ConsumeClaim(session sarama.ConsumerGroupSession, claim sarama.ConsumerGroupClaim) error {
+	for m := range claim.Messages() {
+		// Wrap in anonymouse function for easier defer.
```
pacejackson, line 112: typo here:

```suggestion
		// Wrap in anonymous function for easier defer.
```

pacejackson (approved):

kafkabp/consumer.go
```diff
@@ -34,27 +68,29 @@ type consumer struct {
...
 		return nil, err
 	}
 
+	if cfg.GroupID != "" {
+		return newGroupConsumer(cfg, sc)
+	}
+
 	kc := &consumer{
 		cfg:    cfg,
 		sc:     sc,
```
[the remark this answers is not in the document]
fishy: Good point, but I feel the name `newTopicConsumer` might be a little bit misleading... but then `newConsumer` is misleading as well so 🤷‍♂️

kafkabp/consumer.go
```diff
@@ -100,6 +139,7 @@ func (kc *consumer) reset() error {
 
 		partitions, err := c.Partitions(kc.cfg.Topic)
 		if err != nil {
+			c.Close()
```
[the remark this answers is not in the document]
fishy: It is. But I think it never/rarely happens so we are mostly fine.

kafkabp/consumer.go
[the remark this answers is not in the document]
pacejackson: Ya, I'm not married to that name (I was more or less just guessing what would be an appropriate name) if you have one that you think is better.

pacejackson (approved):

PR #304, opened 2020-11-24
runtimebp: print to stderr instead of stdout

konradreiche:
This replaces the `fmt.Println` call in `runtimebp/config.go` with printing to stderr like we do in `runtimebp/cpu.go`.

In certain scenarios there is a need to have log messages by baseplate.go filtered out. Since the only occasions where it prints without logger is in runtimebp, they should both print to stderr, allowing it to be filter, i.e. through the user of `2> /dev/null`

---
### Deprecated: previous suggestion

I noticed we still have a place where we call `fmt.Println` instead of relying on `log.Wrapper`. This passes the logger in via the configuration.

Should we extend the function to pass in `context.Context`? I relied on the comment:
```
// Not all Wrapper implementations take advantage of context object passed in,
// but the caller should always pass it into Wrapper if they already have one,
// or just use context.Background() if they don't have one.
```
to decide passing in `context.Background()` for now.

### DIFF
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -2,6 +2,7 @@ package runtimebp
 
 import (
 	"fmt"
+	"os"
 )
 
 // Config is the configuration struct for the runtimebp package.
@@ -36,5 +37,5 @@ func InitFromConfig(cfg Config) {
 		min = cfg.NumProcesses.Min
 	}
 	prev, current := GOMAXPROCS(min, max)
-	fmt.Println("GOMAXPROCS:", prev, current)
+	fmt.Fprintf(os.Stderr, "GOMAXPROCS: %d %d\n", prev, current)
 }


fishy (changes requested):
This one is a bit tricky and is kind of an exception.

First, if people want to use baseplate.go logger for the wrapper here, they need to make sure that they initialize log package before initialize runtimebp package. But in most cases they should init runtimebp first as setting GOMAXPROCS should be the first thing they do in the process before other things, and we do runtimebp before log in baseplate package: https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/baseplate.go#L252-L257

And you need to set the Logger to something in that call from `baseplate.New` to make sure it's logged, probably using `log.StdWrapper` as the zap logger was not initialized yet and will go nowhere, but using `log.StdWrapper` will make it more or less the same as using `fmt.Println` directly.

Also adding a Logger to the config, but then it's not really a config, just an arg to the `InitFromConfig` call, feels weird.

All in all I feel it's probably easier to just make this an exception and use `fmt.Println` (or if you'd like, change it to `fmt.Fprintln` with `os.Stderr`),

(also you need to run `./script/bazel_cleanup.sh`)

konradreiche:
After spending some more time with the way this code is used I came to similar realizations. After all, there is also the stderr print code in `cpu.go`.

The reason for me trying to come up with a change was mainly motivated to get less verbose output for services that run `baseplate.New` in production or in test. The latter more than the former, I try to have my test output to be as silent as possible with focus on the failure but at the same time have end-to-end type of tests that run most of the setup code and that includes `baseplate.New`.

One way to deal with this is to just make most of the calls but not to `runtimebp` manually if I feel like it's not needed but I'd rather have it go through there.

If it gets printed to stderr, is there really a way to filter it out? I think stderr would ensure it always gets printed, especially in test I assume?

fishy:
For local test, you do actually want this to be in stdout (the current `fmt.Println` way) rather than stderr. As the zap logs will go to stderr (I believe), so you can just redirect stdout to `/dev/null` to filter this particular one out.

konradreiche:
That would help but there might be occasions where I would want read out stdout, also `cpu.go` will still write to stderr. I do believe it should always be possible, even if it's not by default, being able to control the logging/output of a library.

The question is how this can be achieved. `log.Wrapper` might not be the right approach but `runtimebp` does take a config that could control this behavior :thinking:

fishy:
If you really want to make those logs controllable, then I think the approach should be that add `log.Wrapper` to config and also `NumCPU` and `GOMAXPROCS` functions (so it would be a breaking change, and it's probably also time to change them to take struct arg), and document it that at the time of calling them zap logger is likely not initialized yet so people might want to use `StdWrapper` for this one.

But I do think this is overkill. This is one log (with up to additional 2 from NumCPU call) happen once during startup and not happening again. I don't really think this is a big problem for local debugging to worth all the complexity added.

konradreiche:
Definitely not bothered about a printout at the start of a service. Just to give an example how it can affect the console output, this is when I run a benchmark test:

```
go test -bench . -run=BenchmarkMyHeavyOperation 
NumCPU: falling back to use shares: -1e-05, <nil>
GOMAXPROCS: 12 1
goos: linux
goarch: amd64
BenchmarkMyHeavyOperation/NumWriters=1-12                     1        1562751095 ns/op
NumCPU: falling back to use shares: -1e-05, <nil>
GOMAXPROCS: 12 1
BenchmarkMyHeavyOperation/NumWriters=10-12                    1        1311611402 ns/op
NumCPU: falling back to use shares: -1e-05, <nil>
GOMAXPROCS: 12 1
BenchmarkMyHeavyOperation/NumWriters=100-12                   1        1344755001 ns/op
NumCPU: falling back to use shares: -1e-05, <nil>
GOMAXPROCS: 12 1
BenchmarkMyHeavyOperation/NumWriters=250-12                   1        1304691574 ns/op
NumCPU: falling back to use shares: -1e-05, <nil>
GOMAXPROCS: 12 1
BenchmarkMyHeavyOperation/NumWriters=500-12                   1        1297288703 ns/op
NumCPU: falling back to use shares: -1e-05, <nil>
GOMAXPROCS: 12 1
BenchmarkMyHeavyOperation/NumWriters=1000-12                  1        1420122199 ns/op
PASS
```

fishy:
> Definitely not bothered about a printout at the start of a service. Just to give an example how it can affect the console output, this is when I run a benchmark test:
> 
> ```
> go test -bench . -run=BenchmarkMyHeavyOperation 
> NumCPU: falling back to use shares: -1e-05, <nil>
> GOMAXPROCS: 12 1
> goos: linux
> goarch: amd64
> BenchmarkMyHeavyOperation/NumWriters=1-12                     1        1562751095 ns/op
> NumCPU: falling back to use shares: -1e-05, <nil>
> GOMAXPROCS: 12 1
> BenchmarkMyHeavyOperation/NumWriters=10-12                    1        1311611402 ns/op
> NumCPU: falling back to use shares: -1e-05, <nil>
> GOMAXPROCS: 12 1
> BenchmarkMyHeavyOperation/NumWriters=100-12                   1        1344755001 ns/op
> NumCPU: falling back to use shares: -1e-05, <nil>
> GOMAXPROCS: 12 1
> BenchmarkMyHeavyOperation/NumWriters=250-12                   1        1304691574 ns/op
> NumCPU: falling back to use shares: -1e-05, <nil>
> GOMAXPROCS: 12 1
> BenchmarkMyHeavyOperation/NumWriters=500-12                   1        1297288703 ns/op
> NumCPU: falling back to use shares: -1e-05, <nil>
> GOMAXPROCS: 12 1
> BenchmarkMyHeavyOperation/NumWriters=1000-12                  1        1420122199 ns/op
> PASS
> ```

But why would you call `GOMAXPROCS` inside a benchmark test? is it that you want to test how different `GOMAXPROCS` value affects the performance? If that's the case, I'd suggest just change the stderr in NumCPU to stdout so you can redirect stdout to /dev/null in `go test -bench`

konradreiche:
It's just a side effect of calling `baseplate.New` which is part of the setup for the benchmark to run.

fishy:
Verified that `go test -bench` actually uses stdout, so you can just change `InitFromConfig` to use `fmt.Fprintln(os.Stderr, ...)` and then you can run `go test -bench ... 2> /dev/null` to eliminate those logs.

fishy (approved):
Can you also update the pr's title and description?

pacejackson (approved):

PR #305, opened 2020-11-25
log: Make Wrapper to implement encoding.TextUnmarshaler

fishy:
log.Wrapper is one type we frequently put into config structs, and
making it to implement encoding.TextUnmarshaler means that it can be
initialized from yaml config directly.


log/wrapper.go
```diff
@@ -84,6 +88,48 @@ func (w Wrapper) ToThriftLogger() thrift.Logger {
...
+//
+// - zap:level: ZapWrapper with given level, for example "zap:error" means
+// ZapWrapper on Error level.
+//
+// - sentry: ErrorWithSentryWrapper.
+func (w *Wrapper) UnmarshalText(text []byte) error {
+	// Special handling for "zap:level" case
+	const zapLevelPrefix = "zap:"
+	if strings.HasPrefix(string(text), zapLevelPrefix) {
+		*w = ZapWrapper(Level(text[len(zapLevelPrefix):]))
```
konradreiche, line 112: :no_bell: Not sure if it is worth protecting against typos but we could lowercase the parsed zap level to ensure
```
zap:info
zap:Info
```
will both work.

konradreiche (approved):
This will be super useful, I am currently setting some of these loggers and levels manually in tests.

log/wrapper.go
fishy: I'm on the fence. The lower case values are defined as constants (https://pkg.go.dev/github.com/reddit/baseplate.go@v0.7.0/log#Level), so doing `ToLower` here means that in the future if we change those constants to have mixed cases (for example `"info"` -> `"Info"`) this will break silently, because we don't have a good way to test that the arg passed into ZapWrapper is correct.

But then that is a moderate breaking change anyways, so it's very unlikely to happen in the first place ¯\\\_(ツ)\_/¯

pacejackson (approved):

PR #306, opened 2020-11-27
timebp: rename time variable for avoid collision with time import

alonsopf:
- Add nanoseconds support
- remove warning: rename time variable for avoid collision with time imported package


timebp/nanosecond.go
```diff
@@ -0,0 +1,116 @@
...
+
+	us, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		return err
+	}
+	*ts = TimestampNanosecond(NanosecondsToTime(us))
+	return nil
+}
+
+// NanosecondsToTime converts milliseconds since EPOCH to time.Time.
```
konradreiche, line 57: I suppose this should be `nanoseconds`

konradreiche (approved):

timebp/nanosecond.go
alonsopf: yeah, i missed renamed the variable "us" to "ns"

fishy (commented):
With nanoseconds the max int64 only goes to year 2262: https://play.golang.org/p/VuemhSwFrjE

What's the use case of this (e.g. in what scenario do we/you need to use int64 nanoseconds since EPOCH as the timestamp)? We defined `TimestampMicrosecond` because it's used in zipkin json format and `TimestampMillisecond` because that's wildly used especially when you are talking to java programs. But I've never saw anything using nanoseconds like this. (I'm not against the idea, just want to understand more about the use case behind it).

alonsopf:
> With nanoseconds the max int64 only goes to year 2262: https://play.golang.org/p/VuemhSwFrjE
> 
> What's the use case of this (e.g. in what scenario do we/you need to use int64 nanoseconds since EPOCH as the timestamp)? We defined `TimestampMicrosecond` because it's used in zipkin json format and `TimestampMillisecond` because that's wildly used especially when you are talking to java programs. But I've never saw anything using nanoseconds like this. (I'm not against the idea, just want to understand more about the use case behind it).

Yeah, you right, my main interest is get knowledge about the baseplate, so i think to the best is close this PR without merge, do you think that is good idea if I tried to remove some warnings (that show me GoLand), such as unhandled errors? or what do you think about one Good-First-Issue for me? 😅

timebp/nanosecond.go
alonsopf: @fishy thanks for aclaration, please see my response below

fishy:
>do you think that is good idea if I tried to remove some warnings

If you mean the shadowing of `time` in the tests, sure, although I don't think those are big deals (our canonical linters are `golint` and `go vet`, see https://github.com/reddit/baseplate.go/blob/master/Style.md#use-tools). You can do that in this PR, but please update the PR title and description accordingly.

If you mean other warnings, can you give some examples?

alonsopf:
> If you mean other warnings, can you give some examples?
for example:

https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/errorsbp/suppressor.go#L34-L36

i understand that is for supress the error, but what if is useful? so i think that can be add a  fmt.Println(err)


https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/filewatcher/filewatcher.go#L205

maybe change the defer with f.Close() with error handling

also:

https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/edgecontext/init_test.go#L23

alonsopf:
Above warnings does not matter, i guess

another tiny deals:

errors collide with import, do you think that not matter much? 

https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/errorsbp/batch.go#L178

fishy:
> > If you mean other warnings, can you give some examples?
> > for example:
> 
> https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/errorsbp/suppressor.go#L34-L36
> 
> i understand that is for supress the error, but what if is useful? so i think that can be add a fmt.Println(err)

`SuppressNone` is an implementation of `Suppressor` and logging errors is not the job of suppressors.

> 
> https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/filewatcher/filewatcher.go#L205
> 
> maybe change the defer with f.Close() with error handling
> 
> also:
> 
> https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/edgecontext/init_test.go#L23

A general rule of thumb is handle errors on `Close` calls for writes because that usually also includes flushing, but for reads there's not much you can do and not much loss either if close really fails.

> Above warnings does not matter, i guess
> 
> another tiny deals:
> 
> errors collide with import, do you think that not matter much?
> 
> https://github.com/reddit/baseplate.go/blob/aa32a66a724be5021965aa4c386bf82976fb0856/errorsbp/batch.go#L178

This is similar to the `time`s in this PR. Those only matters if there's code actually use those imports in the remaining of the shadowed scopes, and that's not the case for all of them. That's why I said they are not big deals.

There's a reason the official `golint` and `go vet` tools, being as opinionated as they are, don't complain about those "issues".

fishy (approved):

pacejackson (approved):

PR #308, opened 2020-12-01
kafkabp: Add experimental rack id support

fishy:


kafkabp/config.go
```diff
@@ -105,7 +121,7 @@ func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
 		return nil, ErrClientIDEmpty
 	}
 
-	var version sarama.KafkaVersion
+	version := defaultSaramaConfig.Version
```
pacejackson, line 124: Rather than doing this, couldn't we just set `c.Version = version` in the if statement checking that a version is set?

pacejackson (changes requested):
Any particular reason for adding this now? It doesn't sound like something that is really used right now and I would bias towards not adding things like this that we don't have any plans to use.

pacejackson (approved):

kafkabp/config.go
fishy: yea we can also do that. but I _slightly_ prefer this approach because:

1. We want to return all errors early, before constructing the actual sarama config, as much as possible.
2. With 1 in mind, this approach avoids a second `if cfg.Version != ""` block, so it's slightly better.

konradreiche (changes requested):
Could we add some more documentation around Rack awareness? Usually the user should go to the documentation of the respective technology to learn about this, but when I search for `rack awareness kafka` the first hit is an elaborate answer in a community forum which tells me there would be value in giving some guidance to our users too.

konradreiche (approved):

PR #309, opened 2020-12-01
timebp: Cleanup incorrect comments

fishy:
Those comments only apply when we use nanoseconds since EPOCH as
timestamp, not milliseconds/microseconds. I can't remember why I added
them in the first place...

https://play.golang.org/p/u-9jx4H7qZU

### DIFF
--- a/timebp/microsecond.go
+++ b/timebp/microsecond.go
@@ -59,8 +59,6 @@ func MicrosecondsToTime(us int64) time.Time {
 	if us == 0 {
 		return time.Time{}
 	}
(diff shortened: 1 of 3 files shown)


konradreiche (approved):

pacejackson (approved):

fishy:
Oh I remembered. The comments were there because the initial implementation is to multiply the timestamp passed in by 1000 (for us) or 1000000 (for ms) as the nanosec arg of `time.Unix` and always set the sec arg to 0. The new implementation doesn't have the same limitation.

PR #311, opened 2020-12-02
kafkabp: mock producer to test group consumer

konradreiche:
This adds a test for `groupConsumer` in `kafkabp/group_consumer.go`. It mocks out the underlying `sarama.GroupConsumer` interface with `mockGroupConsumer` that simulates a producer. `testConsumer` provides the sample implementation of a group consumer and the test checks if the messages are all processed and if closing the consumer works as expected.

There is some complexity in managing the different states of those types that could potentially be simplified.

### DIFF
--- a/kafkabp/group_consumer_test.go
+++ b/kafkabp/group_consumer_test.go
@@ -1,0 +1,169 @@
+package kafkabp
+
+import (
+	"context"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/Shopify/sarama"
+)
+
+func TestGroupConsumer(t *testing.T) {
+	// mock producer side by injecting a mocked consumer group
+	mockProducer := newMockConsumerGroup()
+
+	// test implemenation of a consumer
+	testConsumer := newTestConsumer()
+
+	// inject mocked consumer group
+	gc := &groupConsumer{
+		consumer: mockProducer,
+	}
+
+	// start group consumer in the background and collect all errors
+	errors := make(chan error)
+	go func() {
+		errors <- gc.Consume(testConsumer.Consume, testConsumer.Error)
+	}()
+
+	// expect two messages
+	testConsumer.expectedMessages = 2
+
+	// producer will delivery messages to test consumer
+	mockProducer.messages <- &sarama.ConsumerMessage{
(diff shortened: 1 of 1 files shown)


kafkabp/group_consumer_test.go
```diff
@@ -0,0 +1,169 @@
...
+	// test implemenation of a consumer
+	testConsumer := newTestConsumer()
+
+	// inject mocked consumer group
+	gc := &groupConsumer{
+		consumer: mockProducer,
+	}
+
+	// start group consumer in the background and collect all errors
+	errors := make(chan error)
```
fishy, line 25: Since you are sending one message to this channel, I would suggest make it buffered with size 1, so the sender won't get blocked (it would cause the goroutine to hang forever if there's a bug causing there's no reader to consume from this channel, because sender is blocked).

kafkabp/group_consumer_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	mockProducer.messages <- &sarama.ConsumerMessage{
+		Value: []byte(`{"data":"b"}`),
+	}
+	mockProducer.claimsAssigned <- true
+
+	// wait for processing to finish but fail if it blocks too long
+	select {
+	case <-testConsumer.processed:
+	case <-time.After(1 * time.Second):
```
fishy, line 45: 🔕 I would prefer a lower timeout for unit tests, mainly for making sure that running the unit test won't take too long. But I understand that using a lower timeout has the risk of making the tests more flaky.

kafkabp/group_consumer_test.go
```diff
@@ -0,0 +1,169 @@
...
+		t.Fatal("consumer did not process all messages")
+	}
+
+	expected := []string{
+		`{"data":"a"}`, `{"data":"b"}`,
```
fishy, line 50: style nit: one per line.

fishy (commented):
lgtm in general but have some nitpicks.

fishy (changes requested):
cleaning up my harold dashboard. please re-request my review when this PR is ready to be reviewed.

konradreiche:
Closing for now.

PR #312, opened 2020-12-04
Use forked zap

fishy:
It's intentionally split into 2 commits, so we can revert the first one (and also update upstream zap to newer version) once the patch is accepted by upstream.


konradreiche (approved):

pacejackson (approved):

fishy:
Close as we have https://github.com/reddit/baseplate.go/pull/314 instead.

PR #313, opened 2020-12-04
thriftbp: add a helper method for creating new baseplate.Error objects

pacejackson:
Creating `baseplate.Error` objects is kind of annoying in Go because of the type conversions and need to use pointers, this adds a small helper function that you can use to create a new `baseplate.Error` without having to worry about that.

### DIFF
--- a/thriftbp/errors.go
+++ b/thriftbp/errors.go
@@ -59,3 +59,13 @@ func BaseplateErrorFilter(codes ...int32) retrybp.Filter {
 		return next(err)
 	}
 }
+
+// NewBaseplateError is a helper function for creating baseplate.Error thrift objects.
+func NewBaseplateError(code baseplatethrift.ErrorCode, message string) *baseplatethrift.Error {
+	asInt32 := int32(code)
+	return &baseplatethrift.Error{
+		Code:    &asInt32,
+		Message: &message,
+		Details: make(map[string]string),
+	}
+}
--- a/thriftbp/errors.go
+++ b/thriftbp/errors.go
@@ -63,10 +63,10 @@ func BaseplateErrorFilter(codes ...int32) retrybp.Filter {
 }
 
 // NewBaseplateError is a helper function for creating baseplate.Error thrift
-// objects.
+// objects to avoid manual type-conversion that the generated Thrift code requires.
 //
 // detailsKeysValues is used to populate baseplate.Error.Details and should have
-// an even number of values, if it has an odd number of values, the last value
+// an even number of values. If it has an odd number of values the last value
 // will be discarded. The first entry in each pair of values will be the key and
 // the second entry will be the value. If no values are provided, the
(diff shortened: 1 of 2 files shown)


thriftbp/errors.go
```diff
@@ -59,3 +59,13 @@ func BaseplateErrorFilter(codes ...int32) retrybp.Filter {
 		return next(err)
 	}
 }
+
+// NewBaseplateError is a helper function for creating baseplate.Error thrift objects.
+func NewBaseplateError(code baseplatethrift.ErrorCode, message string) *baseplatethrift.Error {
+	asInt32 := int32(code)
+	return &baseplatethrift.Error{
+		Code:    &asInt32,
```
fishy, line 67: Why is the type defined as `optional i32` instead of `optional ErrorCode` in thrift :shakes-fist:

This actually can be inlined as `thrift.Int32Ptr(int32(code))`.

thriftbp/errors.go
```diff
@@ -59,3 +59,13 @@ func BaseplateErrorFilter(codes ...int32) retrybp.Filter {
...
+		Message: &message,
+		Details: make(map[string]string),
```
fishy, line 69: I think this behavior should be documented, e.g. we don't expect caller to return this error directly, we expect them to fill in the details into the error returned then return the error.

thriftbp/errors.go
pacejackson: So devs can add their own error codes.

thriftbp/errors.go
```diff
@@ -59,3 +61,25 @@ func BaseplateErrorFilter(codes ...int32) retrybp.Filter {
 		return next(err)
 	}
 }
+
+// NewBaseplateError is a helper function for creating baseplate.Error thrift
```
konradreiche, line 65: What you wrote in your PR description I would expect here, something along the lines of:

```
NewBaseplateError is a helper function for creating baseplate.Error thrift objects to avoid manual type-conversion that the generated Thrift code requires.
```

konradreiche (approved):

thriftbp/errors.go
pacejackson: updated

pacejackson:
Hmmmmm, I'm not sure this actually works because this is returning the "Error" type from the generated code in "baseplate.go" while services will have their own generated versions... 😭

fishy:
> Hmmmmm, I'm not sure this actually works because this is returning the "Error" type from the generated code in "baseplate.go" while services will have their own generated versions... 😭

Yes :(

Also there really isn't that much of boilerplate code. Service could inline everything:

```go
return &baseplatethrift.Error{
  Code:    thrift.Int32Ptr(404),
  Message: thrift.StringPtr("not found"),
  Details: map[string]string{
    "key1": "value1",
    "key2": "value2",
  },
}
```

PR #318, opened 2020-12-18
randbp: Small tweak on the GetSeed implementation

fishy:
When reading int64 from crypto/rand failed:

1. Log the error to stderr
2. In case of partial failures (read some bytes but not enough for
   int64), xor the bytes read against time so that time is not the only
   entropy of the fallback seed.

### DIFF
--- a/randbp/seed.go
+++ b/randbp/seed.go
@@ -3,18 +3,39 @@ package randbp
 import (
 	"crypto/rand"
 	"encoding/binary"
+	"fmt"
+	"io"
+	"os"
 	"time"
 )
 
-// GetSeed returns a seed for pseudo-random generator.
+// Define as variables so they can be mocked in unit tests.
+var (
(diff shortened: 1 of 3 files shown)


konradreiche (approved):

randbp/seed.go
```diff
@@ -3,18 +3,39 @@ package randbp
...
+	"fmt"
+	"io"
+	"os"
 	"time"
 )
 
-// GetSeed returns a seed for pseudo-random generator.
+// Define as variables so they can be mocked in unit tests.
+var (
+	cryptoReader               = rand.Read
```
pacejackson, line 14: Calling this “cryptoReader” seems misleading since it’s using rand rather than crypto

fishy: it's `crypto/rand` (there's also `math/rand`)

pacejackson (approved):

PR #320, opened 2020-12-30
Add redisx and redispipebp, deprecate redisbp

pacejackson:
Package redisx wraps https://github.com/joomcode/redispipe in an interface where
rather than returning the result, you pass in a pointer to the variable you want
to put the result into  and it uses reflection to do that.

It is inspired by https://github.com/jmoiron/sqlx and  https://github.com/mediocregopher/radix


pacejackson:
This is mostly the same as the internal package this will replace, the differences are:

1. `CircuitBreaker` interface is in `breakerbp` now.
2. Added `redispipebp/baseplate.go` with methods for creating standard "baseplate" clients.
3. `Syncx.SendMany` and `Syncx.SendTransaction` are variadic functions now rather than taking a list of `Request`s as input.

breakerbp/failure_ratio.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/reddit/baseplate.go/metricsbp"
 )
 
+// CircuitBreaker is the interface that baseplate expects a circuit breaker to implement.
+type CircuitBreaker interface {
+	// Execute should wrap the given function call in circuit breaker logic and return the result.
+	Execute(func() (interface{}, error)) (interface{}, error)
+}
```
fishy, line 19: nit: maybe move it to its own file (e.g. `interface.go`) so it makes more sense if we are adding other implementations in the future.

redis/redispipebp/BUILD.bazel
```diff
@@ -0,0 +1,56 @@
...
+        "@com_github_joomcode_redispipe//redis:go_default_library",
+        "@com_github_joomcode_redispipe//rediscluster:go_default_library",
+        "@com_github_joomcode_redispipe//redisconn:go_default_library",
+        "@com_github_opentracing_opentracing_go//:go_default_library",
+    ],
+)
+
+go_test(
+    name = "go_default_test",
+    srcs = [
```
fishy, line 32: nit: add `size = "small",` between the `name` and `srcs` lines.

redis/redispipebp/config.go
```diff
@@ -0,0 +1,202 @@
...
+	// DB is the database number in Redis.
+	//
+	// Maps to redisconn.Opts.DB
+	DB *int `yaml:"db"`
+
+	// WritePause is the duration of time the write loop should wait to collect
+	// commands to flush.
+	//
+	// Maps to redisconn.Opts.WritePause
+	WritePause *time.Duration `yaml:"writePause"`
```
fishy, line 118: 🔕 Maybe for durations, using non-pointers and checking for `>0` instead of `!= nil` instead is better?

redis/redisx/syncx.go
```diff
@@ -0,0 +1,144 @@
+package redisx
+
+import (
+	"context"
+
+	"github.com/joomcode/redispipe/redis"
+	"github.com/reddit/baseplate.go/errorsbp"
```
fishy, line 7: style nit: add a blank line between them.

fishy (commented):
it's not clear to me whether do we want our users to use `redisx` or `redispipebp`, or both. maybe we should say more in the package doc comments?

redis/redispipebp/BUILD.bazel
pacejackson: No problem but what does that do?

fishy: without it bazel will print out a warning when running with `--test_verbose_timeout_warnings` arg, and we enforced that arg in [our `.bazelrc` file](https://github.com/reddit/baseplate.go/blob/master/.bazelrc).

basically bazel test uses the size (or similar tags) info to decide what the timeout for this test should be (larger tests will have longer timeouts before declare failure), and this info also helps bazel to figure out a strategy/policy to run tests in parallel, etc.

external.bzl
```diff
@@ -16,15 +16,16 @@ def go_dependencies():
...
-        name = "com_github_alicebob_miniredis",
-        importpath = "github.com/alicebob/miniredis",
-        sum = "h1:yBHoLpsyjupjz3NL3MhKMVkR41j82Yjf3KFv7ApYzUI=",
-        version = "v2.5.0+incompatible",
+        name = "com_github_alicebob_miniredis_v2",
+        importpath = "github.com/alicebob/miniredis/v2",
+        sum = "h1:QWoo2wchYmLgOB6ctlTt2dewQ1Vu6phl+iQbwT8SYGo=",
+        version = "v2.14.3",
     )
+
```
fishy, line 28: nit, and this really is a bug in gazelle that it's inconsistent I should report: when it first generate this file it doesn't add those blank lines, and then when it makes changes it adds those blank lines after every block it changed. currently I usually just manually remove those blank lines.

redis/redispipebp/baseplate.go
```diff
@@ -0,0 +1,88 @@
+package redispipebp
```
fishy, line 1: 🔕 lol love the filename.

redis/redispipebp/errors.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	if target, ok := v.(**errorx.Error); ok {
+		*target = e.Errorx
+		return true
+	}
+	return errors.As(e.Unwrap(), v)
+}
+
+// Unwrap implements helper interface for errors.Unwrap.
+// Unwraps the underlying errorx.Error by calling Cause() and wrapping
```
fishy, line 36: blank comment line before this line.

redis/redispipebp/monitored.go
```diff
@@ -0,0 +1,144 @@
+package redispipebp
+
+import (
+	"context"
+	"errors"
+
+	"github.com/joomcode/errorx"
+	"github.com/joomcode/redispipe/redis"
+	"github.com/opentracing/opentracing-go"
+	"github.com/reddit/baseplate.go/tracing"
```
fishy, line 10: nit: tracing should be in the 3rd group.

redis/redispipebp/monitored_test.go
```diff
@@ -0,0 +1,276 @@
+package redispipebp_test
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/joomcode/redispipe/redis"
+	"github.com/reddit/baseplate.go/tracing"
```
fishy, line 10: same here.

redis/redispipebp/retry.go
```diff
@@ -0,0 +1,97 @@
+package redispipebp
+
+import (
+	"context"
+	"errors"
+
+	"github.com/avast/retry-go"
+	"github.com/joomcode/redispipe/redis"
+	"github.com/reddit/baseplate.go/retrybp"
```
fishy, line 9: same here.

redis/redisx/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package redisx wraps https://github.com/joomcode/redispipe in an interface where
+// rather than returning the result, you pass in a pointer to the variable you want
+// to put the result into  and it uses reflection to do that.
+// It is inspired by https://github.com/jmoiron/sqlx and  https://github.com/mediocregopher/radix
```
fishy, line 4: blank comment line before this line.

fishy (approved):
mostly only nitpicks on doc comments left.

external.bzl
fishy: what I meant is that it would be great if you can manually remove those added blank lines 😅

fishy (approved):

external.bzl
pacejackson: image

My brain didn't look closely enough at this and I thought you were talking about a newline at the end of the file...

konradreiche (approved):

PR #321, opened 2021-01-08
thriftbp: Add header to write header list in SetDeadlineBudget

fishy:
We only set the header in ctx, but not add it to write header list, and
that's a bug causing we don't really add the header to the wire.

### DIFF
--- a/thriftbp/client_middlewares.go
+++ b/thriftbp/client_middlewares.go
@@ -229,7 +229,7 @@ func SetDeadlineBudget(next thrift.TClient) thrift.TClient {
 					ms = 1
 				}
 				value := strconv.FormatInt(ms, 10)
-				ctx = thrift.SetHeader(ctx, HeaderDeadlineBudget, value)
+				ctx = AddClientHeader(ctx, HeaderDeadlineBudget, value)
 			}
 
 			return next.Call(ctx, method, args, result)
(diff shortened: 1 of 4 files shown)


thriftbp/client_middlewares.go
```diff
@@ -230,6 +230,10 @@ func SetDeadlineBudget(next thrift.TClient) thrift.TClient {
 				}
 				value := strconv.FormatInt(ms, 10)
 				ctx = thrift.SetHeader(ctx, HeaderDeadlineBudget, value)
+
+				headers := thrift.GetWriteHeaderList(ctx)
+				headers = append(headers, HeaderDeadlineBudget)
+				ctx = thrift.SetWriteHeaderList(ctx, headers)
```
pacejackson, line 236: Should we have a helper method for this?

pacejackson (approved):

thriftbp/client_middlewares.go
fishy: this can be written in one line if you really wanted to:

```go
ctx = thrift.SetWriteHeaderList(ctx, append(thrift.GetWriteHeaderList(ctx), HeaderDeadlineBudget))
```

so... maybe?

pacejackson: Less about the exact lines of code it takes and more just that there's this exact series of steps you have to follow when you want to do this, it's easy to forget to do this, having a helper might help with that?

fishy: good point. done.

pacejackson (approved):

konradreiche (approved):

PR #326, opened 2021-01-25
thriftbp.ReportPayloadSizeMetrics: Add proto id to metrics tags

fishy:


thriftbp/server_middlewares.go
```diff
@@ -316,3 +318,14 @@ func (countingTransport) Read(_ []byte) (n int, err error) { return }
...
 func (countingTransport) Open() (err error) { return }
+
+func tHeaderProtocol2String(proto thrift.THeaderProtocolID) string {
+	switch proto {
+	default:
+		return fmt.Sprintf("%v", proto)
+	case thrift.THeaderProtocolCompact:
+		return "compact"
+	case thrift.THeaderProtocolBinary:
+		return "binary"
```
pacejackson, line 329: what's wrong with just using `%v` for these two?

pacejackson (approved):

thriftbp/server_middlewares.go
fishy: so we didn't implement `THeaderProtocolID.String()`, so if we just use `%v` it will be `header-0` and `header-2`, which is not very readable/user-friendly.

konradreiche (approved):

PR #329, opened 2021-01-27
thriftbp: Fix client pool bug

fishy:
This fixes a bug introduced in 8354a04 causing us to never close broken
connections and always return them back to the pool.

Also take advantage of the newly added TException types to help
determine whether to close the connection or return to the pool.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -481,8 +481,9 @@ func (p *clientPool) wrapCalls(middlewares ...thrift.ClientMiddleware) {
 //
 // This is not called directly, but is rather the inner "Call" wrapped by
 // wrapCalls, so it runs after all of the middleware.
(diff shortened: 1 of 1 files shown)


thriftbp/client_pool.go
```diff
@@ -521,6 +522,20 @@ func shouldCloseConnection(err error) bool {
 	if err == nil {
 		return false
 	}
+	var te thrift.TException
+	if errors.As(err, &te) {
+		switch te.TExceptionType() {
+		case thrift.TExceptionTypeApplication, thrift.TExceptionTypeProtocol, thrift.TExceptionTypeTransport:
+			// We definitely should close connectino on TTransportException.
```
pacejackson, line 529: small typo:

```suggestion
			// We definitely should close the connection on TTransportException.
```

pacejackson (approved):

fishy:
Konrad is OOO so I'm gonna just merge it.

PR #330, opened 2021-01-28
thriftbp: Add closed connections counter to client pools

fishy:
The bug fixed in 6bfad2e highlights the fact that this is a part of the
code that has very limited external visibility, and when a bug happens
there it's very hard to debug (it took me almost 2 days to find it).

So adding a counter to provide external visibility, and if a similar bug
happens in the future, it would be more obvious that this counter
dropped while the actual failure rate remains the same.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -385,6 +385,9 @@ func newClientPool(
 		poolExhaustedCounter: metricsbp.M.Counter(
 			cfg.ServiceSlug + ".pool-exhausted",
 		).With(tags...),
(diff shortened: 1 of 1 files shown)


konradreiche (approved):

pacejackson (approved):

fishy:
:haircut: Dropped "bad" from the counter and renamed the field. Although most of the closed connections are indeed bad it's not necessarily 100% bad.

pacejackson (approved):

konradreiche (approved):

PR #332, opened 2021-01-29
retrybp: Implement retry-after logic in CappedExponentialBackoff

fishy:
Define retrybp.RetryAfterError interface and add support to it to
CappedExponentialBackoff, and implement it in httpbp as
httpbp.ClientError.

### DIFF
--- a/httpbp/errors.go
+++ b/httpbp/errors.go
@@ -3,9 +3,14 @@ package httpbp
 import (
 	"fmt"
 	"html/template"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
+
+	"github.com/reddit/baseplate.go/errorsbp"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


fishy:
Note: before this change we mainly focused on http server in `httpbp` package, and the added APIs to `httpbp` package are focused on the client side. I'm not sure if we want to support them as our API. If we don't feel like support them, I can move that into an example to retrybp.RetryAfterError instead.

httpbp/errors.go
```diff
@@ -633,3 +638,112 @@ func ErrorForCode(code int) *ErrorResponse {
...
+// The caller can (optionally) choose to read the body and feed that back into
+// ClientError.AdditionalInfo (it's always empty string when returned), e.g.:
+//
+//     resp, err := http.Get("https://www.reddit.com")
+//     if err != nil {
+//       return err
+//     }
+//     defer httpbp.DrainAndClose(resp.Body)
+//     err = httpbp.ClientErrorFromResponse(resp)
+//     if err != nil && errors.As(err, &ce) {
```
pacejackson, line 695: `ce` doesn't exist yet in this example.

fishy: :doh: I forgot to delete the `&& ...` part when splitting the if.

konradreiche (approved):
Generally fine with supporting this as part of our API since in reality many of our code bases will need this and I am not sure how else to incorporate this code.

pacejackson (approved):

httpbp/errors.go
fishy: Actually I moved this to testable example so it's guaranteed to be correct code :)

PR #335, opened 2021-02-09
Add ecinterface package

fishy:
And switch everything else to use it instead of edgecontext package.

This is the first step of splitting edgecontext implementation to its
own repo. After this change, nothing in Baseplate.go depends on
edgecontext package any more, so we could move it to its own repo and
remove it from Baseplate.go in the next step.

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -11,7 +11,7 @@ import (
 	yaml "gopkg.in/yaml.v2"
 
 	"github.com/reddit/baseplate.go/batchcloser"
-	"github.com/reddit/baseplate.go/edgecontext"
+	"github.com/reddit/baseplate.go/ecinterface"
 	"github.com/reddit/baseplate.go/errorsbp"
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/metricsbp"
@@ -51,7 +51,7 @@ type Baseplate interface {
 	io.Closer
 
 	Config() Config
-	EdgeContextImpl() *edgecontext.Impl
+	EdgeContextImpl() ecinterface.Interface
 	Secrets() *secrets.Store
 }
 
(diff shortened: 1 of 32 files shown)


konradreiche (approved):
I could see a different naming scheme for edge context, interface and impl help differentiate this concepts more but I would not assume this to be a blocker since it is not what the API users interact with on a daily basis.

baseplate.go
```diff
@@ -242,8 +256,8 @@ func DecodeConfigYAML(reader io.ReadSeeker, serviceCfg interface{}) (Config, err
 // will also decode the config file at the path to set it up.  This can be used
 // to parse additional, service specific config values from the same config
 // file.
```
pacejackson, line 258: We should probably remove this chunk since this moved into `NewArgs`

baseplate.go
```diff
@@ -338,11 +349,11 @@ func (bp impl) Close() error {
 //
 // NewTestBaseplate only returns a Baseplate, it does not initialize any of
 // the monitoring or logging frameworks.
-func NewTestBaseplate(cfg Config, store *secrets.Store) Baseplate {
+func NewTestBaseplate(cfg Config, store *secrets.Store, ecImpl ecinterface.Interface) Baseplate {
```
pacejackson, line 352: Should we also use an "Args" struct for this?

pacejackson (approved):
Looks good overall, one question and one docstring that needs cleaning up was all I saw

baseplate.go
```diff
@@ -242,8 +256,8 @@ func DecodeConfigYAML(reader io.ReadSeeker, serviceCfg interface{}) (Config, err
 // will also decode the config file at the path to set it up.  This can be used
 // to parse additional, service specific config values from the same config
 // file.
-func New(ctx context.Context, path string, serviceCfg interface{}) (context.Context, Baseplate, error) {
-	cfg, err := ParseConfig(path, serviceCfg)
+func New(ctx context.Context, args NewArgs) (context.Context, Baseplate, error) {
```
fishy, line 259: Actually this doesn't work. The edgecontext implementation requires an already initialized secret store, but they don't have that before calling `New` 🤦

pacejackson: :doh:

httpbp/example_server_test.go
```diff
@@ -94,9 +93,9 @@ var (
 func ExampleNewBaseplateServer() {
 	var cfg config
 	ctx, bp, err := baseplate.New(context.Background(), baseplate.NewArgs{
-		EdgeContextImpl: ecinterface.Mock(), // In real code this should be a real implementation instead.
-		ConfigPath:      "example.yaml",
-		ServiceCfg:      &cfg,
+		ConfigPath:         "example.yaml",
+		EdgeContextFactory: nil, // In real code this should be the factory from the actual implementation.
```
pacejackson, line 97: Note, I think having this be nil will cause `baseplate.New` to panic, so probably not what we want to do in an example 😅

fishy: I don't feel like providing a mocked factory, though 😅 the only place the mocked factory will be used is in example code.

fishy: Is it better now?

baseplate.go
```diff
@@ -235,10 +235,11 @@ func DecodeConfigYAML(reader io.ReadSeeker, serviceCfg interface{}) (Config, err
 
 // NewArgs defines the args used in New functino.
 type NewArgs struct {
-	EdgeContextImpl ecinterface.Interface
-
 	ConfigPath string
 
+	// The factory used to create edge context implementation.
```
pacejackson, line 240: We should probably say that this is required and `baseplate.New` will panic if it is not set?

pacejackson (commented):
Ya, maybe we should just note that `baseplate.New` will panic if the Factory is nil? I actually think that that behavior is fine since we don't want to accidentally set up a service without one, just should note it.

httpbp/example_server_test.go
```diff
@@ -92,7 +93,13 @@ var (
 // Baseplate HTTP service.
 func ExampleNewBaseplateServer() {
 	var cfg config
-	ctx, bp, err := baseplate.New(context.Background(), "example.yaml", &cfg)
+	// In real code this MUST be replaced by the factory from the actual implementation.
+	var ecFactory ecinterface.Factory
```
pacejackson, line 97: 👍  this seems like a reasonable way to handle this

pacejackson (approved):

konradreiche (approved):

PR #337, opened 2021-02-11
tracing: Support UUID

fishy:
Add support to UUID and option UseUUID to Tracer's Config.

Also change the log key for trace id from trace_id to traceID.

### DIFF
--- a/tracing/tracer.go
+++ b/tracing/tracer.go
@@ -2,12 +2,15 @@ package tracing
 
 import (
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"strconv"
 	"time"
 
+	"github.com/gofrs/uuid"
 	opentracing "github.com/opentracing/opentracing-go"
 
 	"github.com/reddit/baseplate.go/log"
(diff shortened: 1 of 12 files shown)


fishy:
cc @spladug

konradreiche (approved):

tracing/tracer.go
```diff
@@ -294,6 +305,27 @@ func (t *Tracer) Extract(format interface{}, carrier interface{}) (opentracing.S
...
+				context.Background(),
+				fmt.Sprintf("Failed to generate uuid: %v", err),
+			)
+			// On modern linux kernel, after the system entropy pool is initialized
+			// it's guaranteed to be able to get up to 256 bytes in one call without
+			// error [1], so this should never happen.
+			// If this really happens we don't really have a good way to deal with it
+			// anyways.
+			// [1]: https://man7.org/linux/man-pages/man2/getrandom.2.html
+			panic(err)
```
fishy, line 322: I wonder if we should just return empty string in this case? @pacejackson @konradreiche what do you think?

konradreiche: It would be interesting to learn about it but this would be outside of the initialization path of the program so I would be definitely for returning an empty string instead of panic.

spladug (approved):
sweet. thanks! lgtm modulo the panic

tracing/tracer.go
pacejackson: image

Agree on it probably being better to return an empty string, I don't think we want initializing spans to potentially cause requests to fail.

fishy:
:haircut: I changed to use fake uuid instead of panicking.

konradreiche (approved):

spladug (approved):

pacejackson (approved):

PR #340, opened 2021-02-18
tracing+metricsbp: New tracing metrics

fishy:
Emit new tracing metrics according to baseplate spec, and also keep
emitting old ones for now to not break all the dashboards. The old ones
are marked deprecated and legacy and will be removed in 0.9.0.

Also remove LocalComponentName from local spans. Services still need
that could set the tag "lc" manually.


fishy:
cc @spladug

konradreiche (approved):

pacejackson (approved):

fishy:
:haircut:

I realized that we don't really need set/generics support to do the tags carry over.

pacejackson (approved):

konradreiche (approved):

PR #341, opened 2021-02-25
[WIP] Add sqlbp

jstrate-reddit:


sqlbp/client.go
```diff
@@ -0,0 +1,190 @@
...
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"time"
+
+	"github.com/go-kit/kit/metrics"
+	"github.com/jmoiron/sqlx"
+	"github.com/luna-duclos/instrumentedsql"
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/reddit/baseplate.go/metricsbp"
```
fishy, line 13: style nit: add a blank line before this line. see: https://github.com/reddit/baseplate.go/blob/master/Style.md#import-groups

sqlbp/client.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+type tracer struct {
+	prefix string
+}
+
+type span struct {
+	tracer
+	parent opentracing.Span
```
fishy, line 24: style nit: add a blank line between embedded fields and other fields

sqlbp/client.go
```diff
@@ -0,0 +1,190 @@
...
+func NewTracer(prefix string) instrumentedsql.Tracer {
+	return tracer{
+		prefix: prefix,
+	}
+}
+
+// GetSpan returns a span from the context, or creates one.
+func (t tracer) GetSpan(ctx context.Context) instrumentedsql.Span {
+
+	if ctx == nil {
```
fishy, line 38: while technically you can pass in a nil ctx, this is considered the caller's fault, so in reality we never check for it. see: https://go.googlesource.com/go/+/go1.16/src/context/context.go#37

sqlbp/client.go
```diff
@@ -0,0 +1,190 @@
...
+		parent := opentracing.StartSpan(
+			t.prefix,
+			tracing.SpanTypeOption{
+				Type: tracing.SpanTypeClient,
+			})
```
fishy, line 43: style nit: `},\n)`

sqlbp/client.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// GetSpan returns a span from the context, or creates one.
+func (t tracer) GetSpan(ctx context.Context) instrumentedsql.Span {
+
+	if ctx == nil {
+		parent := opentracing.StartSpan(
+			t.prefix,
+			tracing.SpanTypeOption{
+				Type: tracing.SpanTypeClient,
```
fishy, line 42: 🔕 In `NewChild` function you also make the child span a client span, so making both parent and child span client spans sounds a little bit weird to me. Maybe here it should be a local span instead? I don't feel strongly either way, though.

sqlbp/client.go
```diff
@@ -0,0 +1,190 @@
...
+		tracer: s.tracer,
+	}
+}
+
+func (s *span) SetLabel(k, v string) {
+	s.parent.SetTag(k, v)
+}
+
+func (s *span) SetError(err error) {
+	if err == nil || err == driver.ErrSkip {
```
fishy, line 78: the second one probably should be `errors.Is(err, driver.ErrSkip)` instead.

sqlbp/client.go
```diff
@@ -0,0 +1,190 @@
...
+		return
+	}
+	s.err = err
+}
+
+func (s *span) Finish() {
+	s.parent.FinishWithOptions(tracing.FinishOptions{
```
fishy, line 85: the name `parent` starts to confuse me. why do you call `parent.FinishWithOptions` in `Finish` implementation? does `parent` actually mean this span, instead of its parent span?

PR #345, opened 2021-03-09
ecinterface: Add global implementation

fishy:
And change all other packages taking ecinterface.Interface as an arg to
fallback to the global one when the arg is nil.

### DIFF
--- a/edgecontext/edgecontext.go
+++ b/edgecontext/edgecontext.go
@@ -100,12 +100,15 @@ func Factory(cfg Config) ecinterface.Factory {
 }
 
 // Init intializes an Impl.
+//
+// It also calls ecinterface.Set to store the implementation created globally.
 func Init(cfg Config) *Impl {
 	impl := &Impl{
 		store:  cfg.Store,
 		logger: cfg.Logger,
 	}
 	impl.store.AddMiddlewares(impl.validatorMiddleware)
+	ecinterface.Set(impl)
 	return impl
 }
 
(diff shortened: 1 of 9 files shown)


edgecontext/edgecontext.go
```diff
@@ -100,12 +100,15 @@ func Factory(cfg Config) ecinterface.Factory {
...
 // Init intializes an Impl.
+//
+// It also calls ecinterface.Set to store the implementation created globally.
 func Init(cfg Config) *Impl {
 	impl := &Impl{
 		store:  cfg.Store,
 		logger: cfg.Logger,
 	}
 	impl.store.AddMiddlewares(impl.validatorMiddleware)
+	ecinterface.Set(impl)
```
pacejackson, line 111: I'm not sure we actually want to combine these? What if you (for some reason) actually _want_ to have 2 Impls?

fishy: this is the reason I didn't remove the `ecinterface.Interface` arg from where they are used to only use the global one, if people really want to use a different one in certain places they can by using that arg.

not doing this will result in people upgrading to baseplate.go v0.8.0 would need to remember to call `Set` after the init, which could be easily forgot. the downside is that if for some reason they call `edgecontext.Init` twice, the second call will overwrite the first one for the global value, but as long as they saved the first one in a variable they can still use that one to do overrides.

httpbp/middlewares.go
```diff
@@ -161,6 +162,9 @@ func InitializeEdgeContextFromTrustedRequest(
 		args.Logger.Log(ctx, "Error while parsing EdgeRequestContext: "+err.Error())
 		return ctx
 	}
+	if args.EdgeContextImpl == nil {
+		args.EdgeContextImpl = ecinterface.Get()
```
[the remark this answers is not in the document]
fishy: right I forgot that one 😅

pacejackson (approved):

konradreiche (approved):

PR #346, opened 2021-03-09
tracing: Improve race condition in tests

fishy:
This change improves the race condition detected in
https://cloud.drone.io/reddit/baseplate.go/1690/2/2.

To completely fix that we would need to change the globalTracer into an
atomic value, which is actually overkill as this race condition would
almost only happen in tests and is not a real threat to production code.


pacejackson (approved):
If I'm understanding correctly the "race" condition is... If you called `InitGlobalTracer` concurrently which should never happen outside of tests right?

fishy:
> If I'm understanding correctly the "race" condition is... If you called `InitGlobalTracer` concurrently which should never happen outside of tests right?

yes.

PR #347, opened 2021-03-10
tracing: Add UUIDIntact config to tracer

fishy:
Default to false and remove hyphens from generated uuids, as that could
break zipkin collectors.


konradreiche (approved):

spladug (approved):

pacejackson (approved):

fishy:
Confirmed that this works on our staging service.

PR #354, opened 2021-03-18
Add healthcheck binary and library

fishy:
This allows go services to get healthcheck without depending on
Baseplate.py. They can either build the healthcheck binary from
github.com/reddit/baseplate.go/cmd/healthcheck, or create a main package
in their repo and use the library version.

The implementation matches the behavior (including command line flags)
of the Baseplate.py implementation [1], with the exception of unix
domain socket, and added --timeout flag.

Also change the default timeout from Baseplate.py's 30s to 1s, as 30s
for a healthcheck timeout is too high.

[1]: https://github.com/reddit/baseplate.py/blob/develop/baseplate/server/healthcheck.py

### DIFF
--- a/cmd/lib/healthcheck/healthcheck_test.go
+++ b/cmd/lib/healthcheck/healthcheck_test.go
@@ -1,0 +1,273 @@
+package healthcheck_test
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net"
+	"net/http"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/apache/thrift/lib/go/thrift"
+
+	"github.com/reddit/baseplate.go/cmd/lib/healthcheck"
+	"github.com/reddit/baseplate.go/httpbp"
+	"github.com/reddit/baseplate.go/internal/gen-go/reddit/baseplate"
+	"github.com/reddit/baseplate.go/thriftbp"
+)
+
+type healthyMap = map[baseplate.IsHealthyProbe]bool
+
+var (
+	allHealthy = healthyMap{
+		baseplate.IsHealthyProbe_READINESS: true,
+		baseplate.IsHealthyProbe_LIVENESS:  true,
+		baseplate.IsHealthyProbe_STARTUP:   true,
+	}
+	allUnhealthy = healthyMap(nil)
+)
+
+type service struct {
+	addr string
+	up   func(t *testing.T)
+	down func(t *testing.T)
+}
+
+type thriftHandler struct {
+	healthy healthyMap
+}
+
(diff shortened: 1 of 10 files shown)


konradreiche (approved):

cmd/lib/healthcheck/oneof.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (o *oneof) Set(v string) error {
+	if _, ok := o.choices[v]; ok {
+		o.value = v
+		return nil
+	}
+	return fmt.Errorf("%q is not one of the choices of %s", v, o.printChoices())
+}
+
+func (o oneof) printChoices() string {
```
pacejackson, line 33: 🔕 minor nit, this doesn't actually _print_ anything, maybe something like `choicesStr` would be a more appropriate name?

cmd/lib/healthcheck/healthcheck_test.go
```diff
@@ -0,0 +1,261 @@
...
+				baseplate.IsHealthyProbe_LIVENESS:  true,
+				baseplate.IsHealthyProbe_STARTUP:   false,
+			}),
+		},
+		{
+			label: "startup-unhealthy-thrift-2",
+			err:   false, // This one checks readiness probe so it should report healthy
+			service: thriftService(healthyMap{
+				baseplate.IsHealthyProbe_READINESS: true,
+				baseplate.IsHealthyProbe_LIVENESS:  true,
```
pacejackson, line 173: Should we set this to false as well?

cmd/lib/healthcheck/healthcheck_test.go
```diff
@@ -0,0 +1,261 @@
...
+			}),
+		},
+		{
+			label: "liveness-unhealthy-http-2",
+			args:  []string{"--type", "wsgi"},
+			err:   false, // This one checks readiness probe so it should report healthy
+			service: httpService(healthyMap{
+				baseplate.IsHealthyProbe_READINESS: true,
+				baseplate.IsHealthyProbe_LIVENESS:  false,
+				baseplate.IsHealthyProbe_STARTUP:   true,
```
pacejackson, line 205: should we set this to false as well?

pacejackson (approved):
Just a couple small comments, lgtm overall though

pacejackson (approved):
Kewl, would we consider deprecating the Python implementation in the medium to long term and just have everyone use this since, in theory, they can just install it as a binary and run it?

fishy:
After talked to @spladug I changed the default timeout from 30s to 1s.

spladug (approved):

PR #359, opened 2021-04-06
thriftbp: Add doc regarding the concurrency of ClientPool

fishy:
It could be confusing on whether the client created from ClientPool is
concurrent safe, and that could lead to footguns, so adding some doc to
explicitly call it out.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -236,6 +236,13 @@ type Client interface {
 
 // ClientPool defines an object that implements thrift.TClient using a pool of
 // Client objects.
+//
+// A ClientPool is safe to be shared across different goroutines,
+// but a concrete thrift client created on top of it is not.
+// A concrete thrift client is the one you created from thrift compiled go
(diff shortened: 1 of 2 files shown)


konradreiche (approved):

pacejackson (approved):

pacejackson (changes requested):
So it seems like this is actually a regression due to a change in Thrift that caused these clients to become silently non-concurrent safe and I think we should either try to address the regression or there are more things that need to be changed.

The baseplate client pool was meant to make it so you _could_ use the same client, that's why it pulls a new TClient from the pool for you before each call and puts it back.

pacejackson:
Note, I would prefer if we could address the regression itself as it is not intuitive that you need to new create a new client before each request and makes the API much more cumbersome than it was before.

fishy:
As discussed offline, to address the regression we would need to introduce a breaking change, so that will be done later, and this change is just the short term doc update.

pacejackson (approved):

konradreiche (approved):

PR #360, opened 2021-04-07
thriftbp: Add WrapBaseplateError and BaseplateErrorWrapper

fishy:
Thrift compiler generated baseplate.Error (from Error exception in
baseplate.thrift) does not provide meaningful messages when logged, so
implement WrapBaseplateError to wrap it to another error with meaningful
error message, and provide BaseplateErrorWrapper client middleware to
always do that automatically.


pacejackson (changes requested):
Can you add a test to show that:

```
var be &baseplate.Error
if errors.As(myWrappedError, *be)
```

works?

pacejackson (approved):

konradreiche (approved):

PR #368, opened 2021-04-19
Add HTTP client and middleware support to httpbp

konradreiche:
This adds support for HTTP clients by providing a monitored client and a set of middlewares. The way I see it being used is that a configuration is added by the user that could look like this:

```yaml
solr:
  slug: solr
  maxConnections: 100,
  retries: 2
  circuitBreaker:
    minRequests: 1
    maxRequest: 10
  loadBalancer:
    hosts:
      - https://solr-1:8983/solr
      - https://solr-2:8983/solr
      - https://solr-3:8983/solr
```

This is primarily a port from the HTTP client abstraction we have been using internally with the modification of having the client middleware be based on `http.RoundTripper`. Some questions that are worth discussing:
- What middlewares if any besides tracing should be default?
- Should there be an abstraction that provides the HTTP methods as Go methods?


httpbp/client_middlewares.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func (c *circuitBreaker) setupBreaker(host string) {
+	// this code path is executed rarely and usually at the start of a process
+	c.mu.RUnlock()
+	c.mu.Lock()
+	c.breakers[host] = gobreaker.NewCircuitBreaker(c.settings)
+	c.mu.Unlock()
+	c.mu.RLock()
+}
```
konradreiche, line 97: Can someone think of a better way than this locking Shashlik? The circuit breaker should be able to support a circuit breaker per host being queried and I can see how it makes sense to have it initialize lazy. In most cases we just want a read-lock and this is the path for initializing it requiring a write lock.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,239 @@
...
+			next: next,
+			slug: config.Slug,
+		}
+	}
+}
+
+func (m *monitorClient) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+	span, ctx := opentracing.StartSpanFromContext(
+		req.Context(),
+		m.slug,
```
konradreiche, line 225: I believe this should be `m.slug + ".request"` to match the baseplate.py implementation.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,239 @@
...
+// middlewares plus any additional client middlewares passed into this
+// function.
+func NewClient(config Config, middlewares ...ClientMiddleware) *http.Client {
+	transport := &http.Transport{
+		MaxConnsPerHost: config.MaxConnections,
+	}
+	defaults := []ClientMiddleware{
+		MonitorClient(config),
+		CircuitBreaker(config.CircuitBreaker),
+		Retries(config),
```
konradreiche, line 29: Any thoughts on default middlewares?

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,239 @@
...
+
+import (
+	"net/http"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/reddit/baseplate.go/breakerbp"
+	"github.com/reddit/baseplate.go/tracing"
```
fishy, line 11: nit: import groups.

httpbp/client_middlewares.go
fishy: I think we should have another one that calls `ClientErrorFromResponse` before and calls `DrainAndClose(resp.Body)` after (or just defer it).

httpbp/client_middlewares.go
fishy: I wrote https://pkg.go.dev/go.yhsif.com/defaultdict, which is basically pairing a `sync.Map` with `sync.Pool`.

fishy: but then again, do we really need this breaker to be per-host?

we also have `slug` for the client, which kinds of imply that we don't really use a single client for everything (like the default `http.Client`), but we actually create a client for every upstream. In that case, a single breaker should be sufficient?

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,239 @@
...
+	return transport
+}
+
+type circuitBreaker struct {
+	next     http.RoundTripper
+	settings gobreaker.Settings
+
+	mu       sync.RWMutex
+	breakers map[string]breakerbp.CircuitBreaker
+}
```
fishy, line 52: 🔕 I feel that those structs could be avoided if you just define a lambda type:

```go
type roundtripper func(req *http.Request) (*http.Response, error)

func (f roundtripper) RoundTrip(req *http.Request) (*http.Response, error) {
  return f(req)
}
```

this way all those structs can be actually inlined.

httpbp/client_middlewares.go
konradreiche: Possibly, the one scenario I can think of here is the load balancer, i.e. we are talking to different Solr read replicas and we'd want the breaker to sort out a faulty endpoint.

httpbp/client_middlewares.go
konradreiche: :+1: that would also be in-line with our other middleware implementations.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+}
+
+type retries struct {
+	next    http.RoundTripper
+	retries int
+}
+
+// Retries provides a retry middleware.
+func Retries(config Config) ClientMiddleware {
```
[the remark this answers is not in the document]
fishy: See https://github.com/reddit/baseplate.go/blob/v0.8.1/thriftbp/client_middlewares.go#L249 :)

httpbp/client_middlewares.go
konradreiche: The only caveat is that it cannot be passed directly to `http.Client` as transport as it expects `http.RoundTripper` but I think this can be covered by having an abstraction that will take care of delegating it to the `roundTripper` type.

konradreiche: Currently trying to apply this approach: would this require a `func type` definition for each middleware?

konradreiche: I refactored the client middleware interface, let me know if this is what you had in mind or if there is another way to reduce the complexity further.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+
+		resp, err := breaker.(*gobreaker.CircuitBreaker).Execute(func() (interface{}, error) {
+			resp, err := f(req)
+			if err != nil {
+				return nil, err
+			}
+			// circuit break on any HTTP 5xx code
+			if resp.StatusCode >= http.StatusInternalServerError {
+				err = resp.Body.Close()
```
gogsbread, line 97: You may want to both read and close the body for the transport to be reused. [go doc](https://golang.org/pkg/net/http/#Client.Do) seems to specify it that way

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,228 @@
...
+				if err != nil {
+					return nil, err
+				}
+				return nil, err
```
gogsbread, line 101: believe I asked this question before in the other PR as well, but not sure if it has been answered. Wouldn't the `err` here be `nil` since the body closure was fine, but we still it to have a error since we consider all 5xx to potentially break the circuit?

fishy: If we don't add a middleware with [`httpbp.ClientErrorFromResponse`](https://pkg.go.dev/github.com/reddit/baseplate.go/httpbp#ClientErrorFromResponse) after this one, then yes go's stdlib http client would only return an error when there's network level error. But we should add such a middleware so this should still work (even without that, it still works for network errors).

httpbp/client_middlewares.go
konradreiche: @fishy I updated it based on your Go playground snippet. This is a lot shorter now.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,194 @@
...
+func Retries(config retry.Config) ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripper(func(req *http.Request) (resp *http.Response, err error) {
+			retrybp.Do(req.Context(), func() error {
+				resp, err = next.RoundTrip(req)
+				if err != nil {
+					return err
+				}
+				return nil
+			}, retry.Attempts(1))
```
fishy, line 116: we should only add `retry.Attemps(1)` if there's no other options specified in the config, like what we did in `thriftbp`: https://github.com/reddit/baseplate.go/blob/88dcf6c19330435fa8c6766f1b97c9c40adfc75a/thriftbp/client_middlewares.go#L110-L112

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+			return resp, nil
+		})
+	}
+}
+
+// LoadBalancer implements a round-robin load balancer.
+func LoadBalancer(config Config) ClientMiddleware {
+	attempts := 0
+	offset := uint64(0)
```
fishy, line 128: style nit: prefer:

```suggestion
	var (
		attempts int
		offset   uint64
	)
```

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,194 @@
...
+
+			remaining := attempts
+			for remaining > 0 {
+				i := atomic.AddUint64(&offset, 1)
+				hostIndex := i % uint64(len(hosts))
+				req.Host = hosts[hostIndex]
+				req.URL.Host = hosts[hostIndex]
+
+				resp, err := next.RoundTrip(req)
+				if err == gobreaker.ErrOpenState || err == gobreaker.ErrTooManyRequests {
```
fishy, line 141:
```suggestion
				if errors.Is(err, gobreaker.ErrOpenState) || errors.Is(err, gobreaker.ErrTooManyRequests) {
```

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,194 @@
...
+					remaining--
```
fishy, line 142: so the current logic is that, if the host picked by loadbalancer has circuit breaker open, we try the previous one instead, and if the first one still has loadbalancer open, we try it again and return the error (line 147). this means that in the case of circuit breaker is open, we actually put more pressure on the first host.

I think this logic needs more work, and since it looks like the requirement for load balancer is a quite unique requirement that's not really needed by most services, I'd suggest we just remove this from the pr, and y'all can just implement it in your code base instead.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,194 @@
...
+			return next.RoundTrip(req)
+		})
+	}
+}
+
+// MaxConcurrency is a middleware to ensure that there will be only a maximum
+// number of requests concurrently in-flight at any given time.
+func MaxConcurrency(config Config) ClientMiddleware {
+	maxConcurrency := int64(config.MaxConcurrency)
+	activeRequests := int64(0)
```
fishy, line 156: style nit: prefer:
```suggestion
	var activeRequests int64
```
(see https://github.com/uber-go/guide/blob/master/style.md#use-var-for-zero-value-structs)

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,194 @@
...
+				}
+			}
+			return next.RoundTrip(req)
+		})
+	}
+}
+
+// MaxConcurrency is a middleware to ensure that there will be only a maximum
+// number of requests concurrently in-flight at any given time.
+func MaxConcurrency(config Config) ClientMiddleware {
```
fishy, line 154: 🔕 I'm not really sure this middleware is useful, go's stdlib http client can handle concurrency pretty well.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,194 @@
...
+
+// MonitorClient is a HTTP client middleware that wraps HTTP requests in a
+// client span.
+func MonitorClient(config Config) ClientMiddleware {
+	slug := config.Slug
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripper(func(req *http.Request) (resp *http.Response, err error) {
+			span, ctx := opentracing.StartSpanFromContext(
+				req.Context(),
+				slug,
```
fishy, line 178: for client spans, we do expect the name to be in the format of `client.endpoint`.

since for http requests the "endpoint" part is often ambiguous, I'd suggest either:

1. hardcode an endpoint for everything, e.g. `slug + ".http"`

or

2. convert the uri to endpoint, e.g. `slug + "." + sanitize(req.URL.Path)` (`sanitize` should replace `/`, `.`, and a few others to `_` or `-`).

fishy (changes requested):

httpbp/client_middlewares.go
konradreiche: That's a great gotcha!

httpbp/client_middlewares.go
konradreiche: Agreed, it's very specific and depends on what the user wants to accomplish with it.

httpbp/client_middlewares.go
konradreiche: +1 on that but I think this is something we should coordinate with baseplate.py? https://github.com/reddit/baseplate.py/blob/48bcb0b099608aa8b8eecbf97cb929537d68753d/baseplate/clients/requests.py#L203

fishy: that's basically my proposal 1, but with `request` instead of `http` as the hardcoded endpoint name :)

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,196 @@
...
+func NewClient(config ClientConfig, middlewares ...ClientMiddleware) *http.Client {
+	transport := &http.Transport{
+		MaxConnsPerHost: config.MaxConnections,
+	}
+
+	defaults := []ClientMiddleware{
+		ClientErrorWrapper(),
+		MonitorClient(config),
+		Retries(),
+		MaxConcurrency(config),
```
fishy, line 40: since we cannot add `DrainAndClose` into a middleware, `MaxConcurrency` could be a footgun, so I think we should remove it from default default middlewares.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// ClientErrorWrapper applies ClientErrorFromResponse to the returned response
+// ensuring a HTTP status response outside the range [200, 400) is wrapped in
+// an error relieving users from the need to check the status response.
+func ClientErrorWrapper() ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripper(func(req *http.Request) (resp *http.Response, err error) {
+			resp, err = next.RoundTrip(req)
+			err = ClientErrorFromResponse(resp)
```
fishy, line 58: sorry. this means that we could return both non-nil resp and non-nil err, and users need to `defer httpbp.DrainAndClose(resp.Body)` even if err is non-nil, which can also be a footgun. for that reason, I think we should not make this a default middleware (but still provide it), and also document that this could return both non-nil resp and non-nil err, so if user opt to use it they need to handle the draining of the response body even if err is non-nil.

konradreiche: I added a comment. The behavior around drain and close could be good reasons to a bigger interface that actually wraps the HTTP requests instead of being applied on the transport layer which would allow us to take care of this for the user. At the same time, even with just the standard library the user needs to make sure to adhere to practices like closing the response body when using the standard `http` package.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// ClientErrorWrapper applies ClientErrorFromResponse to the returned response
+// ensuring a HTTP status response outside the range [200, 400) is wrapped in
+// an error relieving users from the need to check the status response.
+//
+// Adding this middleware means that both, non-nil response and non-nil error,
+// can be returned. If configured the caller needs to call
+// httpbp.DrainAndClose(resp.Body) to ensure the underlying TCP connection can
+// be re-used.
```
fishy, line 56: it's probably worth adding an example here, something like:

```
// Instead of doing this:
//
//     // bad example, DO NOT USE
//     resp, err := client.Do(req)
//     if err != nil {
//       // TODO: handle error
//       return err
//     }
//     defer httpbp.DrainAndClose(resp.Body)
//     // TODO: handle resp
//
// Do this:
//
//     resp, err := client.Do(req)
//     if resp != nil {
//       defer httpbp.DrainAndClose(resp.Body)
//     }
//     if err != nil {
//       // TODO: handle error
//       return err
//     }
//     // TODO: handle resp
```

httpbp/client_middlewares.go
fishy: having a middleware controlling the whole lifecycle of the response will likely be too much and anti-pattern. I don't think that will work :(

one alternative is that we just do `DrainAndClose(resp.Body)` if `ClientErrorFromResponse(resp)` returns non-nil, but that also means the caller would only know the status code, without knowing any body and other header in the response.

oh and also, this line should be:

```suggestion
			if err == nil {
				err = ClientErrorFromResponse(resp)
			}
```

httpbp/client_middlewares.go
fishy:
```suggestion
				slug + ".request",
```

httpbp/errors.go
```diff
@@ -64,6 +65,11 @@ const (
 	DefaultErrorTemplateName = "httpbp/error"
 )
 
+// ClientMiddleware errors are returned if a known error is occuring during the middleware processing.
```
fishy, line 68: grammar nit: remove "are"?

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,199 @@
...
+func ClientErrorWrapper() ClientMiddleware {
```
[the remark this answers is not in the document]
fishy: `io.LimitReader` doesn't really give you that feature, and I'd argue that it's really not that important to warrant all the effort to fork `io.LimitReader` to add that feature.

httpbp/client_middlewares.go
konradreiche: I'll add that. Should I remove `TODO: handle error` since returning the error is a way of delegating the error handler to the caller or do we keep it in there to remind people to handle errors eventually?

httpbp/errors.go
fishy: No it's required by client spans: https://pkg.go.dev/github.com/reddit/baseplate.go/tracing#SpanTypeOption.Type

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,199 @@
...
+type roundTripper func(req *http.Request) (*http.Response, error)
+
+func (f roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req)
+}
+
+// NewClient returns a standard HTTP client wrapped with the default
+// middlewares plus any additional client middlewares passed into this
+// function.
+func NewClient(config ClientConfig, middlewares ...ClientMiddleware) *http.Client {
```
kylelemons, line 31: I am unsure of this formulation of a `NewClient` API.  This makes it unnecessarily tied to ClientConfig, for example.

The "business end" of this PR is the notion of middleware.  Let that be its legacy:

```
func WrapTransport(transport http.Transport, middleware ...ClientMiddleware) http.Transport {
  if base == nil {
    transport = http.DefaultTransport
  }
  for _, mw := range middleware {
    transport = mw(transport)
  }
  return transport
}

func (cc ClientConfig) Client() *http.Client {
  return &http.Client{
    Transport: WrapTransport(
      &http.Transport{
        MaxConnsPerHost: cc.MaxConnections,
      },
      ClientMonitor(cc.Slug),
      ClientRetry(),
      ClientCircuitBreaker(cc.CircuitBreaker),
    ),
  }
}
```

This makes the middleware usable even in contexts that don't have a ClientConfig, and it makes the ClientConfig responsible for knowing how to translate itself into appropriate middleware.

httpbp/client_middlewares.go
kylelemons: Examples should be done in the _test.go with an Example function; this will show up in godoc but it will also be compiled to ensure it won't drift.

kylelemons: Why not
```
resp, err := client.Do(req)
defer httpbp.DrainAndClose(resp)
if err != nil {
  // TODO: handle error
}
```

This lets you check for both nil responses and bodies inside DrainAndClose.

httpbp/client_middlewares.go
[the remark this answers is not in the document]
kylelemons: Hmm, I might be missing something, but why does any reading need to happen in the middleware?  ClientErrorFromResponse only requires the header and the response code and doesn't touch the body, right?

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,199 @@
...
+		return roundTripper(func(req *http.Request) (*http.Response, error) {
+			host := req.URL.Hostname()
+
+			// new circuit breakers should rarely get allocated: the pool reduces
+			// garbage collector overhead
+			newBreaker := pool.Get()
+			breaker, loaded := breakers.LoadOrStore(host, newBreaker)
+			if loaded {
+				pool.Put(newBreaker)
+			}
```
kylelemons, line 94: This seems dangerous... I assume that in this case `newBreaker == breaker`, right?  If so, then it could get pulled out and used concurrently.

Did you mean `defer pool.Put` or something?

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,199 @@
...
+				if resp.StatusCode >= http.StatusInternalServerError {
+					// read & close to ensure underlying RoundTripper
+					// (http.Transport) is able to re-use the persistent TCP
+					// connection.
+					_, _ = io.ReadAll(resp.Body)
+					err = resp.Body.Close()
+					if err != nil {
+						return nil, err
+					}
+					return nil, err
```
kylelemons, line 111:
```suggestion
					_, _ = io.ReadAll(resp.Body)
					return nil, resp.Body.Close()
```

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,199 @@
...
+				return nil, err
+			}
+			return resp, nil
+		})
+	}
+}
+
+// MaxConcurrency is a middleware to ensure that there will be only a maximum
+// number of requests concurrently in-flight at any given time.
+func MaxConcurrency(config ClientConfig) ClientMiddleware {
```
kylelemons, line 157: As alluded to above, I wouldn't make these top-level functions take in a ClientConfig.  That's passing in far more than is necessary.  These should either be methods on ClientConfig, to make their coupling clear (i.e. so nobody will think they can or should use one without the other) or, even better, make these accept the relevant parameters so they are decoupled from ClientConfig.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,199 @@
...
+			}, retryOptions...)
+			if err != nil {
+				return nil, err
+			}
+			return resp, nil
+		})
+	}
+}
+
+// MaxConcurrency is a middleware to ensure that there will be only a maximum
```
kylelemons, line 155: To me this sounds like it has a semaphore that will stall new requests if there are too many in flight.  Returning an error is surprising to me.

I would either:
1. Make it explicit: say that ErrConcurrencyLimit is returned, or
2. Use a semaphore (ideally a `make(chan struct{}, count)`)

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,199 @@
...
+			span.SetTag("http.method", req.Method)
+			span.SetTag("http.url", req.URL)
+
+			defer func() {
+				span.FinishWithOptions(tracing.FinishOptions{
+					Ctx: req.Context(),
+					Err: err,
+				}.Convert())
+			}()
+			req = req.WithContext(ctx)
```
kylelemons, line 195: Overwriting locals is enough of a cognitive burden that I think it's worth just inlining this on the following line

httpbp/client_middlewares.go
konradreiche: That sounds good to me if you think we could have it in the reverse way where there is still a `NewClient(config Config)` but like you said have the middlewares not be coupled to the config struct but instead its primitives.

I like the idea to instantiate a `http.Client` from a config but I believe that is unlike any of our other APIs in baseplate.go

httpbp/client_middlewares.go
fishy: since we are going to handle the body in this middleware, that example in the comments is no longer needed.

kylelemons: You can put the "don't do" in the example, something like

```
if false {
  // Don't do this:
}

// Do this:
```

httpbp/client_middlewares.go
```diff
@@ -54,11 +69,43 @@ func NewClient(config ClientConfig, middlewares ...ClientMiddleware) *http.Clien
...
+//       return err
+//     }
+//     defer httpbp.DrainAndClose(resp.Body)
+//     // TODO: handle resp
+//
+// Do this:
+//
+//     resp, err := client.Do(req)
+//     if resp != nil {
+//       defer httpbp.DrainAndClose(resp.Body)
```
kylelemons, line 88: you now drain and close in the middleware, right?

httpbp/client_middlewares.go
```diff
@@ -54,11 +69,43 @@ func NewClient(config ClientConfig, middlewares ...ClientMiddleware) *http.Clien
...
 	return func(next http.RoundTripper) http.RoundTripper {
-		return roundTripper(func(req *http.Request) (resp *http.Response, err error) {
+		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
 			resp, err = next.RoundTrip(req)
-			err = ClientErrorFromResponse(resp)
+			if err != nil {
+				return resp, err
+			}
+
+			defer DrainAndClose(resp.Body)
```
kylelemons, line 104: 🔕 I am still unsure why this needs to be done here, and I think it deserves a comment here that explains the necessity.

httpbp/errors.go
```diff
@@ -65,7 +65,7 @@ const (
 	DefaultErrorTemplateName = "httpbp/error"
 )
 
-// ClientMiddleware errors are returned if a known error is occuring during the middleware processing.
+// ClientMiddleware errors returned if a known error is occuring during the middleware processing.
```
kylelemons, line 68: Hmm, this symbol name doesn't match the var?

Do you mean

```
// Well-known errors for middleware layers.
var (
  // ErrConcurrencyLimit is returned by Foo if there are too many requests in flight
  ErrConcurrencyLimt = ...
)
```

kylelemons (approved):

httpbp/client_middlewares.go
konradreiche: This is definitely wrong. Just tripped over it while writing tests, we only want to drain and close if `ClientErrorFromResponse` returns an error, in that instance it seems to be okay to close since the API user should handle the error first before attempting to read from the body.

httpbp/errors.go
konradreiche: Something I borrowed from the way GoDoc is provided for errors in `thriftbp` but worth adding a separate comment: https://github.com/reddit/baseplate.go/blob/master/thriftbp/errors.go#L47-L52

konradreiche:
@fishy @kylelemons Added tests and addressed the comments. The retry middleware might need some more thought on how to interact with retryable HTTP errors since those do not get translated to an error via `ClientErrorWrapper`.

Also liking less and less the thought of a default HTTP client with pre-configured middleware but have it rather be something that users pick by their own merits.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	// add middleware in reverse so the first in the list is the outermost
+	for i := len(middleware) - 1; i >= 0; i-- {
+		transport = middleware[i](transport)
+	}
+	return transport
+}
+
+// ClientErrorWrapper applies ClientErrorFromResponse to the returned response
+// ensuring a HTTP status response outside the range [200, 400) is wrapped in
```
fishy, line 74: grammar nit: s/a/an/

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,222 @@
...
+// an error relieving users from the need to check the status response.
+//
+// Adding this middleware means that both, non-nil response and non-nil error,
+// can be returned. If configured the caller needs to call
+// httpbp.DrainAndClose(resp.Body) to ensure the underlying TCP connection can
+// be re-used.
+func ClientErrorWrapper() ClientMiddleware {
+	const limit = 1024
```
fishy, line 82: we probably want to make this configurable.

and since making it configurable later would be a breaking change, we probably want to make this configurable now :/

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// ClientErrorWrapper applies ClientErrorFromResponse to the returned response
+// ensuring a HTTP status response outside the range [200, 400) is wrapped in
+// an error relieving users from the need to check the status response.
+//
+// Adding this middleware means that both, non-nil response and non-nil error,
+// can be returned. If configured the caller needs to call
+// httpbp.DrainAndClose(resp.Body) to ensure the underlying TCP connection can
+// be re-used.
```
fishy, line 80: I think this part of the doc needs to be updated to reflect the current implementation.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,222 @@
...
+			return resp, err
+		})
+	}
+}
+
+// CircuitBreaker is a middleware that prevents sending requests that are likely
+// to fail based on a configurable failure ratio based on total failures and
+// requests.
+func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
+	pool := &sync.Pool{
```
fishy, line 114: super nit: this doesn't really need to be a pointer.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,222 @@
...
+		New: func() interface{} {
+			return breakerbp.NewFailureRatioBreaker(config)
+		},
+	}
+	breakers := &sync.Map{}
```
fishy, line 119: style nit: this doesn't need to be a pointer either, and as a result prefer `var breakers sync.Map` (https://github.com/uber-go/guide/blob/master/style.md#use-var-for-zero-value-structs)

httpbp/client_middlewares.go
fishy: now since we can return both non-nil resp and err, this should be `return resp, err` instead.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,222 @@
...
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+			if len(retryOptions) == 0 {
+				retryOptions = []retry.Option{retry.Attempts(1)}
+			}
+
+			err = retrybp.Do(req.Context(), func() error {
+				// include ClientErrorWrapper to ensure retry is applied for
+				// some HTTP 5xx responses
+				resp, err = ClientErrorWrapper()(next).RoundTrip(req)
```
fishy, line 167: since we added both `ClientErrorWrapper` and `Retries` into default middlewares list, we don't want to do this here? (do `ClientErrorWrapper` twice is likely harmless, but also wasteful)

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,222 @@
...
+// roundTripperFunc adapts closures and functions to implement http.RoundTripper.
+type roundTripperFunc func(req *http.Request) (*http.Response, error)
+
+func (f roundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req)
+}
+
+// NewClient returns a standard HTTP client wrapped with the default
+// middleware plus any additional client middleware passed into this
+// function.
```
fishy, line 31: we should document what are the default middlewares here.

fishy (changes requested):
The remaining comments are all minor.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+}
+
+// Retries provides a retry middleware.
+func Retries(retryOptions ...retry.Option) ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+			if len(retryOptions) == 0 {
+				retryOptions = []retry.Option{retry.Attempts(1)}
+			}
```
fishy, line 162: Another thing about retry http requests I just realized this morning: you can't really retry requests with a body (e.g. you can only retry get requests).

The reason is that [`http.Request.Body`](https://pkg.go.dev/net/http#Request.Body) is an `io.ReadCloser`, so you can only read it once. On retries, the read will return EOF immediately, and as a result, those retried requests will *silently* be corrupt (using empty body instead of the original body intended).

I think we should document that, *and* force no retry inside this middleware if `req.Body` is non-nil (we probably want to do some tests to make sure that `req.Body != nil` is a reliable way to distinguish the situation).

fishy: actually never mind. I think that case is already handled because when redirect happens you still need to read the request body multiple times, so there's [`http.Request.GetBody`](https://pkg.go.dev/net/http#Request.GetBody).

we probably want to double check that this works, though.

httpbp/client_middlewares.go
konradreiche: TIL: it depends how you [pronounce](https://www.quora.com/A-http-or-an-http-which-is-the-correct-grammatical-usage) HTTP. I'll go with the majority `an HTTP` :smiley:

httpbp/client_middlewares.go
konradreiche: It is an annoyance to have it run twice but wouldn't it also mean the Retry middleware isn't really useful outside of this scope? I would argue it would be good to leave it in case a user wants to make use of just the retry middleware for their HTTP client.

httpbp/client_middlewares.go
konradreiche: Added a test to check if it reads the body twice in case of a POST request.

httpbp/client_middlewares.go
fishy: that's fine to me, but if we go that route:

1. We should document both in `Retries` and `ClientErrorWrapper` that `Retries` included `ClientErrorWrapper`.
2. We should remove `ClientErrorWrapper` from default middlewares.
3. The limit size arg we are gonna add to `ClientErrorWrapper` need to be added to `Retries` as well.

konradreiche: Ah true, definitely didn't realize we can remove `ClientErrorWrapper` from the default middleware in this case.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,225 @@
...
+
+	defaults := []ClientMiddleware{
+		MonitorClient(config.Slug),
+		Retries(
+			config.MaxErrorReadAhead,
+			retry.Attempts(1),
+			retrybp.Filters(
+				retrybp.NetworkErrorFilter,
+				retrybp.RetryableErrorFilter,
+			),
```
fishy, line 50: 1. `RetryableErrorFilter` should be the first in the filter chain, see [its documentation](https://pkg.go.dev/github.com/reddit/baseplate.go/retrybp#RetryableErrorFilter) for why
2. We shouldn't be adding a default, hardcoded filter chain. This should be part of `ClientConfig`, and if it's unset (default value), we should only add `retry.Attempts(1)` (so users can control per request retry behavior via context object).

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,225 @@
...
+// as transitive middleware through Retries.
+func NewClient(config ClientConfig, middleware ...ClientMiddleware) *http.Client {
+	transport := &http.Transport{
+		MaxConnsPerHost: config.MaxConnections,
+	}
+
+	defaults := []ClientMiddleware{
+		MonitorClient(config.Slug),
+		Retries(
+			config.MaxErrorReadAhead,
```
fishy, line 45: we need to check for zero value and fallback to default one before doing this.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,225 @@
...
+	retry "github.com/avast/retry-go"
+	"github.com/opentracing/opentracing-go"
+
+	"github.com/reddit/baseplate.go/breakerbp"
+	"github.com/reddit/baseplate.go/retrybp"
+	"github.com/reddit/baseplate.go/tracing"
+)
+
+const (
+	defaultMaxErrorReadAhead = 1024
```
fishy, line 19: We usually export our default constants, this way in the config struct we can point the users to the default value which would appear on pkg.go.dev.

httpbp/client_middlewares.go
fishy: ^ this comment is not addressed

fishy (changes requested):

httpbp/client_middlewares.go
konradreiche: For 2: do you think it's sufficient to have a if `retries > 1` then apply the filters? Or have the config take an `[]retry.Option` similar to how we do it for Thrift

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,240 @@
...
+		ClientErrorWrapper(config.MaxErrorReadAhead),
+	}
+	// apply retries only if retry options are set but replace
+	// ClientErrorWrapper since Retries already wraps ClientErrorWrapper
+	if len(config.RetryOptions) > 0 {
+		defaults[1] = Retries(
+			config.MaxErrorReadAhead,
+			config.RetryOptions...,
+		)
+	}
```
fishy, line 67: We should not be adding `ClientErrorWrapper` to default middlewares, and we should always add `Retries` to default middlewares, even if `config.RetryOptions` is not set.

Can you follow [`thriftbp`'s example](https://pkg.go.dev/github.com/reddit/baseplate.go/thriftbp#BaseplateDefaultClientMiddlewares)?

fishy (changes requested):
You also have a deadlock in your test: https://github.com/reddit/baseplate.go/pull/368/checks?check_run_id=2606164139 (the test timed out at 10min, so it must be a deadlock).

httpbp/client_middlewares.go
konradreiche: Apparently, round trippers are not supposed to pass a response if `err` is non-nil https://github.com/golang/go/blob/master/src/net/http/client.go#L256 :slightly_frowning_face:

konradreiche:
@fishy I believe the deadlock happened in the concurrency test I wrote where I don't close the response body and it runs out of connections.

The retry logic default is now applied similar to what is done in [`client_middlewares.go#L112-L114`].(https://github.com/reddit/baseplate.go/blob/master/thriftbp/client_middlewares.go#L112-L114)

The fact that a correct RoundTripper implementation should not return both, non-nil error and non-nil response, should be a problem:
```
2021/05/24 20:53:14 RoundTripper returned a response & error; ignoring response
```
The only place we do this is in `ClientErrorWrapper`. What was the reason again for returning the response in face of an error? :thinking:

fishy:
> @fishy I believe the deadlock happened in the concurrency test I wrote where I don't close the response body and it runs out of connections.
> 
> The retry logic default is now applied similar to what is done in [`client_middlewares.go#L112-L114`].(https://github.com/reddit/baseplate.go/blob/master/thriftbp/client_middlewares.go#L112-L114)
> 
> The fact that a correct RoundTripper implementation should not return both, non-nil error and non-nil response, should be a problem:
> 
> ```
> 2021/05/24 20:53:14 RoundTripper returned a response & error; ignoring response
> ```
> 
> The only place we do this is in `ClientErrorWrapper`. What was the reason again for returning the response in face of an error? 🤔

:sigh: In that case I guess our best option is to just drop the response (`return nil, err`).

The reason we wanted to preserve response is that we only read the first N bytes into the error, but not the headers, and the user might need some useful information from the headers.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,239 @@
...
+					var ce *ClientError
+					if !errors.As(err, &ce) {
+						return resp, err
+					}
+					body, e := io.ReadAll(io.LimitReader(resp.Body, int64(limit)))
+					if e != nil {
+						return resp, e
+					}
+					ce.AdditionalInfo = string(body)
+					return resp, ce
```
fishy, line 119: we need to change `resp` to nil on lines 112, 116, 119, and do

```go
if err != nil {
  return nil, err
}
return resp, nil
```

on line 122.

fishy (commented):
You also need to run `./scripts/bazel_cleanup.sh` and commit the change it made.

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,239 @@
...
+			if err != nil {
+				return nil, err
+			}
+			return resp, nil
+		})
+	}
+}
+
+// MaxConcurrency is a middleware to limit the number of concurrent in-flight
+// requests at any given time by returning an error if the maximum is reached.
```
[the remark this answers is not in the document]
konradreiche: I would add it to the default middlewares but I would have a hard time to figure out what a reasonable limit would be or are you saying: add it to the default middleware but do not apply if limit is not set (equivalent to have it be 0)?

httpbp/errors.go
```diff
@@ -64,6 +65,20 @@ const (
...
+	// ErrConcurrencyLimit is returned by the max concurrency middleware if
+	// there are too many requests in-flight.
+	ErrConcurrencyLimit = errors.New("hit concurrency limit")
+)
+
+// ClientConfig errors are returned if the configuration validation fails.
+var (
+	errConfigMissingSlug              = errors.New("slug cannot be empty")
+	errConfigInvalidMaxErrorReadAhead = errors.New("maxErrorReadAhead value needs to be positive")
+	errConfigInvalidMaxConnections    = errors.New("maxConnections value needs to be positive")
```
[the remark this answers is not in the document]
konradreiche: yes! I always get frustrated with APIs hiding their errors not allowing me to work with them :slightly_smiling_face:

fishy (approved):

httpbp/client_middlewares.go
```diff
@@ -0,0 +1,246 @@
...
+func MonitorClient(slug string) ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+			span, ctx := opentracing.StartSpanFromContext(
+				req.Context(),
+				slug+".request",
+				tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
+			)
+			span.SetTag("http.method", req.Method)
+			span.SetTag("http.url", req.URL)
```
pacejackson, line 235: are there any concerns about cardinality in this tag?

pacejackson (approved):

httpbp/client_middlewares.go
fishy: this is span tag, not metrics tag (unless someone adds this into the global allow-list). there's no cardinality concerns on span tags.

PR #370, opened 2021-04-21
thriftbp: Listen on localhost in unit tests

kylelemons:
On Mac OS, binding to a non-localhost address will ask the user for approval. Since unit tests are rebuilt each time, this approval is required each time.

After this change, `go test ./...` does not require any approvals on my Mac.

### DIFF
--- a/thriftbp/client_pool_test.go
+++ b/thriftbp/client_pool_test.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	addr = ":0"
+	addr = "localhost:0"
 )
 
 func TestSingleAddressGenerator(t *testing.T) {


thriftbp/client_pool_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	addr = ":0"
+	addr = "localhost:0"
```
fishy, line 18: 🔕 I wonder which is the more universally appropriate one between "localhost" and "127.0.0.1"

fishy (approved):
Please wait until all codeowners' approvals before merging.

thriftbp/client_pool_test.go
kylelemons: I went with localhost so that it's not assuming IPv4 or IPv6 availability.  (I've worked with both IPv4-only and IPv6-only servers in the past)

fishy: does "ipv6-only" means only have ipv6 for non-local traffic, or does that mean even 127.0.0.1 is not available?

I believe "localhost" is not something provided by the kernel, but the `/etc/hosts` file (or equivalent) instead, although in 99.99% cases it does exist (but I could be wrong)

konradreiche: This is an interesting paragraph in a [blog post](https://www.pixelstech.net/article/1538275121-Difference-between-localhost-and-127-0-0-1) about this which I would love to verify but I could not find any further mentions of this:

> It's very important to know that it will not go through network card when using localhost while it will when using 127.0.0.1, this means localhost will not be affected by network card configurations and firewall settings, all ports are open as well.

konradreiche (approved):

thriftbp/client_pool_test.go
kylelemons: > does "ipv6-only" mean only have ipv6 for non-local traffic, or does that mean even 127.0.0.1 is not available?

In this case ipv6-only meant that you could not call the `bind` syscall with an IPv4 (or even IPv4-mapped-IPv6) address.  The `lo` interface did not have 127.0.0.1 bound (at least, as far as the applications could see).

> interesting paragraph in a blog post

I am 99% sure that the blog post is mistaken.  localhost will be resolved to 127.0.0.1 or ::1 and then a route will be selected.  On a normal system it will select a route with `dev lo` which will cause it to go through the kernel's loopback interface (which, among other things, can have a giant MTU, like 65k or something, instead of 1500) and thus bypass the NIC.

> in 99.99% cases it does exist

Most linux-like systems will be horrifically broken without a localhost entry in /etc/hosts, and if memory serves it's one of the few things that is provided in the distroless docker images for this reason.  But yeah, there might be niche cases in which it won't work, but I hope they don't overlap with environments where someone will be running `go test`...

pacejackson (approved):

PR #372, opened 2021-05-06
thriftbp: Fix WrapBaseplateError

fishy:
I just realized that in the previous WrapBaseplateError implementation,
we unwrap the error passed in into baseplate error, and only keep the
unwrapped baseplate error, but drops everything originally wrapped
around it.

Store both the original error and unwrapped baseplate error into the
unexported error struct, so that its Unwrap can return the original
error, but Error can use the original baseplate error to construct for
better error message.


pacejackson (approved):

konradreiche (approved):
:+1:

PR #374, opened 2021-05-11
Add gRPC tracing & edge context support

konradreiche:
It might be worth taking a look how to integrate gRPC rather sooner than later. I put this code together for a proof of concept but could potentially be adopted in baseplate.go since the middleware structure appears to be very close to what we do in thriftbp.

For this change I think there are two questions worth discussing:

1.) gRPC headers map keys to a list of values, i.e. [here](https://github.com/reddit/baseplate.go/compare/master...konradreiche:grpcbp#diff-3ca33bbd31feadb9de782c346470b0cc99ce6ebcec524551c675980697e23085R63) we can see that it pulls out the first value. I am not aware how to make use of a list of those values. This could be ignored for now or handled differently.

```go
	if values, ok := md[HeaderTracingTrace]; ok {
		headers.TraceID = values[0]
	}
```

2.) How to do testing. The gRPC middleware repository uses a test protobuf file to generate Go files unlike what we do in `thriftbp` testing where it is using a bare Thrift server. We could a) look into attempting to do something similar, b) provide our own test protobuf file that we own or c) use the one in [go-grpc-ecosystem](https://github.com/grpc-ecosystem/go-grpc-middleware/).


grpcbp/headers.go
```diff
@@ -0,0 +1,56 @@
...
+	// If not present, defaults to false.
+	HeaderTracingSampled = "Sampled"
+	// Trace flags, a 64-bit integer encoded in decimal.
+	// If not present, defaults to null.
+	HeaderTracingFlags = "Flags"
+)
+
+// HeaderTracingSampledTrue is the header value to indicate that this trace
+// should be sampled.
+const HeaderTracingSampledTrue = "1"
```
fishy, line 34: 🔕 don't feel great that we now duplicate them in two packages, maybe we should move this part to a common package and imported by both `thriftbp` and `grpcbp`.

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,77 @@
...
+		return ctx
+	}
+	values, ok := md[HeaderEdgeRequest]
+	if !ok {
+		return ctx
+	}
+
+	ctx, err := impl.HeaderToContext(ctx, values[0])
+	if err != nil {
+		log.Error("Error while parsing EdgeRequestContext: " + err.Error())
```
fishy, line 48: (I know this is copied from the thriftbp code, that should be fixed as well,)

if we are not going to use a `log.Wrapper` here (which is debatable, as passing in a `log.Wrapper` all the way here just for this line might be considered overkill), then we should at least use zap logger properly:

```
log.C(ctx).Errorw("Error while parsing EdgeRequestContext", "err", err)
```

grpcbp/headers.go
```diff
@@ -0,0 +1,56 @@
+package grpcbp
+
+import (
+	"context"
+
+	"github.com/reddit/baseplate.go/ecinterface"
+	"google.golang.org/grpc/metadata"
```
fishy, line 7: nit: import group

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,77 @@
+package grpcbp
+
+import (
+	"context"
+
+	"github.com/reddit/baseplate.go/ecinterface"
+	"github.com/reddit/baseplate.go/log"
+	"github.com/reddit/baseplate.go/tracing"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
```
fishy, line 10: nit: import group

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,77 @@
...
+func StartSpanFromGRPCContext(ctx context.Context, name string) (context.Context, *tracing.Span) {
+	var headers tracing.Headers
+	var sampled bool
+
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return tracing.StartSpanFromHeaders(ctx, name, headers)
+	}
+
+	if values, ok := md[HeaderTracingTrace]; ok {
```
fishy, line 62: I _think_ the recommended way is to use `Get` over accessing the map directly? so this should be something like:
```
if values := md.Get(HeaderTracingTrace); len(values) > 0 {
```

looking at the doc of `MD`, it does mention that it will automatically convert uppercase keys into lowercase, so accessing the uppercase keys directly might result in always getting not ok here.

grpcbp/tracing.go
```diff
@@ -0,0 +1,40 @@
...
+const ()
+
+func CreateGRPCContextFromSpan(ctx context.Context, span *tracing.Span) context.Context {
+	ctx = metadata.AppendToOutgoingContext(ctx,
+		HeaderTracingTrace, span.TraceID(),
+		HeaderTracingSpan, span.ID(),
+		HeaderTracingFlags, strconv.FormatInt(span.Flags(), 10),
+	)
+	if span.ParentID() != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx,
```
fishy, line 20: style nit: put `ctx,` in its own line as well.

grpcbp/tracing.go
```diff
@@ -0,0 +1,40 @@
...
+	"strconv"
+
+	"github.com/reddit/baseplate.go/tracing"
+	"google.golang.org/grpc/metadata"
+)
+
+const ()
+
+func CreateGRPCContextFromSpan(ctx context.Context, span *tracing.Span) context.Context {
+	ctx = metadata.AppendToOutgoingContext(ctx,
```
fishy, line 14: style nit: put `ctx,` in its own line.

grpcbp/tracing.go
```diff
@@ -0,0 +1,40 @@
...
+		HeaderTracingSpan, span.ID(),
+		HeaderTracingFlags, strconv.FormatInt(span.Flags(), 10),
+	)
+	if span.ParentID() != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx,
+			HeaderTracingParent, span.ParentID(),
+		)
+	} else {
+		md, _ := metadata.FromIncomingContext(ctx)
+		delete(md, HeaderTracingParent)
```
fishy, line 25: same key canonicalization issue, that direct `delete` probably won't work.

looking at the doc, I think the way to unset a header is to explicitly set `""` as the value:
```
md.Set(HeaderTracingParent, "")
```

grpcbp/tracing.go
```diff
@@ -0,0 +1,40 @@
...
+		ctx = metadata.NewOutgoingContext(ctx, md)
+	}
+	if span.Sampled() {
+		ctx = metadata.AppendToOutgoingContext(
+			ctx,
+			HeaderTracingSampled, HeaderTracingSampledTrue,
+		)
+	} else {
+		md, _ := metadata.FromIncomingContext(ctx)
+		delete(md, HeaderTracingSampled)
```
fishy, line 35: same here

fishy (changes requested):

grpcbp/headers.go
konradreiche: Maybe a package called `transport`?

grpcbp/tracing.go
kylelemons: nah, having ctx or format strings on the first line is fine

konradreiche:
Applied the following changes:
- replace direct map access with `Get`
- replace `delete` with `Set(key, "")`
- moved _some_ of the thriftbp edge context related headers into a `transport` package

Any thoughts so far on:

> 2.) How to do testing. The gRPC middleware repository uses a test protobuf file to generate Go files unlike what we do in thriftbp testing where it is using a bare Thrift server. We could a) look into attempting to do something similar, b) provide our own test protobuf file that we own or c) use the one in go-grpc-ecosystem.

Otherwise I would probably prefer b) which seems to give us the most control over the source code and still keep the test code readable and easy to manipulate to our needs.

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+func InitializeEdgeContext(ctx context.Context, impl ecinterface.Interface) context.Context {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return ctx
+	}
+	values, ok := md[transport.HeaderEdgeRequest]
+	if !ok {
```
fishy, line 47:
```suggestion
	values := md.Get(transport.HeaderEdgeRequest)
	if len(values) == 0 {
```

grpcbp/tracing.go
```diff
@@ -0,0 +1,39 @@
+package grpcbp
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/reddit/baseplate.go/tracing"
+	"github.com/reddit/baseplate.go/transport"
+	"google.golang.org/grpc/metadata"
```
fishy, line 9: nit import groups.

grpcbp/tracing.go
fishy: well we didn't write an exception for that in the style guide so I'd prefer to either fix the style guide or stick with it.

transport/headers.go
```diff
@@ -0,0 +1,31 @@
+package transport
+
+// Edge request context propagation related headers, as defined in
+// https://pages.github.snooguts.net/reddit/baseplate.spec/component-apis/thrift#edge-request-context-propagation
```
fishy, line 4: we probably should mention in the comment that those headers are only used in thrift and grpc (for http we use different headers, usually with `x-` prefix).

fishy (changes requested):

grpcbp/server_middlewares.go
konradreiche: We should probably provide an abstraction when accessing the gRPC metadata that takes care of retrieving the first element and just returns a bool to check if it exists to make this code easier to read.

konradreiche:
Got around to finally address the client middleware test. Going to spend some time on the server middleware test and missing documentation for exported methods but this is ready for another review.

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,94 @@
...
+		return ctx
+	}
+	values := md.Get(transport.HeaderEdgeRequest)
+	if len(values) == 0 {
+		return ctx
+	}
+
+	ctx, err := impl.HeaderToContext(ctx, values[0])
+	if err != nil {
+		log.C(ctx).Errorw("Error while parsing EdgeRequestContext: " + err.Error())
```
fishy, line 60: this should be fixed in [`thriftbp`](https://github.com/reddit/baseplate.go/blob/21c209724c85ade8443509934da6543d5ad6744e/thriftbp/server_middlewares.go#L174) as well, but since we are using log directly (not using `log.Wrapper`), it should be:

```suggestion
		log.C(ctx).Errorw("Error while parsing EdgeRequestContext", "err", err)
```

instead.

grpcbp/server_middlewares.go
fishy: yes we should. this pattern is repeated so many times in the code.

fishy (changes requested):
Please group the imports [correctly](https://github.com/reddit/baseplate.go/blob/master/Style.md#import-groups), and resolve the CI issues (for the go one, you need to first do `go mod tidy && git add go.mod go.sum`, as we need to revert the changes done by gazelle to go.sum to [workaround a bug in gazelle](https://github.com/reddit/baseplate.go/blob/6e19f3fdbd6e35a75a72bea34dd17fbcd7363b4f/scripts/bazel_cleanup.sh#L9-L10)).

We should also refactor two functions out as they are repeated across the code:

```go
func methodSlug(method string) string { //this one probably do not need to be exported
  split := strnigs.Split(method, "/")
  return strings.ToLower(split[len(split)-1])
}

func GetFromMD(md metadata.MD, key string) (value string, ok bool) { // this one likely need to be exported
  if values := md.Get(key); len(values) > 0 {
    return values[0], true
  }
  return "", false
}
```

go.mod
```diff
@@ -59,3 +59,12 @@ require (
 	golang.org/x/tools v0.1.3-0.20210608163600-9ed039809d4c // indirect
 	honnef.co/go/tools v0.2.0 // indirect
 )
+
+require (
```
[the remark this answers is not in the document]
konradreiche: Apparently 1.17 attempts to group them into direct and indirect require blocks but I also notices that it sometimes fails to do so.

grpcbp/client_middlewares.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/reddit/baseplate.go/ecinterface"
+	"github.com/reddit/baseplate.go/tracing"
+	"google.golang.org/grpc"
+)
+
+type MonitorClientArgs struct {
+	ServiceSlug string
+}
+
+func MonitorInterceptor(args MonitorClientArgs) grpc.UnaryClientInterceptor {
```
kylelemons, line 17: MonitorInterceptorArgs?

grpcbp/client_middlewares.go
```diff
@@ -0,0 +1,65 @@
...
+			}.Convert())
+		}()
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+}
+
+func ForwardEdgeRequestContext(ecImpl ecinterface.Interface) grpc.UnaryClientInterceptor {
+	if ecImpl == nil {
+		ecImpl = ecinterface.Get()
+	}
```
kylelemons, line 51: This has bit me before.  If nil is not a valid `ecinterface.Interface` (which it isn't) then it should be an error to pass it as one.  There should be exactly one place (presumably baseplate.New) that applies things like default values, if any.

grpcbp/client_middlewares.go
kylelemons: I think "Monitor" is an incorrect name for this.  This does _tracing_ which is not monitoring.  Monitor, to me at least, implies metrics.

TracingInterceptorUnary?
Or, even better, TraceUnary?

grpcbp/client_middlewares.go
```diff
@@ -0,0 +1,65 @@
...
+			span.FinishWithOptions(tracing.FinishOptions{
+				Ctx: ctx,
+				Err: err,
+			}.Convert())
+		}()
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+}
+
+func ForwardEdgeRequestContext(ecImpl ecinterface.Interface) grpc.UnaryClientInterceptor {
```
kylelemons, line 48: ForwardEdgeContextUnary?

grpcbp/client_middlewares.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"strings"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/reddit/baseplate.go/ecinterface"
+	"github.com/reddit/baseplate.go/tracing"
+	"google.golang.org/grpc"
+)
+
```
kylelemons, line 12: What about streaming interceptors?

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/reddit/baseplate.go/mqsend"
+	"github.com/reddit/baseplate.go/tracing"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/test/bufconn"
+)
+
+func TestMonitorInterceptor(t *testing.T) {
+	// create test server
+	l, _ := initServer(t)
```
kylelemons, line 21: `init` to me implies initializing a global.  I prefer `setup` for "make and give me a"

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// create test client with monitoring interceptor
+	conn := initClient(t, l, grpc.WithUnaryInterceptor(
+		MonitorInterceptor(
+			MonitorClientArgs{
+				ServiceSlug: "test",
+			},
+		),
+	))
```
kylelemons, line 30: creating a test server almost universally implies the need for a test client, and in the few instances where it won't it's nbd to `_` it away.

So have `setupServer` return a client too.

In a past life, we returned a struct that had the client, server, and address in it, so that we could add additional things over time without affecting client code, and that worked really well.

grpcbp/client_middlewares_test.go
kylelemons: don't use single letter identifiers, especially ones that don't clearly evoke what they are supposed to stand for :)

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// instantiate gRPC client
+	ctx, mmq := initServerSpan(t)
+	client := pb.NewTestServiceClient(conn)
+
+	t.Run("span-success", func(t *testing.T) {
+		_, err := client.Ping(ctx, &pb.PingRequest{})
+		if err != nil {
```
kylelemons, line 38: `if _, err := ...; err != nil {` to make it clear that you don't need the err later

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+			t.Fatal(err)
```
kylelemons, line 39: at least give it a tiny bit of context, e.g. `Ping: %s`

(same below)

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+		// drain recorder to validate spans
+		msg, err := drainRecorder(mmq)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		var trace tracing.ZipkinSpan
+		err = json.Unmarshal(msg, &trace)
+		if err != nil {
```
kylelemons, line 50: `if err := ...; err != nil {`

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		t.Error("got success span, want: error span")
+	})
+}
+
+func initServer(t *testing.T, opts ...grpc.ServerOption) (*bufconn.Listener, *mockService) {
+	t.Helper()
+
+	// create in-memory connection to reproduce network behavior but without
+	// the need to clean up OS-level resources
```
kylelemons, line 105: 🔕 I don't understand the necessity of this.  Loopback connections are highly reliable and can catch very real bugs.  I strongly encourage using them.  Especially where buffering and timeouts are concerned, it seems far safer to use the OS primitives than trust that the logic in a test-only helper is accurate enough.

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+	return l, service
+}
+
+func initClient(t *testing.T, l *bufconn.Listener, opts ...grpc.DialOption) *grpc.ClientConn {
+	// configure required dial option and let test add any further options
+	bufDialer := func(context.Context, string) (net.Conn, error) {
+		return l.Dial()
+	}
+	opts = append([]grpc.DialOption{
+		grpc.WithContextDialer(bufDialer),
```
kylelemons, line 126: just `l.DialContext`?

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+		if err := s.Serve(l); err != nil {
+			t.Error(err)
+		}
+	}()
+	t.Cleanup(s.Stop)
+
+	return l, service
+}
+
+func initClient(t *testing.T, l *bufconn.Listener, opts ...grpc.DialOption) *grpc.ClientConn {
```
kylelemons, line 120: 🔕 fwiw, I use `lis` not `l` where I don't use the full word `listener`

grpcbp/client_middlewares_test.go
```diff
@@ -0,0 +1,174 @@
...
+	}, opts...)
+
+	// create connection to be used by gRPC client
+	conn, err := grpc.DialContext(
+		context.Background(),
+		"bufnet",
+		opts...,
+	)
+	if err != nil {
+		t.Fatal(err)
```
kylelemons, line 137: add context, and consider `t.Helper`ing this func

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/reddit/baseplate.go/transport"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+func InjectServerSpanInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (_ interface{}, err error) {
+		split := strings.Split(info.FullMethod, "/")
+		m := split[len(split)-1]
```
kylelemons, line 19: method

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,94 @@
...
+		ctx, span := StartSpanFromGRPCContext(ctx, m)
+
+		if md, ok := metadata.FromIncomingContext(ctx); ok {
+			if values := md.Get(transport.HeaderTracingTrace); len(values) > 0 {
+				span.SetTag(tracing.TagKeyPeerService, values[0])
+			}
+		}
```
kylelemons, line 26: metadata.MD is readable as its zero value, so you just need

`md, _ := metadata.FromIncomingContext(ctx)`

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+func InjectServerSpanInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (_ interface{}, err error) {
+		split := strings.Split(info.FullMethod, "/")
+		m := split[len(split)-1]
+		ctx, span := StartSpanFromGRPCContext(ctx, m)
+
+		if md, ok := metadata.FromIncomingContext(ctx); ok {
+			if values := md.Get(transport.HeaderTracingTrace); len(values) > 0 {
```
kylelemons, line 23: you might consider using `== 1` instead of `> 0`

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,94 @@
...
+func StartSpanFromGRPCContext(ctx context.Context, name string) (context.Context, *tracing.Span) {
+	var (
+		headers tracing.Headers
+		sampled bool
+	)
+
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return tracing.StartSpanFromHeaders(ctx, name, headers)
+	}
```
kylelemons, line 74: same as above, I don't think you need to care much about the `!ok` case (on the read side)

grpcbp/server_middlewares.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if values := md.Get(transport.HeaderTracingTrace); len(values) > 0 {
```
kylelemons, line 76: 🔕 consider `== 1` for these as well.  If someone provides two, it's probably a bug.

Ideally, you might even have a metric for malformed headers and a helper to retrieve them and do the counting.

grpcbp/server_middlewares_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"testing"
+	"time"
+
+	pb "github.com/grpc-ecosystem/go-grpc-middleware/testing/testproto"
+	"github.com/reddit/baseplate.go/ecinterface"
+	"github.com/reddit/baseplate.go/mqsend"
+	"github.com/reddit/baseplate.go/tracing"
+	"github.com/reddit/baseplate.go/transport"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
```
kylelemons, line 16: import groups

grpcbp/server_middlewares_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// set up recorder to validate span recording
+	mmq := initTracing(t)
+
+	t.Run("span-success", func(t *testing.T) {
+		// send request to server
+		_, err := client.Ping(context.Background(), &pb.PingRequest{})
+		if err != nil {
+			t.Fatal(err)
+		}
```
kylelemons, line 41: same here and below, use `if err := ...` when you don't need any of the values past the if chain

kylelemons (changes requested):

grpcbp/client_middlewares.go
konradreiche: I initially named it `TracingInterceptor`. Tracing is the better name but we use `Monitor[ed]` across the baseplate repository. Maybe we can keep it consistent and rename them all or is there a more subtle difference with other monitor middlewares?
I also okay calling this tracing right away since this is a new package.

I assume we would call it Unary to allow for a Stream implementation too?

grpcbp/server_middlewares.go
konradreiche: I think we need to handle the case when there is more than one, i.e. reporting somehow that those headers are used incorrectly?

grpcbp/client_middlewares.go
konradreiche: For me it's at least a questionable side effect. Should we just remove it?

grpcbp/server_middlewares.go
konradreiche: That would mean the method exits silently though without creating a span.

grpcbp/server_middlewares.go
konradreiche: I would say this should probably go to Sentry.

grpcbp/client_middlewares.go
konradreiche: I think that should be added too but I would prefer to focus on getting something end-to-end working first unless we already have a great use case for streaming endpoints.

grpcbp/client_middlewares_test.go
konradreiche: Does the context help though if the user needs to go and look at the test at this line anyway? I remember we had a similar discussion about this. I almost feel the test code is easier to read without all the string interpolations to give additional context.

Other than that I am happy to standardize is going forward. How would we do it for instance with:
```go
if err = json.Unmarshal(msg, &trace); err != nil {
			t.Fatal(err)
		}
```
Should it be `Unmarshal: %s` or `json.Unmarshal: %s`? :thinking:

grpcbp/client_middlewares_test.go
konradreiche: +1 on having end-to-end tests that test the real thing. Though I will say that in my experience I do see a good bunch of sporadic port not available errors on CIs when picking an available port via `:0`. They also make the tests a bit slower overall.

grpcbp/client_middlewares_test.go
konradreiche: `bufconn.Listener` does not provide that method. Speaks for your point to test the real thing :wink:

grpcbp/client_middlewares_test.go
fishy: I think when it's in a function without `t.Helper()` the line number alone should be enough. But we do need context if it's called inside a function that called `t.Helper()` (as that will remove the line numbers from within that function).

But I won't mind if we decided to force always adding the context to `t.Fatal`.

grpcbp/client_middlewares_test.go
fishy: failure to listen on `:0` means there's no port available at all on the machine, I think? that doesn't sound like something that should happen at all (if the machine exhausted all ports it's already in a very very bad state). are you sure it's listing on `:0` failed, not some hardcoded port?

they _might_ be slower, but that's negligible.

grpcbp/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package grpcbp provides Baseplate specific gRPC related helpers.
```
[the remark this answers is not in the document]
konradreiche: Good point, I need to extend on the documentation here, currently not having a `doc.go` will make `staticcheck` fail.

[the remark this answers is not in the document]
fishy: I disagree. It's not always obvious which file has the most exported API, always just do the package level documentation in a separated `doc.go` seems easier style to follow (and enforce if we want to enforce it).

if the package only has one non-test .go file I'm fine with writing the package doc there without a separated doc.go, but as long as it has more than one .go files I'd prefer to have a separated doc.go instead of trying to decide "which is the most important .go file to write package doc".

thriftbp/headers.go
```diff
@@ -6,40 +6,41 @@ import (
...
 
 	"github.com/reddit/baseplate.go/ecinterface"
+	"github.com/reddit/baseplate.go/transport"
 )
 
 // Edge request context propagation related headers, as defined in
 // https://pages.github.snooguts.net/reddit/baseplate.spec/component-apis/thrift#edge-request-context-propagation
 const (
-	HeaderEdgeRequest = "Edge-Request"
+	HeaderEdgeRequest = transport.HeaderEdgeRequest
```
fishy, line 15: please also mark them as deprecated if possible, that way staticcheck will complain if people tries to use them.

I'm not super sure if `// Deprecated: ...` on the whole block works or you have to put it on every constant. but you can verify that by trying to use them and see if staticcheck complains.

fishy (commented):
I filed https://github.com/bazelbuild/rules_go/issues/2969 for the bazel failure :sigh:

go.mod
```diff
@@ -4,26 +4,11 @@ go 1.17
...
-	github.com/joomcode/errorx v1.0.3
-	github.com/joomcode/redispipe v0.9.4
-	github.com/opentracing/opentracing-go v1.2.0
-	github.com/sony/gobreaker v0.4.1
-	go.uber.org/zap v1.15.0
-	gopkg.in/fsnotify.v1 v1.4.7
-	gopkg.in/yaml.v2 v2.3.0
-)
-
-require (
```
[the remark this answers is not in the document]
konradreiche: Here is a reference to how the likely intended this to work:
> Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies.

https://golang.org/doc/go1.17#go-command

We could sort it to list the indirect ones in a separate require block.

thriftbp/headers.go
konradreiche: Do we have an example where we used this in the past, I couldn't find anything searching through `git log -p`. I tried several versions but staticcheck would not complain that it is still being used within `thriftbp` or does it only apply if you import a deprecated type?

i.e. with
```go
// Deprecated: HeaderUserAgent related headers.
const HeaderUserAgent = "User-Agent"
```
it would result in
```
thriftbp/headers.go:46:1: comment on exported const HeaderUserAgent should be of the form "HeaderUserAgent ..." (ST1022)
```

fishy: you can have both:
```go
// HeaderUserAgent is the user-agent header.
//
// Deprecated: use transport.HeaderUserAgent instead
const HeaderUserAgent = "User-Agent"
```

or if that doesn't work, you can always add linter directive (see https://staticcheck.io/docs/configuration#maintenance-of-linter-directives):
```go
// Deprecated: HeaderUserAgent related headers.
//lint:ignore ST1022 this is deprecated
const HeaderUserAgent = "User-Agent"
```

fishy (commented):
I'm mostly happy with the current state. things need to be sorted before merging:

1. go.sum file that CI complains about
2. mark the thriftbp headers as deprecated (but if we can merge this in before 0.9.0 release we don't need to do that)

I'm willing to remove all bazel related files from this repo in this PR if we can't get the rules_go bug fixed (it looks like we are highly unlikely to get that fixed soon enough).

things that's nice to have but can wait until a followup PR, since this PR is already pretty big.
1. switch from bufconn to a loopback address for testing
2. streaming middlewares.

go.mod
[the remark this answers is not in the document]
kylelemons: I generally don't think people should be babysitting this file.  If the go tool spits it out, it's fine with me.

kylelemons (approved):

PR #375, opened 2021-05-21
Add yaml serialization tags to breaker config

marcoferrer:
This change allows introduces yaml tags to the fields of the circuit breaker config. Allowing config to be sourced from a common file and removing the need for a code change when making adjustments

### DIFF
--- a/breakerbp/failure_ratio.go
+++ b/breakerbp/failure_ratio.go
@@ -26,7 +26,7 @@ type FailureRatioBreaker struct {
 // Config represents the configuration for a FailureRatioBreaker.
 type Config struct {
 	// Minimum requests that need to be sent during a time period before the breaker is eligible to transition from closed to open.
-	MinRequestsToTrip int `yaml:"primary"`
+	MinRequestsToTrip int `yaml:"minRequestsToTrip"`
 
 	// Percentage of requests that need to fail during a time period for the breaker to transition from closed to open.
 	// Represented as a float in [0,1], where .05 means >=5% failures will trip the breaker.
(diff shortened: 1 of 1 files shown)


breakerbp/failure_ratio.go
```diff
@@ -26,32 +26,32 @@ type FailureRatioBreaker struct {
 // Config represents the configuration for a FailureRatioBreaker.
 type Config struct {
 	// Minimum requests that need to be sent during a time period before the breaker is eligible to transition from closed to open.
-	MinRequestsToTrip int
+	MinRequestsToTrip int `yaml:"primary"`
```
fishy, line 29: why is this "primary"?

fishy (changes requested):
Note that this is technically a breaking change, as if someone is already using it with yaml, they need to change the keys of their yaml config for their code to continue to work, so after approved we'll only merge this after we are done with all the non-breaking 0.8.x changes.

breakerbp/failure_ratio.go
konradreiche: Ideally this can be mapped to the actual field name, e.g.
```go
`yaml:"minRequestsToTrip"`
```

marcoferrer: Thats was a miss. I agree `yaml:"minRequestsToTrip"` is the correct value.
- [ ] Fix min requests tag

marcoferrer:
@fishy Would changing the keys to pascal case make this change backwards compatible with any existing serialization usages?

fishy:
@marcoferrer I have lost track of what "pascal case" is 😅 , but the only true backward compatible way is to use all lower case (e.g. `minrequeststotrip` for field `MinRequestsToTrip`), which is the same as don't specify yaml tag at all.

also I'm not saying we should not make breaking changes. I'm just saying we need to slightly delay the merging of approved breaking changes as we don't want to maintain a `0.8.x` branch :)

fishy (approved):

pacejackson (approved):

konradreiche (approved):

PR #376, opened 2021-05-21
httpbp&thriftbp: Tweak default list for retryable codes

fishy:
There's no guarantee that there's no side-effects in 500, 502, 504, and
507 errors. So be more conservative and don't mark them as retryable by
default.

Also tweak httpbp.ClientError.Retryable logic, to return 0 instead of -1
for codes not in the default list. This way users can implement their
own retrybp.Filter to mark certain codes not in the default list
retryable. On thriftbp the list is already customizable and also we
defer to the next filter instead of mark the error non-retryable when
it's not in the list, so this actually brings us feature parity (kinda)
between thriftbp and httpbp.

The current codes in default retryable list for both httpbp and thriftbp
are:

* 425 (too early)
* 429 (too many requests)
* 503 (service unavailable)

### DIFF
--- a/httpbp/errors.go
+++ b/httpbp/errors.go
@@ -672,8 +672,8 @@ func (ce ClientError) RetryAfterDuration() time.Duration {
 
 // Retryable implements retrybp.RetryableError.
 //
-// It returns no decision (0) if the status code is unknown (0),
-// true (1) on any of the following conditions and false (-1) otherwise:
+// It returns true (1) on any of the following conditions and no decision (0)
+// otherwise:
 //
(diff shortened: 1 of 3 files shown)


fishy:
This is a breaking change. Will only be merged after we are done with all non-breaking 0.8.x merges.

pacejackson (approved):

fishy:
sidenote: I _believe_ 503 should be used by proxies to mean that they can't talk to the actual server, so it should be safe to assume no side-effects, but please correct me if I'm wrong 😅

konradreiche (approved):

Issue #366, opened 2021-04-07
Enable strict yaml parsing


fishy:
In `baseplate.New`:

1. If the passed in extra config is nil, turn on strict mode on the parsing of `baseplate.Config`
2. If the passed in extra config is non-nil, allow user to pass in another boolean arg to indicate whether they want to enable strict parsing for their extra config. If they pass in true for it, they must add `baseplate.Config` into their config struct with `yaml:",inline"` tag so it could pass strict mode.

PR #377, opened 2021-05-21
baseplate: Implement strict yaml parsing

fishy:
Closes https://github.com/reddit/baseplate.go/issues/366.

Always enable strict yaml parsing, and require users to embed
baseplate.Config with yaml tag of `yaml:",inline"` in their custom
config structs.

Also unexport parseConfig.

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -20,7 +20,19 @@ import (
 	"github.com/reddit/baseplate.go/tracing"
 )
 
-// Config is a general purpose config for assembling a Baseplate server
+// Configer defines the interface that allows you to extend Config with your own
+// configurations.
+type Configer interface {
+	GetConfig() Config
+}
+
+var (
+	_ Configer = Config{}
+)
+
+// Config is a general purpose config for assembling a Baseplate server.
+//
+// It implements Configer.
 type Config struct {
 	// Addr is the local address to run your server on.
 	//
@@ -46,11 +58,17 @@ type Config struct {
 	Tracing tracing.Config   `yaml:"tracing"`
 }
 
+// GetConfig implements Configer.
+func (c Config) GetConfig() Config {
+	return c
+}
+
 // Baseplate is the general purpose object that you build a Server on.
 type Baseplate interface {
 	io.Closer
 
-	Config() Config
+	Configer
+
 	EdgeContextImpl() ecinterface.Interface
 	Secrets() *secrets.Store
 }
@@ -130,7 +148,7 @@ func Serve(ctx context.Context, args ServeArgs) error {
 			//
 			// If one is set, we will only wait for that duration for the server to
 			// stop gracefully and will exit after the deadline is exceeded.
-			timeout := server.Baseplate().Config().StopTimeout
+			timeout := server.Baseplate().GetConfig().StopTimeout
 
 			// Default to 30 seconds if not set.
 			if timeout == 0 {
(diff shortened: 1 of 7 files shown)


baseplate.go
```diff
@@ -213,22 +213,48 @@ func ParseConfig(path string, serviceCfg interface{}) (Config, error) {
...
+// `yaml:",inline"` tag to pass strict mode. For example:
+//
+//     type myServiceConfig struct {
+//       // Not really used, but must present and must be exported to pass
+//       // strict mode.
+//       Unused baseplate.Config `yaml:",inline"`
+//
+//       // Actual configs
+//       // ...
+//     }
```
pacejackson, line 240: This feels really... unintuitive and I'm not really a fan of having to add extra cruft like this just to allow strict yaml parsing.

What if instead, rather than just passing in an `interface{}` for the service config, we instead have an interface:

```golang
type WithBaseplateConfig interface {
 func BaseplateConfig() Config
}
```

`baseplate.Config` could even implement this itself and then you would just have to embed it and I think it would work fine? Then we can do a single pass to get the entire config and enable strict parsing in a way that I think is clearer?

pacejackson (changes requested):

baseplate.go
fishy: I don't think that will work (if you can make it work, please provide a working example :) )

How yaml parsing works is to use `reflect` to get the fields of the struct, if the field is defined as an interface (like the case of embedding), it won't work because yaml parsing cannot get its concrete type there so it cannot fill its data.

I know this is not ideal, that's why this is opt-in (user need to explicitly set the boolean field to true) rather than opt-out. As far as I know this is the only way we could enable strict parsing and strict parsing is valuable for users who want it, so it's worth providing this not ideal opt-in option for them.

pacejackson: This works:

```go
type Config struct {
	Addr string
}

func (c Config) GetConfig() Config {
	return c
}

type Configer interface {
	GetConfig() Config
}

type config struct {
	Config `yaml:",inline"`

	Message string
}

func parseConfig(cfg Configer) error {
	f, err := os.Open("config.yaml")
	if err != nil {
		return err
	}
	defer f.Close()

	decoder := yaml.NewDecoder(f)
	decoder.SetStrict(true)
	if err = decoder.Decode(cfg); err != nil {
		return err
	}
	return nil
}
```

fishy: oh you are embedding `Config` the struct, not `Configer` the interface.

so this is essentially the same as just remove `Unused ` part from my example (which is more or less the same as just renaming that field from `Unused` to `Config`). I can make that change, but I don't think the interface is necessary/useful?

fishy: oh I guess the point of it is to force users to embed `Config` so we just enable strict mode without an option? the problem of that is embedding the field is only half of it, the other important half is `yaml:",inline"` tag which can still be a big footgun.

pacejackson: So there's another part which I think was missed, we just have everyone do this and update the cookiecutter to support this. Then we're also avoiding the dual pass (because we just do the whole config in one go and get the `baseplate.Config` from there rather than creating it from scratch and parsing it) and you don't need to do anything special to enable strict parsing.

pacejackson: @konradreiche @kylelemons thoughts?

fishy: cookiecutter can help new projects, yes. but the transition from old projects can still cause a lot of footguns 😬 

but I think this is an acceptable compromise.

pacejackson: I think that's probably fine?

konradreiche: In general it would be ideal to not hide some of the internal functionality in the code documentation, e.g. a user should not necessarily need to read the code documentation to use the API correctly. While it will always be required to a degree it would be desirable to avoid this.

For this concrete example, is this something that be improved by providing a second yaml decoding function?

konradreiche (changes requested):

baseplate.go
fishy: @konradreiche that won't work, unless we split the yaml config content into 2 files (one for baseplate and one for service-custom)

konradreiche: Doesn't that sound more appropriate? I am thinking this would create more flexibility. It might seem redundant here in the library code but from the user perspective it would help to just use what they specifically need for their use case rather than piggybacking it through the same function? :thinking:

fishy: That's really up to debate.

Yes that removes a footgun, but on the other side now you need to maintain two config files for your deploys (multiply that by 2 when you need to maintain 2 versions, e.g. prod + local), that can get nightmare-ish quickly.

With cookiecutter update, the footgun is mostly limited to existing services upgrading (new service will have the correct config by default, and they need to go out their way to break it), which is mostly acceptable.

fishy:
:haircut: @pacejackson

pacejackson (approved):
lgtm, only question I have is whether we want to make strict parsing optional (I'm fine with making it mandatory FWIW)

fishy:
> lgtm, only question I have is whether we want to make strict parsing optional (I'm fine with making it mandatory FWIW)

with the new approach I would really prefer to make it mandatory. strict mode eliminates a lot of silent typo issues in the config.

pacejackson:
I like it too, just wanted to raise the discussion point. 😄

baseplate.go
```diff
@@ -20,7 +20,19 @@ import (
 	"github.com/reddit/baseplate.go/tracing"
 )
 
-// Config is a general purpose config for assembling a Baseplate server
+// Configer defines the interface that allows you to extend Config with your own
+// configurations.
+type Configer interface {
+	GetConfig() Config
```
konradreiche, line 26: :no_bell: just
```go
Config() Config
```
?

konradreiche (approved):

baseplate.go
fishy: this won't work for the structs embedding it (because the field is also named `Config`, because embedding)

PR #379, opened 2021-06-07
baseplate: Add Drainer

fishy:
Inspired by the Baseplate.py change[1], add Drainer implementation on
top-level package so services can use it to fail their READINESS checks.

Also change kafkabp.Consumer.IsHealthy function signature to add context
arg, so it implements the newly added baseplate.Healthyer interface. The
context arg is used by neither of them right now, but we might want to
start using them in other implementations in the future so adding the
arg now can avoid a future breaking change.

[1]: https://github.com/reddit/baseplate.py/commit/c4a815d2729c3e232849b1e8079d3fb698f621ef

### DIFF
--- a/drainer.go
+++ b/drainer.go
@@ -1,0 +1,49 @@
+package baseplate
+
+import (
+	"context"
+	"io"
+	"sync/atomic"
+)
+
+// HealthChecker defines an interface to report healthy status.
+type HealthChecker interface {
+	IsHealthy(ctx context.Context) bool
+}
+
+// HealthCheckCloser is the combination of Healthyer and io.Closer.
+type HealthCheckCloser interface {
+	HealthChecker
+	io.Closer
+}
+
+type drainer struct {
+	closed int64
+}
+
+func (d *drainer) IsHealthy(_ context.Context) bool {
+	return atomic.LoadInt64(&d.closed) == 0
+}
+
+func (d *drainer) Close() error {
+	atomic.StoreInt64(&d.closed, 1)
+	return nil
+}
+
+// Drainer creates a HealthCheckCloser implementation that can be used to drain
+// service during graceful shutdown.
+//
+// The HealthCheckCloser returned would start to report healthy once created,
+// and start to report unhealthy as soon as its Close is called.
+//
+// Please refer to the example on how to use this feature in your service.
+// Basically you should create a Drainer in your main function,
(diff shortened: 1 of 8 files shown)


fishy:
cc @spladug

spladug (approved):

drainer.go
```diff
@@ -0,0 +1,49 @@
+package baseplate
+
+import (
+	"context"
+	"io"
+	"sync/atomic"
+)
+
+// Healthyer defines an interface to report healthy status.
+type Healthyer interface {
```
konradreiche, line 10: Since this looks a little bit like a misspelling of healthier I wonder if we should have a different convention for interface names when the method is bound to a method of the form `Is[adjective]` especially when ending on -y to something like `type Health interface` instead.

konradreiche (approved):

drainer.go
pacejackson: What about calling this `HealthChecker`?

kafkabp/consumer.go
```diff
@@ -43,13 +43,15 @@ type ConsumeMessageFunc func(ctx context.Context, msg *sarama.ConsumerMessage)
...
+//
+// It's also a superset of (implements) baseplate.Healthyer.
 type Consumer interface {
 	io.Closer
 
 	Consume(ConsumeMessageFunc, ConsumeErrorFunc) error
 
 	// IsHealthy returns false after Consume returns.
-	IsHealthy() bool
+	IsHealthy(ctx context.Context) bool
```
pacejackson, line 54: Should this just embed the `Healthyer` interface instead?

pacejackson (approved):

drainer.go
fishy: between the 2 I certainly like `HealthChecker` more

fishy: Renamed to `HealthChecker` and `HealthCheckCloser`.

kafkabp/consumer.go
fishy: There are two reasons:

1. Avoid the dependency
2. Since all the consumer implementations are unexported, this is where we document the behavior of the `IsHealthy` function, so changing this to embedding will make the document less useful, as people would need to jump to that definition to see how it actually looks like (it's not a problem for `io.Closer` as that one is "well-known").

pacejackson (approved):

konradreiche (approved):

PR #383, opened 2021-07-01
Enable staticcheck in CI

fishy:
And also fix all the things staticcheck complained about.

While I'm here, also upgrade the go version from 1.16.2 to 1.16.5.


pacejackson (approved):

konradreiche (approved):
:tada:

PR #385, opened 2021-07-02
Update baseplate.go to meet upcoming requirements for environment substitution in configs

kylelemons:
# API CHANGE

This commit introduces a breaking API change:
* baseplate.ParseConfigYAML(file, ptr) -> baseplate.ParseConfigYAML(ptr)

    According to the new baseplate spec, the config file location should be
    determined based on the environment variable BASEPLATE_CONFIG_PATH

There is also a new piece of functionality:
* YAML configuration files interpolate $ENV and ${ENV} from the environment

    This is also a new addition to the baseplate spec.

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -232,7 +232,8 @@ func Serve(ctx context.Context, args ServeArgs) error {
 //       log.Fatalf("Parsing config: %s", err)
 //     }
 //     ctx, bp, err := baseplate.New(baseplate.NewArgs{
-//       ServiceCfg: cfg,
+//       EdgeContextFactory: edgecontext.Factory(...),
+//       ServiceCfg:         cfg,
 //     })
 //
 // If you do have customized configurations to decode from YAML,
@@ -250,7 +251,8 @@ func Serve(ctx context.Context, args ServeArgs) error {
 //       log.Fatalf("Parsing config: %s", err)
 //     }
 //     ctx, bp, err := baseplate.New(baseplate.NewArgs{
-//       ServiceCfg: cfg,
+//       EdgeContextFactory: edgecontext.Factory(...),
+//       ServiceCfg:         cfg,
 //     })
 //
 // Environment variable references (e.g. $FOO and ${FOO}) are substituted into the
--- a/baseplate.go
+++ b/baseplate.go
@@ -2,15 +2,13 @@ package baseplate
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"os"
 	"time"
 
-	"gopkg.in/yaml.v2"
-
 	"github.com/reddit/baseplate.go/batchcloser"
+	"github.com/reddit/baseplate.go/configbp"
 	"github.com/reddit/baseplate.go/ecinterface"
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/metricsbp"
@@ -217,26 +215,28 @@ func Serve(ctx context.Context, args ServeArgs) error {
 	return <-shutdownChannel
 }
 
-// ParseConfigYAML parses path as YAML into cfg.
+// ParseConfigYAML loads the baseplate config into the structed pointed to by cfgPointer.
 //
-// It enables strict parsing,
-// if there are yaml tags not defined in cfg passed in,
-// that will cause an error.
+// The configuration file is located based on the $BASEPLATE_CONFIG_PATH
+// environment variable.
+//
+// To avoid easy mistakes, "strict" mode is enabled while parsing the file,
+// which means that all values in the YAML config must have a matching
+// struct or value during decoding.
 //
 // If you don't have any customized configurations to decode from YAML,
 // you can just pass in a *pointer* to baseplate.Config:
 //
 //     var cfg baseplate.Config
-//     err := baseplate.ParseConfigYAML(reader, &cfg)
-//     // TODO: handle err
+//     if err := baseplate.ParseConfigYAML(&cfg); err != nil {
+//       log.Fatalf("Parsing config: %s", err)
+//     }
 //     ctx, bp, err := baseplate.New(baseplate.NewArgs{
-//       EdgeContextFactory: ...,
 //       ServiceCfg: cfg,
 //     })
 //
 // If you do have customized configurations to decode from YAML,
-// the recommended way of passing the strict yaml parsing is to embed
-// baseplate.Config with `yaml:",inline"` yaml tags, for example:
+// embed a baseplate.Config with `yaml:",inline"` yaml tags, for example:
 //
 //     type myServiceConfig struct {
 //       // The yaml tag is required to pass strict parsing.
@@ -246,30 +246,20 @@ func Serve(ctx context.Context, args ServeArgs) error {
 //       FancyName string `yaml:"fancy_name"`
 //     }
 //     var cfg myServiceCfg
-//     err := baseplate.ParseConfigYAML(reader, &cfg)
-//     // TODO: handle err
+//     if err := baseplate.ParseConfigYAML(&cfg); err != nil {
+//       log.Fatalf("Parsing config: %s", err)
+//     }
 //     ctx, bp, err := baseplate.New(baseplate.NewArgs{
-//       EdgeContextFactory: ...,
 //       ServiceCfg: cfg,
 //     })
 //
-func ParseConfigYAML(path string, cfg Configer) error {
-	if path == "" {
-		return errors.New("baseplate.ParseConfigYAML: no config path given")
+// Environment variable references (e.g. $FOO and ${FOO}) are substituted into the
+// YAML from the process-level environment before parsing the configuration.
+func ParseConfigYAML(cfgPointer Configer) error {
+	if configbp.BaseplateConfigPath == "" {
+		return fmt.Errorf("no $BASEPLATE_CONFIG_PATH specified, cannot load config")
 	}
-
-	f, err := os.Open(path)
-	if err != nil {
-		return fmt.Errorf("baseplate.ParseConfigYAML: %w", err)
-	}
-	defer f.Close()
-
-	decoder := yaml.NewDecoder(f)
-	decoder.SetStrict(true)
-	if err := decoder.Decode(cfg); err != nil {
-		return fmt.Errorf("baseplate.ParseConfigYAML: %w", err)
-	}
-	return nil
+	return configbp.ParseStrictFile(configbp.BaseplateConfigPath, cfgPointer)
 }
 
 // NewArgs defines the args used in New functino.
--- a/configbp/config.go
+++ b/configbp/config.go
@@ -13,6 +13,7 @@ import (
 	"go.uber.org/zap"
 	"gopkg.in/yaml.v2"
 
+	"github.com/reddit/baseplate.go/internal/limitopen"
 	"github.com/reddit/baseplate.go/log"
 )
 
@@ -21,7 +22,7 @@ var BaseplateConfigPath = os.Getenv("BASEPLATE_CONFIG_PATH")
 
 type envsubstReader struct {
 	buffer bytes.Buffer
-	lines  bufio.Scanner
+	lines  *bufio.Scanner
 }
 
 func (r *envsubstReader) Read(buf []byte) (int, error) {
@@ -47,7 +48,7 @@ func (r *envsubstReader) Read(buf []byte) (int, error) {
 // Environment variables (e.g. $FOO and ${FOO}) are substituted from the environment before parsing.
 // The configuration is parsed into each of the targets, which will typically be pointers to structs.
 func ParseStrictFile(path string, ptr interface{}) error {
-	f, err := os.Open(path)
+	f, _, err := limitopen.Open(path)
 	if err != nil {
 		return err // contains filename
 	}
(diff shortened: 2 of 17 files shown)


configbp/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/reddit/baseplate.go/log"
+	"gopkg.in/yaml.v2"
```
pacejackson, line 12: style nit, this should be:

```suggestion
	"gopkg.in/yaml.v2"

	"github.com/reddit/baseplate.go/log"
```

configbp/config.go
```diff
@@ -0,0 +1,67 @@
...
+	rawData := new(strings.Builder)
+	if _, err := io.Copy(rawData, reader); err != nil {
+		return fmt.Errorf("reading config: %w", err)
+	}
+
+	data := ExpandConfig(rawData.String())
+
+	for _, ptr := range ptrs {
+		dec := yaml.NewDecoder(strings.NewReader(data))
+		dec.SetStrict(true)
```
pacejackson, line 59: 🔕  Doesn't strict parsing kind of defeat the purpose of passing in multiple `ptrs`? Each one would have to support every field in the YAML IIRC.

pacejackson (approved):

configbp/config.go
kylelemons: ah, right... I made this CL initially before the strict parsing was added and missed that I'm no longer using the varargs. thanks for the call-out

configbp/config.go
kylelemons: Yep.  In reconciling the editor config I might've accidentally nuked my settings ><

pacejackson (approved):

configbp/config.go
```diff
@@ -0,0 +1,69 @@
...
+//   - YAML (i.e. ".yaml") files are decoded according to ParseYAML.
+func ParseFile(path string, ptr interface{}) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return err // contains filename
+	}
+	defer f.Close() // safe to blindly close read-only files
+
+	switch ext := filepath.Ext(path); ext {
+	case ".yaml":
```
fishy, line 42: what if the file is ending in `.yml`, which I believe is also widely used for yaml files?

since we only support one format that is yaml, I don't really think this check is really needed.

configbp/config.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+// ParseYAML parses YAML read from the given Reader with strict decoding enabled.
+//
+// Environment variables (e.g. $FOO and ${FOO}) are substituted from the environment before parsing.
+// The configuration is parsed into the target, which will typically be a pointer to a struct.
+func ParseYAML(reader io.Reader, ptr interface{}) error {
+	rawData := new(strings.Builder)
+	if _, err := io.Copy(rawData, reader); err != nil {
```
fishy, line 55: at the very least, please use `io.LimitReader` with a sane limit to wrap `reader`.

I think it's doable to fork `os.Expand` to wrap an `io.Reader` instead of converting strings, but that's probably too hairy and not really worth doing.

configbp/config.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/reddit/baseplate.go/log"
+)
+
+// ExpandConfig performs baseplate-conformant configuration expansion.
+//
+// Currently, this includes expanding $FOO and ${FOO} with their respective environment variables.
+func ExpandConfig(rawConfig string) string {
+	expanded := os.ExpandEnv(rawConfig)
+	if expanded != rawConfig {
+		log.Debugf("Expanded configuration:\n%s", expanded)
```
fishy, line 22: 🔕 in general I'm against using direct log functions in a library code, but consider the use case of this library this is probably ok.

fishy (changes requested):

configbp/config.go
kylelemons: This helps avoid accidents like the one you described in the baseplate.spec conversation :)

also, added .yml and downcasing.

configbp/config.go
kylelemons: This kind of thing has led to things failing in production before, as things cross arbitrary thresholds.  I'd rather not.

configbp/config.go
kylelemons: I'd kinda like to get my ctxlog package into baseplate at some point, and that could potentially let us do it a bit more cleanly, but yeah.  Not my favorite but these kinds of things have been life-savers in the past.

configbp/config.go
fishy: there are a lot of ways to handle that:

1. Define it as an exported global variable instead of a const, so if a service know that they are using a huge config file, they can change the value of the limit in their main function, before calling `baseplate.New`.
2. Implement `io.LimitReader` by ourselves, using `bufio.Reader`, and return a different error than eof when we hit the limit and there are still more data in the original reader. This way if the yaml parser really reads beyond the limit it will fail, and that failure will be output to the log.
3. The line-buffer idea you talked about is also worth considering, but I don't believe there's a limit to the line either, so it might still oom when reading from for example `/dev/zero`.

So you probably need 1+2 or 3, but really, don't just copy the whole file into memory.

fishy (changes requested):

configbp/config.go
fishy: 2: https://github.com/reddit/baseplate.go/pull/388

eaceaser:
@fishy just paging this stuff in. Can you elaborate a bit on the use case here? i.e. is it only for FileWatching?

configbp/config.go
konradreiche: We are using ` fmt.Fprintln(os.Stderr` in some other [places](https://github.com/reddit/baseplate.go/blob/master/runtimebp/config.go#L40) though I find debug level preferable as it is easier to suppress noisy messages when you are aware of them.

konradreiche (approved):

configbp/config.go
fishy: @konradreiche there are two reasons we used `fmt.Fprintln(os.Stderr)` there:

1. `runtimebp` is a standalone library, it does not need to be tied with our logging library (people can choose to use it without using any other baseplate.go libraries)
2. at the time `log.Debug/Info/Error/etc.` goes nowhere before we initialized them (that was fixed in efa21f216085c6239e587f65563def72fc51bd88)

.editorconfig
```diff
@@ -1,13 +0,0 @@
-# https://EditorConfig.org
```
fishy, line 1: why remove the editorconfig file?

.gitignore
```diff
@@ -6,3 +6,7 @@
 
 # Binaries
 /cmd/healthcheck/healthcheck
+
+# Config files
+.idea
+.editorconfig
```
fishy, line 12: the point of `.editorconfig` is to sync it with every developer so the code has a consistent tabstop (it's also supported by github so when you view the files on github it picks tabstop from `.editorconfig`)

fishy (changes requested):
Now that f16633383d is merged, please rebase on top of latest master branch and use `limitopen` to open the config file and check for size limit.

.gitignore
kylelemons: Developers should be free to configure their editors however is comfortable for them.
Ensuring that basic standards are maintained in the codebase should be left up to linters.

You could probably convince me to leave the bazel ones in there but I would rather defer to buildifier, which is enforceable, rather than .editorconfig.

fishy: for go code, even if tab is enforced, the size of the tab actually affects where do people start to wrap the long lines.

fishy: no one uses `.editorconfig` to config their local editor, if they want to do that they will do that in the editor instead (vimrc, vscode config, etc.)

the whole point of `.editoconfig` is to sync the config across the project.

baseplate.go
```diff
@@ -246,30 +246,20 @@ func Serve(ctx context.Context, args ServeArgs) error {
 //       FancyName string `yaml:"fancy_name"`
 //     }
 //     var cfg myServiceCfg
-//     err := baseplate.ParseConfigYAML(reader, &cfg)
-//     // TODO: handle err
+//     if err := baseplate.ParseConfigYAML(&cfg); err != nil {
+//       log.Fatalf("Parsing config: %s", err)
+//     }
 //     ctx, bp, err := baseplate.New(baseplate.NewArgs{
-//       EdgeContextFactory: ...,
```
fishy, line 252: this is still required so we probably should keep this line?

kylelemons:
Since this is now targeting 0.9.x it might be worth a breaking API change, so I've included that here.

PTAL

configbp/config.go
kylelemons: I have introduced a line-based environment substitutor.

configbp/config.go
kylelemons: I've changed around how the logging is done -- please take a look and let me know what you think

.gitignore
kylelemons: Reverted.

configbp/config.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+// ParseStrictYAML parses YAML read from the given Reader.
+//
+// Environment variables (e.g. $FOO and ${FOO}) are substituted from the environment before parsing.
+// The configuration is parsed into each of the targets, which will typically be pointers to structs.
+func ParseStrictYAML(reader io.Reader, ptr interface{}) error {
+	reader = &envsubstReader{
+		lines: *bufio.NewScanner(reader),
```
fishy, line 70: 🔕 I would slightly prefer just to define `envsubstReader.lines` with type `*bufio.Scanner` so we don't have to do the pointer dereference here, but I don't feel strongly.

configbp/config.go
```diff
@@ -0,0 +1,94 @@
...
+	// Return some data to satisfy the reader
+	return r.buffer.Read(buf)
+}
+
+// ParseStrictFile parses configuration from the file at the given path.
+//
+// Environment variables (e.g. $FOO and ${FOO}) are substituted from the environment before parsing.
+// The configuration is parsed into each of the targets, which will typically be pointers to structs.
+func ParseStrictFile(path string, ptr interface{}) error {
+	f, err := os.Open(path)
```
fishy, line 50: I think we still want to use `limitopen.Open` instead of `os.Open` here. Although you implemented the line reader (nice!), if the file is mapped to `/dev/zero` we could still have a single line with unlimited length that blows out memory.

Between checking for size and checking for extension (we can do both, sure, but that's overkill imo) I still slightly prefer to check size. 

We cannot use `limitopen.OpenWithLimit` directly here because the gauge and the sentry are not initialized yet, but we do can use `limitopen.Open` here.

configbp/config_test.go
```diff
@@ -0,0 +1,68 @@
+package configbp_test // to break dep cycle
```
fishy, line 1: 🔕 even without the potential of dep cycle we should still use external test whenever possible (e.g. we don't need to access any unexported things in the tests) :)

fishy (approved):

configbp/config_test.go
kylelemons: We went around and around on this at Google and in the end the consensus is that external tests should only be used when required.

1. There is no consensus that external API tests are actually better, particularly for regression testing. The simplest test that verifies behavior is usually the right one, and sometimes that means using an unexported API.
1. You can still test the exported API only even if you're not in an external test, but the reverse is not true.
1. When using Bazel, if someone decides to make an internal test, it changes what gazelle will generate -- the internal test now takes the "foo_test" target name and shoves the external test to "foo_x_test" which can be confusing, especially if you are tracking things related to the test target over time.

We didn't prevent people from writing external tests, or change them when we found them, but it was not encouraged.

configbp/config.go
kylelemons: bufio.Scanner has an internal maximum token size of 64KiB

https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/bufio/scan.go;l=81;bpv=1;bpt=1

kylelemons: oh, but we will still carry it over.  I see.

configbp/config.go
kylelemons: I did it this way to avoid a _tiny_ bit of pointer chasing.  Shouldn't really matter here though, and agree that it's a bit weird.

baseplate.go
kylelemons: added back, though I'm also throwing in the `edgecontext.Factory(...)`

configbp/config_test.go
fishy: I believe 3 is no longer true with new version of gazelle.

I'm not suggesting to use external-only tests. I'm suggesting to use external tests as much as possible, and only use internal tests when you need to test something that's not possible in external tests (so mix them if needed). e.g. use external tests to test public api, plus internal tests to test internal implementations that's not exposed.

it's likely less an issue at google, but I do see people not exposing the api correctly, and not realize that with internal test (because in internal test you still have access to everything). those problems can be avoided with external tests because external tests put you into the api users' shoes.

fishy (approved):

configbp/config_test.go
kylelemons: Unfortunately even external tests still have access to `internal/` packages that a user won't have access to.  There is no silver bullet here, so the simplest approach (package-internal tests that can choose to test the external API if they wish) wins out.

konradreiche: Especially for unit tests it would prevent you from working with the internal state of a type which you often need to do to test the desired behavior, where as integration tests could live in a different space to begin with.

konradreiche (approved):

PR #386, opened 2021-07-02
Optimize imports

kylelemons:
This applies the following import style:
  - All imports are in exactly one block
  - Imports are organized into 3 groups
     "stdlib"
     "external"
     "same-module"
  - Redundant renamed imports are removed

### DIFF
--- a/Style.md
+++ b/Style.md
@@ -38,22 +38,10 @@ import (
 )
 ```
 
-### Exceptions
-
-In whole file examples
-(see the end of https://pkg.go.dev/testing?tab=doc#hdr-Examples),
-merge Groups 2&3.
-The reason is that whole file examples are presented in package users' point of
-view,
-and in that point of view everything under `github.com/reddit/baseplate.go` are
-also third-party packages.
-
-Please note that this exception does not include tests and normal examples.
-
 ## Rename imports
 
-When an imported package name is different from the last element of the import
-path, always rename import it to make it explicit.
+When an imported package name is different from what is implied by the
+final element(s) of the import path, rename the import it to make it explicit.
 
 Example:
 
@@ -66,11 +54,15 @@ import (
 
 ### Exceptions
 
-As recommended by go modules version design,
-in the example of `"github.com/go-redis/redis/v7"`,
-`v7` is the major version number,
-so we consider `redis` as the actual last element of import path.
-As a result there's no need to rename import it.
+The implied package name of all of the following is still "foo", and thus
+they do not need to be renamed:
+
+* `.../foo.v1`
+* `.../foo/v1`
+* `.../foo-go`
+* `.../foo.go`
+
+Most import tools already understand these conventions.
 
 ## One-per-line style
 
--- a/baseplate.go
+++ b/baseplate.go
@@ -8,7 +8,7 @@ import (
 	"os"
 	"time"
 
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 
 	"github.com/reddit/baseplate.go/batchcloser"
 	"github.com/reddit/baseplate.go/ecinterface"
--- a/baseplate_test.go
+++ b/baseplate_test.go
@@ -12,7 +12,7 @@ import (
 	"testing"
 	"time"
 
-	baseplate "github.com/reddit/baseplate.go"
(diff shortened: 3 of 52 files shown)


redis/redispipebp/breaker_test.go
```diff
@@ -7,9 +7,10 @@ import (
 	"time"
 
 	"github.com/joomcode/redispipe/redis"
-	"github.com/reddit/baseplate.go/breakerbp"
 	"github.com/sony/gobreaker"
 
+	"github.com/reddit/baseplate.go/breakerbp"
+
```
pacejackson, line 13: We got _oooone_ leftover extra newline here

pacejackson (approved):

fishy:
We are already doing the first 2, so this is only for the 3rd one, right? can you also update `Style.md` as we currently have an exception there that you are changing?

(also, regarding 1 `All imports are in exactly one block`, there's actually an exception of `imporet "C"` which should always be separated, but we don't use any cgo here so it doesn't matter that much)

regarding 3, the reason I added that exception into the style guide is because this helps devs to quickly find what an identifier in the code is referring to. e.g. when they see `retry.Option` and then look into the imports they see `import retry "github.com/avast/retry-go"` they will know that `retry` is the import, but if the import is only `import "github.com/avast/retry-go"` they would only know that this is likely it but they do need a full ide/tools to be sure about it.

I know that some tools try to auto remove those "redundant" rename imports, and I would argue that this is actually a bug that should be fixed in those tools because those "redundant" rename imports provides value. but I don't feel too strongly about that ¯\\\_(ツ)\_/¯

redis/redispipebp/breaker_test.go
kylelemons: hmmmmmmmmmmmmmmmm goland wachu doin

kylelemons:
> We are already doing the first 2, so this is only for the 3rd one, right? can you also update Style.md as we currently have an exception there that you are changing?

Updated

> I know that some tools try to auto remove those "redundant" rename imports, and I would argue that this is actually a bug that should be fixed in those tools because those "redundant" rename imports provides value. but I don't feel too strongly about that ¯\_(ツ)_/¯

The `-go`, `.go` and `/vN` conventions are understood by most tools.  If I had `gopkg.in/foo.v1` import itself as `bar`, most of those tools would auto-add the `bar` name.

Style.md
```diff
@@ -67,11 +55,15 @@ import (
...
-`v7` is the major version number,
-so we consider `redis` as the actual last element of import path.
-As a result there's no need to rename import it.
+The implied package name of all of the following is still "foo", and thus
+they do not need to be renamed:
+
+* `.../foo.v1`
+* `.../foo/v1`
+" `.../foo-go`
+" `.../foo.go`
```
fishy, line 64: `"` should be `*` instead (the current markdown rendering is wrong: https://github.com/reddit/baseplate.go/blob/8817556713b648bed2c4bf1fd844265b01546c50/Style.md#exceptions-1)

Style.md
```diff
@@ -67,11 +55,15 @@ import (
 
 ### Exceptions
 
-As recommended by go modules version design,
-in the example of `"github.com/go-redis/redis/v7"`,
-`v7` is the major version number,
-so we consider `redis` as the actual last element of import path.
-As a result there's no need to rename import it.
+The implied package name of all of the following is still "foo", and thus
```
fishy, line 58: >The implied package name of all of the following is still "foo"

🔕 citation needed :)

fishy: actually nevermind me, I read the sentence wrong.

fishy (changes requested):
Please fix the markdown: https://github.com/reddit/baseplate.go/pull/386#discussion_r663232273

konradreiche (approved):

Style.md
```diff
@@ -4,9 +4,10 @@
 
 Make sure your code passes the following tools:
 
-- `gofmt -s`
-- `golint`
 - `go vet`
+- `gofmt -s`
+- `golint` (via `go install golang.org/x/lint/golint@latest`)
```
fishy, line 9: can you rebase on latest master? `golint` is already removed there :)

fishy (approved):

fishy (approved):

PR #387, opened 2021-07-02
Improve Sentry error legibility & update sentry-go dependency

konradreiche:
Currently Sentry errors are reported with this title:
```
*errors.errorString github.com/reddit/baseplate.go/log in ErrorWithSentry
```
Sentry extracts the top frame from the stack trace to display where the error originated. Since we wrap Sentry error logging in a method in the `log` package it will always be this frame.

This change improves the legibility of Sentry errors by:
* marking stack trace frames from baseplate.go as not `in-app`
* swap exception type and value to display the error message as the header instead (see https://github.com/getsentry/sentry-go/issues/307#issuecomment-737871530)


log/sentry.go
```diff
@@ -67,16 +70,24 @@ type SentryConfig struct {
...
+	var (
+		sampleRate float64        = 1
+		beforeSend beforeSendFunc = dropStackTraceFrame
+	)
 	if cfg.SampleRate != nil && *cfg.SampleRate >= 0 && *cfg.SampleRate <= 1 {
 		sampleRate = *cfg.SampleRate
 	}
+	if cfg.BeforeSend != nil {
+		beforeSend = cfg.BeforeSend
+	}
```
fishy, line 82: since `sentry.ClientOptions.BeforeSend` can only take a single function, what if the user want to override `BeforeSend` to do something else, but they also want to keep this default behavior that they want to drop the first frame?

I think we should export `dropStackTraceFrame` so callers can call it in their implementation, and probably also provide a way to merge in multiple `BeforeSend` functions.

konradreiche: I wonder if there is a way with less additional types to deal with. The way I saw this implemented in the Bugsnag library is to have an option called skip levels that just defines a number of frames to skip.

What if we export the whole configuration but we also apply it but wrap the `BeforeSend` method ourselves? Too much magic?

fishy: what I meant is just to export `dropStackTraceFrame` to `DropStackTraceFrame`, so if the user want to do something else they don't have to reimplement `dropStackTraceFrame`:

```
func myBeforeSend(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
  event = log.DropStackTraceFrame(event, hint)
  // do extra work
}
```

log/log_test.go
```diff
@@ -12,3 +14,15 @@ func TestZapLogger(t *testing.T) {
...
 }
+
+func TestInitSentry(t *testing.T) {
+	c, err := InitSentry(SentryConfig{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	ErrorWithSentry(context.Background(), "", errors.New("what"))
+	err = c.Close()
+	if err != nil {
```
kylelemons, line 25: if err := ...; err != nil

log/sentry.go
kylelemons: you can kinda already do

orig := cfg.BeforeSend
cfg.BeforeSend = func(...) {
  orig()
  ...
}

and you could make a helper to do that

log/sentry.go
```diff
@@ -169,3 +181,21 @@ func extractKeyValuePairs(keysAndValues []interface{}, f func(key, value string)
 	}
 	return false
 }
+
+type beforeSendFunc func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event
```
kylelemons, line 185: not worth naming this, really

kylelemons (approved):

konradreiche:
After consulting with Sentry the preferable option is to mark frames that are not in-app. I tested it and it will ensure that for the event title in Sentry it uses the first frame that is not from `github.com/reddit/baseplate.go/log.ErrorWithSentry` but will at the same time preserve the frame as part of the stack trace. 

https://docs.sentry.io/product/data-management-settings/event-grouping/stack-trace-rules/#mark-in-app-frames

What do you think? @pacejackson @fishy @kylelemons

log/log_test.go
```diff
@@ -12,3 +14,14 @@ func TestZapLogger(t *testing.T) {
...
 	log(globalLogger)
 }
+
+func TestInitSentry(t *testing.T) {
+	c, err := InitSentry(SentryConfig{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	ErrorWithSentry(context.Background(), "", errors.New("what"))
+	if err = c.Close(); err != nil {
```
kylelemons, line 24: :=

kylelemons (approved):
I like it! This definitely feels cleaner.

log/sentry.go
```diff
@@ -71,12 +74,32 @@ func InitSentry(cfg SentryConfig) (io.Closer, error) {
...
 		sampleRate = *cfg.SampleRate
 	}
+
+	// ensure first stack trace frame that is part of this package will be
+	// marked as not in-app to improve the error display in Sentry
+	beforeSend := func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
+		for i, exception := range event.Exception {
+			for j := len(exception.Stacktrace.Frames) - 1; j > 0; j-- {
+				frame := exception.Stacktrace.Frames[j]
+				if frame.Function == "ErrorWithSentry" && frame.Module == "github.com/reddit/baseplate.go/log" {
```
fishy, line 84: We also have `ErrorWithSentryWrapper`, which has [a lambda in it](https://github.com/reddit/baseplate.go/blob/v0.8.3/log/wrapper.go#L205), which won't satisfy this loop.

Maybe we should remove the function name check and only check for module?

log/sentry.go
```diff
@@ -71,12 +74,32 @@ func InitSentry(cfg SentryConfig) (io.Closer, error) {
...
+					event.Exception[i].Stacktrace.Frames[j].InApp = false
+					break
```
fishy, line 86: and if we only check for the module, we could have multiple frames that's not in-app, so we can't break on the first one either.

fishy (commented):
This is definitely much more cleaner than the previous approach.

Can you also upgrade the version of sentry in baseplate.go's go.mod? I think the latest version is several versions newer than the one in go.mod, which results in different services using different version of sentry library, and gives us inconsistent results (there are behavior changes between those versions).

log/sentry.go
konradreiche: Great point. I think it should be okay to remove it, alternatively we could do
```go
strings.HasPrefix(frame.Function, "ErrorWithSentry")
```
which captures:
```
ErrorWithSentry
ErrorWithSentryWrapper.func1
```

The only reason to include a check on the function name would be to ensure we are not missing other Sentry errors that originate from the baseplate package. But then again… we will still have the information in Sentry, so not a big deal.

log/sentry.go
konradreiche: True, it should be safe to mark them all as not in-app.

log/sentry.go
fishy: If nothing is lost, just marked differently on the ui, I'd prefer to do prefix check on the module to mark everything under `github.com/reddit/baseplate.go` as not in-app, over only excluding log package.

fishy: maybe we want to do:

```
const base = "github.com/reddit/baseplate.go"
if frame.Module == base || strings.HasPrefix(frame.Module, base + "/") {
...
```

in case we are gonna do `github.com/reddit/baseplate.go-upgrader` in the future :)

konradreiche: Agreed, this should give us the best default behavior and catch all cases. Also updated the comment.

konradreiche: @fishy wouldn't we want to mark something like `github.com/reddit/baseplate.go-upgrader` as not in-app too? Though generally we probably shouldn't mark as not in-app outside of the domain of this package.

fishy: no `github.com/reddit/baseplate.go-upgrader` could be an app in itself, and things under that prefix should be in-app.

but this is just a very hypothetical scenario and I don't feel strongly either way.

log/sentry.go
```diff
@@ -71,12 +75,33 @@ func InitSentry(cfg SentryConfig) (io.Closer, error) {
...
+
+	// Improve legibility of Sentry errors by using the error message as header
+	// instead of the error type and marking stack trace frame from
+	// baseplate.go as not in-app.
+	const base = "github.com/reddit/baseplate.go"
+	beforeSend := func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
+		for i, exception := range event.Exception {
+			event.Exception[i].Type, event.Exception[i].Value = event.Exception[i].Value, event.Exception[i].Type
+			for j, frame := range exception.Stacktrace.Frames {
+				if frame.Module == base || strings.HasPrefix(frame.Module, base+"/") {
```
fishy, line 87: super nit: we probably want to define `const prefix = base + "/"` as well, as doing the concatenation in every for loop could be wasteful.

log/sentry.go
```diff
@@ -71,12 +75,33 @@ func InitSentry(cfg SentryConfig) (io.Closer, error) {
...
 		sampleRate = *cfg.SampleRate
 	}
+
+	// Improve legibility of Sentry errors by using the error message as header
+	// instead of the error type and marking stack trace frame from
+	// baseplate.go as not in-app.
+	const base = "github.com/reddit/baseplate.go"
+	beforeSend := func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
+		for i, exception := range event.Exception {
+			event.Exception[i].Type, event.Exception[i].Value = event.Exception[i].Value, event.Exception[i].Type
```
fishy, line 85: probably add a comment here, with links to the sentry issue on github, to explain what this does.

fishy (approved):

pacejackson (approved):

log/sentry_test.go
```diff
@@ -125,3 +128,119 @@ func TestExtractKeyValuePairs(t *testing.T) {
 		)
 	}
 }
+
+func TestSentryBeforeSend(t *testing.T) {
+	t.Run("mark frames with explicit error log", func(t *testing.T) {
```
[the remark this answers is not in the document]
konradreiche: I think it just replaces whitespace with underscores, I thought that's the common conversion when you want to run a test manually.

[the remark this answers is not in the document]
fishy: yea but the problem is that when someone want to run a specific test, use a proper label will mean that they can just copy paste the name from the code. For example if we use "explicit-error-log" here then someone can run `go test TestSentryBeforeSend/explicit-error-log`, but with the current name they need to figure out what's the normalized name first before they can do that.

I don't think that's a blocker (I already approved it) but I do prefer to have shorter names without spaces :)

fishy (approved):

log/sentry_test.go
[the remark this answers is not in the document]
konradreiche: Yes, let's unify it across the repository :+1:

PR #388, opened 2021-07-07
Add internal/limitreader package

fishy:
This implements something similar to io.LimitReader, but reports error
when it's read beyond the limit.

It's under internal as this is a special case that we don't want to make
as part of our public API. We can move it out of internal in the future
if there are wider demands for it.

Switch filewatcher to use that instead of io.LimitReader. It could also
be used in https://github.com/reddit/baseplate.go/pull/385.

### DIFF
--- a/filewatcher/filewatcher_test.go
+++ b/filewatcher/filewatcher_test.go
@@ -298,7 +298,6 @@ func limitedParser(t *testing.T, expectedSize int) filewatcher.Parser {
 	return func(f io.Reader) (interface{}, error) {
 		buf, err := io.ReadAll(f)
 		if err != nil {
-			t.Error(err)
 			return nil, err
 		}
 		if len(buf) != expectedSize {
@@ -313,17 +312,32 @@ func limitedParser(t *testing.T, expectedSize int) filewatcher.Parser {
 	}
 }
 
+type logWrapper struct {
+	called int64
+}
+
+func (w *logWrapper) wrapper(tb testing.TB) log.Wrapper {
+	return func(_ context.Context, msg string) {
+		tb.Helper()
+		tb.Logf("logger called with msg: %q", msg)
+		atomic.AddInt64(&w.called, 1)
+	}
+}
+
+func (w *logWrapper) getCalled() int64 {
+	return atomic.LoadInt64(&w.called)
+}
+
 func TestParserSizeLimit(t *testing.T) {
 	interval := time.Millisecond
 	filewatcher.InitialReadInterval = interval
 	writeDelay := interval * 10
 	timeout := writeDelay * 20
 
-	const size = 3
 	payload1 := []byte("Hello, world!")
-	expectedPayload1 := payload1[:size]
-	payload2 := []byte("Bye, world!")
-	expectedPayload2 := payload2[:size]
+	size := len(payload1)
+	expectedPayload := payload1[:size]
+	payload2 := []byte("Bye bye, world!")
 
 	dir, err := os.MkdirTemp("", "filewatcher_test_")
 	if err != nil {
@@ -346,6 +360,7 @@ func TestParserSizeLimit(t *testing.T) {
 		}
 	}()
 
+	var wrapper logWrapper
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 	data, err := filewatcher.New(
@@ -353,15 +368,15 @@ func TestParserSizeLimit(t *testing.T) {
 		filewatcher.Config{
 			Path:        path,
 			Parser:      limitedParser(t, size),
-			Logger:      log.TestWrapper(t),
-			MaxFileSize: size,
+			Logger:      wrapper.wrapper(t),
+			MaxFileSize: int64(size),
 		},
 	)
 	if err != nil {
 		t.Fatal(err)
 	}
 	defer data.Stop()
-	compareBytesData(t, data.Get(), expectedPayload1)
+	compareBytesData(t, data.Get(), expectedPayload)
 
 	func() {
 		f, err := os.OpenFile(path, os.O_WRONLY|os.O_TRUNC, 0644)
@@ -375,7 +390,15 @@ func TestParserSizeLimit(t *testing.T) {
 	}()
 	// Give it some time to handle the file content change
 	time.Sleep(time.Millisecond * 500)
-	compareBytesData(t, data.Get(), expectedPayload2)
+	// We expect the second parse would fail because of the data is beyond the
+	// limit, so the data should still be expectedPayload
+	compareBytesData(t, data.Get(), expectedPayload)
+	// Since we expect the second parse would fail, we also expect the logger to
+	// be called once.
+	const expectedCalled = 1
+	if called := wrapper.getCalled(); called != expectedCalled {
+		t.Errorf("Expected log.Wrapper to be called %d times, actual %d", expectedCalled, called)
+	}
 }
 
 func TestMockFileWatcher(t *testing.T) {
--- a/internal/limitreader/reader.go
+++ b/internal/limitreader/reader.go
@@ -1,0 +1,48 @@
+package limitreader
+
+import (
+	"bufio"
+	"errors"
+	"io"
+)
+
+// ErrReadLimitExceeded is the error returned when there's more data on the
+// underlying reader and a read operation is trying to read beyond that.
+var ErrReadLimitExceeded = errors.New("limitreader: read limit exceeded")
+
+// New wraps reader with a limit.
+//
+// It's similar to io.LimitReader,
+// but instead of always returning io.EOF when read beyond the limit,
+// it would return a different error (which would cause the read operation to
+// fail) when there's more data on the underlying reader.
+func New(r io.Reader, limit int64) io.Reader {
+	return &reader{
+		reader:    bufio.NewReaderSize(r, 16),
+		remaining: limit,
+	}
+}
+
+type reader struct {
+	reader    *bufio.Reader
+	remaining int64
+}
+
+func (r *reader) Read(p []byte) (int, error) {
+	// EOF handling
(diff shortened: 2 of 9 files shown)


internal/limitreader/reader_test.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+			if c.err == nil {
+				if err != nil {
+					t.Errorf("Did not expect error, got %v", err)
+				}
+			} else {
+				if !errors.Is(err, c.err) {
+					t.Errorf("Expected error %v, got %v", c.err, err)
+				}
+			}
```
pacejackson, line 73: 🔕  I think you can just check `if !errors.Is(err, c.err)` and it will still work if they are both nil?

fishy: Done (it feels a little bit too magical to me, but I don't feel strongly either way)

pacejackson (approved):

filewatcher/filewatcher_test.go
```diff
@@ -313,17 +312,40 @@ func limitedParser(t *testing.T, expectedSize int) filewatcher.Parser {
...
 	filewatcher.InitialReadInterval = interval
 	writeDelay := interval * 10
 	timeout := writeDelay * 20
 
-	const size = 3
 	payload1 := []byte("Hello, world!")
+	size := len(payload1)
 	expectedPayload1 := payload1[:size]
-	payload2 := []byte("Bye, world!")
+	_ = expectedPayload1
```
kylelemons, line 345: use or remove

filewatcher/filewatcher_test.go
```diff
@@ -313,17 +312,40 @@ func limitedParser(t *testing.T, expectedSize int) filewatcher.Parser {
...
+	payload2 := []byte("Bye bye, world!")
 	expectedPayload2 := payload2[:size]
+	_ = expectedPayload2
```
kylelemons, line 348: use or remove

filewatcher/filewatcher_test.go
```diff
@@ -375,7 +400,10 @@ func TestParserSizeLimit(t *testing.T) {
 	}()
 	// Give it some time to handle the file content change
 	time.Sleep(time.Millisecond * 500)
-	compareBytesData(t, data.Get(), expectedPayload2)
+	// We expect the second parse would fail because of the data is beyond the
+	// limit, so the data should still be expectedPayload1
+	compareBytesData(t, data.Get(), expectedPayload1)
+	wrapper.check(1)
```
kylelemons, line 406: 🔕 return the value and compare in the test

internal/limitreader/reader.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"bufio"
+	"errors"
+	"io"
+)
+
+// ErrReadBeyondLimit is the error returned when there's more data on the
+// underlying reader and a read operation is trying to read beyond that.
+var ErrReadBeyondLimit = errors.New("limitreader: read beyond limit")
```
kylelemons, line 11: 🔕 ErrReadLimitExceeded?

internal/limitreader/reader.go
```diff
@@ -0,0 +1,49 @@
...
+		return 0, ErrReadBeyondLimit
+	}
+
+	if remaining := r.limit - r.read; len(p) > int(remaining) {
+		p = p[:remaining]
+	}
+	read, err := r.reader.Read(p)
+	r.read += int64(read)
+	return read, err
+}
```
kylelemons, line 49: I don't think this should introduce buffering.  Leave that to the user.  I think that simplifies this to:

```suggestion
func New(r io.Reader, limit int64) io.Reader {
	return &reader{
		reader: r,
		limit:  limit,
	}
}

type reader struct {
	reader io.Reader
	limit  int64
}

func (r *reader) Read(p []byte) (int, error) {
  if r.limit <= 0 {
    return 0, ErrReadLimitExceeded
  }
	if r.limit > len(p) {
	  p = p[:r.limit]
	}
	n, err := r.reader.Read(p)
	r.limit -= n
	return n, err
}
```

kylelemons (changes requested):

internal/limitreader/reader.go
kylelemons: I would also 🔕 recommend exporting the LimitReader struct and returning that instead of io.Reader, since that would allow for some amount of inlining at call-sites, though it's unlikely to be relevant for us.

fishy: The suggestion you proposed is essentially the same as `io.LimitReader`, except that it returns a different error than `io.EOF`. The problem with that approach is that it cannot distinguish between the corner case of the underlying reader has exactly limit bytes of data, vs. the underlying reader has more than limit bytes of data. The reason is that in most of the `io.Reader` implementations, when you read the last n bytes of data, it won't return `n, io.EOF`. They would return `n, nil`, and then return `0, io.EOF` on the next `Read` call. (see https://play.golang.org/p/s8GEFnbn-T- for an example).

We can, of course, always add the limit by 1 in `New`, so that it can distinguish between the two cases. But that will result in this implementation actually could return limit+1 data. That's not a big problem for our use cases, and probably fine if we keep this implementation under `internal/` forever, but that is not something I would ever expose as a public API in the future.

The `bufio` approach is just more correct, although the correctness is not super important here. In the worst case scenario, that the underlying reader has a chunk of exactly limit bytes of data, and the next chunk would block for a bit before returning, both approaches (`bufio` or just use `limit+1`) would hit that blocking Read anyways in order to return the error. `bufio` will not block for longer even if the next chunk does not have enough data to fill the buffer size. The only downside of `bufio` approach is that it uses slightly more memory for the buffer, but the default buffer size is only 4096 bytes (https://cs.opensource.google/go/go/+/refs/tags/go1.16.5:src/bufio/bufio.go;drc=refs%2Ftags%2Fgo1.16.5;l=19).

filewatcher/filewatcher_test.go
fishy: that would require us to compare the logged text in a test, so changing the text (or the error message) in the future would break the test.

kylelemons: I mean the number (i.e. 1)

internal/limitreader/reader.go
kylelemons: Sure, if you want to deal with the boundary condition, it looks like this:

https://play.golang.org/p/GbaDhtZOu_4
```
func (r *reader) Read(p []byte) (int, error) {
	if r.limit <= 0 {
		n, err := r.reader.Read(p[:1])
		if !errors.Is(err, io.EOF) {
			err = ErrReadLimitExceeded
		}
		return n, err
	}
	if r.limit < int64(len(p)) {
		p = p[:r.limit]
	}
	n, err := r.reader.Read(p)
	r.limit -= int64(n)
	return n, err
}
```

That will try to read a test byte on the off-chance that precisely the amount of data is read.

kylelemons (changes requested):
see thread about removing buffering

internal/limitreader/reader.go
fishy: That's still an incorrect implementation. (it's correct "enough" for our use case). If someone keeps calling `Read` after reached the limit, those `Read` calls would eventually return `io.EOF` instead of `ErrReadLimitExceeded`. With `bufio` approach after the every `Read` call is guaranteed to return `ErrReadLimitExceeded` after the first one returning that error.

fishy: https://play.golang.org/p/Urf1103SLo_8 demonstrates the "correctness" of the different approaches.

I use `io.LimitReader` as the baseline, as that's what we are replacing. We really only want to replace the error returned, not the actual content returned (your approach will read one more byte).

kylelemons: The reason that I return the extra byte is because it is the more correct thing to do, because the extra byte is read from the underlying reader.  With your limit reader, any data caught in the buffer is completely lost, which is a far larger sin than returning one extra byte.

Observe: https://play.golang.org/p/KPfLanNM7Ul
```
original content: "123456789", limit: 5

  io.LimitReader: read: "12345", err: <nil>
                  then: "6789", err: <nil>
       no buffer: read: "123456", err: read limit exceeded
                  then: "789", err: <nil>
           bufio: read: "12345", err: read limit exceeded
                  then: "", err: <nil>
```

In any case, the fixes for the two issues you state are trivial:
* Lie about the peeked byte, like you do in the buffered version (but honestly, this is _bad_ and you shouldn't do it)
* Save returned errors so repeated calls are noops (this is only needed if your callers are doing weird things)
https://play.golang.org/p/U9G2mQgWimO

kylelemons: It is not at all uncommon to depend on this kind of behavior, either.  For example, if we e.g. read the first 1k of a file, do envsubst on it, and then combine with the rest, using the presence of the error to report to the user a warning that their config was too long.  You can do that with the stdlib limit reader (without the warning), or my limit reader, but not yours because yours discards data.  Readers should not discard data in general, it's a loaded footgun.

kylelemons (changes requested):
see thread

fishy:
>For example, if we e.g. read the first 1k of a file, do envsubst on it, and then combine with the rest, using the presence of the error to report to the user a warning that their config was too long

This is not a realistic scenario. An env var can be just at the boundary. For example you have this: `$ENVVAR` while the `$ENV` part is before boundary and `VAR` is after, and if you do that you will substitute `$ENV` and combine that with literal `VAR`, which is wrong.

And if we _really_ want to do that, stdlib is sufficient and we don't need to build our own:

```
func envVarSubstitutionMaxN(r io.Reader, n int) io.Reader {
  var buf bytes.Buffer
  _, err := io.CopyN(&buf, r, n)
  // handle err, do env var substitution inside buf
  return io.MultiReader(buf, r)
}
```

fishy:
Replaced by https://github.com/reddit/baseplate.go/pull/390

PR #389, opened 2021-07-12
Add thriftbp.RecoverPanic middleware

pacejackson:
This middleware recovers from panics and reports the error to Sentry.

### DIFF
--- a/thriftbp/server_middlewares.go
+++ b/thriftbp/server_middlewares.go
@@ -64,14 +64,16 @@ type DefaultProcessorMiddlewaresArgs struct {
 // 4. AbandonCanceledRequests
 //
 // 5. ReportPayloadSizeMetrics
+//
+// 6. RecoverPanic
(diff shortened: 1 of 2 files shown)


thriftbp/server_middlewares.go
```diff
@@ -337,3 +340,38 @@ func tHeaderProtocol2String(proto thrift.THeaderProtocolID) string {
...
+						rErr = asErr
+					} else {
+						rErr = fmt.Errorf("panic in %q: %+v", name, r)
+					}
+					log.ErrorWithSentry(
+						ctx,
+						"recovered from panic:",
+						rErr,
+						"endpoint", name,
+					)
```
pacejackson, line 362: Should this be a logger that is passed in rather than `log.ErrorWithSentry`?

fishy (approved):

thriftbp/server_middlewares.go
konradreiche: As an option maybe but it is the right default.

konradreiche (approved):

Issue #365, opened 2021-04-07
Change `log.Wrapper.UnmarshalText` to use `ErrorWithSentryWrapper` as the default one


fishy:
In order to support #364, we also need to change `log.Wrapper.UnmarshalText` to use `ErrorWithSentryWrapper` instead of `NopWrapper` as the default one.

fishy:
even if we changed the default of `UnmarshalText` to sentry, it won't help when the config line is not present in the yaml.
for example, say we have this config struct:

```go
type Config struct {
  Logger log.Wrapper `yaml:"logger"`
  OtherThing int `yaml:"otherThing"`
  ...
}
```

for this yaml it works that we parse it as sentry:

```yaml
logger:
otherThing: 3
```

for this yaml it doesn't work (the log.Wrapper will be nil):

```yaml
otherThing: 3
```

so we need to change `log.Wrapper`'s default to sentry as well.

PR #390, opened 2021-07-14
Add internal/limitopen package

fishy:
Replaces https://github.com/reddit/baseplate.go/pull/388.


filewatcher/filewatcher_test.go
```diff
@@ -50,7 +50,11 @@ func TestFileWatcher(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer os.RemoveAll(dir)
+	t.Cleanup(func() {
```
kylelemons, line 53: use `t.TempDir`, it will clean it up automatically

internal/limitopen/limitopen.go
```diff
@@ -0,0 +1,101 @@
...
+//    read from it indefinitely)
+//
+// It would never return both non-nil r and err.
+// When err is nil it's the caller's responsibility to close r returned.
+func Open(path string) (r io.ReadCloser, size int64, err error) {
+	var f *os.File
+	f, err = os.Open(path)
+	if err != nil {
+		return
+	}
```
kylelemons, line 37: * Don't use bare returns, especially if it requires predeclaring variables
* Add context to errors, and use %w if you want them to be visible

```suggestion
	f, err := os.Open(path)
	if err != nil {
		return nil, 0, fmt.Errorf("opening file: %w", err)
	}
```

internal/limitopen/limitopen.go
```diff
@@ -0,0 +1,101 @@
...
+// it will also use log.ErrorWithSentry to report it.
+// When hardLimit > 0 and the size of the path as reported by the os is larger,
+// it will close the file and return an error directly.
+func OpenWithLimit(path string, softLimit, hardLimit int64) (io.ReadCloser, error) {
+	r, size, err := Open(path)
+	if err != nil {
+		return nil, err
+	}
+
+	metricsbp.M.Histogram("limitopen.size").With(
```
fishy, line 72: actually I'm thinking maybe gauge is better than histogram here.

internal/limitopen/limitopen.go
```diff
@@ -0,0 +1,101 @@
...
+//
+// It would never return both non-nil r and err.
+// When err is nil it's the caller's responsibility to close r returned.
+func Open(path string) (r io.ReadCloser, size int64, err error) {
+	var f *os.File
+	f, err = os.Open(path)
+	if err != nil {
+		return
+	}
+
```
kylelemons, line 38: instead of forcing the programmer to remember to close:
```
func Open(...) (..., rerr error)
...
// after opening:
defer func() {
  if rerr != nil {
     f.Close()
  }
}()
```

internal/limitopen/limitopen.go
```diff
@@ -0,0 +1,101 @@
...
+	var stats fs.FileInfo
+	stats, err = f.Stat()
+	if err != nil {
+		f.Close()
+		return
+	}
```
kylelemons, line 44:
```suggestion
	stats, err := f.Stat()
	if err != nil {
		return nil, 0, fmt.Errorf("getting file size: %w", err)
	}
```

internal/limitopen/limitopen.go
```diff
@@ -0,0 +1,101 @@
...
+// When hardLimit > 0 and the size of the path as reported by the os is larger,
+// it will close the file and return an error directly.
+func OpenWithLimit(path string, softLimit, hardLimit int64) (io.ReadCloser, error) {
+	r, size, err := Open(path)
+	if err != nil {
+		return nil, err
+	}
+
+	metricsbp.M.Histogram("limitopen.size").With(
+		"path", path,
```
kylelemons, line 73: you probably want to only do base name for cardinality reasons?

filewatcher/filewatcher.go
```diff
@@ -103,7 +108,7 @@ func (r *Result) watcherLoop(
 			case fsnotify.Create, fsnotify.Write:
 				// Wrap with an anonymous function to make sure that defer works.
 				func() {
-					f, err := os.Open(path)
+					f, err := limitopen.OpenWithLimit(path, softLimit, hardLimit)
 					if err != nil {
 						logger.Log(context.Background(), "filewatcher: I/O error: "+err.Error())
```
kylelemons, line 113: The soft limit is logged to sentry, but the hard limit won't be?

internal/limitopen/limitopen.go
```diff
@@ -0,0 +1,101 @@
...
+		return
+	}
+
+	var stats fs.FileInfo
+	stats, err = f.Stat()
+	if err != nil {
+		f.Close()
+		return
+	}
+
```
kylelemons, line 45: Since you already have the stat, you might as well check that it's a regular file

kylelemons (changes requested):

filewatcher/filewatcher.go
fishy: hard limit violation is a hard failure, which is already handled (for the first parser call it will be returned to the caller, for later calls it's handled by logger passed in, the logger passed into filewatcher is usually `log.ErrorWithSentryWrapper`, see https://github.com/reddit/baseplate.go/issues/365).

internal/limitopen/limitopen.go
```diff
@@ -0,0 +1,104 @@
...
+//    beyond the size returned (for example, if you use this function to open
+//    /dev/zero, the system would return 0 as the size, and as a result when
+//    reading from r you would get EOF immediately, instead of being able to
+//    read from it indefinitely)
+//
+// It would never return both non-nil r and err.
+// When err is nil it's the caller's responsibility to close r returned.
+func Open(path string) (r io.ReadCloser, size int64, err error) {
+	var f *os.File
+	f, err = os.Open(path)
```
kylelemons, line 34: you can use `f, err :=` now and don't need the var

internal/limitopen/limitopen.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	defer func() {
+		if err != nil {
+			f.Close()
+		}
+	}()
+
+	var stats fs.FileInfo
+	stats, err = f.Stat()
```
kylelemons, line 46: same

kylelemons (approved):

internal/limitopen/limitopen.go
fishy: the named returns are for documentation purpose (if we just make the function signature `(io.ReadCloser, int64, error)` it won't be obvious what the `int64` return is for).

pacejackson (approved):

konradreiche (approved):

PR #392, opened 2021-07-22
tracing: UUID -> Hex

fishy:
The major open source tracing libraries/standards, including Zipkin,
OpenTracing, and OpenTelemetry are all using 128-bit hex id for traces
and 64-bit hex id for spans. So change our uuid trace/span id to hex id.

cc: @spladug @praxist

### DIFF
--- a/tracing/BUILD.bazel
+++ b/tracing/BUILD.bazel
@@ -26,7 +26,6 @@ go_library(
         "//runtimebp",
         "//timebp",
         "@com_github_getsentry_sentry_go//:sentry-go",
-        "@com_github_gofrs_uuid//:uuid",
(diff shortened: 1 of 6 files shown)


konradreiche (approved):
Is that essentially almost equivalent to a UUID and just changes the representation format or is it more nuanced?

fishy:
> Is that essentially almost equivalent to a UUID and just changes the representation format or is it more nuanced?

For the trace id, yes (when we used uuid we have to chop off the hyphens so they are essentially just 128-bit hex)

For the span ids, they are half of what we used before in UUID version (when we used full uuid version the zipkin collector we used will chop them in half anyways).

Issue #362, opened 2021-04-07
`thriftbp.ClientPool` should not implements `thrift.TClient`


fishy:
As discussed with #359, making `thriftbp.ClientPool` implementing `thrift.TClient` leads to footgun that services creating a concrete service client out of it and share that across multiple goroutines, which would lead to data races.

Adding a function like `thriftbp.ClientPool.TClient` (name tbd) to return an `thrift.TClient` implementation instead, so forcing services sharing the created service client to fix their code.

PR #393, opened 2021-07-23
thriftbp: Change ClientPool from implementing TClient to return it

fishy:
Fixes https://github.com/reddit/baseplate.go/issues/362.

Currently there's a semi-widespread footgun of services sharing the
concrete thrift client rather than the client pool between goroutines.
As described in #362, this breaking change hopefully can force them to
fix the issue.

### DIFF
--- a/thriftbp/doc_client_test.go
+++ b/thriftbp/doc_client_test.go
@@ -43,6 +43,16 @@ func NewMyServiceClient(_ thrift.TClient) MyService {
 
 // END THRIFT GENERATED CODE SECTION
 
+// MyServiceClientFactory is a thin wrapper around the ClientPool to help
+// avoiding the reuse of the concrete client.
+type MyServiceClientFactory struct {
+	pool thriftbp.ClientPool
+}
+
+func (f MyServiceClientFactory) Client() MyService {
+	return NewMyServiceClient(f.pool.TClient())
+}
+
 func LoggingMiddleware(next thrift.TClient) thrift.TClient {
 	return thrift.WrappedTClient{
 		Wrapped: func(ctx context.Context, method string, args, result thrift.TStruct) (thrift.ResponseMeta, error) {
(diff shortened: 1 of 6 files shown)


pacejackson (approved):
I wonder if it would be a good idea to suggest a simple wrapper for devs to use like this:

```go
type MyServiceClientFactory struct {
	Pool thriftbp.ClientPool
}

// Client returns a new client.
// Example:
//	myService := MyServiceClientFactory{p: pool}
//	resp, err := myService.Client().MyEndpoint(ctx, resp)
func (f MyServiceClientFactory) Client() myservice.MyServiceClient {
	return myservice.NewMyServiceClient(f.Pool.TClient())
}
```

fishy:
@pacejackson I think the problem is that currently services don't use a client factory, and also they didn't realize that `thriftbp.ClientPool` is actually a client factory (maybe we should rename it to `thriftbp.ClientPoolFactory` :) )

In the release note of `v0.9.0` I would likely write something similar to this to explain it (but instead of naming it as -Factory, just say they should store the pool into their service object and create the concrete client every time before using).

pacejackson:
So my reason for suggesting this is just that I think it's a good suggestion to make the code a little more ergonomic. 

```go
type MyService struct {
	Pool thrifbp.ClientPool
}

func (ms *MySerice) MyEndpoint(ctx context.Context, req MyRequest) error {
	...
	client := otherservice.NewOtherService(ms.Pool.TClient())
	resp, err := client.MyEndpoint(ctx, req)
	...
}
```

vs:

```go
type OtherService struct {
	Pool thrifbp.ClientPool
}

func (os OtherService) Client() otherservice.Client {
	return otherservice.NewOtherService(os.Pool.TClient())
}


type MyService struct {
	OtherService OtherService
}

func (ms *MySerice) MyEndpoint(ctx context.Context, req MyRequest) error {
	...
	resp, err := ms.OtherService.Client().MyEndpoint(ctx, req)
	...
}
```

fishy:
While I agree with you, I think this is better done on thrift compiler instead (so for every service, in additional to generate `NewServiceClient`, it also generates `NewServiceClientFactory`).

But sure I can add it the example to the doc comment.

fishy:
New client factory example added.

pacejackson (approved):

thriftbp/doc_client_test.go
```diff
@@ -43,6 +43,16 @@ func NewMyServiceClient(_ thrift.TClient) MyService {
 
 // END THRIFT GENERATED CODE SECTION
 
+// MyServiceClientFactory is a think wrapper around the ClientPool to help
```
konradreiche, line 46: thin wrapper?

konradreiche (approved):

thriftbp/doc_client_test.go
fishy: yes >.>

PR #401, opened 2021-08-09
Add wait function for cluster client

andizzle:
Add a WAIT function for redis cluster client that can actually wait for the targeted client.


external.bzl
```diff
@@ -1,11 +1,24 @@
...
+        sum = "h1:bi/1aS/5W00E2ny5q65w9SnKpWEF/UIOqDYBILpo9rA=",
+        version = "v0.1.1",
+    )
+    go_repository(
+        name = "cc_mvdan_xurls_v2",
+        importpath = "mvdan.cc/xurls/v2",
+        sum = "h1:NSZPykBXJFCetGZykLAxaL6SIpvbVy/UFEniIfHAa8A=",
+        version = "v2.2.0",
+    )
+
```
fishy, line 16: please manually remove this blank line and all other blank lines in this file added by gazelle (see: https://github.com/bazelbuild/bazel-gazelle/issues/1088)

redis/deprecated/redisbp/monitored_client.go
```diff
@@ -2,12 +2,19 @@ package redisbp
...
 	"time"
 
 	"github.com/go-redis/redis/v8"
+
 	"github.com/reddit/baseplate.go/metricsbp"
 )
 
+// ErrReplicationFactorFailed returns when the cluster client wait function returns replica reached count
+// that is less than desired replication factor
+var ErrReplicationFactorFailed = errors.New("redis: failed to meet the requested replication factor")
```
fishy, line 16: nit: this should be `redisbp:` instead of `redis:` (use the name of the package for the error prefix).

redis/deprecated/redisbp/monitored_client.go
```diff
@@ -30,12 +37,40 @@ func NewMonitoredFailoverClient(name string, opt *redis.FailoverOptions) *redis.
...
 }
 
+// ClusterClient extends redis cluster client with a functional Wait function
+type ClusterClient struct {
+	*redis.ClusterClient
+}
+
+// Wait blocks the current client until all the previous write commands are
+// successfully transferred and acknowledged by at least the specified number of replicas.
+func (cc *ClusterClient) Wait(ctx context.Context, key string, numSlaves int, timeout time.Duration) (replicas int64, err error) {
```
fishy, line 47: 🔕 I guess this is tied to redis command so very unlikely to change in the future, but I'd still prefer to convert the args into a struct so we can avoid breaking changes if we need to add/remove args. e.g.

```
Wait(ctx context.Context, args WaitArgs) (replicas int64, err error)
```

fishy (changes requested):
Please also run `go mod tidy` and commit the changes.

fishy:
also cc: @pacejackson

redis/deprecated/redisbp/monitored_client.go
andizzle: Ok sure

andizzle:
@fishy I'm unsure how I introduced the lint issue, I ran the `./scripts/bazel_cleanup.sh` that introduced a number of dependency updates, but this change does not introduces new deps. Should I manually reset those changes?

fishy:
`./scripts/bazel_cleanup.sh` does not run `go mod tidy`, you need to run both (`go mod tidy` then `./scripts/bazel_cleanup.sh`).

redis/deprecated/redisbp/monitored_client.go
```diff
@@ -30,12 +37,40 @@ func NewMonitoredFailoverClient(name string, opt *redis.FailoverOptions) *redis.
...
+	if numSlaves <= 0 {
+		return 0, nil
+	}
+
+	client, err := cc.ClusterClient.MasterForKey(ctx, key)
+	if err != nil {
+		return 0, err
```
fishy, line 54: 🔕 should this error also be wrapped?

andizzle:
Umm I ran `go mod tidy && ./scripts/bazel_cleanup.sh` and committed everything, but still get the liner error.

fishy:
oh right you actually need to do `go mod tidy && git add go.mod go.sum && ./go-mod-bazel.sh`, because of https://github.com/reddit/baseplate.go/blob/1ad14606c0418ea2ce050c94f82f35a737686f90/scripts/bazel_cleanup.sh#L6-L7

redis/deprecated/redisbp/monitored_client.go
```diff
@@ -30,12 +37,47 @@ func NewMonitoredFailoverClient(name string, opt *redis.FailoverOptions) *redis.
...
+type ClusterClient struct {
+	*redis.ClusterClient
+}
+
+// WaitArgs enclose inputs for Wait command into a struct
+type WaitArgs struct {
+	Key       string
+	NumSlaves int
+	Timeout   time.Duration
+}
```
pacejackson, line 50: I'm not sure that you're really getting anything out of using an `arg` struct, all of these are required arguments from what I can tell.

redis/deprecated/redisbp/monitored_client.go
```diff
@@ -30,12 +37,47 @@ func NewMonitoredFailoverClient(name string, opt *redis.FailoverOptions) *redis.
...
 
+// ClusterClient extends redis cluster client with a functional Wait function
+type ClusterClient struct {
+	*redis.ClusterClient
+}
+
+// WaitArgs enclose inputs for Wait command into a struct
+type WaitArgs struct {
+	Key       string
+	NumSlaves int
```
pacejackson, line 48: Unless you want to call this "ReplicationFactor", it probably doesn't make sense to use that term in the errors.

pacejackson: also, `NumReplicas` might be a better name for this arg.

pacejackson (changes requested):

redis/deprecated/redisbp/monitored_client.go
fishy: I suggested to use an struct not for optional args, but for making this a non-breaking change in the future if we need to add/remove args. The main reason is that we already have 3 args (besides context.Context) and history tells us that at that number it's very likely we'll try to add/remove args in the future 😅

redis/deprecated/redisbp/monitored_client.go
```diff
@@ -30,12 +37,47 @@ func NewMonitoredFailoverClient(name string, opt *redis.FailoverOptions) *redis.
...
+	client, err := cc.ClusterClient.MasterForKey(ctx, args.Key)
+	if err != nil {
+		return 0, fmt.Errorf("redisbp: error while trying to retrieve master from key, %w", err)
+	}
+
+	replicas, err = client.Wait(ctx, args.NumSlaves, args.Timeout).Result()
+	if err != nil {
+		return 0, fmt.Errorf("redisbp: error while trying to apply replication factor, %w", err)
+	} else if int(replicas) < args.NumSlaves {
+		return replicas, ErrReplicationFactorFailed
```
konradreiche, line 68: Should this also just be `return 0, ErrReplicationFactorFailed`?

konradreiche (approved):

redis/deprecated/redisbp/monitored_client.go
andizzle: This still returns the number of replicas it reached, it could be 0 < r < numSlaves.

external.bzl
```diff
@@ -373,6 +373,7 @@ def go_dependencies():
         sum = "h1:GOZbcHa3HfsPKPlmyPyN2KEohoMXOhdMbHrvbpl2QaA=",
         version = "v0.1.0",
     )
+
```
fishy, line 376: nit: Please remove those added blank lines. (actually just this one)

redis/deprecated/redisbp/monitored_client.go
```diff
@@ -30,12 +37,47 @@ func NewMonitoredFailoverClient(name string, opt *redis.FailoverOptions) *redis.
...
+
+	client, err := cc.ClusterClient.MasterForKey(ctx, args.Key)
+	if err != nil {
+		return 0, fmt.Errorf("redisbp: error while trying to retrieve master from key, %w", err)
+	}
+
+	replicas, err = client.Wait(ctx, args.NumReplicas, args.Timeout).Result()
+	if err != nil {
+		return 0, fmt.Errorf("redisbp: error while trying to apply replication factor, %w", err)
+	} else if int(replicas) < args.NumReplicas {
```
fishy, line 67: 🔕  hmm I'm not sure why none of the linters caught this, since you already did return in the other branch you shouldn't need to use `else` here:

```suggestion
	}
	if int(replicas) < args.NumReplicas {
```

pacejackson (approved):

redis/deprecated/redisbp/monitored_client.go
fishy: 🔕 It's probably more useful to add how many replicas we got into the error message, as long as you use `%w` caller can still use `errors.Is` to check it:

```suggestion
		return replicas, fmt.Errorf("%w: %d/%d", ErrReplicationFactorFailed, replicas, args.NumReplicas)
```

redis/deprecated/redisbp/monitored_client.go
```diff
@@ -30,12 +37,47 @@ func NewMonitoredFailoverClient(name string, opt *redis.FailoverOptions) *redis.
...
+	}
+
+	client, err := cc.ClusterClient.MasterForKey(ctx, args.Key)
+	if err != nil {
+		return 0, fmt.Errorf("redisbp: error while trying to retrieve master from key, %w", err)
+	}
+
+	replicas, err = client.Wait(ctx, args.NumReplicas, args.Timeout).Result()
+	if err != nil {
+		return 0, fmt.Errorf("redisbp: error while trying to apply replication factor, %w", err)
```
fishy, line 66: super nit: use `:` instead of `,` in error message.

fishy (approved):

fishy (approved):

PR #406, opened 2021-08-13
errorsbp: Export PrefixError and PrefixedError

fishy:
They used to be used inside Batch.AddPrefix, but that won't be usable in
the use case that the user need to run some workers in parallel and add
the errors to a batch. Export PrefixError and PrefixedError to support
that use case.

### DIFF
--- a/errorsbp/prefix.go
+++ b/errorsbp/prefix.go
@@ -1,0 +1,48 @@
+package errorsbp
+
+// PrefixError appends prefix to err.
+//
+// If err is nil, nil will be returned.
+// If prefix is empty, err will be returned as-is.
+// Otherwise, the returned error will be of type *PrefixedError,
+// and the error message would be:
+//
+//     "prefix: err.Error()"
+//
+// NOTE: A reason to use PrefixError over fmt.Errorf(prefix + ": %w", err)
+// is that prefix could contain format verbs, e.g. prefix = "foo%sbar".
+func PrefixError(prefix string, err error) error {
+	if err == nil {
+		return nil
+	}
+
+	if prefix == "" {
+		return err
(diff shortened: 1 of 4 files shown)


errorsbp/prefix.go
```diff
@@ -0,0 +1,48 @@
...
+// If err is nil, nil will be returned.
+// If prefix is empty, err will be returned as-is.
+// Otherwise, the returned error will be of type *PrefixedError,
+// and the error message would be:
+//
+//     "prefix: err.Error()"
+//
+// NOTE: A reason to use PrefixError over fmt.Errorf(prefix + ": %w", err)
+// is that prefix could contain format verbs, e.g. prefix = "foo%sbar".
+func PrefixError(prefix string, err error) error {
```
kylelemons, line 14: What is the benefit of `PrefixError` over using `fmt.Errorf` with `%w`? I am generally a fan of using the stdlib features over making our own.

kylelemons (changes requested):

errorsbp/prefix.go
fishy: with `fmt.Errorf` you can't blindly do `errs <- fmt.Errorf("prefix: %w", worker())`, because that will convert a nil-error into a non-nil error.

fishy: another benefit is that if you want to check which worker failed, you can (by `errors.As(err, new(errorbp.PrefixedError))` then check its `Prefix()`).

kylelemons: If you're looking to enable figuring out which worker failed, make this a WorkerError.  Don't conflate a prefix with the worker that failed.

PR #407, opened 2021-08-14
Consolidate config types and YAML annotations

### LINKED ISSUE #364: Remove the `Config` types that's solely for yaml parsing
We currently have a few `Config` types defined in packages that their sole purpose if for yaml parsing, for example `tracing.Config`. We should remove them and just make the actual config type used by the code (for example, `tracing.TracerConfig` in this case) for yaml parsing.

konradreiche:
This change consolidates `tracing.Config` and `tracing.TracerConfig` into one type with corresponding YAML tags which would allow allow us configure the tracing logger. In addition, this provides YAML tags for any other configuration type.

Closes #364

### DIFF
--- a/filewatcher/filewatcher.go
+++ b/filewatcher/filewatcher.go
@@ -132,9 +132,11 @@ var (
 )
 
 // Config defines the config to be used in New function.
+//
+// Can be deserialized from YAML.
 type Config struct {
 	// The path to the file to be watched, required.
-	Path string
+	Path string `yaml:"path"`
 
 	// The parser to parse the data load, required.
 	Parser Parser
@@ -143,7 +145,7 @@ type Config struct {
 	// either returned by parser or by the underlying file system watcher.
 	// Please note that this does not include errors returned by the first parser
 	// call, which will be returned directly.
-	Logger log.Wrapper
+	Logger log.Wrapper `yaml:"logger"`
 
 	// Optional. When <=0 DefaultMaxFileSize will be used instead.
 	//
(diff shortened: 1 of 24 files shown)


kylelemons (approved):

filewatcher/filewatcher.go
```diff
@@ -132,9 +132,11 @@ var (
...
 // Config defines the config to be used in New function.
+//
+// Can be deserialized from YAML.
 type Config struct {
 	// The path to the file to be watched, required.
-	Path string
+	Path string `yaml:"path"`
 
 	// The parser to parse the data load, required.
 	Parser Parser
```
fishy, line 142: Since this field, which is required, cannot be deserialized from yaml, does it actually make sense to deserialize this config from yaml? how do we use this config in the yaml right now?

tracing/config.go
```diff
@@ -5,32 +5,75 @@ import (
...
+
+	// In test code,
+	// this field can be used to set the message queue the tracer publishes to,
+	// usually an *mqsend.MockMessageQueue.
+	//
+	// This field will be ignored when QueueName is non-empty,
+	// to help avoiding footgun prod code.
+	//
+	// DO NOT USE IN PROD CODE.
+	TestOnlyMockMessageQueue mqsend.MessageQueue
```
fishy, line 76: for this one we should add `yaml:"-"`, I think?

fishy (changes requested):
I think you also missed `metricsbp.Config` vs. `metricsbp.StatsdConfig`

fishy:
Oh and we also need to add yaml tags to `log.SentryConfig`, as it's currently used in `baseplate.Config`: https://pkg.go.dev/github.com/reddit/baseplate.go#Config.Sentry

filewatcher/filewatcher.go
fishy: I guess with two-step `baseplate.New` we no longer have this problem.

fishy (approved):

PR #408, opened 2021-08-17
baseplate: Implement two-step New

fishy:
Remove NewArgs.ConfigPath in favor of two-step New, and change
DecodeConfigYAML into ParseConfigYAML to support two-step New.

Two-step New gives service code a chance to override some of the parsed
config before using them to initialize Baseplate packages. For example
Config.Sentry.BeforeSend cannot be deserialized from yaml so this is the
only chance for service code to set it.

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -217,22 +217,7 @@ func Serve(ctx context.Context, args ServeArgs) error {
 	return <-shutdownChannel
 }
 
-// parseConfig parses from the YAML file at the given path into cfg.
-func parseConfig(path string, cfg Configer) error {
-	if path == "" {
-		return errors.New("baseplate.ParseConfig: no config path given")
-	}
-
-	f, err := os.Open(path)
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-
-	return DecodeConfigYAML(f, cfg)
-}
-
-// DecodeConfigYAML decods the YAML read from the given Reader into cfg.
+// ParseConfigYAML parses path as YAML into cfg.
 //
 // It enables strict parsing,
 // if there are yaml tags not defined in cfg passed in,
@@ -242,7 +227,12 @@ func parseConfig(path string, cfg Configer) error {
 // you can just pass in a *pointer* to baseplate.Config:
 //
 //     var cfg baseplate.Config
-//     baseplate.DecodeConfigYAML(reader, &cfg)
+//     err := baseplate.ParseConfigYAML(reader, &cfg)
+//     // TODO: handle err
+//     ctx, bp, err := baseplate.New(baseplate.NewArgs{
+//       EdgeContextFactory: ...,
+//       ServiceCfg: cfg,
+//     })
 //
 // If you do have customized configurations to decode from YAML,
 // the recommended way of passing the strict yaml parsing is to embed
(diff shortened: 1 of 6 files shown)


baseplate.go
```diff
@@ -301,15 +314,22 @@ type NewArgs struct {
...
+	if args.ConfigPath == "" && args.ServiceCfg == nil {
+		return nil, nil, errors.New("baseplate.New: at least one of ConfigPath and ServiceCfg must be set")
 	}
-	err := parseConfig(args.ConfigPath, cfger)
-	if err != nil {
-		return nil, nil, err
+
+	cfger := args.ServiceCfg
+	if args.ConfigPath != "" {
+		if cfger == nil {
```
kylelemons, line 323: I think it's clearer to put this outside the configPath != "" check.  It requires a lot of squinting and thinking to prove to oneself that line 334 cannot panic because cfger can't be nil at that point.

Checking after line 321 makes it very clear and ensures that a future change to line 317 or nearby can't introduce that as a bug.

baseplate.go
```diff
@@ -325,21 +345,33 @@ func New(ctx context.Context, args NewArgs) (context.Context, Baseplate, error)
 	closer, err := log.InitSentry(cfg.Sentry)
 	if err != nil {
 		bp.Close()
-		return nil, nil, err
+		return nil, nil, fmt.Errorf(
+			"baseplate.New: failed to init sentry from config %#v: %w",
+			cfg.Sentry,
+			err,
+		)
```
kylelemons, line 352: 🔕 I usually print the error on the first line and then the rest on a subsequent line.  ala
```
fmt.Errorf("failed to init sentry from config: %s\n%s", err, marshalForError(cfg.Sentry))
```

(I also don't rely on the %#v output since it's pretty rough to parse as a reader, also 🔕 )

kylelemons: same for below errors

baseplate_test.go
```diff
@@ -229,7 +230,24 @@ type serviceConfig struct {
...
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			t.Fatalf("Failed to close tmp config file %q: %v", path, err)
+		}
+	}()
+	if _, err := io.Copy(f, strings.NewReader(content)); err != nil {
+		t.Fatalf("Failed to write to tmp config file %q: %v", path, err)
+	}
+}
```
kylelemons, line 248: 1. use os.WriteFile
2. do the tempdir call inside the helper (it'll be unique for each call) and return the path

```suggestion
func tempConfig(t *testing.T, content string) string {
	t.Helper()
  path := filepath.Join(t.TempDir(), "config.yaml"))
	if _, err := os.WriteFile(path, []byte(content)); err != nil {
		t.Fatalf("SETUP: Failed to write to tmp config file %q: %v", path, err)
	}
	return path
}
```

kylelemons (changes requested):
Will this come with a change to baseplate-cookiecutter too?

baseplate.go
fishy: I consider `\n` an abnormal in error messages and should be avoided. Sure if you have a very very long error message maybe an `\n` is inevitable 🤷 , but it should be avoided as much as possible.

Both golint and staticcheck will treat capitalized error message or error messages ending with `.` as an error, which is inline with "don't put `\n` into your error messages".

baseplate.go
fishy: on second thought I should remove `args.ConfigPath` instead of deprecating it, so this will be irrelevant.

baseplate_test.go
fishy: Changed to `os.WriteFile`, and since that we no longer to wrap another helper function for the tests.

baseplate.go
kylelemons: The first rule of error messages is that they should be useful.  Splatting a config into the error with %#v almost always violates this because it's horribly opaque, the user didn't write the config in Go syntax, it doesn't follow pointers, and it's all smashed into one unbroken blob.  So, it's "less bad" to stuff it in the second line.

The alternative is to log the bad config and then return an error saying to scroll up and look at the error, which isn't a great look, but is still serviceable.

kylelemons (changes requested):

baseplate.go
fishy: are you ok with `%+v`?

kylelemons: +v doesn't really solve the problem though.  It's not Go syntax, sure, but it's still a condensed opaque syntax that doesn't match what the user typed in

kylelemons: and I don't think +v follows pointers any more than #v, right?

fishy: I would say dumping the config here (with either `%#v` or `%+v`, but `%v` is not good enough) is providing useful information. it will make it obvious to the user if the error is caused by the parsed config is actually empty (so they might have a wrong filename passed in, or maybe a typo in the yaml file, or something like that).

But I agree with you that it's less important than the error itself, so I swapped the order of them (by using `baseplate.New: ...: %w (config: %#v)` format string)

kylelemons (approved):

baseplate.go
```diff
@@ -217,22 +217,7 @@ func Serve(ctx context.Context, args ServeArgs) error {
...
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-
-	return DecodeConfigYAML(f, cfg)
-}
-
-// DecodeConfigYAML decods the YAML read from the given Reader into cfg.
+// ParseConfigYAML parses path as YAML into cfg.
```
konradreiche, line 220: Now where this is called explicitly, do we need to call out the fact that this is supposed to be called before passing it to `baseplate.New` or do we not necessarily care as it should be given through the baseplate-cookiecutter? Also trying to think if any steps are necessary to prevent problems during upgrading if someone misses the release notes about the change.

konradreiche (approved):

baseplate.go
fishy: yes there will be a cookiecutter update after the release, and this will also be noted in the release note.

people can't blindly upgrade, `baseplate.NewArgs.ConfigPath` is removed so their code won't compile :)

PR #410, opened 2021-08-23
Upgrade go to 1.17

fishy:
Also update thrift to 0.14.2, and switch to use our own bazelisk docker
image for bazel on CI.


go.mod
```diff
@@ -1,32 +1,58 @@
 module github.com/reddit/baseplate.go
 
-go 1.16
+go 1.17
```
kylelemons, line 3: can you run `go mod tidy -go=1.17` on this? I think it will add a second block to facilitate dep graph pruning and lazy loading

(it should only need to be run once, I believe)

kylelemons (approved):

go.mod
fishy: I already did that (it's the same as running `go mod tidy` with a `go 1.17` line in `go.mod`), but merged the two `require` blocks together afterwards.

But looks like 1.17's default behavior is to group indirects together, so I changed `go.mod` file to that.

konradreiche (approved):

PR #411, opened 2021-08-24
retrybp: Add BreakerErrorFilter

fishy:
Add BreakerErrorFilter, and an example on how to use retrybp with
breakerbp.

Also fix an issue that we are not taking advantage of retry.Context
option to end retries as soon as ctx is canceled.

### DIFF
--- a/retrybp/doc_test.go
+++ b/retrybp/doc_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/avast/retry-go"
 
+	"github.com/reddit/baseplate.go/breakerbp"
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/metricsbp"
 	"github.com/reddit/baseplate.go/retrybp"
@@ -66,13 +67,13 @@ func Example() {
 				return true
 			},
 		),
-		retry.OnRetry(func(n uint, err error) {
+		retry.OnRetry(func(attempts uint, err error) {
 			if err != nil {
 				metricsbp.M.Counter("operation.failure").Add(1)
 				log.Errorw(
 					"operation failed",
 					"err", err,
-					"attempt", n,
+					"attempt", attempts,
 				)
 			} else {
 				metricsbp.M.Counter("operation.succeed").Add(1)
@@ -80,8 +81,61 @@ func Example() {
 		}),
 	)
 
-	// TODO: check error
+	// TODO: In real code, you need to check error and use the result here.
 	_ = err
-	// TODO: use result
+	_ = result
+}
+
+// This example demonstrates how to use retrybp with breakerbp.
+func ExampleBreakerErrorFilter() {
+	const timeout = time.Millisecond * 200
+
+	// TODO: use the actual type of your operation.
+	type resultType = int
+
+	// In real code this should be your actual operation you need to retry.
+	// Here we just use a placeholder for demonstration purpose.
+	operation := func() (resultType, error) {
+		return 0, nil
+	}
+
+	breaker := breakerbp.NewFailureRatioBreaker(breakerbp.Config{
+		// TODO: Add breaker configs.
+	})
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	const defaultRetries = 2
+	var result resultType
+	err := retrybp.Do(
+		ctx,
+		func() error {
+			// This is the main function call,
+			// wrapped with breaker.Execute to apply circuit breaker on it.
+			_, err := breaker.Execute(func() (interface{}, error) {
+				var err error
+				result, err = operation()
+				return result, err
+			})
+			return err
+		},
+
+		// Retry options.
+		retry.Attempts(defaultRetries+1), // retries + the first attempt
+		retrybp.CappedExponentialBackoff(retrybp.CappedExponentialBackoffArgs{
+			InitialDelay: time.Millisecond * 10,
+			MaxJitter:    time.Millisecond * 5,
+		}),
+		retrybp.Filters(
+			retrybp.RetryableErrorFilter, // this should always be the first filter.
+			retrybp.BreakerErrorFilter,   // this is the filter to handle errors returned by the breaker.
+			retrybp.ContextErrorFilter,
+			retrybp.NetworkErrorFilter, // optional: only use this if the requests are idempotent.
+		),
+	)
+
+	// TODO: In real code, you need to check error and use the result here.
+	_ = err
 	_ = result
 }
--- a/retrybp/filters.go
+++ b/retrybp/filters.go
@@ -6,6 +6,7 @@ import (
 	"net"
 
 	retry "github.com/avast/retry-go"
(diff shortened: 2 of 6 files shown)


retrybp/doc_test.go
```diff
@@ -85,3 +86,70 @@ func Example() {
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	const defaultRetries = 2
+	var result resultType
+	err := retrybp.Do(
+		ctx,
+		func() error {
+			// This is the main function call,
+			// wrapped with breaker.Execute to apply circuit breaker on it.
```
kylelemons, line 116: "wrapped with breaker.Execute" is that outdated? or maybe should be "which will be implicitly wrapped by breaker.Execute" or something?

retrybp/doc_test.go
```diff
@@ -85,3 +86,70 @@ func Example() {
...
+			res, err := breaker.Execute(func() (interface{}, error) {
+				ret, err := operation()
+				return ret, err
+			})
+			result, _ = res.(resultType)
```
kylelemons, line 121: this is a bad pattern:
* don't leak bad values into outer scopes (i.e. ensure err != nil first before assigning)
* don't discard errors (check the OK value or let it panic)

retrybp/doc_test.go
```diff
@@ -85,3 +86,70 @@ func Example() {
...
+			return err
+		},
+
+		// Default retry options, can be overridden by ctx passed in.
+		retry.Attempts(defaultRetries+1), // retries + the first attempt
+		retrybp.CappedExponentialBackoff(retrybp.CappedExponentialBackoffArgs{
+			InitialDelay: time.Millisecond * 10,
+			MaxJitter:    time.Millisecond * 5,
+		}),
+		retrybp.Filters(
```
kylelemons, line 131: 🔕 "Filters" here doesn't explain to me what it does, honestly.  Is this "Retry if it _matches_" or "Retry _except_ if it matches"? I can't tell without looking at the docs.

retrybp/doc_test.go
```diff
@@ -85,3 +86,70 @@ func Example() {
...
+			retrybp.RetryableErrorFilter, // this should always be the first filter.
+			retrybp.BreakerErrorFilter,
+			retrybp.ContextErrorFilter,
+			retrybp.NetworkErrorFilter, // optional: only use this if the requests are idempotent.
+		),
+		retry.OnRetry(func(n uint, err error) {
```
kylelemons, line 137: 🔕  `n` -> `attempt`

retrybp/doc_test.go
```diff
@@ -85,3 +86,70 @@ func Example() {
...
+			} else {
+				metricsbp.M.Counter("operation.succeed").Add(1)
+			}
+		}),
+	)
+
+	// TODO: check error
+	_ = err
+	// TODO: use result
+	_ = result
```
kylelemons, line 154: Instead of saying "TODO", say "In real code, you will need to check your error and use your results"

retrybp/filters.go
```diff
@@ -139,6 +140,26 @@ func RetryableErrorFilter(err error, next retry.RetryIfFunc) bool {
 	return next(err)
 }
 
+// BreakerErrorFilter is a Filter implementations that retries circuit-breaker
+// errors from gobreaker/breakerbp.
+//
+// It should only be used when you are using circuit breaker,
```
kylelemons, line 146: I feel like it can be used even if you aren't, it just won't ever match.  Perhaps

```
This is only useful if you are using...
```

retrybp/retry.go
```diff
@@ -48,10 +48,19 @@ func GetOptions(ctx context.Context) (options []retry.Option, ok bool) {
...
 	options, _ := GetOptions(ctx)
-	options = append(defaults, options...)
-	err := retry.Do(fn, options...)
+	mergedOptions := make([]retry.Option, 1, 1+len(defaults)+len(options))
+	// Always do this as the first option, to allow overriding it in either
+	// defaults or options from the ctx.
+	mergedOptions[0] = retry.Context(ctx)
+	mergedOptions = append(mergedOptions, defaults...)
+	mergedOptions = append(mergedOptions, options...)
+	err := retry.Do(fn, mergedOptions...)
```
kylelemons, line 63: ❗  This feels like a footgun.

There are a few options, including at least:
* Make it a type-safe positional param.
* Make it able to find it regardless of parameter, usually by making options an interface.
* Apply the options and then update the resulting option struct, and then pass that as a single option.

retrybp/doc_test.go
```diff
@@ -85,3 +86,70 @@ func Example() {
...
+		func() error {
+			// This is the main function call,
+			// wrapped with breaker.Execute to apply circuit breaker on it.
+			res, err := breaker.Execute(func() (interface{}, error) {
+				ret, err := operation()
+				return ret, err
+			})
+			result, _ = res.(resultType)
+			return err
+		},
```
pacejackson, line 123: couldn't you avoid the type conversion in here?:

```go
	var result resultType
	err := retrybp.Do(
		ctx,
		func() error {
			// This is the main function call,
			// wrapped with breaker.Execute to apply circuit breaker on it.
			_, err := breaker.Execute(func() (interface{}, error) {
				var e error
				result, e = operation()
				return result, e
			})
			return err
		},
```

kylelemons (changes requested):

retrybp/doc_test.go
fishy: Neither. Filters make individual decisions (it could be both), and defer to the next filter in the chain if they cannot make the decision.

Those are explained by the doc of Filters and Filter. I don't think it makes sense to repeat the same doc here. When people are looking at examples they are already on pkg.go.dev so I think it's reasonable to expect them to look at the doc of the types/functions if they want to learn more.

retrybp/filters.go
fishy: here the "should only" is emphasizing on the "and have proper backoff policy set" part. I just added the line breaker after comma because this is like markdown that it doesn't care about line breakers so I prefer to put line breakers after commas and periods instead of purely based on line length.

the reason it should only be used with proper backoff policy set is that without proper backoff policy, it's highly likely the retries will just fail and not really doing anything.

retrybp/retry.go
fishy: this is limited by the upstream api: https://pkg.go.dev/github.com/avast/retry-go#Do. also this is a common pattern used by go libraries (I don't particularly love this pattern, fwiw).

kylelemons: Sorry, I missed that it was just adding the first option, I misread it as _changing_ the first option, assuming it was always something.

🔕

kylelemons (approved):
FYI, it makes it much harder to do subsequent rounds when you rewrite history -- I recommend (and the forthcoming code review guidelines will also recommend) resolving comments in subsequent commits so their diffs can be reviewed separately.

pacejackson (approved):

konradreiche (approved):

PR #413, opened 2021-08-26
Enable all checks on staticcheck and remove golint

fishy:
Staticcheck with all checks is a superset of golint.


konradreiche (approved):
How will we take care of skipping generated code? Currently, getting a lot of warnings for Thrift Go code.

PR #414, opened 2021-08-30
thriftbp: Add support for User-Agent THeader

fishy:
Add SetUserAgent client middleware, and handle User-Agent THeader in
InjectServerSpan server middleware.

### DIFF
--- a/thriftbp/client_middlewares.go
+++ b/thriftbp/client_middlewares.go
@@ -84,6 +84,12 @@ type DefaultClientMiddlewareArgs struct {
 	//
 	// If it's not set, the global one from ecinterface.Get will be used instead.
 	EdgeContextImpl ecinterface.Interface
+
+	// The name for the server to identify this client,
+	// via the "User-Agent" (HeaderUserAgent) THeader.
+	//
+	// Optional. If this is empty, no "User-Agent" header will be sent.
+	ClientName string
 }
 
 // BaseplateDefaultClientMiddlewares returns the default client middlewares that
@@ -105,9 +111,11 @@ type DefaultClientMiddlewareArgs struct {
 //
 // 5. MonitorClient - This creates the spans of the raw client calls.
 //
-// 6. BaseplateErrorWrapper
+// 6. SetClientName(clientName)
 //
-// 7. SetDeadlineBudget
+// 7. BaseplateErrorWrapper
+//
+// 8. SetDeadlineBudget
 func BaseplateDefaultClientMiddlewares(args DefaultClientMiddlewareArgs) []thrift.ClientMiddleware {
 	if len(args.RetryOptions) == 0 {
 		args.RetryOptions = []retry.Option{retry.Attempts(1)}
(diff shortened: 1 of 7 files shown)


thriftbp/server_middlewares.go
```diff
@@ -128,6 +128,10 @@ func StartSpanFromThriftContext(ctx context.Context, name string) (context.Conte
 // it will be changed to IDLExceptionSuppressor instead.
 // Please use errorsbp.SuppressNone explicitly instead if that's what's wanted.
 //
+// If User-Agent (HeaderUserAgent) THeader is,
```
kylelemons, line 131: ... is set,

thriftbp/client_middlewares.go
```diff
@@ -84,6 +84,11 @@ type DefaultClientMiddlewareArgs struct {
 	//
 	// If it's not set, the global one from ecinterface.Get will be used instead.
 	EdgeContextImpl ecinterface.Interface
+
+	// The value to be set on the User-Agent THeader.
+	//
+	// Optional. If this is empty, no User-Agent header will be sent.
+	UserAgent string
```
kylelemons, line 91: Should we call this "PeerService" or something, to make the "intent" clear rather than the implementation?

Do we have any way to learn this or set it automatically?

kylelemons (approved):

kylelemons:
Adding in @spladug for a quick look to compare w/ python and spec and general context around the change

thriftbp/client_middlewares.go
spladug: Agree on calling it something less tied to the specific header (since we'll use other headers in other protocols) and that auto-detection would be really useful if possible. Python-wise it's almost all auto-detected.

spladug (approved):
LGTM modulo the one comment about the config name and auto-detection

thriftbp/client_middlewares.go
fishy: there's not a good way to do the auto detection in go. we don't have a single name in [the global config](https://pkg.go.dev/github.com/reddit/baseplate.go#Config) (we do have separated `Metrics.Namespace` and `Tracing.Namespace`), and even if we do, it's not passed here. Having a global variable storing the global name of the current running service and use that for metrics, tracing, and here _might_ make sense but I consider that out of scope for now.

as for the name, the thing on the wire is called `User-Agent`, if we ever use it for things other than `peer.service` calling it `PeerService` might be misleading?

fishy: how about calling it `ClientName` so it's tied to neither the wire header nor the span tag?

thriftbp/client_middlewares.go
```diff
@@ -85,10 +85,10 @@ type DefaultClientMiddlewareArgs struct {
 	// If it's not set, the global one from ecinterface.Get will be used instead.
 	EdgeContextImpl ecinterface.Interface
 
-	// The value to be set on the User-Agent THeader.
+	// The value to be set on the "User-Agent" (HeaderUserAgent) THeader.
```
kylelemons, line 88: I would say something more like "The value that the server will receive as the PeerName for connections initiated by this client." or something.

Whigh might mean that we want a name like "OriginPeer" or "SendPeerService" or something?

kylelemons (approved):

thriftbp/client_middlewares.go
kylelemons: *PeerService

fishy: how about something like "The name for the server to identify this client, via the "User-Agent" (HeaderUserAgent) THeader."?

PR #418, opened 2021-09-07
log: No longer swap sentry error type and message by default

fishy:
Doing so can cause sentry to no longer cluster errors with slightly
different error messages as the same error. Change it to opt-in instead.

### DIFF
--- a/log/sentry.go
+++ b/log/sentry.go
@@ -85,9 +85,7 @@ func InitSentry(cfg SentryConfig) (io.Closer, error) {
 	)
 	beforeSend := func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
 		for i, exception := range event.Exception {
-			// swap source location and of error type as title in issue list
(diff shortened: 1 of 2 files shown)


kylelemons (changes requested):
I think I would rather explore this from the Sentry angle.  Grouping by error type and grouping by error message are both somewhat flawed...

fishy:
> Grouping by error type and grouping by error message are both somewhat flawed...

while I agree with that part, grouping by error type is at least usable without extra work, but grouping by error message is not (sentry keeps spamming slack channel with "unique"  error messages)

konradreiche (approved):
Unblocking for now for the 0.9.0 release, being able to opt-in gives us at least some flexibility until we figure out the right approach with fingerprinting.

kylelemons (approved):
Approving for now, but in general I think we should be taking opinionated stances and not allowing opt-in.  Every opt-in we add multiplies the number of the situations a user could find themselves in, and makes it harder to understand and debug services at scale.

PR #419, opened 2021-09-08
Update opentracing-go to v1.2.0

foreverest:
A library I imported in my code depends on `opentracing-go@v1.2.0` which results in the v1.1.0 (which baseplate depends on) being overwritten.

The problem is they have changed the error log key in v1.2.0 which is a breaking change since it results in the `Span.FinishWithOptions` function ignoring errors.

Can we update the version to v1.2.0 ([release notes](https://github.com/opentracing/opentracing-go/releases/tag/v1.2.0))?
Or, otherwise, we might want to respect both `error` and `error.object` keys.

### DIFF
--- a/tracing/finish_option.go
+++ b/tracing/finish_option.go
@@ -11,8 +11,8 @@ const (
 	ctxKey = "context"
 
 	// Note that this must be the same as
-	// https://github.com/opentracing/opentracing-go/blob/v1.1.0/log/field.go#L128
-	errorKey = "error"
+	// https://github.com/opentracing/opentracing-go/blob/v1.2.0/log/field.go#L128
+	errorKey = "error.object"
 )
 
 // FinishOptions are the options to be converted into opentracing.FinishOptions.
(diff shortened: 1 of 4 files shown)


fishy (approved):

konradreiche (approved):

tracing/finish_option.go
```diff
@@ -11,8 +11,8 @@ const (
 	ctxKey = "context"
 
 	// Note that this must be the same as
-	// https://github.com/opentracing/opentracing-go/blob/v1.1.0/log/field.go#L128
-	errorKey = "error"
+	// https://github.com/opentracing/opentracing-go/blob/v1.2.0/log/field.go#L128
+	errorKey = "error.object"
```
kylelemons, line 15: oof... this seems like a breaking change in a minor version bump?

kylelemons (approved):

tracing/finish_option.go
foreverest: Yes!

foreverest:
The CI failure is unrelated to this change, is it?

fishy:
> The CI failure is unrelated to this change, is it?

it's unrelated.

foreverest:
Can someone merge this, please?

PR #420, opened 2021-09-08
thriftbp: Fix bug in ttl client

fishy:
In the current implementation ttlClient.expiration could never be nil,
so when user set MaxConnectionAge to a negative value to disable TTL it
doesn't really work as intended.

Also add a test for it.

### DIFF
--- a/thriftbp/ttl_client_test.go
+++ b/thriftbp/ttl_client_test.go
@@ -26,3 +26,23 @@ func TestTTLClient(t *testing.T) {
 		t.Error("Expected IsOpen call after sleep to return false, got true.")
 	}
 }
+
+func TestTTLClientNegativeTTL(t *testing.T) {
+	trans := thrift.NewTMemoryBuffer()
+	factory := thrift.NewTBinaryProtocolFactoryConf(nil)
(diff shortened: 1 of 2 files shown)


konradreiche (approved):

thriftbp/ttl_client_test.go
```diff
@@ -26,3 +26,23 @@ func TestTTLClient(t *testing.T) {
 		t.Error("Expected IsOpen call after sleep to return false, got true.")
 	}
 }
+
+func TestTTLClientNegativeTTL(t *testing.T) {
```
kylelemons, line 30: Is this flaky?  In my experience, timing-related tests usually are.  Does it pass with `--count=1000`?

kylelemons (approved):

thriftbp/ttl_client_test.go
fishy: it's not flaky as `time.Sleep` guarantees to sleep at least the desired amount of time (it would be flaky if `time.Sleep` sleeps slightly less than the desired time, and also in that case only the other test would be flaky, not this one).

PR #421, opened 2021-09-08
Fix doc comment of limitopen.Open

fishy:
The current rendering is quite confusing:
https://pkg.go.dev/github.com/reddit/baseplate.go@v0.9.0-alpha01/internal/limitopen#Open

### DIFF
--- a/internal/limitopen/limitopen.go
+++ b/internal/limitopen/limitopen.go
@@ -22,10 +22,10 @@ import (
 // 2. It returns the size reported by the system to the user.
 //
 // 3. The io.ReadCloser returned would be guaranteed to be not able to read
-//    beyond the size returned (for example, if you use this function to open
-//    /dev/zero, the system would return 0 as the size, and as a result when
-//    reading from r you would get EOF immediately, instead of being able to
-//    read from it indefinitely)
+// beyond the size returned (for example, if you use this function to open
+// /dev/zero, the system would return 0 as the size, and as a result when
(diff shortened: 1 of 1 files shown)


internal/limitopen/limitopen.go
```diff
@@ -22,10 +22,10 @@ import (
...
 //
 // 3. The io.ReadCloser returned would be guaranteed to be not able to read
-//    beyond the size returned (for example, if you use this function to open
-//    /dev/zero, the system would return 0 as the size, and as a result when
-//    reading from r you would get EOF immediately, instead of being able to
-//    read from it indefinitely)
+// beyond the size returned (for example, if you use this function to open
+// /dev/zero, the system would return 0 as the size, and as a result when
+// reading from r you would get EOF immediately, instead of being able to
+// read from it indefinitely)
```
kylelemons, line 28: I usually do it this way:

```
//  1. foo
//     more
//  2. bar
//     more
```

i.e. add an extra space or two before the `1.` so that the whole thing renders verbatim.  Having them broken into paragraphs is also a bit weird.

kylelemons (approved):

internal/limitopen/limitopen.go
kylelemons: 🔕 though.

fishy: making the whole ordered bullets verbatim works in some sense but the rendering can still be less than ideal.

konradreiche (approved):

PR #426, opened 2021-09-18
log: Attach attaches both sentry hub and logger

fishy:
Creating a contextual sentry hub is not very straight forward, and when
user uses the ctx returned from log.Attach with log.ErrorWithSentry,
they would expect the sentry report also have the info they attached.

Also remove tracing/span.InjectTraceContext, as it's really just a
shorthand for log.Attach and it's not really used anywhere else.

### DIFF
--- a/log/context.go
+++ b/log/context.go
@@ -2,7 +2,9 @@ package log
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/getsentry/sentry-go"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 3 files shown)


konradreiche (approved):

log/context.go
```diff
@@ -15,35 +17,49 @@ const (
...
+			scope.SetTag(k, fmt.Sprintf("%v", v))
+		}
+	})
+	ctx = context.WithValue(ctx, sentry.HubContextKey, hub)
 
+	// create and attach the logger
+	const additional = 1 // Number of non-AdditionalPairs fields in AttachArgs struct.
+	kv := make([]interface{}, 0, len(args.AdditionalPairs)*2+additional)
 	if args.TraceID != "" {
 		kv = append(kv, zap.String(traceIDKey, args.TraceID))
```
kylelemons, line 58: this is only appending a key, no value, so won't it cause logger.With to panic?

log/context.go
```diff
@@ -15,35 +17,49 @@ const (
...
+		}
+		for k, v := range args.AdditionalPairs {
+			scope.SetTag(k, fmt.Sprintf("%v", v))
+		}
+	})
+	ctx = context.WithValue(ctx, sentry.HubContextKey, hub)
 
+	// create and attach the logger
+	const additional = 1 // Number of non-AdditionalPairs fields in AttachArgs struct.
+	kv := make([]interface{}, 0, len(args.AdditionalPairs)*2+additional)
```
kylelemons, line 56: shouldn't this be 2*(len(AdditionalPairs)+1) since it needs to add on a {key, value} pair, which is two "values"?

tracing/tracer.go
```diff
@@ -206,7 +206,7 @@ func (t *Tracer) StartSpan(operationName string, opts ...opentracing.StartSpanOp
 	} else {
 		span.trace.traceID = t.newTraceID()
 		span.trace.sampled = randbp.ShouldSampleWithRate(t.sampleRate)
-		initRootSpan(span)
+		initRootSpan(context.Background(), span)
```
kylelemons, line 209: 🔕 I think the context should be plumbed in explicitly since the rule of thumb is "Background only in main() and TestX"

kylelemons (changes requested):

tracing/tracer.go
fishy: what do you mean by "plumbed in explicitly"?

here we are just using the context object in order to reuse the code. basically in `initRootSpan` we call `log.Attach` to create the logger and sentry hub, then we discard the logger (because we don't use it) and get the created hub out of the ctx and attach that to the span. so we need a parent context here (because we should never pass along nil contexts).

kylelemons: 🔕 the context should generally be passed in, rather than assuming context.Background

kylelemons (approved):

PR #427, opened 2021-09-20
Add explicit yaml tag for Thrift circuit breaker

konradreiche:
Thanks to YAML strict parsing I noticed that we do not explicitly map the breaker configuration for Thrift pool clients.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -167,7 +167,7 @@ type ClientPoolConfig struct {
 	// When BreakerConfig is non-nil,
 	// a breakerbp.FailureRatioBreaker will be created for the pool,
 	// and its middleware will be set for the pool.
-	BreakerConfig *breakerbp.Config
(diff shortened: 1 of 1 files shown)


thriftbp/client_pool.go
```diff
@@ -167,7 +167,7 @@ type ClientPoolConfig struct {
 	// When BreakerConfig is non-nil,
 	// a breakerbp.FailureRatioBreaker will be created for the pool,
 	// and its middleware will be set for the pool.
-	BreakerConfig *breakerbp.Config
+	BreakerConfig *breakerbp.Config `yaml:"circuitBreaker"`
```
fishy, line 170: in other yaml tags we usually only change the capitalization, never the word itself, so this should be `breakerConfig` instead. or you can rename the field to `CircuitBreaker`.

fishy (changes requested):

fishy (approved):

kylelemons (approved):

PR #428, opened 2021-09-23
httpbp: Fix CircuitBreaker

fishy:
Fix one bug: when status code >= 500 and closing the body does not error
(it won't error in majority of cases), we do not actually report error
to the circuit breaker.

Also fix a few inefficiencies:

1. When loaded is true, newBreaker is not used, so we can return it back
   to the pool immediately, instead of defer'd.
2. Use DrainAndClose instead of ReadAll and Close, as
   io.Copy(io.Discard) is more efficient than ReadAll.
3. Use pointer for the breaker in the pool and the map.

Also avoid dealing with interface{} from breaker.Execute, and add a test
to CircuitBreaker.

### DIFF
--- a/httpbp/client_middlewares.go
+++ b/httpbp/client_middlewares.go
@@ -137,7 +137,8 @@ func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
 	var breakers sync.Map
 	pool := sync.Pool{
 		New: func() interface{} {
-			return breakerbp.NewFailureRatioBreaker(config)
+			breaker := breakerbp.NewFailureRatioBreaker(config)
+			return &breaker
 		},
 	}
 
@@ -150,28 +151,28 @@ func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
 			newBreaker := pool.Get()
 			breaker, loaded := breakers.LoadOrStore(host, newBreaker)
 			if loaded {
-				defer pool.Put(newBreaker)
+				pool.Put(newBreaker)
+				newBreaker = nil
 			}
 
-			resp, err := breaker.(breakerbp.FailureRatioBreaker).Execute(func() (interface{}, error) {
-				resp, err := next.RoundTrip(req)
+			var resp *http.Response
+			_, err := breaker.(*breakerbp.FailureRatioBreaker).Execute(func() (interface{}, error) {
+				r, err := next.RoundTrip(req)
 				if err != nil {
 					return nil, err
 				}
+				resp = r
 				// circuit break on any HTTP 5xx code
 				if resp.StatusCode >= http.StatusInternalServerError {
-					// read & close to ensure underlying RoundTripper
-					// (http.Transport) is able to re-use the persistent TCP
-					// connection.
-					_, _ = io.ReadAll(resp.Body)
-					return nil, resp.Body.Close()
+					DrainAndClose(resp.Body)
+					return nil, ClientError{
+						Status:     resp.Status,
+						StatusCode: resp.StatusCode,
+					}
 				}
-				return resp, nil
+				return nil, nil
 			})
-			if resp != nil {
-				return resp.(*http.Response), err
-			}
-			return nil, err
+			return resp, err
 		})
 	}
 }
--- a/httpbp/BUILD.bazel
+++ b/httpbp/BUILD.bazel
@@ -61,5 +61,6 @@ go_test(
         "//secrets",
         "//tracing",
         "@com_github_avast_retry_go//:retry-go",
+        "@com_github_sony_gobreaker//:gobreaker",
     ],
 )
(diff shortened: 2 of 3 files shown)


httpbp/client_middlewares.go
```diff
@@ -150,28 +150,27 @@ func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
 			newBreaker := pool.Get()
 			breaker, loaded := breakers.LoadOrStore(host, newBreaker)
 			if loaded {
-				defer pool.Put(newBreaker)
+				pool.Put(newBreaker)
```
kylelemons, line 153: maybe also zero out newBreaker (defensively)?

httpbp/client_middlewares.go
```diff
@@ -150,28 +150,27 @@ func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
...
 			}
 
-			resp, err := breaker.(breakerbp.FailureRatioBreaker).Execute(func() (interface{}, error) {
-				resp, err := next.RoundTrip(req)
+			var resp *http.Response
+			_, err := breaker.(breakerbp.FailureRatioBreaker).Execute(func() (interface{}, error) {
+				var err error
+				resp, err = next.RoundTrip(req)
```
kylelemons, line 159: Don't leak values out of the current scope before the error has been checked.  This has resulted in hours or days of debugging in the past, since a partial value can leak FAR from where it is created before it causes an issue, and the change that triggers it is almost always very far from where it gets detected.

```
r, err := ...
if err != nil { ... }
resp = r
```

kylelemons (changes requested):

httpbp/client_middlewares.go
fishy: everything in the breaker is unexported so there's no way to zero it out. also I'd rather trust sync.Map.LoadOrStore is implemented correctly.

kylelemons: I mean `newBreaker = nil` so that if someone uses it, it will fail instead of causing a race

kylelemons (approved):

httpbp/client_middlewares.go
fishy: that totally defeats the purpose of using a pool.

kylelemons: you're missing my point, I think.

If you put newbreaker into the pool, _and then someone accidentally uses it_, it is a race condition.

Either leave the defer, which doesn't have this issue, or `nil` out the _variable_ so that it can't be misused.

kylelemons (changes requested):

httpbp/client_middlewares.go
fishy: you are missing the point of this code.

here we want to get the breaker out of the map (sync.Map), with the requirement that:
1. two calls with the same key should always get the same breaker out of the map.
2. two calls with different keys should always get the different breakers out of the map.

sync.Map.Load does not work, because it will return nil if this is the first ever call to this key, and then you need to initialize the breaker and store it back to the map, and that's no longer atomic.

this is why to use [LoadOrStore](https://pkg.go.dev/sync#Map.LoadOrStore) here. with LoadOrStore, you always need a new breaker standing by. if this is the first call with this key, the new breaker will be stored into the map and returned back to you. if this is not the first call with this key, the breaker stored by the first call will be returned, and the new breaker is not used.

this is why we can put newbreaker back to the pool if loaded is true.

kylelemons: I appreciate your explanation; this is how I understood the code already, so this doesn't change my stance.  Let me try to be more clear:

I am not concerned about the correctness of the code today, I am concerned about potential bugs that could be introduced in the future.

One of the reasons that `defer` is such a powerful language feature is that it can prevent future changes of the code from introducing subtle bugs.  With `defer` here, the value stored in `newBreaker` is not returned to the pool until the function is returning, and so if someone in 6 months accidentally decides to use it, it's a logical bug, but not a data race.  If you replace the value early, then there is a value stored in the `newBreaker` variable that is **still in scope** here **and in the pool** -- so there is potential for misuse, because it could be returned by the pool to another goroutine.

There are few different classes of fix:
* Remove the _value_ from the scope -- i.e. set `newBreaker` to nil after storing its value in the pool
* Remove the _variable_ from the scope -- i.e. make a local closure or anonymous scope to ensure `newBreaker` is only in scope while doing the (get+loadOrStore)
* Keep the `defer` so that a future misuse of the variable does not introduce a race

fishy: oh I didn't realize that you meant set it to nil *after* stored it back to the pool.

fwiw the first 2 options are legit, the 3rd one (`defer`) is not. we should only store unused breakers back to the pool, never used ones. (breakers are safe for concurrent use, so race condition is not the concern, storing a used one is).

fishy: also fwiw I already wrote a library for the enclosed version: https://pkg.go.dev/go.yhsif.com/defaultdict

fishy:
:haircut:

konradreiche (approved):

kylelemons (approved):

PR #429, opened 2021-10-05
log: Add kv support to ZapWrapper

fishy:
This would help people search for logs from certain upstream
dependencies, for example they can search for "from:thrift" for logs
produced from thrift's TSimpleServer.

This is a breaking change. But the majority of the internal usage are
from yaml config parsing, and there are only 19 direct usages of
log.ZapWrapper from internal code search, and the majority of the 19 are
from my team's monorepo.

### DIFF
--- a/log/wrapper.go
+++ b/log/wrapper.go
@@ -113,10 +113,12 @@ func (w Wrapper) ToThriftLogger() thrift.Logger {
 // - "std": StdWrapper with default stdlib logger
 // (log.New(os.Stderr, "", log.LstdFlags)).
 //
-// - "zap": ZapWrapper on default level (Info).
+// - "zap": ZapWrapper on default level (Info) with no kv pairs.
 //
-// - "zap:level": ZapWrapper with given level, for example "zap:error" means
-// ZapWrapper on Error level.
+// - "zap:level:k1=v1,k2=v2...": ZapWrapper with given level and kv pairs, with
+// the ":k=v..." part being optional. For example "zap:error" means ZapWrapper on
+// Error level with no kv pairs, "zap:info:key1=value1" means ZapWrapper on Info
+// level with "key1":"value1" pair.
 //
 // - "sentry": ErrorWithSentryWrapper.
 //
@@ -127,7 +129,32 @@ func (w *Wrapper) UnmarshalText(text []byte) error {
 	// Special handling for "zap:level" case
 	const zapLevelPrefix = "zap:"
 	if strings.HasPrefix(s, zapLevelPrefix) {
-		*w = ZapWrapper(Level(strings.ToLower(s[len(zapLevelPrefix):])))
+		split := strings.Split(s, ":")
+		if len(split) > 3 {
+			return fmt.Errorf(`log.Wrapper.UnmarshalText: malformed input: too many ":": %q`, s)
+		}
+		var pairs map[string]interface{}
+		if len(split) > 2 {
+			kvs := strings.Split(split[2], ",")
+			pairs = make(map[string]interface{}, len(kvs))
+			for _, kv := range kvs {
+				kv = strings.TrimSpace(kv)
+				index := strings.Index(kv, "=")
+				if index < 0 {
+					return fmt.Errorf(`log.Wrapper.UnmarshalText: malformed input: no "=" in kv pair %q`, kv)
+				}
+				key := kv[:index]
+				val := kv[index+1:]
+				if _, ok := pairs[key]; ok {
+					return fmt.Errorf("log.Wrapper.UnmarshalText: malformed input: key %q appeared at least twice", key)
+				}
+				pairs[key] = val
+			}
+		}
+		*w = ZapWrapper(ZapWrapperArgs{
+			Level:   Level(strings.ToLower(split[1])),
+			KVPairs: pairs,
+		})
 		return nil
 	}
 
@@ -141,7 +168,9 @@ func (w *Wrapper) UnmarshalText(text []byte) error {
 	case "std":
 		*w = StdWrapper(stdlog.New(os.Stderr, "", stdlog.LstdFlags))
 	case "zap":
-		*w = ZapWrapper(Level(""))
+		*w = ZapWrapper(ZapWrapperArgs{
+			Level: Level(""),
+		})
 	case "sentry":
 		*w = ErrorWithSentryWrapper()
 	}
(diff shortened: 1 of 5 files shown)


log/wrapper.go
```diff
@@ -113,10 +113,11 @@ func (w Wrapper) ToThriftLogger() thrift.Logger {
 // - "std": StdWrapper with default stdlib logger
 // (log.New(os.Stderr, "", log.LstdFlags)).
 //
-// - "zap": ZapWrapper on default level (Info).
+// - "zap": ZapWrapper on default level (Info) with no kv pairs.
 //
-// - "zap:level": ZapWrapper with given level, for example "zap:error" means
-// ZapWrapper on Error level.
+// - "zap:level:k1:v1:k2:v2...": ZapWrapper with given level and kv pairs, for
```
kylelemons, line 118: why not

```suggestion
// - "zap:level:k1=v1:k2=v2...": ZapWrapper with given level and kv pairs, for
```

kylelemons: Also, I think I would consider having the keyvals be `,`-separated so that you don't preclude yourself from adding additional (optional) configuration after them in the future.  (you can't add arguments after a variadic)

e.g.

```
zap:level:k1=1,k2=2,k3=3:console=true
```

or something (my imagination about what we might want to configure is failing me right now, but you get the idea)

log/wrapper.go
```diff
@@ -180,29 +195,39 @@ func TestWrapper(tb testing.TB) Wrapper {
...
-			f = logger.Panic
+			f = logger.Panicw
 		case FatalLevel:
-			f = logger.Fatal
+			f = logger.Fatalw
+		}
+		kv := make([]interface{}, 0, len(args.KVPairs)*2)
+		for k, v := range args.KVPairs {
+			kv = append(kv, k, v)
 		}
```
kylelemons, line 229: Do this once in the `ZapWrapper`.

1. Huge performance improvement; now you're not leaking a ton of boxed values and slices
2. No need to worry about concurrent modification of the ZapWrapperArgs map.

kylelemons (changes requested):

log/wrapper.go
kylelemons: Oh, also, I would make it explicit that this is still optional by explaiing it separately from level.

```
zap:level
 - docs
zap:level:k1=1,k2=2
 - docs
```

or w/e.

Also, I think this is a non-breaking change as long as nobody uses :s in their level name, which seems probable?

fishy: I think this approach would still have the problem of what if the user does not want to add any kv pairs, but they do want to add `console=true` (in the hypothesis scenario that we add it in the future), so we need to choose between allowing `zap:level:console=true` and `zap:level::console=true` (e.g they must have an explicit empty "kv block").

konradreiche: What is the semantic of `console=true` here? I think I am lacking context on this.

kylelemons: I was making up a hypothetical future thing we might want to configure, sorry.  It's just meant to be "anything that we might want to add in the future"

log/wrapper.go
```diff
@@ -127,7 +129,28 @@ func (w *Wrapper) UnmarshalText(text []byte) error {
...
+				if index < 0 {
+					return errMalform
+				}
+				pairs[kv[:index]] = kv[index+1:]
+			}
+		}
+		*w = ZapWrapper(ZapWrapperArgs{
+			Level:   Level(strings.ToLower(split[1])),
+			KVPairs: pairs,
+		})
```
kylelemons, line 153: It's not saving you anything by "precomputing" this -- you'll only ever compute it once.  Now, however, you also compute this in the happy path.

This also doesn't say _how_ it's malformed.  Each return path should have a unique error message to make it clear both to the reader and someone who is trying to follow the error through the code what is wrong.

```suggestion
		split := strings.Split(s, ":")
		if len(split) > 3 {
			return fmt.Errorf("log.Wrapper.UnmarshalText: malformed input: too many arguments in %q", s)
		}
		var pairs map[string]interface{}
		if len(split) > 2 {
			kvs := strings.Split(split[2], ",")
			pairs = make(map[string]interface{}, len(kvs))
			for _, kv := range kvs {
				kv = strings.TrimSpace(kv)
				index := strings.Index(kv, "=")
				if index < 0 {
					return fmt.Errorf("log.Wrapper.UnmarshalText: malformed input: keyval pair %q has no = in %q", kv, s)
				}
				pairs[kv[:index]] = kv[index+1:]
			}
		}
		*w = ZapWrapper(ZapWrapperArgs{
			Level:   Level(strings.ToLower(split[1])),
			KVPairs: pairs,
		})
```

log/wrapper.go
```diff
@@ -127,7 +129,28 @@ func (w *Wrapper) UnmarshalText(text []byte) error {
...
+		if len(split) > 2 {
+			kvs := strings.Split(split[2], ",")
+			pairs = make(map[string]interface{}, len(kvs))
+			for _, kv := range kvs {
+				kv = strings.TrimSpace(kv)
+				index := strings.Index(kv, "=")
+				if index < 0 {
+					return errMalform
+				}
+				pairs[kv[:index]] = kv[index+1:]
```
kylelemons, line 147: 🔕 you could also make sure the key is unique.

kylelemons (changes requested):

log/wrapper.go
fishy: I'm not sure what your concern is in:

>It's not saving you anything by "precomputing" this -- you'll only ever compute it once. Now, however, you also compute this in the happy path.

but I addressed your other concerns.

konradreiche (approved):
Great way to work around the existing limitations.

log/wrapper.go
kylelemons: I am not advocating against using `=` I'm advocating against using `:` to separate keyval pairs. 
 `zap:level:k1=1,k2=2:somethingelse` is more readable than `zap:level:k1=1:k2=2!!!somethingelse` or whatever we would have to do to work around the fact that we allowed arbitrary number of `:`-separate keyval pairs when we need to extended this in the future.
 
 And I say "when" because if we are having to extend this once, it's very likely we will need to extend it again.  So we should plan for that.

kylelemons (approved):

log/wrapper.go
kylelemons: You fixed it.  I was saying that you were doing the `Errorf` even in non-error cases

PR #430, opened 2021-10-06
thriftbp: Add slug to client pool error logs

fishy:
This will help services with multiple client pools to identify which
pool produced those error logs.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -431,6 +431,8 @@ func newClientPool(
 	pooledClient := &clientPool{
 		Pool: pool,
 
+		slug: cfg.ServiceSlug,
+
 		poolExhaustedCounter: metricsbp.M.Counter(
 			cfg.ServiceSlug + ".pool-exhausted",
 		).With(tags...),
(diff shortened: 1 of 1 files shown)


thriftbp/client_pool.go
```diff
@@ -538,7 +542,12 @@ func (p *clientPool) pooledCall(ctx context.Context, method string, args, result
 		if shouldCloseConnection(err) {
 			p.poolClosedConnectionsCounter.Add(1)
 			if e := client.Close(); e != nil {
-				log.Errorw("Failed to close client", "origErr", err, "closeErr", e)
+				log.C(ctx).Errorw(
+					"Failed to close client",
+					"pool", p.slug,
+					"origErr", err,
```
konradreiche, line 548: Just `err` to match most common key-value pair for querying on errors unless we don't want to change the previous format.

fishy: `err` doesn't really give enough info on which error it's referring to in this context, because we have 2 errors.

kylelemons: 🔕 IMO since we're returning err, it's not the focus of this log message, so we should use "err" for the "error that happened while [failing to close the client]" (i.e. the one this log is about) and optionally can use "orig_err" for "the error that made me want to close"

kylelemons (approved):

Issue #378, opened 2021-05-24
Add Prometheus metrics endpoint


SuperQ:
In order to provide support for services being monitored by Prometheus we need to add the ability to configure a Prometheus `/metrics` endpoint.

The https://github.com/prometheus/client_golang library is recommended.

This would only be useful for registered metrics "tagged" mode.

PR #431, opened 2021-10-12
Add prometheus metrics

JessicaGreben:
This PR is for issue #378 and adds the beginning scaffolding to export metrics for Prometheus.  Ultimately we will fully migrate over to Prometheus instead of using the current Statsd metrics, but in the meantime we will need to support both.

The baseplate spec specifies that an `active_requests` metrics must be (ref: [spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/metrics-tagged.md#runtime-metrics)) emitted so this PR adds that as the first custom metric.

The `/metrics` endpoint has been added to the golang baseplate code via the `baseplate-cookiecutter` repo in this [PR](https://github.snooguts.net/reddit/baseplate-cookiecutter/pull/247). 

The next work that will be done in a different PR is to add the scaffolding that allows applications to generate custom metrics for their own purposes.

### DIFF
--- a/metricsbp/baseplate_hooks.go
+++ b/metricsbp/baseplate_hooks.go
@@ -129,13 +129,13 @@ type countActiveRequestsHook struct {
 
 func (h countActiveRequestsHook) OnPostStart(_ *tracing.Span) error {
 	h.metrics.incActiveRequests()
-	h.prometheusMetrics.incActiveRequests()
+	h.prometheusMetrics.ActiveRequests.Inc()
 	return nil
 }
 
 func (h countActiveRequestsHook) OnPreStop(_ *tracing.Span, _ error) error {
 	h.metrics.decActiveRequests()
-	h.prometheusMetrics.decActiveRequests()
+	h.prometheusMetrics.ActiveRequests.Dec()
 	return nil
 }
 
--- a/metricsbp/prometheus.go
+++ b/metricsbp/prometheus.go
@@ -4,26 +4,21 @@ import (
 	"context"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
-var activeRequests = prometheus.NewGauge(prometheus.GaugeOpts{
+var activeRequests = promauto.NewGauge(prometheus.GaugeOpts{
 	Name: "active_requests",
 	Help: "The number of requests being handled by the service.",
 })
 
 type PrometheusMetrics struct {
+	ActiveRequests prometheus.Gauge
 }
 
 // NewPrometheusMetrics creates a new PrometheusMetrics object that is used to register new metrics for monitoring.
 func NewPrometheusMetrics(ctx context.Context, cfg Config) *PrometheusMetrics {
-	prometheus.MustRegister(activeRequests)
-	return &PrometheusMetrics{}
-}
-
-func (pm *PrometheusMetrics) incActiveRequests() {
-	activeRequests.Inc()
-}
-
-func (pm *PrometheusMetrics) decActiveRequests() {
-	activeRequests.Dec()
+	return &PrometheusMetrics{
+		ActiveRequests: activeRequests,
+	}
 }


metricsbp/prometheus.go
```diff
@@ -0,0 +1,24 @@
...
+	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var activeRequests = promauto.NewGauge(prometheus.GaugeOpts{
+	Name: "active_requests",
+	Help: "The number of requests being handled by the service.",
+})
```
fishy, line 13: in statsd this is a "runtime gauge" instead of a plain gauge. a "runtime gauge" comes with two tags, [the hostname and the pid, and auto add "runtime." prefix](https://github.com/reddit/baseplate.go/blob/df2b3cb40d2bf1f2e9677300b3e5e265c29ba3ba/metricsbp/runtime_stats.go#L131-L153).

then we have some special handling on telegraf on everything under `runtime.` prefix: we strip the 2 special tags, and calculate the average/max/min/etc. from all the pods reporting this gauge. this is kind of like treating it as a histogram instead of a gauge.

we probably should have an internal spec discussion on how do we want to handle runtime gauges on prometheus first.

metricsbp/config.go
```diff
@@ -63,7 +63,8 @@ type Config struct {
 // with the global tracing hook registry.
 func InitFromConfig(ctx context.Context, cfg Config) io.Closer {
 	M = NewStatsd(ctx, cfg)
-	tracing.RegisterCreateServerSpanHooks(CreateServerSpanHook{Metrics: M})
+	pm := NewPrometheusMetrics(ctx, cfg)
+	tracing.RegisterCreateServerSpanHooks(CreateServerSpanHook{Metrics: M, PrometheusMetrics: pm})
```
fishy, line 67: as a style nit, I feel that the only appropriate case of writing a whole struct in one line is when you only fill zero or one of its fields. anything >1 should be write in one-per-line instead:
```suggestion
	tracing.RegisterCreateServerSpanHooks(CreateServerSpanHook{
		Metrics: M,
		PrometheusMetrics: pm,
	})
```

fishy (changes requested):

metricsbp/prometheus.go
bjk-reddit: This doesn't really exist the same way in Prometheus. This is because Prometheus always includes the instance of the monitored target in every metric. For Go, there is only ever one process, so effectively every metric includes those tags.

metricsbp/prometheus.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var activeRequests = promauto.NewGauge(prometheus.GaugeOpts{
+	Name: "active_requests",
```
bjk-reddit, line 11: Is this specific to a protocol? Typically for this metric we would prefix it with the protocol.

This can be done with the `Namespace`. Something like this:

```
prometheus.GaugeOpts{
	Namespace: "http",
	Name: "active_requests",
...
}
```

metricsbp/prometheus.go
bjk-reddit: That part of the spec needs to be corrected for Prometheus.

metricsbp/baseplate_hooks.go
```diff
@@ -35,21 +38,23 @@ func (h CreateServerSpanHook) OnCreateServerSpan(span *tracing.Span) error {
 // ends, with success=True/False tags for the counter based on whether an error
 // was passed to `span.End` or not.
 type spanHook struct {
-	metrics *Statsd
+	metrics           *Statsd
+	prometheusMetrics *PrometheusMetrics
```
kylelemons, line 42: From what I've seen, the "best practice" for injecting prometheus metrics (especially when there's just one, like "active_requests" is to pass in the `prometheus.Counter` -- that also could allow you to do the "active_requests" per-transport here potentially, using `prometheus.CounterVec.With` to pre-populate the label.

Leaving this out to just do in the per-transport section also sounds fine if you prefer.

metricsbp/prometheus.go
JessicaGreben: sounds like the best thing to do is move this to the per transport server middleware. will make that change.

metricsbp/baseplate_hooks.go
JessicaGreben: thanks for the details. sounds like i will close this PR and add active requests to the per transport PRs.

JessicaGreben:
Closing this to add active_request to the per transport server middleware instead.

PR #433, opened 2021-10-14
Add prometheus metrics for thriftbp server.

JessicaGreben:
This PR adds Prometheus metrics for Thriftbp server based on [preference service](https://github.snooguts.net/reddit/reddit-service-preference-go/blob/master/service/service_rpc_middleware.go) and the baseplate.spec [PR](https://github.snooguts.net/reddit/baseplate.spec/pull/39) and closed PR #431.

The thrift client metrics will be added in a different PR once this is merged.

Background:

- this PR for thrift server metric with some testing helpers (PR #443).
- another PR for adding thrift client metrics
- another PR that adds spec-validation tests
- a PR that adds http client and server prometheus metrics
- a PR that adds grpc client and server prometheus metrics


go.mod
```diff
@@ -61,6 +61,14 @@ require (
 	golang.org/x/text v0.3.6 // indirect
 	golang.org/x/tools v0.1.3-0.20210608163600-9ed039809d4c // indirect
 	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 // indirect
-	google.golang.org/protobuf v1.25.0 // indirect
+	google.golang.org/protobuf v1.26.0-rc.1 // indirect
 	honnef.co/go/tools v0.2.0 // indirect
 )
+
+require (
```
fishy, line 68: you are likely not running go 1.17.2, please upgrade your go to 1.17.2, which fixed this bug.

thriftbp/server_middlewares.go
```diff
@@ -380,3 +383,59 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
...
+				"thrift_baseplate_status":      baseplateStatus,
+				"thrift_baseplate_status_code": baseplateStatusCode,
+			}
+			activeRequests.With(activeRequestLabels).Inc()
+
+			defer func() {
+				successLabel := "true"
+				if !success {
+					successLabel = "false"
+				}
```
fishy, line 407: this is actually not part of the definition of the success label, we can remove these 4 lines (treat it as true as long as err is nil).

either use `successLabel := strconv.FormatBool(suppressor.Suppress(err) == nil)` (if we still allow error suppressing), or just `successLabel := strconv.FormatBool(err == nil)`.

thriftbp/server_middlewares.go
```diff
@@ -380,3 +383,59 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
...
+
+				if err != nil {
+					successLabel = "false"
+					exceptionTypeLabel = strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
+
+					var bpErr baseplateError
+					if errors.As(err, &bpErr) {
+						code := bpErr.GetCode()
+						baseplateStatusCode = fmt.Sprintf("%d", code)
```
fishy, line 416: nit: prefer `strconv.FormatInt` over `fmt.Sprintf`.

thriftbp/server_middlewares.go
```diff
@@ -380,3 +383,59 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
...
+			var successLabel, exceptionTypeLabel, baseplateStatusCode, baseplateStatus string
+			activeRequestLabels := prometheus.Labels{
+				"thrift_service":               service,
+				"thrift_method":                method,
+				"thrift_success":               successLabel,
+				"thrift_exception_type":        exceptionTypeLabel,
+				"thrift_baseplate_status":      baseplateStatus,
+				"thrift_baseplate_status_code": baseplateStatusCode,
+			}
+			activeRequests.With(activeRequestLabels).Inc()
```
fishy, line 401: do we really need all those labels for active requests? isn't it that when we do the `Inc` all those labels are empty strings anyways?

thriftbp/server_middlewares.go
```diff
@@ -380,3 +383,59 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
 		},
 	}
 }
+
+// PrometheusMetrics tracks Prometheus metrics specific to the Thrift service.
+func PrometheusMetrics(service string) thrift.ProcessorMiddleware {
```
fishy, line 388: nit: we might want to rename `service` arg to `slug` to be more consistent with other code.

thriftbp/server_middlewares_test.go
```diff
@@ -332,3 +333,76 @@ func TestPanicMiddleware(t *testing.T) {
 		}
 	})
 }
+
+func TestPrometheusMetricsMiddleware(t *testing.T) {
+	t.Run("bp error", func(t *testing.T) {
```
fishy, line 338: nit: you can convert the 4 subtests into a table test. the fields of the table are:
```
struct {
  err error
  next thrift.WrappedTProcessorFunction
  expectedOK bool
  expectedErr error
}
```
but you are not really checking any actual prometheus metrics in the tests. you are just checking that the middleware does not change ok and error values from next call.

thriftbp/server_middlewares.go
fishy: I believe in the current state `Dec` will use the correct labels, because we saved the whole labels map and they are strings not pointers, but yes this is very error-prone and could easily go wrong in the future.

thriftbp/server_middlewares.go
JessicaGreben: @fishy How do I know if we still allow error suppressing?

go.mod
fishy: yea I found out that this bug is still not fixed in go 1.17.2 :sigh:

you can just manually remove line 66-68 (so merge the two indirect require blocks), and rerun `go mod tidy` so it sorts that block correctly for you.

thriftbp/server_middlewares.go
```diff
@@ -380,3 +383,59 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
...
+			activeRequests.With(activeRequestLabels).Inc()
+
+			defer func() {
+				successLabel := "true"
+				if !success {
+					successLabel = "false"
+				}
+
+				if err != nil {
+					successLabel = "false"
```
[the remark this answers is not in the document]
JessicaGreben: so should I set success just based on error then instead of the "success"?

thriftbp/server_middlewares.go
```diff
@@ -380,3 +383,50 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
...
+				"thrift_service": serviceSlug,
+				"thrift_method":  method,
+			}
+			activeRequests.With(activeRequestLabels).Inc()
+
+			defer func() {
+				var exceptionTypeLabel, baseplateStatusCode, baseplateStatus string
+				successLabel := strconv.FormatBool(err == nil)
+				if err != nil {
+					exceptionTypeLabel = strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
```
fishy, line 402: 🔕 I _think_ the trimming of `*` could be a premature optimization, as we rarely use an error type both as reference and as pointer, but I don't feel strongly either way

(also if `*` in labels will cause troubles in prometheus then yea we absolutely need to do that)

thriftbp/prometheus.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+var (
+	thriftLabels = []string{
+		"thrift_service",
+		"thrift_method",
+		"thrift_success",
+		"thrift_exception_type",
+		"thrift_baseplate_status",
+		"thrift_baseplate_status_code",
```
fishy, line 15: nit: the same string literals are used both here and in `PrometheusMetricMiddleware`, although this is unlikely to happen but it could still potentially happen that in the future someone changed the string in one place and forgot to change it in the other place and that will cause issues, so I'd prefer to define them as constants to avoid that issue.

thriftbp/server_middlewares_test.go
```diff
@@ -332,3 +333,36 @@ func TestPanicMiddleware(t *testing.T) {
...
+		wantOK  bool
+	}{
+		{"bp error", thrift.NewTProtocolExceptionWithType(thrift.PROTOCOL_ERROR, thriftbp.WrapBaseplateError(errors.New("test"))), false},
+		{"application error", thrift.NewTApplicationException(thrift.UNKNOWN_METHOD, "unknown err mdsg"), false},
+		{"compile error", baseplate.NewError(), false},
+		{"success", nil, true},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
```
fishy, line 350: 🔕 we are still not checking whether we emitted the correct prometheus metrics in the test, is that something doable in prometheus go library?

thriftbp/prometheus.go
JessicaGreben: good idea. let me fix that.

thriftbp/server_middlewares.go
fishy: regexp is a good point, it's fine to keep it

thriftbp/server_middlewares.go
```diff
@@ -380,3 +383,50 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
...
 	}
 }
+
+// PrometheusMetricMiddleware returns middleware to track Prometheus metrics specific to the Thrift service.
+func PrometheusMetricMiddleware(serviceSlug string) thrift.ProcessorMiddleware {
+	return func(method string, next thrift.TProcessorFunction) thrift.TProcessorFunction {
+		process := func(ctx context.Context, seqID int32, in, out thrift.TProtocol) (success bool, err thrift.TException) {
+			start := time.Now()
+			activeRequestLabels := prometheus.Labels{
+				"thrift_service": serviceSlug,
```
kylelemons, line 393: @fishy Is there any way to extract the notion of the "thrift service" from the notion of the "server"?  For example, for awhile we were serving the same thrift service from both pref-go and pref-py -- maybe we could just use the struct name for the thrift service name?

thriftbp/server_middlewares_test.go
kylelemons: You can start the test with `m.Reset()` and then `testutil.GatherAndCompare` or ToFloat64

https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil#GatherAndCompare

I think we'll be wanting some of our own, better testing helpers over time... we had some nice ones before that could do things like measure deltas for a metric + labels across the test case, but it's probably overkill for this.

thriftbp/prometheus.go
```diff
@@ -0,0 +1,38 @@
...
+		Name:    "thrift_server_handling_seconds",
+		Help:    "RPC latencies",
+		Buckets: prometheus.ExponentialBuckets(0.0001, 1.5, 26), // 100us ~ 2.5s
+	}, thriftLabels)
+
+	rpcStatusCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "thrift_server_handled_total",
+		Help: "Total RPC request count",
+	}, thriftLabels)
+
```
kylelemons, line 28: 🔕 It might be helpful to separate these into two var blocks as a way of scoping the labels and the metrics that share them:

```suggestion
)

var (
```

kylelemons (approved):

thriftbp/server_middlewares.go
fishy: we don't have a canonical way to do that right now, and I don't think just use the struct name is a good solution to that, but it could potentially be a good default fallback (e.g. when the config didn't specify the name of the service, we extract it from the struct name).

thriftbp/prometheus.go
```diff
@@ -0,0 +1,47 @@
...
+	serviceLabel             = "thrift_service"
+	methodLabel              = "thrift_method"
+	successLabel             = "thrift_success"
+	exceptionLabel           = "thrift_exception_type"
+	baseplateStatusLabel     = "thrift_baseplate_status"
+	baseplateStatusCodeLabel = "thrift_baseplate_status_code"
+)
+
+var (
+	ThriftLabels = []string{
```
fishy, line 18: do we really need to export these variables? what are their use cases outside of this package? if you only need to access them in the test, you can convert the test to internal test instead (move the test for `PrometheusMetricMiddleware` to a new `_test.go` file and don't add `_test` to the package name of the test file).

thriftbp/server_middlewares_test.go
JessicaGreben: @kylelemons I added some testutil checks, I used `collectAndCount` and `ToFloat64`, but I was having trouble `ToFloat64` for the histogram type. Do you know of a way to test the values for the historgram?

thriftbp/prometheus.go
JessicaGreben: yes they are only exported for tests. I will move the test file to fix.

pacejackson: we might want to actually export the `const`-s with documentation comments so they show up in the documentation?

go.mod
```diff
@@ -45,9 +46,13 @@ require (
 	github.com/jcmturner/gokrb5/v8 v8.4.2 // indirect
 	github.com/jcmturner/rpc/v2 v2.0.3 // indirect
 	github.com/klauspost/compress v1.12.2 // indirect
-	github.com/kr/logfmt v0.0.0-20140226030751-b84e30acd515 // indirect
+	github.com/matttproud/golang_protobuf_extensions v1.0.1 // indirect
 	github.com/mediocregopher/radix.v2 v0.0.0-20181115013041-b67df6e626f9 // indirect
 	github.com/pierrec/lz4 v2.6.0+incompatible // indirect
+	github.com/prometheus/client_golang v1.11.0
```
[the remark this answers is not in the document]
fishy: as much as I hate babysitting this file, let's move this line to the first block manually (and rerun `go mod tidy` to sort them properly)

thriftbp/server_middlewares.go
```diff
@@ -380,3 +383,64 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
 		},
 	}
 }
+
+// PrometheusMetricMiddleware returns middleware to track Prometheus metrics specific to the Thrift service.
+// It emits the following prometheus metrics:
```
fishy, line 388: nit: add a blank line (`//`) between these 2 lines, and do some line-wrapping on the long lines in the comment.

thriftbp/prometheus_test.go
```diff
@@ -0,0 +1,97 @@
...
+	testCases := []struct {
+		name          string
+		wantErr       thrift.TException
+		wantOK        bool
+		baseplateCode string
+	}{
+		{"bp error", thrift.NewTProtocolExceptionWithType(thrift.PROTOCOL_ERROR, WrapBaseplateError(errors.New("test"))), false, ""},
+		{"application error", thrift.NewTApplicationException(thrift.UNKNOWN_METHOD, "unknown err msg"), false, ""},
+		{"compile error", baseplate.NewError(), false, "0"},
+		{"success", nil, true, ""},
```
fishy, line 26: also style nit: use [one-per-line style](https://github.com/reddit/baseplate.go/blob/master/Style.md#one-per-line-style) and don't use unkeyed struct constructors.

JessicaGreben: i would prefer to leave as it if thats ok.

thriftbp/prometheus.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const (
+	serviceLabel             = "thrift_service"
+	methodLabel              = "thrift_method"
+	successLabel             = "thrift_success"
+	exceptionLabel           = "thrift_exception_type"
+	baseplateStatusLabel     = "thrift_baseplate_status"
+	baseplateStatusCodeLabel = "thrift_baseplate_status_code"
```
bjk-reddit, line 14: There is no need to prefix labels like this. The metric is `thrift`, therefore any label naming implies thrift.

Department of redundancy department. 😄 

Although, `service` is possibly something we should reserve for outside of application naming.

JessicaGreben: these label names are defined in the [baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#thrift), so if they need to change I think they should be updated there. Thoughts?

prometheusbp/metrics.go
```diff
@@ -0,0 +1,7 @@
+package prometheusbp
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var DefaultBuckets = prometheus.ExponentialBuckets(0.0001, 1.5, 26) // 100us ~ 2.5s
```
JessicaGreben, line 7: @bjk-reddit do you think these bucket values are ok?

thriftbp/prometheus.go
bjk-reddit: Yes, I think we should fix the spec. I'll send out a PR.

bjk-reddit: Nevermind, I'm not going to argue this anymore.

prometheusbp/metrics.go
bjk-reddit: I'm wondering if this is a bit too granular to have as a default. Looking at one of our first services, this generates 19,000 metrics due to the number of methods/pods/etc.

What we we start out with something reduced to 16 buckets, with a range of 100us to ~15s with this:

```suggestion
var DefaultBuckets = prometheus.ExponentialBuckets(0.0001, 2.5, 14) // 100us ~ 14.9s
```

This will give us reasonable granularity from very fast responses to our front-end timeout. This will be more useful for services that depends on slow backends.

fishy: 🔕 @bjk-reddit: will 100us be small enough? it probably is for rpc calls, but we might be using some local spans in the future and those can be much faster. not sure if we want the same default buckets to work for all.

bjk-reddit: @fishy I think 100us is probably fine for a default lower bound. We're mostly concerned about things being slower, so counting things in buckets when it's longer is the more important part.  If you think about it from an SLO perspective (which is what the metrics are about), knowing the difference between 50us and 100us is "more than fast enough" from a full RPC reply.

Here's an example of the data from the new preference service.

Screen Shot 2021-11-03 at 17 19 38
 
There are a few requests under 100us, but only sub 10 RPS compared to the 100,000 RPS in the 300us-2ms range.

fishy: I'm mainly talking about some of the local spans that the majority (or all) of the data points are below 100us. Those usually don't have an SLA/SLO, but developers still care about latency data.

bjk-reddit: Those are probably not something we would include in metrics, only tracing spans. The main goal of metrics and these buckets are for real-time monitoring for our SLO/SLI use cases.

prometheusbp/testutil.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+)
+
+// PrometheusMetricTest stores information about a metric to use for testing.
+type PrometheusMetricTest struct {
+	t           *testing.T
```
fishy, line 12: in general, prefer to use `tb testing.TB` so that this struct can also be used in benchmark tests (although it probably doesn't make sense to use this in benchmark tests)

fishy (approved):
please fix the things staticcheck complains about, otherwise lgtm (as long as ben is satisfied on the prometheus buckets/etc. :) )

prometheusbp/testutil.go
JessicaGreben: @fishy  thats a good idea. How about for now, leave as it, but if this ends up being used in benchmark tests int he future, we can change it to `tb testing.TB` then. Does that sound ok?

fishy: that's fine, but this is a trivial change so I don't understand the push back.

JessicaGreben: oops sorry wasn't trying to push back at all. im totally happy to make the change. i think i misunderstood your comment... I didn't know if you wanted me to make the change or not because of this part "(although it probably doesn't make sense to use this in benchmark tests)".  But I don't feel strongly either way and I'm totally happy to change it.

fishy (approved):

PR #438, opened 2021-10-28
thriftbp: Add background refresh for ttlClients

fishy:
For every ttlClient that has a ttl, add a background goroutine to
refresh the connection when ttl hits.

This adds one goroutine per connection in the pool, until the connection
fails to refresh.

Both IsOpen and Call are protected by a locked state now. When ttlClient
is used, we always call IsOpen first then Call, and in theory these 2
calls might not be using the same connection (e.g. IsOpen is called
right before the connection expires), but even in that scenario, Call
would use a freshly refreshed connection, so things should still work.

### DIFF
--- a/thriftbp/ttl_client.go
+++ b/thriftbp/ttl_client.go
@@ -1,12 +1,17 @@
 package thriftbp
 
 import (
+	"context"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/go-kit/kit/metrics"
 
+	"github.com/reddit/baseplate.go/metricsbp"
 	"github.com/reddit/baseplate.go/randbp"
 )
+
+type ttlClientGenerator func() (thrift.TClient, thrift.TTransport, error)
 
 // DefaultMaxConnectionAge is the default max age for a Thrift client connection.
 const DefaultMaxConnectionAge = time.Minute * 5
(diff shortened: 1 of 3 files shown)


fishy:
:haircut: ready for review now.

thriftbp/ttl_client.go
```diff
@@ -19,26 +23,48 @@ var _ Client = (*ttlClient)(nil)
 
 // ttlClient is a Client implementation wrapping thrift's TTransport with a TTL.
 type ttlClient struct {
-	thrift.TClient
-
-	transport thrift.TTransport
+	// configs needed for refresh to work
+	generator ttlClientGenerator
+	ttl       time.Duration
+	jitter    float64
```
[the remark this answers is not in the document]
kylelemons: Saving the jittered TTL seems right to me yeah.

thriftbp/ttl_client.go
```diff
@@ -19,26 +25,50 @@ var _ Client = (*ttlClient)(nil)
...
+	if c.timer != nil {
+		c.timer.Stop()
+	}
 	return c.transport.Close()
 }
 
+func (c *ttlClient) Call(ctx context.Context, method string, args, result thrift.TStruct) (thrift.ResponseMeta, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	return c.client.Call(ctx, method, args, result)
```
kylelemons, line 60: it might be nice to check closed here too, to return a nice error message.

🔕 Also, it would be _great_ to annotate any error messages that get returned from this with the slug, since thrift doesn't do that it just gives bare IPs...

thriftbp/ttl_client.go
```diff
@@ -19,26 +25,50 @@ var _ Client = (*ttlClient)(nil)
...
-// If TTL has passed, it closes the underlying TTransport and returns false.
-// Otherwise it just calls the underlying TTransport's IsOpen function.
+// It checks underlying TTransport's IsOpen first,
+// if that returns false, it returns false.
+// Otherwise it checks TTL,
+// returns false if TTL has passed and also close the underlying TTransport.
 func (c *ttlClient) IsOpen() bool {
+	c.lock.Lock()
+	defer c.lock.Unlock()
 	if !c.transport.IsOpen() {
```
kylelemons, line 72: `return !c.closed && c.transport.IsOpen()`?

thriftbp/ttl_client.go
```diff
@@ -49,18 +79,70 @@ func (c *ttlClient) IsOpen() bool {
...
+		// close the old transport before replacing it, to avoid connection leaks.
+		c.transport.Close()
+	}
+	c.transport = transport
+	c.replaceCounter.With("success", "True").Add(1)
+}
+
+// renew updates expiration and timer base on the given timestamp.
+//
+// caller of this function MUST already held the lock.
```
kylelemons, line 114: 🔕 🔕 (very optional)

This "must hold the lock" is when I pivot to using a 1-buffered channel as a mutex with all of the mutex-guarded state inside it.

Then you attach "must hold the lock" methods on that state struct.  Now it is much harder to call it wrong.

kylelemons (approved):

thriftbp/ttl_client.go
fishy: do you have an example? I know how to use 1-buffer channel to replace a lock, but I'm not sure how to use it in a function that you are supposed to be already holding the lock.

thriftbp/ttl_client.go
fishy: also we can't return directly here, that doesn't handle the case that the connection is still good but the ttl was hit (`time.AfterFunc` could be off by a short period of time, or the case of background refresh failed but the original connection was still good).

konradreiche (approved):

thriftbp/ttl_client.go
fishy: I guess https://github.com/reddit/baseplate.go/pull/438/commits/0cecf4d4f707b426e630011e802159b5e1047e3a is what you meant?

thriftbp/ttl_client.go
```diff
@@ -17,50 +22,135 @@ const DefaultMaxConnectionAgeJitter = 0.1
...
+	if data.transport != nil {
+		// close the old transport before replacing it, to avoid connection leaks.
+		data.transport.Close()
+	}
+	data.transport = transport
+	c.replaceCounter.With("success", "True").Add(1)
+}
+
+// renew updates expiration and timer in data base on the given timestamp.
+func (c *ttlClient) renew(now time.Time, data *ttlClientLockedData) {
```
kylelemons, line 122: I typically reverse this -- `*ttlClientLockedData` is the receiver, and then it can take in a *ttlClient param

thriftbp/ttl_client.go
```diff
@@ -17,50 +22,135 @@ const DefaultMaxConnectionAgeJitter = 0.1
 
 var _ Client = (*ttlClient)(nil)
 
+type ttlClientLockedData struct {
```
kylelemons, line 25: 🔕 (nit, naming) I typically call it "fooState" -- so in this case "ttlClientState"

and then the field I typically name "state"

but the names you have are perfectly readable, so 🤷

thriftbp/ttl_client.go
```diff
@@ -17,50 +22,135 @@ const DefaultMaxConnectionAgeJitter = 0.1
...
+	// configs needed for refresh to work
+	generator ttlClientGenerator
+	ttl       time.Duration
 
-	transport thrift.TTransport
+	replaceCounter metrics.Counter
 
-	// if expiration is zero, then the client will be kept open indefinetly.
-	expiration time.Time
+	// data guarded by lock (buffer-1 channel)
```
kylelemons, line 41: 🔕 I typically put this as an EOL comment on the field

thriftbp/ttl_client.go
```diff
@@ -17,50 +22,135 @@ const DefaultMaxConnectionAgeJitter = 0.1
...
 
 // Close implements Client interface.
 //
 // It calls underlying TTransport's Close function.
 func (c *ttlClient) Close() error {
-	return c.transport.Close()
+	data := <-c.locked
+	defer func() {
+		c.locked <- data
+	}()
```
kylelemons, line 52: 🔕 gofmt will let you have this as a one-liner

kylelemons (approved):

thriftbp/ttl_client.go
fishy: no it doesn't :) (we have gofmt in CI and it didn't complain)

PR #442, opened 2021-11-09
Add prometheus metrics for Thrift client

JessicaGreben:
This PR adds Prometheus metrics for Thriftbp client based on the [baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#thrift).

Background:

- this PR for thrift server metric with some testing helpers (PR [#433](https://github.com/reddit/baseplate.go/pull/433)).
- another PR for adding thrift client metrics (PR #[442](https://github.com/reddit/baseplate.go/pull/442) )
- another PR that adds spec-validation tests
- a PR that adds http client and server prometheus metrics
- a PR that adds grpc client and server prometheus metrics

### DIFF
--- a/thriftbp/client_middlewares.go
+++ b/thriftbp/client_middlewares.go
@@ -342,21 +342,21 @@ var (
 //     as a string if present (e.g. 404), or the empty string
 //   - thrift_baseplate_status_code: the human-readable status code, e.g.
 //     NOT_FOUND, or the empty string
-func PrometheusClientMiddleware(serviceSlug, serverSlug string) thrift.ClientMiddleware {
+func PrometheusClientMiddleware(localServiceSlug, remoteServiceSlug string) thrift.ClientMiddleware {
 	return func(next thrift.TClient) thrift.TClient {
 		return thrift.WrappedTClient{
 			Wrapped: func(ctx context.Context, method string, args, result thrift.TStruct) (_ thrift.ResponseMeta, err error) {
 				start := time.Now()
 				activeRequestLabels := prometheus.Labels{
-					serviceLabel: serviceSlug,
-					methodLabel:  method,
-					slugLabel:    serverSlug,
+					localServiceLabel:      localServiceSlug,
+					methodLabel:            method,
+					remoteServiceSlugLabel: remoteServiceSlug,
 				}
 				clientActiveRequests.With(activeRequestLabels).Inc()
 
 				defer func() {
 					var exceptionTypeLabel, baseplateStatusCode, baseplateStatus string
-					successLbl := strconv.FormatBool(err == nil)
+					success := strconv.FormatBool(err == nil)
 					if err != nil {
 						exceptionTypeLabel = strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
 
@@ -364,24 +364,30 @@ func PrometheusClientMiddleware(serviceSlug, serverSlug string) thrift.ClientMid
 						if errors.As(err, &bpErr) {
 							code := bpErr.GetCode()
 							baseplateStatusCode = strconv.FormatInt(int64(code), 10)
-							baseplateStatus := baseplate.ErrorCode(code).String()
-							if baseplateStatus == "<UNSET>" {
-								baseplateStatus = ""
+							if status := baseplate.ErrorCode(code).String(); status != "<UNSET>" {
+								baseplateStatus = status
 							}
 						}
 					}
 
-					labels := prometheus.Labels{
-						serviceLabel: serviceSlug,
-						methodLabel:  method,
-						successLabel: successLbl,
-						slugLabel:    serverSlug,
+					latencyLabels := prometheus.Labels{
+						localServiceLabel:      localServiceSlug,
+						methodLabel:            method,
+						successLabel:           success,
+						remoteServiceSlugLabel: remoteServiceSlug,
 					}
-					clientLatencyDistribution.With(labels).Observe(time.Since(start).Seconds())
-					labels[baseplateStatusCodeLabel] = baseplateStatusCode
-					labels[exceptionLabel] = exceptionTypeLabel
-					labels[baseplateStatusLabel] = baseplateStatus
-					clientRPCRequestCounter.With(labels).Inc()
+					clientLatencyDistribution.With(latencyLabels).Observe(time.Since(start).Seconds())
+
+					rpcCountLabels := prometheus.Labels{
+						localServiceLabel:        localServiceSlug,
+						methodLabel:              method,
+						successLabel:             success,
+						exceptionLabel:           exceptionTypeLabel,
+						baseplateStatusCodeLabel: baseplateStatusCode,
+						baseplateStatusLabel:     baseplateStatus,
+						remoteServiceSlugLabel:   remoteServiceSlug,
+					}
+					clientRPCRequestCounter.With(rpcCountLabels).Inc()
 					clientActiveRequests.With(activeRequestLabels).Dec()
 				}()
 
--- a/thriftbp/client_middlewares.go
+++ b/thriftbp/client_middlewares.go
@@ -2,16 +2,21 @@ package thriftbp
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/avast/retry-go"
 	"github.com/opentracing/opentracing-go"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/reddit/baseplate.go/breakerbp"
 	"github.com/reddit/baseplate.go/ecinterface"
 	"github.com/reddit/baseplate.go/errorsbp"
+	"github.com/reddit/baseplate.go/internal/gen-go/reddit/baseplate"
 	"github.com/reddit/baseplate.go/retrybp"
 	"github.com/reddit/baseplate.go/tracing"
 	"github.com/reddit/baseplate.go/transport"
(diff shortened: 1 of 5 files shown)


thriftbp/client_middlewares.go
```diff
@@ -313,3 +318,75 @@ var (
...
+//
+// * thrift_client_requests_total counter with all labels above plus:
+//
+//   - thrift_exception_type: the human-readable exception type, e.g.
+//     baseplate.Error, etc
+//   - thrift_baseplate_status: the numeric status code from a baseplate.Error
+//     as a string if present (e.g. 404), or the empty string
+//   - thrift_baseplate_status_code: the human-readable status code, e.g.
+//     NOT_FOUND, or the empty string
+func PrometheusClientMiddleware(serviceSlug, serverSlug string) thrift.ClientMiddleware {
```
kylelemons, line 345: `localServiceSlug, remoteServiceSlug`? (it took me awhile to work out that I think this is what they mean, so making the names match might help others)

Also "service" vs "server" is a bit confusing, so unless you have an overriding reason I would use "Service" for both

thriftbp/client_middlewares.go
```diff
@@ -313,3 +318,75 @@ var (
...
+				activeRequestLabels := prometheus.Labels{
+					serviceLabel: serviceSlug,
+					methodLabel:  method,
+					slugLabel:    serverSlug,
+				}
+				clientActiveRequests.With(activeRequestLabels).Inc()
+
+				defer func() {
+					var exceptionTypeLabel, baseplateStatusCode, baseplateStatus string
+					successLbl := strconv.FormatBool(err == nil)
```
kylelemons, line 359: `Lbl`: don't drop letters to save typing.  "Label" is just fine

thriftbp/client_middlewares.go
```diff
@@ -313,3 +318,75 @@ var (
...
+						exceptionTypeLabel = strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
+
+						var bpErr baseplateError
+						if errors.As(err, &bpErr) {
+							code := bpErr.GetCode()
+							baseplateStatusCode = strconv.FormatInt(int64(code), 10)
+							baseplateStatus := baseplate.ErrorCode(code).String()
+							if baseplateStatus == "<UNSET>" {
+								baseplateStatus = ""
+							}
```
kylelemons, line 370: (potential bug)

This doesn't set "baseplateStatus" from line 358 because it's shadowed on line 367.  One way to write this would be

```suggestion
							if str := baseplate.ErrorCode(code).String(); str != "<UNSET>" {
								baseplateStatus = str
							}
```

It might be helpful to make a function that encapsulates this logic to make it easier to test

thriftbp/client_middlewares.go
```diff
@@ -313,3 +318,75 @@ var (
...
+					labels := prometheus.Labels{
+						serviceLabel: serviceSlug,
+						methodLabel:  method,
+						successLabel: successLbl,
+						slugLabel:    serverSlug,
+					}
+					clientLatencyDistribution.With(labels).Observe(time.Since(start).Seconds())
+					labels[baseplateStatusCodeLabel] = baseplateStatusCode
+					labels[exceptionLabel] = exceptionTypeLabel
+					labels[baseplateStatusLabel] = baseplateStatus
```
kylelemons, line 383: 🔕 I don't entirely like that the labels above get a literal and these don't... It increases the "cognitive burden" of the code because you need to track the mutations in your head to figure out the keys that are set and whether anything is overridden.

There's also a chance that the strings have duplicate values -- the literal prevents this bug at compile time, but mutating the map would not.

I would probably just make two maps (names are just for example):

```
histogramLabels := prometheus.Labels{...}
detailLabels := prometheus.Labels{...}
```

thriftbp/prometheus_test.go
```diff
@@ -100,3 +100,81 @@ func TestPrometheusMetricsMiddleware(t *testing.T) {
...
+			}
+			if !tt.wantErr && err != nil {
+				t.Errorf("got err: %v", err)
+			}
+		})
+	}
+}
+
+type mockClient struct {
+	returnErr bool
```
kylelemons, line 170: this can be more flexible by making it `err error` and returning it directly

thriftbp/prometheus_test.go
```diff
@@ -100,3 +100,81 @@ func TestPrometheusMetricsMiddleware(t *testing.T) {
...
+				t.Error("wanted an err, but got nil")
+			}
+			if !tt.wantErr && err != nil {
+				t.Errorf("got err: %v", err)
+			}
+		})
+	}
+}
+
+type mockClient struct {
```
kylelemons, line 169: I strongly recommend mocking the server and using the real thrift client to connect to an in-process TestBaseplate -- that ensures that you are handling the errors off the wire properly.  For example, the `error` returned by the client when a server returns a particular error will have been serialized and deserialized from the wire, so it might not match exactly what you expect.  A mocked client can potentially return values that a real client will never return, which can lead to bugs with the real code.

kylelemons: f.ex https://sourcegraph-staging.build.ue1.snooguts.net/github.snooguts.net/reddit/reddit-service-preference-go/-/blob/service/service_test.go?L333

kylelemons (changes requested):
Awesome!  :strangely_excited:

thriftbp/prometheus.go
```diff
@@ -14,22 +14,23 @@ const (
 	exceptionLabel           = "thrift_exception_type"
 	baseplateStatusLabel     = "thrift_baseplate_status"
 	baseplateStatusCodeLabel = "thrift_baseplate_status_code"
+	slugLabel                = "thrift_slug"
```
fishy, line 17: similar to kyle's comment above, here it's really hard to tell the difference between `slugLabel` and `serviceLabel`.

fishy (changes requested):

thriftbp/client_middlewares.go
JessicaGreben: good catch. ty!

thriftbp/client_middlewares.go
```diff
@@ -313,3 +318,81 @@ var (
...
+					methodLabel:            method,
+					remoteServiceSlugLabel: remoteServiceSlug,
+				}
+				clientActiveRequests.With(activeRequestLabels).Inc()
+
+				defer func() {
+					var exceptionTypeLabel, baseplateStatusCode, baseplateStatus string
+					success := strconv.FormatBool(err == nil)
+					if err != nil {
+						exceptionTypeLabel = strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
```
bjk-reddit, line 361: Hmm, this has the potential to be a huge un-bounded cardinality bomb. I would highly recommend not including something like this.

bjk-reddit (changes requested):

thriftbp/client_middlewares.go
JessicaGreben: oh yeah. good catch. let me look into a fix

bjk-reddit: So I had a chat with @kylelemons. Apparently the `%T` limits it to just the error type. It still concerns me, but maybe it's misplaced.

pacejackson: Kyle pointed this out to me earlier today but this probably isn't actually all that useful, any error that isn't an actual struct will just show up as `errors.errorString`:

https://play.golang.org/p/bEbqYiP4dL1

kylelemons: Ah, OK so Thrift is in a bit of a different place.  Thrift errors _do_ have to be specific error types, because that's how Thrift decides what to send over the wire.  So, it actually _is_ meaningful in this case.  For Thrift, this is basically how you ask what type of Exception was thrown.

kylelemons (approved):

thriftbp/client_middlewares_test.go
```diff
@@ -458,3 +459,103 @@ func TestSetClientName(t *testing.T) {
...
+}
+
+type mockBaseplateService struct {
+	fail bool
+	err  error
+}
+
+func (srv mockBaseplateService) IsHealthy(ctx context.Context, req *baseplatethrift.IsHealthyRequest) (r bool, err error) {
+	return !srv.fail, srv.err
+}
```
JessicaGreben, line 540: @kylelemons or @fishy (or anyone else): For readability, I would like to be explicit about when I'm implementing an interface.  In the past I've done this:
```
var _ baseplatethrift.BaseplateServiceV2 = (*mockBaseplateService)(nil)
```
But I'm reading in Effective Go "Don't do this for every type that satisfies an interface" [here](https://golang.org/doc/effective_go#blank_implements).

Also in Effective Go [docs](https://golang.org/doc/effective_go#generality), it describes "if a type exists only to implement an interface and will never have exported methods beyond that interface, there is no need to export the type itself. Export just the interface...", but this case doesn't apply to all my times when I want to be explicit  about interfaces I've implemented. 

What about adding this?
```
var _ baseplatethrift.BaseplateServiceV2 = mockBaseplateService{}
```

Or should I simply not try to be explicit about when an interface is implemented?

pacejackson: Reading through that part of Effective Go, the next bit it says is:

>  By convention, such declarations are only used when there are no static conversions already present in the code, which is a rare event.

The main goal of blocks like that IMO is not to say what interface something implements, but to verify that it implements it. In this case, your code wouldn't compile if `mockBaseplateService` didn't implement `baseplatethrift.BaseplateServiceV2` so you don't really need that check (because you pass it to `setupFake` which expects handler to implement that interface).

thriftbp/client_middlewares.go
pacejackson: Note this would probably end up having a similar issue since we've been encouraging devs to use the `baseplate.Error` struct rather than defining their own exceptions. On the plus side though, you could see if it _is_ a `baseplate.Error` and extract the error code and maybe turn that into a string and put it in there as `baseplate.Error.${code}`?

thriftbp/client_middlewares_test.go
JessicaGreben: @pacejackson thanks for the response. I get all that and I understand that the compiler will enforce it in this case of  `mockBaseplateService`, but I'm more asking in general. 

My question is more with regards to readability for the programmer ... I want to communicate to the reader that the object they are looking at implements a specific interface.  That make sense? I like how java has an "implements" keyword because its explicit. I know golang interfaces are not explicit, but just curious if there is a way to make them so because I like it more for readability.

pacejackson: Personally, I don't think it actually adds to the readability. I feel like it's usually pretty clear when you're implementing an interface.

fishy (approved):

PR #443, opened 2021-11-15
Deprecate runtimebp.numProcesses yaml config

fishy:
We saw some services confused on what it does, and just set the min
number to some large number and hope that will give them better
concurrency performance.

Since we already don't allow services to customize the formula from yaml
config, it really doesn't make much sense to allow them to customize
min/max from the yaml. Deprecate them and remove them in a future
breaking change.

If a service really need a different GOMAXPROCS from number of
container-aware CPUs, they can call runtime.GOMAXPROCS or
runtimebp.GOMAXPROCS after baseplate.New.

### DIFF
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -2,6 +2,7 @@ package runtimebp
 
 import (
 	"fmt"
+	"math"
 	"os"
 )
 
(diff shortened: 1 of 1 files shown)


pacejackson (approved):

konradreiche (approved):

kylelemons (changes requested):
Is it possible for this to break anyone?

fishy:
> Is it possible for this to break anyone?

that depends on how to define "break".

if someone was using

```yaml
runtime:
  numProcesses:
    min: 48
```

to get GOMAXPROCS to 48, this change will revert them back to num cpu rounding up (likely a small number ~2), and it will have performance impacts.

it's... grey area?

fishy:
searching for "numProcesses file:yaml" in internal sourcegraph only found 2 results, none with `min` set. not sure how much you trust sourcegraph's search :) (the one case I know someone uses  `min: 48` I think they already fixed it).

fishy:
💇 I believe now is the time to revisit this.

I believe previously the only use case of setting those are setting the min to a large number to essentially "pin" GOMAXPROCS to the same large number, which should no longer be needed after 7895979 & 6ddacb1, as when there's no cpu limit set we just fallback to physical number of cpus instead of cpu request. (cc @konradreiche as your team seems to be the only team doing that)

pacejackson (approved):

konradreiche (approved):
We're fine with this since we only set it to get the same behavior that will happen by default with this change :+1:

PR #444, opened 2021-11-15
Fix things shellcheck complained on shell scripts

fishy:
Currently the only shell script we have is linters.sh.


pacejackson (approved):

kylelemons (approved):
<3 shellcheck

konradreiche (approved):

PR #445, opened 2021-11-29
Add prometheus metrics spec test

JessicaGreben:
This PR adds a test to validate the baseplate Prometheus metrics conform to the [baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md) and the Prometheus standards.


prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,80 @@
...
+
+// ValidateSpec validates that the Prometheus metrics being exposed conform to
+// the baseplate spec.
+func ValidateSpec(t *testing.T, metricPrefix string, wantMetricCount int) {
+	if err := validateSpec(metricPrefix, wantMetricCount); err != nil {
+		t.Error(err)
+	}
+}
+
+func validateSpec(metricPrefix string, wantMetricCount int) error {
```
kylelemons, line 28: this should probably return the count, not assert on it itself

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,80 @@
...
+	metricFam, err := prometheus.DefaultGatherer.Gather()
+	if err != nil {
+		return err
+	}
+
+	var gotMetricCount int
+	for _, m := range metricFam {
+		metricName := *m.Name
+
+		if strings.Contains(metricName, metricPrefix) {
```
kylelemons, line 38: * Use `HasPrefix` not `Contains`

* Reduce indentation by continuing early

```go
if !strings.HasPrefix(metricName, metricPrefix) {
  continue
}
```

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,80 @@
...
+			gotMetricCount++
+			if err := validatePromLint(metricName); err != nil {
+				return err
+			}
+			if err := validateName(metricName, metricPrefix); err != nil {
+				return err
+			}
```
kylelemons, line 45: You don't want to return a single error here, as that will result in having to fix one bug at a time.  You want to return _all_ errors that are encountered.

see errorsbp.Batch

kylelemons (changes requested):

prometheusbp/promtest/spec.go
JessicaGreben: I chose to use `Contains` instead of `HasPrefix` to catch if the  intended prefix ended up in a different part of the metric name... but as I write that now maybe its an unnecessary additional step.

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,80 @@
...
+	errPrometheusLint = errors.New("problem with Prometheus GatherAndLint")
+)
+
+// ValidateSpec validates that the Prometheus metrics being exposed conform to
+// the baseplate spec.
+func ValidateSpec(t *testing.T, metricPrefix string, wantMetricCount int) {
+	if err := validateSpec(metricPrefix, wantMetricCount); err != nil {
+		t.Error(err)
+	}
+}
```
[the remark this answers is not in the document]
JessicaGreben: "wantMetricCount" is how many metrics there should be with the metricPrefix. I can add that to the doc string. 

I'm not sure what you mean "wantServiceCount" or even "wantServiceNames ...string".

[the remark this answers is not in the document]
konradreiche: +1 on this sentiment, it becomes clear as soon as I scroll to the validation where it describes in the comment:
```
// validateName checks that:
// 1) the metric name has at least 3 parts separated by "_".
```
explaining why we pass 3 here
```
defer promtest.ValidateSpec(t, "thrift", 3)
```

Not sure how this can be written in a more expressive way. That said, I think those tests often serve as something that protects against any type of regressions so potentially not too important.

konradreiche (approved):

prometheusbp/promtest/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package promtest provides test utilites for testing Prometheus metrics.
+package promtest
```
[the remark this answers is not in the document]
JessicaGreben: 🤷‍♀️ I don't feel strongly either way.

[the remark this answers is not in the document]
JessicaGreben: Thinking about this more, I do want `[prometheusbp.DefaultBuckets](https://github.com/reddit/baseplate.go/blob/prometheus/prometheusbp/metrics.go#L8)` to be useable by services using baseplate.go as well as the `promtest` stuff.  So maybe it would be better to leave where it is. Thoughts @kylelemons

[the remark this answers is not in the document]
kylelemons: I think you want a few packages here, not just one:

* `prometheusbp` -- general-purpose code for use anywhere (e.g. `DefaultBuckets`)
* `prometheusbp/promtest` -- general-purpose test helpers for anyone to use (e.g. the `CheckDelta` stuff)
* `internal/prometheusbp/spectest` -- specific test helper(s) to unify baseplate spec prom compliance logic between other baseplate.go packages

prometheusbp/promtest/spec.go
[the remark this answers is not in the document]
kylelemons: ok, so if `wantMetricCount` is just "go to the spec and count how many metrics are expected", then isn't that already part of the spec? The caller doesn't need to provide that.  Any extra metrics that are not expected shuld result in an error, any missing metrics should result in an error.

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+// ValidateSpec validates that the Prometheus metrics being exposed conform to
+// the baseplate spec. This test will validate all metrics that begin with the
+// metricPrefix and will validate the number of metrics with that prefix match
+// wantMetricCount.
+func ValidateSpec(t *testing.T, metricPrefix string, wantMetricCount int) {
+	var batchErrs errorsbp.Batch
+	gotMetricCount, errs := validateSpec(metricPrefix)
+	if len(errs) > 0 {
+		batchErrs.Add(errs...)
+	}
```
kylelemons, line 31: > If an error is also a Batch, its underlying error(s) will be added instead of the Batch itself.

so `validateSpec` can return a `batch.Compile()` as a single error and you can `batchErrs.AddPrefix("validate spec", err)` here

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+	errCount          = errors.New("wrong metric count for prefix")
+	errPrometheusLint = errors.New("problem with Prometheus GatherAndLint")
+)
+
+// ValidateSpec validates that the Prometheus metrics being exposed conform to
+// the baseplate spec. This test will validate all metrics that begin with the
+// metricPrefix and will validate the number of metrics with that prefix match
+// wantMetricCount.
+func ValidateSpec(t *testing.T, metricPrefix string, wantMetricCount int) {
+	var batchErrs errorsbp.Batch
```
kylelemons, line 27: t.Helper()

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+	gotMetricCount, errs := validateSpec(metricPrefix)
+	if len(errs) > 0 {
+		batchErrs.Add(errs...)
+	}
+	if gotMetricCount != wantMetricCount {
+		batchErrs.Add(fmt.Errorf("%w: got %d, want %d", errCount, gotMetricCount, wantMetricCount))
+	}
+
+	for _, e := range batchErrs.GetErrors() {
+		t.Errorf("err: %v", e.Error())
```
kylelemons, line 37: This is `t.Errorf` and the messages should explain their failure, so the `err: ` prefix should be redundant

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+
+// validateName checks that:
+// 1) the metric name has at least 3 parts separated by "_".
+// The parts are <namespace>_<metric>_<suffix> as described in Prometheus naming conventions.
+// Ref: https://prometheus.io/docs/practices/naming
+// 2) the metric name has the correct prefix.
+func validateName(name, prefix string) []error {
+	const metricPartSeparator = "_"
+	var batchErrs errorsbp.Batch
+	metricNameParts := strings.Split(name, metricPartSeparator)
```
kylelemons, line 77: it's fine for a name to have more parts.  You just want _at least_ three.

SplitN(name, metricPartSeparator, 3)

that will group together all of the ones after the third into the third one

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+	if len(metricNameParts) < 3 {
+		batchErrs.Add(fmt.Errorf("%w: got %d, want > 3", errLength, len(metricNameParts)))
+	}
+	if !strings.HasPrefix(name, prefix+metricPartSeparator) {
```
kylelemons, line 81: if got, want := metricNameParts[0], prefix; got != want {
  errors.Add(fmt.Errorf("%w: metric %q does not have required prefix %q", errPrefix, name, prefix)
}

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// validateName checks that:
+// 1) the metric name has at least 3 parts separated by "_".
+// The parts are <namespace>_<metric>_<suffix> as described in Prometheus naming conventions.
+// Ref: https://prometheus.io/docs/practices/naming
+// 2) the metric name has the correct prefix.
+func validateName(name, prefix string) []error {
+	const metricPartSeparator = "_"
+	var batchErrs errorsbp.Batch
```
kylelemons, line 76: these are local variables, they should probably just be a single word.

`separator` or `sep`
`errors` or `batch`

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+func validateSpec(metricPrefix string) (int, []error) {
+	var metricCount int
+	var batchErrs errorsbp.Batch
+
+	metricFam, err := prometheus.DefaultGatherer.Gather()
+	if err != nil {
+		return metricCount, batchErrs.GetErrors()
+	}
+
+	for _, m := range metricFam {
```
kylelemons, line 50: I would make a map of the required metrics and delete them as you see them, and then add an error for each missing metric at the end.  Right now you are just checking for a count, which means that it would be easy to get the wrong metrics and not fail the spec test.

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		metricCount++
+		if errs := validatePromLint(metricName); len(errs) > 0 {
+			batchErrs.Add(errs...)
+		}
+		if errs := validateName(metricName, metricPrefix); len(errs) > 0 {
+			batchErrs.Add(errs...)
+		}
+	}
```
kylelemons, line 64: One of the most important parts of the spec is the metric labels.  I think it is important to make sure that the right labels are applied.

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+	var metricCount int
+	var batchErrs errorsbp.Batch
+
+	metricFam, err := prometheus.DefaultGatherer.Gather()
+	if err != nil {
+		return metricCount, batchErrs.GetErrors()
+	}
+
+	for _, m := range metricFam {
+		metricName := *m.Name
```
kylelemons, line 51: similar to below, local variables don't need to contain all possible context in their names.  `name` and `count` are perfectly clear in context here.

prometheusbp/promtest/spec.go
```diff
@@ -0,0 +1,97 @@
...
+
+		if !strings.HasPrefix(metricName, metricPrefix) {
```
kylelemons, line 53: I think you're going to want some more logic here.

The format of a metric name is `{domain}_{client,server}` so I would accept _both_ of those, especially since some of the logic for what is considered "correct" metric-wise is dependent on whether it's the client or server.

kylelemons (changes requested):

prometheusbp/promtest/spec.go
kylelemons: (this assumes you get the prefix and the server/client separately per above)

JessicaGreben: I didn't realize `%q` was a thing. :chefs_kiss:

kylelemons: omg `%q` and `%T` are two of my favorite things about Go.  C only wishes its printf were as useful

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+	switch strings.Join(suffix, separator) {
+	case "_latency_seconds":
+		labelSuffixes = append(labelSuffixes, successLabelSuffix)
+	case "_requests_total":
+		labelSuffixes = append(labelSuffixes, successLabelSuffix, "exception_type", "baseplate_status", "baseplate_status_code")
+	case "_active_requests":
+		// no op
+	default:
+		// todo: error
+	}
```
kylelemons, line 184: You can collapse some of this early logic into
```
switch {
case strings.HasSuffix(name, "_latency_seconds"):
case strings.HasSuffix("_requests_total"):
case strings.HasSuffix("_active_requests")
}
```

And you can accept "prefix" as a param for the purpose of constructing wantLabels below

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// buildMetricNames creates a set of all the expected metrics names for the given prefix
+// that should be registered with Prometheus as defined in the baseplate spec.
+// The prefix will be either thrift, http, or grpc.
+func buildMetricNames(prefix string) map[string]struct{} {
+	suffixes := []string{"_latency_seconds", "_requests_total", "_active_requests"}
+	names := map[string]struct{}{}
+	for _, suffix := range suffixes {
+		names[prefix+"_"+server+suffix] = struct{}{}
```
kylelemons, line 93: 🔕 
I think I would do `prefix + "_" + server + "_" + suffix` (i.e. remove the _ from the suffixes.

If you _really_ want to get fancy you can call

 https://pkg.go.dev/github.com/prometheus/client_golang@v1.11.0/prometheus#BuildFQName

`names[prometheus.BuildFQName(prefix, client, suffix)]`

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+			batch.Add(fmt.Errorf("%w metric %q label %q", errLabelNotFound, name, wantLabel))
+		}
+		delete(wantLabels, wantLabel)
+	}
+
+	if len(wantLabels) > 0 {
+		// todo: err
+	}
+	return batch.Compile()
+}
```
kylelemons, line 143: I would use `cmp.Diff`

```
// var errDiffLabels = fmt.Errorf("labels incorrect")
if diff := cmp.Diff(gotLabels, wantLabels); diff != "" {
  return fmt.Errorf("%w: (-got +want)\n%s", errDiffLabels, diff)
}
```

cmp understands maps and lengths and prints out a nice diff that could be like

```diff
 {
-  "grpc_service": true,
+  "grpc_serice": true,
   "grpc_method": true,
   ...
 }
```

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+	metricFam, err := prometheus.DefaultGatherer.Gather()
+	if err != nil {
+		return batch
+	}
+
+	allMetricNames := buildMetricNames(prefix)
+
+	for _, m := range metricFam {
+		name := *m.Name
+
```
kylelemons, line 54: My slight concern with the construction of the loop (i.e. going through all metrics) is that we might accidentally trip over some additional metrics.  For example, if the grpc library does something like `grpc_registered_proto_service_count` (a possibly-bad example of something that is common between the client and server) then it would fail the `validatePromLint` even though the library could still be spec compliant.

I would probably
* use a loop to construct the desired metric names
* use a loop over the gathered metrics to collect a map of metric name to definition info
* loop through desired names and raise an error if they aren't found in the actual map, and lint and check labels for the ones that are.

kylelemons (changes requested):

internal/prometheusbp/spectest/spec.go
JessicaGreben: this method `buildMetricNames(prefix)` creates a list of desired metrics names so that your concern of `grpc_registered_proto_service_count` getting tripped on shouldn't occur. `validatePromLint` will only be executed if `buildMetricNames(prefix)` returns the metric name.  

Down on line 80 it errors if a metric is not found.

kylelemons: indeed 😅

internal/prometheusbp/spectest/spec.go
JessicaGreben: perfection! thanks.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	if len(allMetricNames) > 0 {
+		notFound := make([]string, 0, len(allMetricNames))
+		for k := range allMetricNames {
+			notFound = append(notFound, k)
+		}
+		batch.Add(fmt.Errorf("%w metrics %v", errNotFound, notFound))
+	}
+
```
kylelemons, line 82: 🔕 
might be convenient to make a helper `batch.Add(missingMetrics(allMetricNames)`

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+func validateSpec(prefix string) errorsbp.Batch {
+	var batch errorsbp.Batch
+
+	metricFam, err := prometheus.DefaultGatherer.Gather()
+	if err != nil {
+		return batch
+	}
+
+	allMetricNames := buildMetricNames(prefix)
```
kylelemons, line 50: 🔕 since this represents a "shrinking" list of metrics, you could call it `remainingMetricNames` or something -- and then the loop at the end reads super nicely

kylelemons (changes requested):

internal/prometheusbp/spectest/spec.go
JessicaGreben: how about `metricNamesUnderTest`?

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+// conform to the baseplate spec.
+func ValidateSpec(t *testing.T, metricPrefix string, wantMetricCount int) {
+	t.Helper()
+
+	var batch errorsbp.Batch
+	batch.AddPrefix("validate spec", validateSpec(metricPrefix))
+
+	if len(batch.GetErrors()) > 0 {
+		t.Errorf(batch.Compile().Error())
+	}
```
fishy, line 40:
```suggestion
	if err := batch.Compile(); err != nil {
		t.Error(err)
	}
```

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func validateSpec(prefix string) errorsbp.Batch {
+	var batch errorsbp.Batch
+
+	metricFam, err := prometheus.DefaultGatherer.Gather()
+	if err != nil {
+		return batch
```
fishy, line 48: do you mean `return err`, or `batch.Add(err); return batch.Compile()` here?

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+	errLength         = errors.New("metric name should have a minimum of three parts")
+	errNotFound       = errors.New("metric not found")
+	errLabelNotFound  = errors.New("label not found")
+	errPrometheusLint = errors.New("prometheus GatherAndLint")
+)
+
+// ValidateSpec validates that the Prometheus metrics being exposed from baseplate.go
+// conform to the baseplate spec.
+func ValidateSpec(t *testing.T, metricPrefix string, wantMetricCount int) {
+	t.Helper()
```
fishy, line 33: also as a general rule of thumb, prefer TB over *T for helper test functions.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,190 @@
...
+	errClientServer   = fmt.Errorf("missing %q or %q as the second part of the metric name", client, server)
+	errLength         = errors.New("metric name should have a minimum of three parts")
+	errNotFound       = errors.New("metric not found")
+	errPrometheusLint = errors.New("prometheus GatherAndLint")
+	errDiffLabels     = errors.New("labels are incorrect")
+)
+
+// ValidateSpec validates that the Prometheus metrics being exposed from baseplate.go
+// conform to the baseplate spec.
+func ValidateSpec(t *testing.T, metricPrefix string, wantMetricCount int) {
```
kylelemons, line 32: 🔕 ah sorry I mentioned this deeper in the call stack and didn't explain that I was imagingin that it would make it up this far.

I kinda imagine that someone might want to use this independently to check client and server metrics -- e.g. when only one of them is implemented 😅 

so I would probably make this API

```go
func ValidateSpec(t *testing.T, metricPrefix, clientOrServer string)
```

As mentioned earlier, the count might be confusing because the spec already catches missing metrics, and this way you can say `ValidateSpec("grpc", "server")` instead of `"grpc_server"` which has a bit more "magic" to it.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+		}
+		batch.Add(validateLabels(name, prefix, labels))
+
+		// after we validate the metric, delete it from the set of all
+		// expected metrics so that we can track if any were not found.
+		delete(metricsUnderTest, name)
+	}
+
+	batch.Add(fmt.Errorf("%w %v", errNotFound, missingMetrics(metricsUnderTest)))
```
kylelemons, line 76:
```suggestion
	batch.Add(fmt.Errorf("%w: %v", errNotFound, missingMetrics(metricsUnderTest)))
```

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return batch
+}
+
+func missingMetrics(metrics map[string]struct{}) []string {
```
kylelemons, line 81: 🔕 now that I read it, maybe "keysFrom"?

```go
fmt.Errorf("%w: %v", errNotFound, keysFrom(metricsUnderTest)))
```

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	return names
+}
+
+// validateName checks the following:
+// 1) the metric name has at least 3 parts separated by "_".
+// The parts are <namespace>_<client/server>_<name>_<suffix>, where suffix is optional.
+// Ref: https://prometheus.io/docs/practices/naming
+// 2) the metric name has the correct prefix.
+// 3) the metric contains either "client" or "server" as the second part.
```
kylelemons, line 109:
```suggestion
// validateName checks the following:
//   1) the metric name has at least 3 parts separated by "_".
//     The parts are <namespace>_<client/server>_<name>_<suffix>, where suffix is optional.
//     Ref: https://prometheus.io/docs/practices/naming
//   2) the metric name has the correct prefix.
//   3) the metric contains either "client" or "server" as the second part.
```

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,190 @@
...
+//   - "<prefix>_success"
+//   - "<prefix>_exception_type"
+//   - "<prefix>_baseplate_status"
+//   - "<prefix>_baseplate_status_code"
+// active_requests metrics expect the following labels:
+//   - "<prefix>_service"
+//   - "<prefix>_method"
+func buildLables(name, prefix string) map[string]struct{} {
+	labelSuffixes := []string{"service", "method"}
+	successLabelSuffix := "success"
```
kylelemons, line 162: 🔕 factoring this out to avoid the same literal twice makes it a bit jarring on line 167 to switch from a var to literals; I'd just use literals the two places this appears

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,203 @@
...
+	switch strings.Join(suffix, separator) {
+	case "_latency_seconds":
+		labelSuffixes = append(labelSuffixes, successLabelSuffix)
+	case "_requests_total":
+		labelSuffixes = append(labelSuffixes, successLabelSuffix, "exception_type", "baseplate_status", "baseplate_status_code")
+	case "_active_requests":
+		// no op
+	default:
```
fishy, line 170: 🔕 this is a personal preference/style thing, but I in general prefer to put `default` before all other `case`s in a `switch`. go's `switch` does not require you to exhaust all the possibilities (like rust does), so with or without `default` is a significant difference and having `default` case at the beginning helps highlight that "this `switch` block has a `default` case`.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,190 @@
...
+		labelSuffixes = []string{}
+	}
+	var wantLabels = map[string]struct{}{}
+	for _, label := range labelSuffixes {
+		wantLabels[prefix+"_"+label] = struct{}{}
+	}
+	return wantLabels
+}
+
+func validatePromLint(metricName string) errorsbp.Batch {
```
kylelemons, line 180: 🔕 lintMetric?

kylelemons (approved):

internal/prometheusbp/spectest/spec.go
kylelemons: this will be something like `maps.KeysOf(metricsUnderTest)` after generics I think

internal/prometheusbp/spectest/spec.go
JessicaGreben: good catch. I will `batch.Add(err); return batch`

internal/prometheusbp/spectest/spec.go
JessicaGreben: oops. forgot. thanks.

internal/prometheusbp/spectest/spec.go
JessicaGreben: I like the `default` case at the end because then it's a similar logic flow as `if else` blocks, but I don't feel too strongly either way. I'm happy to move the `default` case up if you prefer.

internal/prometheusbp/spectest/spec.go
JessicaGreben: will do. also forgot to remove the `wantMetricCount` arg. will fix.

internal/prometheusbp/spectest/spec.go
fishy: this is nitpick/optional, but I would prefer to make this function to return `error` instead of `errorsbp.Batch` (so you just add `.Compile()` to every return inside this function). `batch.AddPrefix` on line 38 can correctly convert an `error` back to a batch when needed.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,198 @@
...
+func validateName(name, prefix, clientOrServer string) errorsbp.Batch {
+	var batch errorsbp.Batch
+
+	const (
+		separator = "_"
+		partCount = 3
+	)
+	parts := strings.SplitN(name, separator, partCount)
+	if len(parts) < partCount {
+		batch.Add(fmt.Errorf("%w metric %q does not have minimum number (%d) of parts", errLength, name, partCount))
```
fishy, line 123: nit: `errLength` already contains (hardcoded) `partCount` info. we should either remove that from `errLength` (and rely on adding that info here), or remove `partCount` here.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,198 @@
...
+		return batch
+	}
+
+	if got, want := parts[0], prefix; got != want {
+		batch.Add(fmt.Errorf("%w metric %q does not have the expected prefix %q", errPrefix, name, prefix))
```
fishy, line 128: (this comment applies to all `fmt.Errorf` calls inside this function)

so the final error message would be:

```
the prefix is not at the beginning of the metric name metric "foo" does not have the expected prefix "bar"
```

which is kinda redundant and hard to read. we should add `:` after `%w`, and simplify the additional info part to not repeat what is already in the `%w` part. so this line should be something like this:

```suggestion
		batch.Add(fmt.Errorf("%w: name: %q prefix: %q", errPrefix, name, prefix))
```

which produces final error message as:
```
the prefix is not at the beginning of the metric name: name: "foo" prefix: "bar"
```

prometheusbp/promtest/testutil.go
```diff
@@ -1,4 +1,4 @@
-package prometheusbp
+package promtest
```
fishy, line 1: github does not allow me to add comments further down this file as they are not changed in this PR, but I think we should add "Use MetricTest to construct it" in the doc comment of `PrometheusMetricTest`, or rename `MetricTest` into `NewPrometheusMetricTest`.

internal/prometheusbp/spectest/spec.go
JessicaGreben: originally I returned error, but when talking with Kyle we changed it to return errorsbp.Batch because it make testing tricky because then I can't test the error itself, but rather have to do string matching which i would like to avoid if possible. ref: https://github.com/reddit/baseplate.go/pull/445/files#diff-8b8f4f423ed61950357bb8113cc700075e8a2a31b64b6a6b6bb878b774fdf5faR288

fishy: OK in that case this is fine.

fishy: but compiled batch errors still implemented `errors.Is`/`errors.As` correctly, so that should not really affect those tests. (e.g. if you change the signature of this function to return `error` instead, the tests should still pass).

internal/prometheusbp/spectest/spec.go
JessicaGreben: the `errLength` message indicates how many parts are required. `partCount` indicates how many there actually are.  I think both are useful, but if you disagree I can remove parCount

fishy: >`partCount` indicates how many there actually are.

that's `len(parts)`, not `partCount`.

JessicaGreben: oops. thanks. let me fix that.

internal/prometheusbp/spectest/spec.go
JessicaGreben: I originally tried `errors.As` but it didn't work, but let me try again since a lot has changed since i was attempting that.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	return batch.Compile()
+}
+
+func validateLabels(name, prefix, clientOrServer string, gotLabels map[string]struct{}) error {
+	var batch errorsbp.Batch
+
+	wantLabels := buildLables(name, prefix, clientOrServer)
+	if diff := cmp.Diff(gotLabels, wantLabels); diff != "" {
+		batch.Add(fmt.Errorf("%w: (-got +want)\n%s", errDiffLabels, diff))
```
fishy, line 142: since this function can return at most one error, I think using a batch here is overkill. this line can simply be:

```go
return fmt.Errorf("%w: (-got +want)\n%s", errDiffLabels, diff)
```

internal/prometheusbp/spectest/spec_test.go
```diff
@@ -0,0 +1,340 @@
...
+func checkErrors(tb testing.TB, gotErr error, wantErrs []error) {
+	tb.Helper()
+
+	switch len(wantErrs) {
+	case 0:
+		if gotErr != nil {
+			tb.Errorf("expected nil err, got: %v", gotErr)
+		}
+	case 1:
+		// no op if only 1 error is expected
```
fishy, line 323: you probably want to verify that `gotErr` is not a batch here, or wait for https://github.com/reddit/baseplate.go/pull/448 and cherry pick that into prometheus branch after it's merged.

internal/prometheusbp/spectest/spec_test.go
```diff
@@ -0,0 +1,340 @@
...
+	default:
+		var be errorsbp.Batch
+		if errors.As(gotErr, &be) {
+			if got, want := len(be.GetErrors()), len(wantErrs); got != want {
+				tb.Errorf("not same number of errors got %d, want %d", got, want)
+			}
+		} else {
+			tb.Error("more than 1 error returned so a batch is expected")
```
fishy, line 331: grammar nit: "more than 1 error returned" implies got > 1, but the actual situation here is that got <= 1 but want > 1.

I would reword this as: `tb.Errorf("expected %d errors, got a single error: %v", len(wantErrs), gotErr)`

internal/prometheusbp/spectest/spec.go
JessicaGreben: good catch. _fixing_

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,195 @@
...
+
+	if len(metricsUnderTest) > 0 {
+		batch.Add(fmt.Errorf("%w: %v", errNotFound, keysFrom(metricsUnderTest)))
+	}
+
+	return batch.Compile()
+}
+
+func keysFrom(metrics map[string]struct{}) []string {
+	missing := []string{}
```
fishy, line 87: 1. prefer `var missing []string`
2. since you actually know the size before hand, prefer `missing := make([]string, 0, len(metrics))`

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,195 @@
...
+	if len(metrics) > 0 {
```
fishy, line 88: this `if` is unnecessary. if `len(metrics) == 0` then the `for` loop will execute zero times are we are still good.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,195 @@
...
+		name := *m.Name
+
+		if _, ok := metricsUnderTest[name]; !ok {
+			continue
+		}
+
+		batch.Add(lintMetric(name))
+		batch.Add(validateName(name, prefix, clientOrServer))
+
+		labels := map[string]struct{}{}
```
fishy, line 66: style nit: prefer `labels := make(map[string]struct{})` to make the intention explicit/clear.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,195 @@
...
+		for _, metric := range m.GetMetric() {
+			for _, label := range metric.Label {
+				labels[*label.Name] = struct{}{}
```
fishy, line 69: nit: since `struct{}{}` is used several times inside this file, I'd prefer to define it as a global variable to improve readability:
```
var dummySetValue = struct{}{}
```
and replaced the usages with `dummySetValue`.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	return missing
+}
+
+// buildMetricNames creates a set of all the expected metrics names for the given prefix
+// that should be registered with Prometheus as defined in the baseplate spec.
+// The prefix will be either thrift, http, or grpc.
+func buildMetricNames(prefix, clientOrServer string) map[string]struct{} {
+	suffixes := []string{"latency_seconds", "requests_total", "active_requests"}
+	names := map[string]struct{}{}
```
fishy, line 101: same style nit here.

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,195 @@
...
+//     The parts are <namespace>_<client/server>_<name>_<suffix>, where suffix is optional.
+//     Ref: https://prometheus.io/docs/practices/naming
+//   2) the metric name has the correct prefix.
+//   3) the metric contains either "client" or "server" as the second part.
+func validateName(name, prefix, clientOrServer string) error {
+	var batch errorsbp.Batch
+
+	const (
+		separator = "_"
+		partCount = 3
```
fishy, line 119: 🔕/super nit: this is the same value in `errLength`. I'd prefer move this const to global and change `errLength` into: `errLength = fmt.Errorf("metric name should have a minimum of %d parts", partCount)`.

also grammar question/nit: should this be `partsCount` instead of `partCount`?

internal/prometheusbp/spectest/spec.go
```diff
@@ -0,0 +1,195 @@
...
+		// no op
+	default:
+		labelSuffixes = []string{}
+	}
+
+	if clientOrServer == client {
+		labelSuffixes = append(labelSuffixes, "slug")
+	}
+
+	var wantLabels = map[string]struct{}{}
```
fishy, line 178: same style nit here

internal/prometheusbp/spectest/spec_test.go
```diff
@@ -0,0 +1,326 @@
...
+		},
+		{
+			name:    "missing",
+			missing: map[string]struct{}{"foo": {}},
+			want:    []string{"foo"},
+		},
+	}
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			if got, want := keysFrom(tt.missing), tt.want; !reflect.DeepEqual(got, want) {
```
fishy, line 33: nit: instead of using `reflect.DeepEqual`, I would prefer to write a helper function to compare string slices:

```
func compareStringSlices(tb testing.TB, got, want []string) {
  ...
}
```

because I believe `reflect.DeepEqual([]string{}, []string(nil))` would return false, but they are both correct (we don't want to fail the tests when in the future the functions change to return `nil` for empty slice).

see: https://go.dev/play/p/02-mlRXai9O

internal/prometheusbp/spectest/spec_test.go
```diff
@@ -0,0 +1,326 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			gotErr := validateSpec(tt.prefix, tt.clientOrServer)
+			checkErrors(t, gotErr, tt.wantErrs)
+		})
+	}
+}
+
+func checkErrors(tb testing.TB, gotErr error, wantErrs []error) {
+	tb.Helper()
+
```
fishy, line 316: nit: add a line here:
```
tb.Logf("got errors %v, want errors %v")
```
so this line will be printed out when the test fails, which provides more info (for example, only printing out the different number of errors (on line 318) is not very useful to figure out what is wrong).

internal/prometheusbp/spectest/spec_test.go
```diff
@@ -0,0 +1,326 @@
...
+	if got, want := errorsbp.BatchSize(gotErr), len(wantErrs); got != want {
+		tb.Errorf("not same number of errors got %d, want %d", got, want)
+	}
+
+	for _, wantErr := range wantErrs {
+		if !errors.Is(gotErr, wantErr) {
+			tb.Fatalf("error mismatch: got %v, want %v", gotErr, wantErr)
```
fishy, line 323: 1. this should be `Errorf` instead of `Fatalf`
2. this is not really a got/want situation. the error message should be something like this: `tb.Errorf("Want error %v not in returned errors", wantErr)` (I did not add want into the error message because the `tb.Logf` part above would print that info).

internal/prometheusbp/spectest/spec.go
JessicaGreben: lol. good catch.

fishy (approved):

PR #446, opened 2021-12-03
add `detach` context library

pacejackson:
This package allows you to get a new context that is `detached` from it's parent (so it is not associated with it's cancelation or deadline) but include values that you have registered to be attached to it.


detach/registry.go
```diff
@@ -0,0 +1,41 @@
...
+func attachAll(dst, src context.Context) (context.Context, context.CancelFunc) {
+	var cleanupFuncs []func(ctx context.Context)
+	for _, wrap := range wrapperRegistry {
+		dst = wrap.Attach(dst, src)
+		if wrap.Cleanup != nil {
+			cleanupFuncs = append(cleanupFuncs, wrap.Cleanup)
+		}
+	}
+	return dst, func() {
+		for _, cleanup := range cleanupFuncs {
```
fishy, line 37: we probably need to run cleanup functions FILO, instead of FIFO.

foreverest:
Curious why don't make it simpler? ([like this](https://github.snooguts.net/reddit/reddit-service-realtime/blob/master/internal/detachedcontext/detachedcontext.go))

detach/registry.go
pacejackson: that makes sense, although the nature of how you actually register these things kinda means that you can't really rely on ordering.

fishy: yea good point.

pacejackson: alternatively, we could lean into that and shuffle `cleanupFuncs`...

detach/detach.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return r.hooks
+}
+
+var registry = new(hooksRegistry)
+
+// Register adds a new Hooks to the library's registry. If all functions in the Hooks are nil, Register will panic.
+func Register(hooks Hooks) {
+	if hooks.Inline == nil && hooks.Async == nil {
+		panic("cannot register Hooks with all nil functions")
```
fishy, line 51: 🔕 since both `Inline` and `Async` would auto skip hooks without the corresponding field, this panic doesn't seem necessary, but I don't feel strongly either way.

detach/detach.go
```diff
@@ -0,0 +1,97 @@
...
+			// loop.
+			next := run
+			run = func(ctx context.Context) {
+				// if you just used 'run' in here, it would always use the last value of "run", so it would call itself over and
+				// over again, creating an infinite loop.
+				hooks.Async(ctx, parent, next)
+			}
+		}
+	}
+	run(ctx)
```
fishy, line 96: 🔕 should this be `go run(ctx)` instead? without that the caller would need to do `go detach.Async(...)`, which sounds redundant (with both `go` and `Async`). but I also don't feel strongly either way.

log/context.go
```diff
@@ -6,12 +6,31 @@ import (
...
+				dst = context.WithValue(dst, contextKey, logger)
+			}
+			return dst
+		},
+		Async: func(dst, src context.Context, next func(ctx context.Context)) {
+			if logger, ok := src.Value(contextKey).(*zap.SugaredLogger); ok && logger != nil {
+				dst = context.WithValue(dst, contextKey, logger)
+			}
+			next(dst)
+		},
```
fishy, line 30: 🔕 we did the same copy in both `Inline` and `Async` here, and that would be a common case for a lot of packages (they need to copy some shared kv pairs in both `Inline` and `Async` hooks). I think we should help avoid this kind of common code duplication from the API level, for example provide a `Base` hook that will be called by both `Inline` and `Async` if it's non-nil?

tracing/tracer.go
```diff
@@ -36,6 +37,21 @@ func init() {
 	opentracing.SetGlobalTracer(&globalTracer)
 	// Init the global allow-list
 	SetMetricsTagsAllowList(nil)
+	// Register a ValueWrap with the detach library
+	detach.Register(detach.Hooks{
+		Inline: func(dst, src context.Context) context.Context {
+			return opentracing.ContextWithSpan(dst, opentracing.SpanFromContext(src))
+		},
+		Async: func(dst, src context.Context, next func(ctx context.Context)) {
+			span, _ := opentracing.StartSpanFromContext(
```
fishy, line 46: 🔕 
1. we probably want to verify that our tracing system is ok with having a child span finishes after the parent span finishes
2. we probably want to check that there's actually a span in `src` before creating the child span? it's possible `detach.Async` is called from a context without any tracing, and we probably don't want to create this top-level span in those cases.

detach/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package detach provides utility function for getting new contexts that are detached from their parent's timeouts and
+// cancel calls but inheriting any values registered with the library.
+package detach
```
fishy, line 3: 🔕 I'm also not sure if `detach` is the best name for this package. The name `detach` itself doesn't say anything specific about context, but renaming it to something like `ctxdetach` is also kinda gross, so ¯\\\_(ツ)\_/¯

fishy (approved):
none of my comments are blocking.

detach/detach.go
pacejackson: I agree that it's not strictly necessary, but I also think that handing in an empty Hooks is a bug or an oversight that we want to fix, which is why I added the panic.

detach/detach.go
fishy: I guess it makes sense for a use case that this is not the whole background goroutine, just a portion of it, e.g.:
```go
go func() {
  // some preparing work
  detach.Async(...)
  // some cleanup work
}
```

so ignore my previous comment 😅

log/context.go
pacejackson: IMO, that's messier than the current state, while there's some duplication, it's minor and only happens in a few libraries. The issue I see with adding something like `Base` is it makes it less straightforward to see what's happening just by reading the `Register` call, you would have to know that if `Base` is defined, it get's called before each of these functions.

detach/detach.go
```diff
@@ -0,0 +1,97 @@
+package detach
+
+import (
+	"context"
+	"sync"
+	"time"
+)
+
+// Hooks are functions that are called with specific detach helpers to attach values from the parent context to the new
```
kylelemons, line 9: 🔕 if you're not going to wrap based on full sentences, 80col is where I would recommend wrapping; this overflows my github review width

detach/detach.go
```diff
@@ -0,0 +1,97 @@
...
+// context. In order to be registered, at least one function in Hooks must be non-nil. Nil functions will be skipped.
+type Hooks struct {
+	// Inline functions are used to attach values from the src context to the dst context in calls to Inline.
+	//
+	// Inline functions should assume that the calling function will wait for the any calls after Inline to return before
+	// it returns, so they should just copy values from src to dst.
```
kylelemons, line 15: so they _may_ just copy values maybe?

detach/detach.go
```diff
@@ -0,0 +1,97 @@
...
+	Inline func(dst, src context.Context) context.Context
+
+	// Async functions wrap the run methods passed to Async. They should copy values from the src context to the dst
```
kylelemons, line 18: they should _derive_ values maybe?

detach/detach.go
kylelemons: yeah agreed

detach/detach.go
```diff
@@ -0,0 +1,97 @@
...
+	return ctx, cancel
+}
+
+// Async wraps run in the Async Hooks registered with the library, creates a new context with the given timeout, and
+// calls the wrapped run function with that context. The wrappers will attach values from the parent context to the new
+// context and perform any cleanup tasks after run completes. The new context will be canceled when Async returns.
+//
+// Async should be used when you want to run an async task as a part of another call, but need to keep some values from
+// the parent context.
+func Async(parent context.Context, timeout time.Duration, run func(ctx context.Context)) {
```
kylelemons, line 78: so, since async tasks might be long-running, I don't think enforcing a timeout makes as much sense.

For example, one prime use-case of this style of API that I have used before is when a user calls an RPC that spins up a background task and then can periodically query its status -- it doesn't necessarily have a timeout since it could last forever, so the caller needs to add cancellation or timeouts _within_ the closure as appropriate.

detach/detach_example_test.go
```diff
@@ -0,0 +1,172 @@
...
+			return dst
+		},
+		Async: func(dst, src context.Context, next func(ctx context.Context)) {
+			if v, ok := src.Value(contextKey).(*ctxVal); ok && v != nil {
+				newVal := &ctxVal{val: v.val}
+				dst = context.WithValue(dst, contextKey, newVal)
+				defer func() {
+					fmt.Printf("newVal.closed() == %v\n", newVal.closed)
+					newVal.close()
+					fmt.Printf("newVal.closed() == %v\n", newVal.closed)
```
kylelemons, line 26: did you mean for one of these prints to be outside of the defer?

kylelemons (changes requested):

detach/detach_example_test.go
pacejackson: No, the point was to show that it wasn't closed even if the one on the original context was and then to show that it was closed.

log/context.go
fishy: then at least we should demonstrate a good example here to avoid code duplication :) something like:
```go
func init() {
  copyContext := func(dst, src context.Context) context.Context {
    ...
  }
  detach.Register(detach.Hooks{
    Inline: func(dst, src context.Context) context.Context {
      return copyContext(dst, src)
    },
    Async: func(dst, src context.Context, next func(ctx context.Context)) {
      next(copyContext(dst, src))
    },
  })
}
```

fishy: random idea: what if we also add the hooks itself to `Async`'s arg list? this way it could be this instead:
```go
detach.Register(detach.Hooks{
  Inline: func(dst, src context.Context) context.Context {
    ...
  },
  Async: func(dst, src context.Context, hook detach.Hooks, next func(ctx context.Context)) {
    next(hook.Inline(dst, src))
  },
})
```
we can also provide that `Async` implementation from `detach` library directly as that works for a good amount of packages, so they only need to implement `Inline` and just use the pre-provided `Async`.

tracing/tracer.go
```diff
@@ -43,13 +43,20 @@ func init() {
 			return opentracing.ContextWithSpan(dst, opentracing.SpanFromContext(src))
 		},
 		Async: func(dst, src context.Context, next func(ctx context.Context)) {
-			span, _ := opentracing.StartSpanFromContext(
-				src,
+			span := AsSpan(opentracing.SpanFromContext(src))
+			if span == nil {
```
fishy, line 47: `AsSpan` would never return `nil`, it's a documented feature 😅 (https://pkg.go.dev/github.com/reddit/baseplate.go/tracing#AsSpan)

but it should just work if you remove `AsSpan()`.

tracing/tracer.go
```diff
@@ -43,13 +43,20 @@ func init() {
...
+			}
+
+			child := opentracing.GlobalTracer().StartSpan(
 				"asyncTask",
+				opentracing.ChildOf(span),
 				SpanTypeOption{Type: SpanTypeLocal},
 			)
-			defer span.Finish()
-			next(opentracing.ContextWithSpan(dst, span))
+			defer child.Finish()
```
fishy, line 57: we would still want to add ctx to finish options:
```suggestion
			defer child.FinishWithOptions(tracing.FinishOptions{
				Ctx: dst,
			}.Convert())
```

log/context.go
```diff
@@ -15,18 +15,16 @@ type contextKeyType struct{}
...
-			}
-			return dst
-		},
+		Inline: copyContext,
 		Async: func(dst, src context.Context, next func(ctx context.Context)) {
-			if logger, ok := src.Value(contextKey).(*zap.SugaredLogger); ok && logger != nil {
-				dst = context.WithValue(dst, contextKey, logger)
-			}
-			next(dst)
+			next(copyContext(dst, src))
```
kylelemons, line 27: should we make this the default behavior if async is nil?

kylelemons (approved):

log/context.go
fishy: default behavior could cause footgun (as packages need to put in an explicit do-nothing `Async` here if they can't copy anything in `Async`), but I proposed to provide an implementation packages can directly reuse in https://github.com/reddit/baseplate.go/pull/446#discussion_r762344054

fishy: but default behavior does have the benefit of not adding hook to the function signature, so it might be worth it :)

pacejackson: > should we make this the default behavior if async is nil?

I don't think so, like yuxuan said, that might not be safe. I really want this part to be explicit since it's really only used by a small number of libraries in framework code so I would rather force that part to be as direct as possible which is what I like about the current interface it provides.

tracing/tracer.go
pacejackson: I fixed this, note I still use `AsSpan` below because our `Span` implements `SpanContext` which is what `ChildOf` wants.

detach/detach_example_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/reddit/baseplate.go/detach"
+)
+
+func init() {
```
fishy, line 11: since `init` function is actually pretty special in go, I'm not sure if an `init` function inside a `_test.go` with `_test` package name will cause it to be executed when the package is imported. I'm 90% sure it should not, but it still makes me extra nervous.

I would prefer to split the 2 examples into 2 different files, and rename the `init` function into `InitInline`/`InitAsync` and call from the `Example*` function explicitly, with comment explaining that in real code they should just be named `init` without explicit call instead. This way the 2 examples will be rendered as full file examples so people can actually see how the init function is written (the current way pkg.go.dev will only render the `Example*` functions and hide everything else in this file).

detach/detach.go
```diff
@@ -0,0 +1,106 @@
...
+
+	r.hooks = append(r.hooks, hooks)
+}
+
+func (r *hooksRegistry) allHooks() []Hooks {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	return r.hooks
+}
```
fishy, line 49: wait I don't think returning `r.hooks` and release the lock makes it safe to be used concurrently? to make it truly thread-safe we likely need to make a shallow copy here, e.g.:
```go
hooks := make([]Hooks, len(r.hooks))
copy(hooks, r.hooks)
return hooks
```

but in reality we expect all the `add` calls happen on initialization so this is not that big a deal.

fishy (approved):

detach/detach.go
pacejackson: I could have sworn this already came up somewhere and I was told not to worry about copying it but I can't find the comment....... 🤔

fishy (approved):

detach/detach.go
pacejackson: ah, @kylelemons and I talked about this outside of the PR,  his point was that since we are just appending, it would be safe to just return it. Thinking about it more, it _seems_ like a good idea to me to not leave that potential footgun around though?

PR #447, opened 2021-12-06
Add prometheus metrics for unary grpc client and server.

JessicaGreben:
This PR adds Prometheus metrics for unary GRPC client and server middleware based on the [baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#grpc).

Streaming support is added here: https://github.com/reddit/baseplate.go/pull/459

Background:

- this PR for thrift server metric with some testing helpers (PR [#433](https://github.com/reddit/baseplate.go/pull/433)).
- another PR for adding thrift client metrics (PR #[442](https://github.com/reddit/baseplate.go/pull/442) )
- another PR that adds spec-validation tests (PR #[445](https://github.com/reddit/baseplate.go/pull/445))
- This PR -  PR that adds grpc client and server prometheus metrics  (PR #[447](https://github.com/reddit/baseplate.go/pull/447))
- a PR that adds http client and server prometheus metrics (PR #451)


grpcbp/client_middlewares.go
```diff
@@ -101,3 +105,134 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+			rpcCountLabels := prometheus.Labels{
+				localServiceLabel:      serviceSlug,
+				methodLabel:            m,
+				typeLabel:              clientStream,
+				successLabel:           success,
+				remoteServiceSlugLabel: serverSlug,
+				codeLabel:              status.Code().String(),
+			}
+			clientRPCRequestCounter.With(rpcCountLabels).Inc()
+			clientActiveRequests.With(activeRequestLabels).Dec()
```
fishy, line 234: I don't think the return of `StreamClientInterceptor` means the request handling is done? it returns a `ClientStreamer`, and my understanding is that the actual request handling happens inside the returned `ClientStreamer`?

grpcbp/server_middlewares_test.go
```diff
@@ -179,5 +182,203 @@ func (t *mockService) PingList(req *pb.PingRequest, c pb.TestService_PingListSer
...
+	}
+}
+
+func TestInjectPrometheusStreamServerClientInterceptor(t *testing.T) {
+	const (
+		serviceName = "testSvc"
+		serverName  = "testServer"
+		method      = "Ping"
+	)
+	// create test server with InjectPrometheusStreamerverInterceptor
```
fishy, line 299: typo: 
```suggestion
	// create test server with InjectPrometheusStreamServerInterceptor
```

grpcbp/server_middlewares.go
```diff
@@ -142,3 +146,113 @@ func StartSpanFromGRPCContext(ctx context.Context, name string) (context.Context
...
+
+			rpcCountLabels := prometheus.Labels{
+				localServiceLabel: serviceSlug,
+				methodLabel:       method,
+				typeLabel:         serverStream,
+				successLabel:      success,
+				codeLabel:         status.Code().String(),
+			}
+			serverRPCRequestCounter.With(rpcCountLabels).Inc()
+			serverActiveRequests.With(activeRequestLabels).Dec()
```
fishy, line 253: same here, I don't think the request handling is done here.

fishy (changes requested):

grpcbp/client_middlewares.go
JessicaGreben: yeah you are right, this returns a `grpc.ClientStream` so the defer won't execute at the correct time. Let me fix that.

grpcbp/server_middlewares.go
JessicaGreben: i've been testing this more to confirm and it does appear the request handling is done here when `handler()` is called. I've been reading the grpc code to confirm, but I was struggling finding the specific code. So in the meantime, I'm adding more test cases to fully convince myself.

grpcbp/client_middlewares.go
```diff
@@ -101,3 +106,191 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+//   - grpc_service, grpc_method, grpc_slug
+//   - grpc_success: "true" if status is OK, "false" otherwise
+//   - grpc_type: type of request, i.e unary
+//   - grpc_code: the human-readable status code, e.g. OK, Internal, etc
+func PrometheusUnaryClientInterceptor(serviceSlug, serverSlug string) grpc.UnaryClientInterceptor {
+	return func(
+		ctx context.Context,
+		method string,
+		req interface{},
+		reply interface{},
```
kylelemons, line 133: 🔕 the opposite of req is resp usually

grpcbp/client_middlewares.go
```diff
@@ -101,3 +106,191 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+	err := s.ClientStream.RecvMsg(m)
+	if err != nil {
+		if err == io.EOF {
+			s.trackMetrics(nil)
+		} else {
+			s.trackMetrics(err)
+		}
+	}
+	return err
+}
```
kylelemons, line 256: 🔕 WDYT about

```suggestion
func (s *monitoredClientStream) RecvMsg(m interface{}) (err error) {
  defer func() {
    if err != nil {
      s.trackMetrics(err)
    }
  }()
	return s.ClientStream.RecvMsg(m)
}
```

and then make trackMetrics understand io.EOF?

grpcbp/client_middlewares.go
```diff
@@ -101,3 +106,191 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+
+func (s *monitoredClientStream) ActiveRequestsInc() {
```
kylelemons, line 258: can this be unexported?

Also 🔕 `incActiveRequests` so it has the verb at the beginning?

grpcbp/client_middlewares.go
```diff
@@ -101,3 +106,191 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+	activeRequestLabels := prometheus.Labels{
+		localServiceLabel:      s.serviceSlug,
+		methodLabel:            s.method,
+		remoteServiceSlugLabel: s.serverSlug,
+	}
+	clientActiveRequests.With(activeRequestLabels).Inc()
+}
+
+func (s *monitoredClientStream) trackMetrics(err error) {
+	success := strconv.FormatBool(err == nil)
```
kylelemons, line 268: (continuing above) `err == io.EOF`?

grpcbp/client_middlewares.go
```diff
@@ -101,3 +106,191 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+	status, _ := status.FromError(err)
```
kylelemons, line 269: it's not worth skipping the `, ok` check -- even though it shouldn't ever be false, if it ever is it'll be much easier to figure out what's going on if we have an explicit check.

grpcbp/server_middlewares_test.go
```diff
@@ -178,6 +183,261 @@ func (t *mockService) PingEmpty(ctx context.Context, req *pb.Empty) (*pb.PingRes
...
+						}
+					}
+					if err == nil {
+						t.Fatal("clientStream Recv: expected err got nil")
+					}
+				}
+
+				if err != nil && tt.success == "true" {
+					t.Fatalf("clientStream Recv: %v", err)
+				}
```
kylelemons, line 438: minor suggestion for improving clarity... I think you can structure it into two conditionals instead of having to rely on nesting:

```
_, err := ...
// ensure that tt.success is set correctly
if got, want := tt.success, fmt.Sprint(err == nil || err == io.EOF); got != want {
  t.Errorf("tt.success = %q, want %q (err: %v)")
}
// handle the actual error
if err != nil {
  // ...
  break
}
count++

```

kylelemons (approved):

grpcbp/client_middlewares.go
JessicaGreben: looks like in grpc they say `reply`, ref: https://pkg.go.dev/google.golang.org/grpc#UnaryClientInterceptor

grpcbp/client_middlewares.go
JessicaGreben: i like this better. will do.

grpcbp/client_middlewares.go
```diff
@@ -101,3 +106,191 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+			serverSlug,
+		)
+		return monitoredStream, nil
+	}
+}
+
+// monitoredClientStream wraps grpc.ClientStream to track Prometheus metrics when streaming starts and finishes.
+type monitoredClientStream struct {
+	grpc.ClientStream
+	start       time.Time
```
fishy, line 221: style nit: add a blank line between embedded, exported, and unexported fields.

grpcbp/client_middlewares.go
```diff
@@ -101,3 +106,191 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+}
+
+func newMonitoredClientStream(stream grpc.ClientStream, method, serviceSlug, serverSlug string) *monitoredClientStream {
+	s := &monitoredClientStream{
+		stream,
+		time.Now(),
+		method,
+		serviceSlug,
+		serverSlug,
+	}
```
fishy, line 234: style nit: use keyed init for readability:
```suggestion
	s := &monitoredClientStream{
		ClientStream: stream,

		start: time.Now(),
		method: method,
		serviceSlug: serviceSlug,
		serverSlug: serverSlug,
	}
```

(🔕 also I still find the names `serviceSlug` vs. `serverSlug` confusing :( )

grpcbp/client_middlewares.go
```diff
@@ -101,3 +106,191 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+	s.ActiveRequestsInc()
```
fishy, line 235: I don't see the corresponding dec active requests for streaming clients. from the reading of the doc I _think_ it should be done with `CloseSend`?

grpcbp/client_middlewares.go
JessicaGreben: will fix serviceSlug vs. serverSlug

grpcbp/client_middlewares.go
JessicaGreben: what should i do if its `!ok`? log something? or set the status to something else?

JessicaGreben:
I moved streaming support into a different PR: https://github.com/reddit/baseplate.go/pull/459

fishy:
oh no

you merged this before you merged the cherry-pick pr, so the squashed pr also contains everything in the cherry-pick PR.

fishy:
I force pushed `prometheus` branch to remove the merged commit (c9b6e86), please recreate this PR.

JessicaGreben:
@fishy oops. sorry. thanks for the fix. i will recreate.

PR #448, opened 2021-12-09
errorsbp: Add BatchSize

fishy:
This is helpful in unit tests to verify that a function returned the
exact number of errors in a batch as expected.

Also add Batch.Len() helper function.

### DIFF
--- a/errorsbp/batch.go
+++ b/errorsbp/batch.go
@@ -40,6 +40,11 @@ func (be Batch) Error() string {
 	return sb.String()
 }
 
+// Len returns the size of the batch.
+func (be Batch) Len() int {
+	return len(be.errors)
+}
+
 // As implements helper interface for errors.As.
 //
 // If v is pointer to either Batch or *Batch,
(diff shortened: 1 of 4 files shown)


fishy:
cc @JessicaGreben

konradreiche (approved):

errorsbp/batch.go
```diff
@@ -209,3 +214,23 @@ func (e *prefixedError) Error() string {
...
+//
+// It's useful in tests,
+// for example to verify that a function indeed returns the exact number of
+// errors that's expected to be returned.
+func BatchSize(err error) int {
+	var be Batch
+	if errors.As(err, &be) {
+		return be.Len()
+	}
+	if err == nil {
```
pacejackson, line 232: 🔕  shouldn't we do this first?

fishy: `errors.As` works as expected for `nil` errors, but sure :)

pacejackson (approved):

PR #449, opened 2021-12-15
Suppress non-5xx HTTPErrors in httpbp.InjectServerSpan

pacejackson:
This will check if the returned error is an HTTPError and will
suppress it if it's not a 5xx error or higher.

### DIFF
--- a/httpbp/middlewares.go
+++ b/httpbp/middlewares.go
@@ -58,7 +58,7 @@ type DefaultMiddlewareArgs struct {
 	EdgeContextImpl ecinterface.Interface
 }
 
-// DefaultMiddleware returns a slice of all of the default Middleware for a
+// DefaultMiddleware returns a slice of all the default Middleware for a
 // Baseplate HTTP server.
 func DefaultMiddleware(args DefaultMiddlewareArgs) []Middleware {
 	if args.TrustHandler == nil {
@@ -110,6 +110,8 @@ func StartSpanFromTrustedRequest(
 	return tracing.StartSpanFromHeaders(ctx, name, spanHeaders)
 }
 
+// HTTPErrorSuppressor is an errorsbp.Suppressor that can be used to suppress
+// HTTPErrors that have a response code under 500.
 func HTTPErrorSuppressor(err error) bool {
 	var httpErr HTTPError
 	if errors.As(err, &httpErr) {
@@ -119,13 +121,19 @@ func HTTPErrorSuppressor(err error) bool {
 }
 
 // InjectServerSpan returns a Middleware that will automatically wrap the
-// HansderFunc in a new server span and stop the span after the function
+// HandlerFunc in a new server span and stop the span after the function
 // returns.
+//
+// Starts the server span before calling the `next` HandlerFunc and stops
+// the span after it finishes.
+// If the function returns an error that's not suppressed by HTTPErrorSuppressor,
+// that will be passed to span.Stop.
 //
 // InjectServerSpan should generally not be used directly, instead use the
 // NewBaseplateServer function which will automatically include InjectServerSpan
 // as one of the Middlewares to wrap your handlers in.
 func InjectServerSpan(truster HeaderTrustHandler) Middleware {
+	// TODO: make a breaking change to allow us to pass in a Suppressor
 	var suppressor errorsbp.Suppressor = HTTPErrorSuppressor
 	return func(name string, next HandlerFunc) HandlerFunc {
 		return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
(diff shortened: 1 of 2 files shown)


httpbp/middlewares.go
```diff
@@ -108,21 +110,32 @@ func StartSpanFromTrustedRequest(
...
+
 // InjectServerSpan returns a Middleware that will automatically wrap the
 // HansderFunc in a new server span and stop the span after the function
 // returns.
 //
 // InjectServerSpan should generally not be used directly, instead use the
 // NewBaseplateServer function which will automatically include InjectServerSpan
 // as one of the Middlewares to wrap your handlers in.
-func InjectServerSpan(truster HeaderTrustHandler) Middleware {
+func InjectServerSpan(truster HeaderTrustHandler, suppressor errorsbp.Suppressor) Middleware {
```
fishy, line 128: this is a breaking change. do we have a way to do it non-breaking? I think we can just always use `HTTPErrorSuppressor` in this version (don't change the API, but change the behavior slightly), and allow customization in a follow-up PR that's a breaking change (and we'll wait to merge that breaking change PR).

fishy (changes requested):

httpbp/middlewares.go
```diff
@@ -108,6 +110,14 @@ func StartSpanFromTrustedRequest(
 	return tracing.StartSpanFromHeaders(ctx, name, spanHeaders)
 }
 
+func HTTPErrorSuppressor(err error) bool {
```
fishy, line 113: Please add doc comment for this function.

httpbp/middlewares.go
```diff
@@ -116,13 +126,14 @@ func StartSpanFromTrustedRequest(
 // NewBaseplateServer function which will automatically include InjectServerSpan
 // as one of the Middlewares to wrap your handlers in.
```
fishy, line 127: Please also update the doc of this function to mention that we auto apply `HTTPErrorSuppressor` here.

httpbp/middlewares.go
```diff
@@ -116,13 +126,14 @@ func StartSpanFromTrustedRequest(
...
 func InjectServerSpan(truster HeaderTrustHandler) Middleware {
+	var suppressor errorsbp.Suppressor = HTTPErrorSuppressor
```
fishy, line 129: 🔕 probably worth adding a TODO here? don't feel strongly either way, though.

fishy (approved):
:fish: after fixed the comments.

fishy (approved):

httpbp/middlewares.go
kylelemons: This doc comment just says what it does, not why I would use it.

"HTTPErrorSuppressor causes metrics to ignore non-5xx errors.  This should be used when dashboards
relying on the builtin metrics assume that client-driven failure modes are excluded."

or something.

That being said, what about just having a new middleware that exposes that 5xx-rate separately, so that we expose both, and then dashboards can use whichever makes the most sense for the service?

httpbp/middlewares_test.go
```diff
@@ -121,17 +135,17 @@ func TestInjectServerSpan(t *testing.T) {
...
 					}
 				}
+				if suppressor.Wrap(c.err) != nil && !hasError {
+					t.Error("error binary annotation was not present.")
+				} else if suppressor.Wrap(c.err) == nil && hasError {
+					t.Error("unexpected error binary annotation")
+				}
 			},
 		)
 	}
```
kylelemons, line 151: can we get an Example of how to use it

kylelemons (changes requested):

httpbp/middlewares.go
pacejackson: > That being said, what about just having a new middleware that exposes that 5xx-rate separately, so that we expose both, and then dashboards can use whichever makes the most sense for the service?

We could (and it's probably a good idea to) add that, but I think we still want this since this affects what get's reported as a "success" or "failure" in the baseplate endpoint metrics.

pacejackson: maybe we should also just un-export this for now since it really isn't something you would use currently, we would export it if we had you pass in an `errorsbp.Suppressor` though.

fishy: yea +1 to unexport it. for http errors server errors and client errors are also different (and this suppressor only works for server errors) which makes the public api more complicated. we can export it later when needed.

httpbp/middlewares_test.go
pacejackson: Did you want an example of using the `HTTPErrorSuppressor`? Since that's unexported now we probably don't need it?

kylelemons (approved):

PR #450, opened 2021-12-16
Add panic recovery middleware to httpbp

pacejackson:
While recovering from panics isn't necessary to avoid the server crashing like it is in thrift, the monitoring from baseplate.go is blind to panics in httpbp. This change adds an unexported middleware that is always applied to http requests that will recover from any panics and bubble those up the call as an HTTPError.

### DIFF
--- a/httpbp/middlewares.go
+++ b/httpbp/middlewares.go
@@ -263,10 +263,9 @@ func recoverPanic(name string, next HandlerFunc) HandlerFunc {
 				} else {
 					rErr = fmt.Errorf("panic in %q: %+v", name, r)
 				}
-				log.ErrorWithSentry(
-					ctx,
+				log.C(ctx).Errorw(
 					"recovered from panic:",
-					rErr,
+					"err", rErr,
 					"endpoint", name,
 				)
 				metricsbp.M.Counter("panic.recover").With(
--- a/httpbp/middlewares.go
+++ b/httpbp/middlewares.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/reddit/baseplate.go/ecinterface"
 	"github.com/reddit/baseplate.go/log"
+	"github.com/reddit/baseplate.go/metricsbp"
 	"github.com/reddit/baseplate.go/tracing"
 )
 
@@ -244,5 +245,38 @@ func SupportedMethods(method string, additional ...string) Middleware {
 			}
 			return next(ctx, w, r)
 		}
+	}
+}
+
+// recoverPanic recovers from any panics, logs them, and sets the returned error
+// to a generic 500 error. recoverPanic is always the last middleware in the
+// middleware chain, so it is the first one when returning, this lets the error
+// bubble up into other middlewares. Since it is always added to the middleware
+// chain is a specific position, it is not exported.
+func recoverPanic(name string, next HandlerFunc) HandlerFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
+		defer func() {
+			if r := recover(); r != nil {
+				var rErr error
+				if asErr, ok := r.(error); ok {
+					rErr = asErr
+				} else {
+					rErr = fmt.Errorf("panic in %q: %+v", name, r)
+				}
+				log.ErrorWithSentry(
+					ctx,
+					"recovered from panic:",
+					rErr,
+					"endpoint", name,
+				)
+				metricsbp.M.Counter("panic.recover").With(
+					"name", name,
+				).Add(1)
+
+				// change named return value to a generic 500 error
+				err = RawError(InternalServerError(), rErr, PlainTextContentType)
+			}
+		}()
+		return next(ctx, w, r)
 	}
 }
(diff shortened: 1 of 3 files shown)


httpbp/server.go
```diff
@@ -45,10 +45,14 @@ type httpHandlerFactory struct {
 }
 
 func (f httpHandlerFactory) NewHandler(endpoint Endpoint) http.Handler {
-	wrappers := make([]Middleware, 0, len(f.middlewares)+len(endpoint.Middlewares)+1)
+	wrappers := make([]Middleware, 0, len(f.middlewares)+len(endpoint.Middlewares)+2)
```
fishy, line 48: probably should add a comment to explain the magic number of 2 😅 (it's for `SupportedMethods` + `recoverPanic`, if I'm reading it correctly)

fishy (approved):

httpbp/server.go
JessicaGreben: personal preference, for magic numbers I think its best to use a `const` and comment the const definition.

httpbp/middlewares.go
```diff
@@ -246,3 +247,36 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
+		defer func() {
+			if r := recover(); r != nil {
+				var rErr error
+				if asErr, ok := r.(error); ok {
+					rErr = asErr
+				} else {
+					rErr = fmt.Errorf("panic in %q: %+v", name, r)
+				}
+				log.ErrorWithSentry(
```
kylelemons, line 266: I'm not a fan of adding new default ways to deplete our sentry quota from backend services.  This should just log the stack trace and increment a metric imo, not hit sentry.

kylelemons (changes requested):

httpbp/middlewares.go
pacejackson: I dunno, unexpected panics seems like the place where we would _want_ to send the error to sentry?

httpbp/middlewares.go
```diff
@@ -246,3 +247,35 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
+		defer func() {
+			if r := recover(); r != nil {
+				var rErr error
+				if asErr, ok := r.(error); ok {
+					rErr = asErr
+				} else {
+					rErr = fmt.Errorf("panic in %q: %+v", name, r)
+				}
+				log.C(ctx).Errorw(
```
kylelemons, line 266: just to triple check, the default logger logs call stacks for errors right?

httpbp/middlewares.go
```diff
@@ -246,3 +247,35 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+// recoverPanic recovers from any panics, logs them, and sets the returned error
+// to a generic 500 error. recoverPanic is always the last middleware in the
+// middleware chain, so it is the first one when returning which lets the error
+// bubble up into other middlewares. Since it is always added to the middleware
+// chain is a specific position, it is not exported.
+func recoverPanic(name string, next HandlerFunc) HandlerFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
+		defer func() {
+			if r := recover(); r != nil {
+				var rErr error
```
kylelemons, line 260: Actually, why do we need to convert this to an error?  "endpoint" is logged as a keyval, so adding the "panic in %q" is redundant.  Can we not just do

```
log.C(ctx).Errorw("recovered from panic",
  "panic", r,
  "endpoint", name,
)
```

Presumably the logger will stringify or jsonify properly?

httpbp/server_test.go
```diff
@@ -392,3 +394,72 @@ func TestNewTestBaseplateServer(t *testing.T) {
...
+								)
+							}
+							if !errors.Is(httpErr, panicErr) {
+								t.Errorf("expected http error to wrap %v, got %v", panicErr, httpErr.Unwrap())
+							}
+						} else {
+							t.Fatalf("unexpected error type %v", err)
+						}
+					}()
+					return next(ctx, w, r)
```
kylelemons, line 445: can you wrap the context in one that will allow you to verify that the right keyvals are logged?

kylelemons (approved):

httpbp/middlewares.go
fishy: yes, it adds stack trace on error and above levels.

httpbp/middlewares.go
fishy: it will try to auto jsonify it for `interface{}`, which might not be the desired outcome if it's an error (which would be the majority of the errors), so I think trying to convert it into an error can still be useful. we could use `r` directly if it's not an error, instead of trying to use `fmt.Errorf`, though.

pacejackson: We still want it to be an error for when we update the returned `err` value.

httpbp/server_test.go
pacejackson: Do we really have a way to test logging?

fishy: we actually have, but it's quite involved and not recommended unless you are writing some core logging feature: https://github.com/reddit/baseplate.go/blob/master/log/zap_wrapper_test.go

PR #451, opened 2021-12-16
Add http prometheus metrics for client and server

JessicaGreben:
This PR adds Prometheus metrics for HTTP client and server based on the [baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#http).

Background:

### DIFF
--- a/grpcbp/server_middlewares_test.go
+++ b/grpcbp/server_middlewares_test.go
@@ -187,18 +187,18 @@ func (t *mockService) PingStream(c pb.TestService_PingStreamServer) error {
 
 func TestInjectPrometheusUnaryServerClientInterceptor(t *testing.T) {
 	const (
-		serviceName = "com.example.reddit.preference.PreferenceService"
-		serverName  = "example-preference-server"
+		serviceName = "mwitkow.testproto.TestService"
+		serverSlug  = "example-preference-server"
 		method      = "Ping"
 	)
 	// create test server with InjectPrometheusUnaryServerInterceptor
 	l, service := setupServer(t, grpc.UnaryInterceptor(
-		InjectPrometheusUnaryServerInterceptor(serviceName),
+		InjectPrometheusUnaryServerInterceptor(),
 	))
 
 	// create test client
 	conn := setupClient(t, l, grpc.WithUnaryInterceptor(
-		PrometheusUnaryClientInterceptor(serviceName, serverName),
+		PrometheusUnaryClientInterceptor(serverSlug),
 	))
 
 	// instantiate gRPC client
@@ -237,46 +237,46 @@ func TestInjectPrometheusUnaryServerClientInterceptor(t *testing.T) {
 			clientActiveRequests.Reset()
 
 			serverLatencyLabels := prometheus.Labels{
-				localServiceLabel: serviceName,
-				methodLabel:       tt.method,
-				typeLabel:         unary,
-				successLabel:      tt.success,
+				serviceLabel: serviceName,
+				methodLabel:  tt.method,
+				typeLabel:    unary,
+				successLabel: tt.success,
 			}
 
 			serverTotalRequestLabels := prometheus.Labels{
-				localServiceLabel: serviceName,
-				methodLabel:       tt.method,
-				typeLabel:         unary,
-				successLabel:      tt.success,
-				codeLabel:         tt.code,
+				serviceLabel: serviceName,
+				methodLabel:  tt.method,
+				typeLabel:    unary,
+				successLabel: tt.success,
+				codeLabel:    tt.code,
 			}
 
 			serverActiveRequestLabels := prometheus.Labels{
-				localServiceLabel: serviceName,
-				methodLabel:       tt.method,
+				serviceLabel: serviceName,
+				methodLabel:  tt.method,
 			}
 
 			clientLatencyLabels := prometheus.Labels{
-				localServiceLabel:      serviceName,
-				methodLabel:            tt.method,
-				typeLabel:              unary,
-				successLabel:           tt.success,
-				remoteServiceSlugLabel: serverName,
+				serviceLabel:    serviceName,
+				methodLabel:     tt.method,
+				typeLabel:       unary,
+				successLabel:    tt.success,
+				serverSlugLabel: serverSlug,
 			}
 
 			clientTotalRequestLabels := prometheus.Labels{
-				localServiceLabel:      serviceName,
-				methodLabel:            tt.method,
-				typeLabel:              unary,
-				successLabel:           tt.success,
-				remoteServiceSlugLabel: serverName,
-				codeLabel:              tt.code,
+				serviceLabel:    serviceName,
+				methodLabel:     tt.method,
+				typeLabel:       unary,
+				successLabel:    tt.success,
+				serverSlugLabel: serverSlug,
+				codeLabel:       tt.code,
 			}
 
 			clientActiveRequestLabels := prometheus.Labels{
-				localServiceLabel:      serviceName,
-				methodLabel:            tt.method,
-				remoteServiceSlugLabel: serverName,
+				serviceLabel:    serviceName,
+				methodLabel:     tt.method,
+				serverSlugLabel: serverSlug,
 			}
 
 			defer promtest.NewPrometheusMetricTest(t, "server latency", serverLatencyDistribution, serverLatencyLabels).CheckExists()
--- a/grpcbp/server_middlewares_test.go
+++ b/grpcbp/server_middlewares_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	pb "github.com/grpc-ecosystem/go-grpc-middleware/testing/testproto"
+	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
@@ -186,18 +187,18 @@ func (t *mockService) PingStream(c pb.TestService_PingStreamServer) error {
 
 func TestInjectPrometheusUnaryServerClientInterceptor(t *testing.T) {
 	const (
-		serviceName = "testSvc"
-		serverName  = "testServer"
-		method      = "Ping"
+		localService      = "testSvc"
+		remoteServiceSlug = "testServer"
+		method            = "Ping"
 	)
 	// create test server with InjectPrometheusUnaryServerInterceptor
 	l, service := setupServer(t, grpc.UnaryInterceptor(
-		InjectPrometheusUnaryServerInterceptor(serviceName),
+		InjectPrometheusUnaryServerInterceptor(localService),
 	))
 
 	// create test client
 	conn := setupClient(t, l, grpc.WithUnaryInterceptor(
-		PrometheusUnaryClientInterceptor(serviceName, serverName),
+		PrometheusUnaryClientInterceptor(localService, remoteServiceSlug),
 	))
 
 	// instantiate gRPC client
@@ -235,55 +236,55 @@ func TestInjectPrometheusUnaryServerClientInterceptor(t *testing.T) {
 			clientRPCRequestCounter.Reset()
 			clientActiveRequests.Reset()
 
-			serverLabelValues := []string{
-				serviceName,
-				tt.method,
-				unary,
-				tt.success,
-				tt.code,
+			serverLatencyLabels := prometheus.Labels{
+				localServiceLabel: localService,
+				methodLabel:       tt.method,
+				typeLabel:         unary,
+				successLabel:      tt.success,
 			}
 
-			serverLatencyLabelValues := []string{
-				serviceName,
-				tt.method,
-				unary,
-				tt.success,
+			serverTotalRequestLabels := prometheus.Labels{
+				localServiceLabel: localService,
+				methodLabel:       tt.method,
+				typeLabel:         unary,
+				successLabel:      tt.success,
+				codeLabel:         tt.code,
 			}
 
-			serverRequestsLabelValues := []string{
-				serviceName,
-				tt.method,
+			serverActiveRequestLabels := prometheus.Labels{
+				localServiceLabel: localService,
+				methodLabel:       tt.method,
 			}
 
-			clientLabelValues := []string{
-				serviceName,
-				tt.method,
-				unary,
-				tt.success,
-				tt.code,
-				serverName,
+			clientLatencyLabels := prometheus.Labels{
+				localServiceLabel:      localService,
+				methodLabel:            tt.method,
+				typeLabel:              unary,
+				successLabel:           tt.success,
+				remoteServiceSlugLabel: remoteServiceSlug,
 			}
 
-			clientLatencyLabelValues := []string{
-				serviceName,
-				tt.method,
-				unary,
-				tt.success,
-				serverName,
+			clientTotalRequestLabels := prometheus.Labels{
+				localServiceLabel:      localService,
+				methodLabel:            tt.method,
+				typeLabel:              unary,
+				successLabel:           tt.success,
+				remoteServiceSlugLabel: remoteServiceSlug,
+				codeLabel:              tt.code,
 			}
 
-			clientRequestsLabelValues := []string{
-				serviceName,
-				tt.method,
-				serverName,
+			clientActiveRequestLabels := prometheus.Labels{
+				localServiceLabel:      localService,
+				methodLabel:            tt.method,
+				remoteServiceSlugLabel: remoteServiceSlug,
 			}
 
-			defer promtest.NewPrometheusMetricTest(t, "server latency", serverLatencyDistribution, serverLatencyLabelValues...).CheckExists()
-			defer promtest.NewPrometheusMetricTest(t, "server rpc count", serverRPCRequestCounter, serverLabelValues...).CheckDelta(1)
-			defer promtest.NewPrometheusMetricTest(t, "server active requests", serverActiveRequests, serverRequestsLabelValues...).CheckDelta(0)
-			defer promtest.NewPrometheusMetricTest(t, "client latency", clientLatencyDistribution, clientLatencyLabelValues...).CheckExists()
-			defer promtest.NewPrometheusMetricTest(t, "client rpc count", clientRPCRequestCounter, clientLabelValues...).CheckDelta(1)
-			defer promtest.NewPrometheusMetricTest(t, "client active requests", clientActiveRequests, clientRequestsLabelValues...).CheckDelta(0)
+			defer promtest.NewPrometheusMetricTest(t, "server latency", serverLatencyDistribution, serverLatencyLabels).CheckExists()
+			defer promtest.NewPrometheusMetricTest(t, "server rpc count", serverRPCRequestCounter, serverTotalRequestLabels).CheckDelta(1)
+			defer promtest.NewPrometheusMetricTest(t, "server active requests", serverActiveRequests, serverActiveRequestLabels).CheckDelta(0)
+			defer promtest.NewPrometheusMetricTest(t, "client latency", clientLatencyDistribution, clientLatencyLabels).CheckExists()
+			defer promtest.NewPrometheusMetricTest(t, "client rpc count", clientRPCRequestCounter, clientTotalRequestLabels).CheckDelta(1)
+			defer promtest.NewPrometheusMetricTest(t, "client active requests", clientActiveRequests, clientActiveRequestLabels).CheckDelta(0)
 
 			ctx := context.Background()
 			if tt.success == "true" {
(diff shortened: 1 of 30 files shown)


httpbp/middlewares.go
```diff
@@ -246,3 +249,47 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+				methodLabel:   method,
+				endpointLabel: endpoint,
+			}
+			serverActiveRequests.With(activeRequestLabels).Inc()
+
+			defer func() {
+				success := strconv.FormatBool(err == nil)
+				code := http.StatusText(http.StatusOK)
+				if err != nil {
+					code = err.Error()
```
JessicaGreben, line 269: anyone know a better way to get the status code here that will be used for `http_response_code` label? ref: [baseplate spec for http](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#http-server-metrics).

kylelemons: you'll need to create a ResponseWriter that intercepts it.

```
type codeInterceptor struct {
  http.ResponseWriter
  code int
}

func (ci *codeInterceptor) WriteHeader(code int) {
  ci.code = code
  ci.ResponseWriter.WriteHeader(code)
}
```

Now you make one of those, stick "w" as the "ResponseWriter" field, pass it to next, and check the .code after it returns.

Note that WriteHeader might not be called, so code might be zero, in which case the implied response code is http.StatusOK

kylelemons (changes requested):

httpbp/middlewares.go
pacejackson: I just did this with statsd metrics as well:

https://github.com/reddit/baseplate.go/pull/453/files#diff-f0f0f7b278ddd73593a069e51e2d62c8b1fcac7d296258c2e724d85d31c8bb9eR278-R291

Note, you also want to check if the error is an `HTTPError` since returning that will set the status code later in the call (after middeware is done): 

https://github.com/reddit/baseplate.go/pull/453/files#diff-f0f0f7b278ddd73593a069e51e2d62c8b1fcac7d296258c2e724d85d31c8bb9eR359-R362

httpbp/middlewares.go
```diff
@@ -246,3 +251,90 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+			code = httpErr.Response().Code
+		} else {
+			code = http.StatusInternalServerError
+		}
+	} else {
+		// if there's no error returned and no call to WriteHeader, assume OK
+		code = http.StatusOK
+	}
+	return code
+}
```
pacejackson, line 340: I like this, since it's just returning the status code, you can also clean it up a little:

```suggestion
func (r *statusCodeRecorder) getCode(err error) int {
	if r.code != 0 {
		// WriteHeader was called explicitly, use that
		return r.code
	}
	if err != nil {
		// something went wrong, check if err is an HTTPErr where we can extract
		// the code, otherwise assume InternalServerError
		var httpErr HTTPError
		if errors.As(err, &httpErr) {
			return httpErr.Response().Code
		}
		return http.StatusInternalServerError

	}
	// if there's no error returned and no call to WriteHeader, assume OK
	return http.StatusOK
}
```

httpbp/middlewares.go
```diff
@@ -246,3 +251,90 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+				}
+				serverLatency.With(labels).Observe(time.Since(start).Seconds())
+				serverRequestSize.With(labels).Observe(float64(r.ContentLength))
+				serverResponseSize.With(labels).Observe(float64(getContentLength(w)))
+
+				totalRequestLabels := prometheus.Labels{
+					methodLabel:   method,
+					successLabel:  success,
+					endpointLabel: endpoint,
+					codeLabel:     strconv.Itoa(code),
```
pacejackson, line 285: Do we want to put the entire code in here or do the `Nxx` thing we do in #453 ?

https://github.com/reddit/baseplate.go/pull/453/files#diff-f0f0f7b278ddd73593a069e51e2d62c8b1fcac7d296258c2e724d85d31c8bb9eR346-R367

JessicaGreben: I'm not sure. The prom baseplate.spec says "numeric status code as a string, e.g. 200" ([ref](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#http-server-metrics)) which seems like it should be the entire code, but for cardinality `Nxx` would be better.  I don't have a strong opinion either way.

pacejackson:
https://github.com/reddit/baseplate.go/pull/453 is merged now

httpbp/prometheus_test.go
```diff
@@ -0,0 +1,185 @@
...
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			serverLatency.Reset()
+			serverTotalRequests.Reset()
+			serverActiveRequests.Reset()
+			clientLatency.Reset()
+			clientTotalRequests.Reset()
+			clientActiveRequests.Reset()
+
+			serverTotalRequestLabels := []string{
```
kylelemons, line 116: I'm starting to think we should swap the APIs to use prometheus.Labels, instead of a slice, since it's a real pain to compare these back and forth with the prometheus.go file and it either automatically detects or we can make it detect errors and give actionable failures. WDYT?

httpbp/prometheus_test.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+
+			if tt.method == http.MethodPost {
+				input := exampleRequest{Input: "foo"}
+				var body bytes.Buffer
+				if err := json.NewEncoder(&body).Encode(input); err != nil {
+					t.Fatal(err)
+				}
+
+				_, err := client.Post(ts.URL+tt.endpoint, "", &body)
```
kylelemons, line 178: if _, err := ...; err != nil

kylelemons (approved):

httpbp/prometheus_test.go
JessicaGreben: good idea. i will make that change.

fishy (changes requested):

httpbp/prometheus.go
```diff
@@ -0,0 +1,115 @@
...
+var (
+	serverLabels = []string{
+		methodLabel,
+		successLabel,
+		endpointLabel,
+	}
+
+	serverLatency = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "http_server_latency_seconds",
+		Help:    "RPC latencies",
```
konradreiche, line 27: nit: I would say HTTP endpoints are not RPCs but just that: HTTP endpoint latencies or just Endpoint latencies.

httpbp/middlewares.go
```diff
@@ -404,3 +407,70 @@ func recordStatusCode(counters counterGenerator) Middleware {
...
+				serverTotalRequests.With(totalRequestLabels).Inc()
+				serverActiveRequests.With(activeRequestLabels).Dec()
+			}()
+
+			return next(ctx, wrapped, r)
+		}
+	}
+}
+
+func getContentLength(w http.ResponseWriter) int64 {
```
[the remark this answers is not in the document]
konradreiche: Avoiding sampling wherever could be useful especially in scenarios where you want to see data in cases when the request rate is low but I am unsure of impact on performance.

httpbp/prometheus.go
```diff
@@ -0,0 +1,115 @@
...
+	clientLatencyLabels = []string{
+		methodLabel,
+		successLabel,
+		endpointLabel,
+		remoteServiceSlugLabel,
+	}
+
+	clientLatency = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "http_client_latency_seconds",
+		Help:    "HTTP latencies",
```
konradreiche, line 88: HTTP request latencies?

httpbp/prometheus_test.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			defer promtest.NewPrometheusMetricTest(t, "server latency", serverLatency, serverSizeLabels).CheckExists()
+			defer promtest.NewPrometheusMetricTest(t, "server total requests", serverTotalRequests, serverTotalRequestLabels).CheckDelta(1)
+			defer promtest.NewPrometheusMetricTest(t, "server active requests", serverActiveRequests, serverActiveRequestLabels).CheckDelta(0)
+			defer promtest.NewPrometheusMetricTest(t, "server request size", serverRequestSize, serverSizeLabels).CheckDelta(float64(tt.size))
+			defer promtest.NewPrometheusMetricTest(t, "server response size", serverResponseSize, serverSizeLabels).CheckDelta(0)
+			defer promtest.NewPrometheusMetricTest(t, "client latency", clientLatency, clientLatencyLabels).CheckExists()
+			defer promtest.NewPrometheusMetricTest(t, "client total requests", clientTotalRequests, clientTotalRequestLabels).CheckDelta(1)
+			defer promtest.NewPrometheusMetricTest(t, "client active requests", clientActiveRequests, clientActiveRequestLabels).CheckDelta(0)
```
konradreiche, line 164: Same question as above.

konradreiche (changes requested):
Only got nits and questions :bow:

httpbp/prometheus.go
JessicaGreben: oops, copy/paste snafu

httpbp/prometheus_test.go
JessicaGreben: i think originally the labels were getting modified later in the test so using a defer was helping use the correct labels. But now that doesn't seem to be occurring in this test.  I don't feel too strongly either way, but I have a slight preference keeping it the way it is now.

httpbp/middlewares.go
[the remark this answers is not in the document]
JessicaGreben: Thanks for the input all. Based on convo, I will wrap `ResponseWriter` via `iobp.CountingSink` to count the bytes written for every request (not sampling). I will try to test performance impact of this and also update the unit tests to confirm behavior (the unit test does currently check, but it asserts the content length is 0, [ref](https://github.com/reddit/baseplate.go/pull/451/files#diff-4e6dfdcedfd78c45607347f564a0d2fc7f1d7a662876d891b91c4156a10b534cR161))

httpbp/client_middlewares.go
```diff
@@ -245,3 +248,61 @@ func MonitorClient(slug string) ClientMiddleware {
...
+			endpoint := req.URL.Path
+			activeRequestLabels := prometheus.Labels{
+				methodLabel:            method,
+				endpointLabel:          endpoint,
+				remoteServiceSlugLabel: serverSlug,
+			}
+			clientActiveRequests.With(activeRequestLabels).Inc()
+
+			defer func() {
+				success := strconv.FormatBool(err == nil && resp.StatusCode == http.StatusOK)
```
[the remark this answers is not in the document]
JessicaGreben: good catch, i will fix this and add all 2xx codes

httpbp/middlewares.go
```diff
@@ -498,21 +499,31 @@ func getContentLength(w http.ResponseWriter) int64 {
...
+	rr.responseCode = code
+}
+
+func (r *responseRecorder) setContentLength() {
+	r.ResponseWriter.Header().Set(contentLengthHeader, strconv.Itoa(r.bytesWritten))
 }
 
-func (r *contentLengthRecorder) setContentLength() {
-	r.ResponseWriter.Header().Set(contentLengthHeader, strconv.Itoa(int(r.Size())))
+func (r *responseRecorder) getCode(err error) int {
```
kylelemons, line 527: Don't leak "business logic" into `responseRecorder`.  It should be very dumb, just tracking what it receives.  Take the code it recorded and use it elsewhere instead.

httpbp/middlewares.go
```diff
@@ -471,7 +472,7 @@ func PrometheusServerMetrics(serverSlug string) Middleware {
 			}()
 
 			n := next(ctx, wrapped, r)
-			reqLenRecorder.setContentLength()
+			wrapped.setContentLength()
```
kylelemons, line 475: Mutating the Headers after the response has been written will have no effect.

kylelemons (changes requested):

konradreiche (approved):

grpcbp/server_middlewares_test.go
```diff
@@ -186,18 +187,18 @@ func (t *mockService) PingStream(c pb.TestService_PingStreamServer) error {
 
 func TestInjectPrometheusUnaryServerClientInterceptor(t *testing.T) {
 	const (
-		serviceName = "testSvc"
-		serverName  = "testServer"
-		method      = "Ping"
+		localService      = "testSvc"
+		remoteServiceSlug = "testServer"
```
kylelemons, line 191: maybe it would be clear if you kept the variable names but swapped the values to to

```suggestion
		serviceName = "com.example.reddit.preference.PreferenceService"
		serverName  = "example-preference-server"
```

httpbp/client_middlewares.go
```diff
@@ -245,3 +248,61 @@ func MonitorClient(slug string) ClientMiddleware {
...
+func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+			start := time.Now()
+			method := req.Method
+			endpoint := req.URL.Path
+			activeRequestLabels := prometheus.Labels{
+				methodLabel:            method,
+				endpointLabel:          endpoint,
+				remoteServiceSlugLabel: serverSlug,
```
kylelemons, line 278: this is mixing "server" vs "service" -- this would be "remoteServerSlugLabel" or probably just "remoteSlugLabel"

httpbp/middlewares.go
```diff
@@ -404,3 +417,102 @@ func recordStatusCode(counters counterGenerator) Middleware {
...
+//
+// * http_server_requests_total counter with all labels above plus:
+//
+//   - http_response_code: numeric status code as a string, e.g. 200
+func PrometheusServerMetrics(serverSlug string) Middleware {
+	return func(name string, next HandlerFunc) HandlerFunc {
+		return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
+			start := time.Now()
+			method := r.Method
+			endpoint := r.URL.Path
```
kylelemons, line 442: So, here is the interesting part of the HTTP metrics... per the spec, this _must_ not just use `r.URL.Path` -- if someone requests `/user/foo` and our endpoint is `/user/{userid}` the endpoint could be `"/user/{userid}"` or whatever, but must not include `foo`.  I believe that what you actually want here is `name`

httpbp/client_middlewares.go
```diff
@@ -245,3 +248,61 @@ func MonitorClient(slug string) ClientMiddleware {
...
+//
+// * http_client_requests_total counter with all labels above plus:
+//
+//   - http_response_code: numeric status code as a string, e.g. 200
+func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+			start := time.Now()
+			method := req.Method
+			endpoint := req.URL.Path
```
kylelemons, line 274: We might need to remove `endpoint` and revisit it later, since I don't think the API can support this presently in a way that doesn't potentially introduce unbounded metrics.

kylelemons (changes requested):

httpbp/middlewares.go
```diff
@@ -362,6 +369,12 @@ func statusCodeFamily(code int) string {
 	return families[family]
 }
 
+// isSuccessStatusCode takes an http status code and returns true if its in the
+// success status code family, i.e "2xx".
+func isSuccessStatusCode(code int) bool {
+	return code/100 == 2
+}
```
fishy, line 376: I think by definition any http code < 400 should be considered as "success" (e.g. a 3xx redirect is not a failure, nor is 101 switch protocol).

@kylelemons ?

httpbp/middlewares.go
```diff
@@ -404,3 +417,102 @@ func recordStatusCode(counters counterGenerator) Middleware {
...
+	http.ResponseWriter
+
+	bytesWritten int
+	responseCode int
+}
+
+func (rr *responseRecorder) Write(b []byte) (n int, err error) {
+	n, err = rr.ResponseWriter.Write(b)
+	rr.bytesWritten += n
+	rr.ResponseWriter.Header().Set(contentLengthHeader, strconv.Itoa(rr.bytesWritten))
```
fishy, line 511: I don't think we want to do this? this actually changes the behavior from what the underlying `http.ResponseWriter.Write` does.

fishy (changes requested):

httpbp/middlewares.go
JessicaGreben: I must have misunderstood...i was trying to implement "I think we should wrap ResponseWriter to actually count the bytes written" from this comment https://github.com/reddit/baseplate.go/pull/451#discussion_r774737561

fishy: the part we need to do here is only to count the bytes, nothing extra.

if `ResponseWriter.Write(b)` writes the content length header, then it's already taken care of on 509 and you do not need to do it again here. if `ResponseWriter.Write(b)` does not write the content length header, then doing it here changes that behavior.

fishy: and I'm like 80% sure when using `ResponseWriter.Write` go actually uses chunked responses without content-length header.

httpbp/client_middlewares.go
JessicaGreben: should i remove the label all together or just set it to `""` for now?

httpbp/middlewares.go
JessicaGreben: i see. makes sense. thanks. i will fix.

httpbp/middlewares.go
JessicaGreben: i agree. at a minimum the baseplate.spec says 2xx or 3xx  so i will add that

httpbp/client_middlewares.go
JessicaGreben: update from a discussion offline:  we are removing the endpoint label in http baseplate.go for now since there isn't great support to do it in a low cardinality way. Will consider adding after baseplate v2 this year.

httpbp/middlewares.go
kylelemons: If you get a 3xx from the http client (i.e. not from RoundTrip) it should be considered a failure because it is not a terminal status code.  Same with 1xx, I believe; 101 switching protocols is nonterminal.

kylelemons: Nevermind, this is used in a RoundTripper, so considering `code >= 100 && code < 400` successful sounds fine to me

httpbp/middlewares.go
```diff
@@ -323,9 +326,13 @@ func (r *statusCodeRecorder) WriteHeader(code int) {
 }
 
 func (r *statusCodeRecorder) getCode(err error) int {
-	if r.code != 0 {
+	return processCodeErr(err, r.code)
+}
+
+func processCodeErr(err error, code int) int {
```
kylelemons, line 332: This name has nagged at me but I haven't really come up with why.  Maybe an API like `errorCodeForMetrics(explicitCode int, requestError err) int` would read more clearly?

httpbp/middlewares.go
fishy: 1. On server side 3xx certainly should be considered success (currently this function is shared between server code and client code), I'm less familiar with 1xx
2. Although go's default client will follow redirections (with 10 max), you can configure the client to not follow redirections. So if the client is configured to follow redirections, you should not get 3xx response codes (if the redirect max is hit you get an error instead, I believe, not 100% sure); but if the client is configured to not follow redirections, the client can get 3xx responses and those should be considered success.

kylelemons (approved):

httpbp/prometheus.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/reddit/baseplate.go/prometheusbp"
+)
+
+const (
+	methodLabel           = "http_method"
+	successLabel          = "http_success"
+	codeLabel             = "http_response_code"
+	remoteServerSlugLabel = "http_slug"
```
fishy, line 14: similarly we might want to remove the `remote-` prefix from the name here.

fishy (approved):

httpbp/prometheus.go
JessicaGreben: I thought I caught them all. Those darm "remote" and "local" keep popping up. Let me remove this one too.

PR #453, opened 2021-12-16
add a middleware to record the status code return by an http server

pacejackson:


httpbp/middlewares.go
```diff
@@ -246,3 +255,103 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+		}
+		return "nan"
+	}
+}
+
+// CounterGenerator is used by RecordStatusCode to create counters for recording
+// http response codes set by the server.
+type CounterGenerator interface {
+	Counter(name string) metrics.Counter
+}
```
pacejackson, line 309: I really just added this to make it easier to test `RecordStatusCode`. I can remove it if we prefer to keep it cleaner and just use metricsbp.M or something different?

fishy: I would rather either move this interface definition into `metricsbp` package, or just use `metricsbp.M`. I think I slightly prefer to just use `metricsbp.M`, as when we moved to prometheus it won't give us `metrics.Counter` anyways so this doesn't really give us any flexibility we wanted.

For testing you could have an unexported interface and an unexported version (`recordStatusCode(counters counterGenerator)`), and the exported version just do `return recordStatusCode(metricsbp.M)`.

httpbp/middlewares.go
```diff
@@ -266,3 +274,103 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+					// something went wrong, check if err is an HTTPErr where we can extract
+					// the code, otherwise assume InternalServerError
+					var httpErr HTTPError
+					if errors.As(err, &httpErr) {
+						code = httpErr.Response().Code
+					} else {
+						code = http.StatusInternalServerError
+					}
+				} else {
+					// if there's no error returned and no call to WriteHeader, assume OK
```
fishy, line 366: slight nit on the "assume" part: this is actually clearly documented that if `WriteHeader` was never called it will be 200, so we can just link the doc here in the comment: https://pkg.go.dev/net/http#ResponseWriter.WriteHeader

kylelemons (changes requested):
Jessica was working on something _super_ similar to this, but it seems like it might end up duplicating logic -- is there a way we can combine the two? (or is that he goal and I am not caught up)

httpbp/middlewares.go
```diff
@@ -266,3 +274,103 @@ func SupportedMethods(method string, additional ...string) Middleware {
...
+	case 8:
+		return "8xx"
+	case 9:
+		return "9xx"
+	default:
+		if family < 1 {
+			return "-nan"
+		}
+		return "nan"
+	}
```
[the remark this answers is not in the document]
pacejackson: well I'll be...

```
// goos: darwin
// goarch: amd64
// pkg: github.com/reddit/baseplate.go/httpbp
// cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
// BenchmarkGetStatusCodeFamilyArray
// BenchmarkGetStatusCodeFamilyArray-16      867824292           1.309 ns/op
func BenchmarkGetStatusCodeFamilyArray(b *testing.B) {
	codes := genCodes(b.N)
	b.ResetTimer()
	for _, code := range codes {
		_ = httpbp.GetStatusCodeFamilyArray(code)
	}
}

// goos: darwin
// goarch: amd64
// pkg: github.com/reddit/baseplate.go/httpbp
// cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
// BenchmarkGetStatusCodeFamily
// BenchmarkGetStatusCodeFamily-16    	100000000	        10.51 ns/op
func BenchmarkGetStatusCodeFamily(b *testing.B) {
	codes := genCodes(b.N)
	b.ResetTimer()
	for _, code := range codes {
		_ = httpbp.GetStatusCodeFamily(code)
	}
}
```

pacejackson:
> Jessica was working on something super similar to this, but it seems like it might end up duplicating logic -- is there a way we can combine the two? (or is that he goal and I am not caught up)

I think we both put up our PRs at around the same time 😅

Combining them is probably what we want, but we might want to get this merged in first and then it can be worked in with @JessicaGreben 's PR since this is for the current metrics while Jessicas work is on a different branch to add prometheus stuff?

JessicaGreben:
@pacejackson Agree, merging this first seems good then I can add to my PR. In the meantime, I'm going to copy your changes into my PR #451 and test it out to see how it works. Will report back.

httpbp/middlewares.go
pacejackson: Good call

httpbp/middlewares.go
```diff
@@ -299,3 +306,103 @@ func recoverPanic(name string, next HandlerFunc) HandlerFunc {
...
+						code = httpErr.Response().Code
+					} else {
+						code = http.StatusInternalServerError
+					}
+				} else {
+					// if there's no error returned and no call to WriteHeader, Go will
+					// return OK.
+					// https://pkg.go.dev/net/http#ResponseWriter.WriteHeader
+					code = http.StatusOK
+				}
```
JessicaGreben, line 380: maybe we could move this logic from ln362-380 into a method on `statusCodeRecorder`, that way I can reuse in the prometheus middleware like so: https://github.com/reddit/baseplate.go/pull/451/files#diff-f0f0f7b278ddd73593a069e51e2d62c8b1fcac7d296258c2e724d85d31c8bb9eR306

httpbp/middlewares.go
```diff
@@ -299,3 +306,105 @@ func recoverPanic(name string, next HandlerFunc) HandlerFunc {
...
+}
+
+var families = [10]string{
+	"nan",
+	"1xx",
+	"2xx",
+	"3xx",
+	"4xx",
+	"5xx",
+	"6xx",
```
fishy, line 353: 🔕  there's nothing in the rfc currently for values >= 600. now with the table approach I'd suggest we just put everything >= 600 into nan as well (we can add new codes when they are added to the rfc)

fishy (approved):

httpbp/middlewares.go
```diff
@@ -299,3 +306,101 @@ func recoverPanic(name string, next HandlerFunc) HandlerFunc {
...
+	"nan",
+	"1xx",
+	"2xx",
+	"3xx",
+	"4xx",
+	"5xx",
+}
+
+// statusCodeFamily takes an http status code and returns it as an "Nxx"
+// string. Returns "nan" if code < 100 or code > 999.
```
fishy, line 356: this comment also needs to be updated now:
```suggestion
// string. Returns "nan" if code < 100 or code > 599.
```

httpbp/middlewares.go
```diff
@@ -299,3 +306,101 @@ func recoverPanic(name string, next HandlerFunc) HandlerFunc {
...
+		return http.StatusInternalServerError
+
+	}
+	// if there's no error returned and no call to WriteHeader, Go will
+	// return OK.
+	// https://pkg.go.dev/net/http#ResponseWriter.WriteHeader
+	return http.StatusOK
+}
+
+var families = [6]string{
```
fishy, line 346: 🔕 is it really necessary to make this an array instead of slice? e.g. this should give you the same result?
```
var families = []string{
```
(unless there's performance implications)

httpbp/middlewares.go
pacejackson: Ok, I don't feel too strong either way, I just went up to `9` because `WriteHeader` lets you write anything between 100 and 999.

httpbp/middlewares.go
pacejackson: I don't know that it's necessary, it was just what @kylelemons example was like:

https://github.com/reddit/baseplate.go/pull/453#discussion_r771585560

kylelemons: Use `[...]` instead of `[6]`.  Using the array turns `len` into a constant value and can improve inlining and bounds check elision.

kylelemons (approved):

httpbp/middlewares.go
pacejackson: > Use [...] instead of [6].

:til:

pacejackson: It actually does seem to be technically faster (albeit the numbers are very small):

slice:

```
goos: darwin
goarch: amd64
pkg: github.com/reddit/baseplate.go/httpbp
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkGetStatusCodeFamily
BenchmarkGetStatusCodeFamily-16    	858927290	         1.402 ns/op
PASS
```

array:
```
goos: darwin
goarch: amd64
pkg: github.com/reddit/baseplate.go/httpbp
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkGetStatusCodeFamily
BenchmarkGetStatusCodeFamily-16    	1000000000	         0.2339 ns/op
PASS
```

PR #454, opened 2021-12-16
fix config override for server TSocket

marcoferrer:
`thriftbp.ServerConfig.Socket` is currently always being overwritten.

### DIFF
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -115,7 +115,6 @@ func NewBaseplateServer(
 		}).ToThriftLogger()
 	}
 	cfg.Addr = bp.GetConfig().Addr
-	cfg.Socket = nil
 	srv, err := NewServer(cfg)
 	if err != nil {
 		return nil, err


thriftbp/server.go
```diff
@@ -115,7 +115,6 @@ func NewBaseplateServer(
 		}).ToThriftLogger()
 	}
 	cfg.Addr = bp.GetConfig().Addr
-	cfg.Socket = nil
```
fishy, line 118: see https://pkg.go.dev/github.com/reddit/baseplate.go/thriftbp#ServerConfig.Socket, especially "used only by NewServer".

this is by design. if you want to use a socket then you have to use `thriftbp.NewServer` instead of `thriftbp.NewBaseplateServer`.

actually the logger override PR merged previously is the same. it's supposed to be overridden in `NewBaseplateServer`.

fishy (changes requested):

thriftbp/server.go
marcoferrer: So there are two main things Im trying to accomplish. Noop logger for thrift server and setting a client connection timeout. This is to resolve the go thrift memory leak issue before the freeze. Otherwise we'll run into periodic ooms

fishy: you can do so by using `thriftbp.NewServer` instead of `thriftbp.NewBaseplateServer`, and add all the default middlewares manually (the rest are more or less the same or you are trying to override anyways).

konradreiche (changes requested):
Echoing Yuxuan's suggestion to reach the desired behavior via `thriftbp.NewServer`

marcoferrer:
closing in favor of https://github.com/reddit/baseplate.go/pull/470

PR #455, opened 2021-12-17
Revert "prevent overwriting server logger config (#452)"

fishy:
This reverts commit bc086bdc11206042e0bde3e7ee5bbd3926a5194d.

As mentioned in [1], this override is supposed to happen.

[1]: https://github.com/reddit/baseplate.go/pull/454#discussion_r770998691

### DIFF
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -106,14 +106,12 @@ func NewBaseplateServer(
 	)
 	middlewares = append(middlewares, cfg.Middlewares...)
 	cfg.Middlewares = middlewares
-	if cfg.Logger == nil {
-		cfg.Logger = log.ZapWrapper(log.ZapWrapperArgs{
-			Level: bp.GetConfig().Log.Level,
-			KVPairs: map[string]interface{}{
-				"from": "thrift",
-			},
-		}).ToThriftLogger()
-	}
+	cfg.Logger = log.ZapWrapper(log.ZapWrapperArgs{
+		Level: bp.GetConfig().Log.Level,
+		KVPairs: map[string]interface{}{
+			"from": "thrift",
+		},
+	}).ToThriftLogger()
 	cfg.Addr = bp.GetConfig().Addr
 	cfg.Socket = nil
 	srv, err := NewServer(cfg)


kylelemons (changes requested):
Can we document on the fields how they are intended to be set?

fishy:
> Can we document on the fields how they are intended to be set?
> 
> Like, "If you need to override this, use the X api" or something

Currently, on the field, the document says "Optional, used only by NewServer." (https://github.com/reddit/baseplate.go/blob/f3396ece11b6c336a5eaae2fe4dd3a9f3425ced2/thriftbp/server.go#L31)

And on the document of the whole `ServerConfig` struct, it says: https://github.com/reddit/baseplate.go/blob/f3396ece11b6c336a5eaae2fe4dd3a9f3425ced2/thriftbp/server.go#L15-L17

I think that should be enough?

marcoferrer:
One issue I have with this is that now users will need to re-implement a significantly important portion of the baseplate API just so that they can have the ability to configure this. One of the main purposes of using a framework like baseplate is so that developers can have access to sane defaults while still being able to configure settings specific to their use case. In removing this setting, in addition to `timeout` and [sockets](https://github.com/reddit/baseplate.go/pull/454#discussion_r770998691) users need to create their own copy of `NewBaseplateServer` and constantly keeps its implementation in sync with the libraries version. This wouldnt be that difficult if the method design allow for delegating to the baseplate impl but overwritten configurations provided makes this impossible.

```go
// NewBaseplateServer returns a new Thrift implementation of a Baseplate
// server with the given config.
func NewBaseplateServer(
	bp baseplate.Baseplate,
	cfg ServerConfig,
) (baseplate.Server, error) {
	middlewares := BaseplateDefaultProcessorMiddlewares(
		DefaultProcessorMiddlewaresArgs{
			EdgeContextImpl:                    bp.EdgeContextImpl(),
			ErrorSpanSuppressor:                cfg.ErrorSpanSuppressor,
			ReportPayloadSizeMetricsSampleRate: cfg.ReportPayloadSizeMetricsSampleRate,
		},
	)
	cfg.Middlewares = middlewares
	if cfg.Logger == nil {
		cfg.Logger = log.ZapWrapper(log.ZapWrapperArgs{
			Level: bp.GetConfig().Log.Level,
			KVPairs: map[string]interface{}{
				"from": "thrift",
			},
		}).ToThriftLogger()
	}
	srv, err := NewServer(cfg)
	if err != nil {
		return nil, err
	}
	return ApplyBaseplate(bp, srv), nil
}
```

fishy:
@marcoferrer `NewServer` is the way to customize however you want, while `NewBaseplateServer` is the opinionated version.

marcoferrer:
> @marcoferrer `NewServer` is the way to customize however you want, while `NewBaseplateServer` is the opinionated version.

But thats the issue. I dont want to customize the server to the nth degree. For me, `NewServer` is too low of an API that requires users to keep track of their own special defaults without being able to reference or at least concat the preferred defaults from the library. It ends up being a foot-gun if theres ever a critical change to how the server is set up and that change isnt back-ported to a users custom `my.ReImplNewBaseplateServer`

fishy:
A line needs to be drawn somewhere between "this is low level stuff, use at your own risk" and "this has a sane default but you are still welcomed to change it". socket and timeout are certainly on the low level stuff side. logger is also on the low level stuff side currently (the approval of the to-be-reverted change was a mistake), but I can be convinced to move it to the other side.

Currently the only reason you need to change it is to add timeout and suppress those logs generated from timeout, which is not a good enough reason to move it to the other side (you can already use `github.com/apache/thrift@master` to fix the memory usage without adding a timeout).

kylelemons (changes requested):
I think having an args struct that has some arguments that aren't consulted tends to be confusing, and I would prefer to have it work as users expect.  If the user overrides a setting that is not supported it should return an error, not silently ignore it.  I also agree that a logger doesn't strike me as too low-level to allow users to configure it.

So I guess I'm slightly in favor of not reverting this.

konradreiche (changes requested):
Part of the problem could be that we are reusing the same configuration type between `NewServer` used for customization and `NewBaseplateServer` used for sane defaults.

fishy:
@kylelemons & @konradreiche this is now a release blocker for v0.9.2 (to include prometheus work).

The change in bc086bd changed `ServerConfig.Logger` to be used by both `NewBaseplateServer` and `NewServer`, but the doc (https://pkg.go.dev/github.com/reddit/baseplate.go@v0.9.2-0.20220112010100-a291dcb5e854/thriftbp#ServerConfig.Logger) was not updated and it still says: `Optional, used only by NewServer.`

We can of course just update the doc to unblock `v0.9.2`. But since this change is actually a piece of a bigger picture of https://github.com/reddit/baseplate.go/pull/463 that's still pending, I think revert it to unblock v0.9.2 and redo it after (probably through #463) properly is the better approach.

kylelemons:
I'm fine with this but I just want to make sure we get an ack from @marcoferrer

kylelemons (approved):

konradreiche (approved):

marcoferrer:
So if youd like to revert this to unblock the release that should be fine. We cant use baseplate in its current state regardless and are having to rely on a fork at the moment since the server defaults are not exported. If we ever get the issues with thrift lib resolved we are most likely still going to rely on using this configuration.

PR #458, opened 2021-12-21
filewatcher: Fix flaky tests

fishy:
When overwriting the file with new content for test, the overwriting was
done incorrectly, via overwriting in place, and can sometimes cause the
filewatcher to be triggered twice (once when opening and truncating the
file, once when the file is closed and writing is done). Add a helper
function to do the writing on a separated file and then use os.Rename
for an atomic move instead.

### DIFF
--- a/filewatcher/filewatcher_test.go
+++ b/filewatcher/filewatcher_test.go
@@ -20,6 +20,26 @@ func parser(f io.Reader) (interface{}, error) {
 	return io.ReadAll(f)
 }
 
+// overwriteFile does atomic overwrite (write to a new file, then use rename to
+// overwrite the old file) instead of in-pleace overwrite (truncate open the
+// file, write to it, close the file).
+//
+// filewatcher is designed to handle atomic overwrites, not in-place overwrites.
+// Doing in-place overwrite will cause the filewatcher to be triggered twice
+// (once the file is truncated, once when closing the file), which would cause
+// some of the tests to fail flakily on CI.
+func overwriteFile(tb testing.TB, path string, content []byte) {
+	tb.Helper()
+
+	tmpPath := filepath.Join(tb.TempDir(), "file")
+	if err := os.WriteFile(tmpPath, content, 0644); err != nil {
+		tb.Fatalf("Unable to write file: %v", err)
+	}
+	if err := os.Rename(tmpPath, path); err != nil {
+		tb.Fatalf("Unable to overwrite file: %v", err)
+	}
+}
+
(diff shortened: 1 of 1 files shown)


filewatcher/filewatcher_test.go
```diff
@@ -79,16 +98,7 @@ func TestFileWatcher(t *testing.T) {
...
-		f, err := os.OpenFile(path, os.O_WRONLY|os.O_TRUNC, 0644)
-		if err != nil {
-			t.Fatal(err)
-		}
-		defer f.Close()
-		if _, err := f.Write(payload2); err != nil {
-			t.Fatal(err)
-		}
-	}()
+	overwriteFile(t, path, bytes.NewReader(payload2))
```
kylelemons, line 101: 🔕 you use bytes.NewReader for both, so you might as well just pass the contents as `[]byte` and use `os.WriteFile`

kylelemons (approved):

filewatcher/filewatcher_test.go
fishy: Looking at the source code of `os.WriteFile`, it tries to do a single `Write`, which I don't really like and it's more likely to fail if we are trying to write large content (although we don't write large content here). While `io.Copy` handles those cases much more thoroughly.

filewatcher/filewatcher_test.go
```diff
@@ -20,6 +20,25 @@ func parser(f io.Reader) (interface{}, error) {
 	return io.ReadAll(f)
 }
 
+func overwriteFile(tb testing.TB, path string, content io.Reader) {
```
pacejackson, line 23: 🔕  might be good to add a comment to this explaining why we have it, at least for future readers learning?

pacejackson (approved):

filewatcher/filewatcher_test.go
fishy: Done.

fishy:
💇 added comment to explain `overwriteFile`.

filewatcher/filewatcher_test.go
kylelemons: You can refactor `overwriteFile` in the future _if_ you end up with a reader.  YAGNI / KISS / premature optimization / etc.

kylelemons (approved):
(to appease harold)

filewatcher/filewatcher_test.go
fishy: Changed to use `os.WriteFile`. Looks like `os.File.Write` itself implemented big write handling, at least for linux.

PR #461, opened 2021-12-21
filewatcher: Fix flaky tests 2

fishy:
Follow up on 846e891, the create-write-close writing can also trigger
filewatcher twice and can make the test flaky when the machine running
it is slow (e.g. when the CI is slow). Change all writes in filewatcher
tests to use the helper to do the atomic write and make sure that they
only trigger filewatcher once.

### DIFF
--- a/filewatcher/filewatcher_test.go
+++ b/filewatcher/filewatcher_test.go
@@ -20,15 +20,15 @@ func parser(f io.Reader) (interface{}, error) {
 	return io.ReadAll(f)
 }
 
-// overwriteFile does atomic overwrite (write to a new file, then use rename to
-// overwrite the old file) instead of in-pleace overwrite (truncate open the
-// file, write to it, close the file).
+// writeFile does atomic write/overwrite (write to a tmp file, then use rename
+// to overwrite the desired path) instead of in-pleace write/overwrite
+// (open/truncate open the file, write to it, close the file).
 //
-// filewatcher is designed to handle atomic overwrites, not in-place overwrites.
-// Doing in-place overwrite will cause the filewatcher to be triggered twice
-// (once the file is truncated, once when closing the file), which would cause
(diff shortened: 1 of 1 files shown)


pacejackson (approved):

kylelemons (approved):
At what point should we consider debouncing / coalescing filewatcher events instead?  I don't think we can reasonably assume that atomic file operations will be the norm forever...

fishy:
> At what point should we consider debouncing / coalescing filewatcher events instead? I don't think we can reasonably assume that atomic file operations will be the norm forever...

This is actually just more a problem in test than in production. In test, the problems are:

1. create-write-close triggers the filewatcher twice, and we write the test in a way that we start a goroutine to delay creating the file, then start a new filewatcher to wait for the initial content. the filewatcher triggered by create could read in zero content and fail the parser/test, that's what caused the test to be flaky. in production we usually use other ways (init container, `container-run.sh`, etc.) to make sure that the initial writing is done before we even start the service code, so it won't have this problem
2. truncate-write-close triggers the filewatcher twice, and the test is designed to make it write bad data to fail the parser, but we expect it to fail once instead of twice. in production if we actually have the zero data (triggered by truncate) that fails the parser then the actual data (triggered by close) passes it, the filewatcher implementations (in both go and py) are designed to be resilient on temporarily failed parsers (we just log it and keep using the last good parser result), so the 2nd one from `close` will still work as expected.

also currently the thing that writes to the files read from filewatcher is baseplate.py's sidecars, ad they already made sure that we always use atomic write/overwrite. I believe this is also part of the baseplate spec. we do want to make sure that any new thing we do that will be read from filewatcher need to make sure that they use atomic write/overwrite, though.

fishy:
also the upcoming vault/secret change will require some changes in filewatcher, so that would be a good chance to visit some of the design decisions.

PR #463, opened 2022-01-04
split thriftbp server default into separate method

marcoferrer:
Exposing defaults from `NewBaseplateServer`

### DIFF
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -69,7 +69,7 @@ type ServerConfig struct {
 // and protocol to serve the given TProcessor which is wrapped with the
 // given ProcessorMiddlewares.
 func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
-	var transport *thrift.TServerSocket
+	var transport thrift.TServerTransport
 	if cfg.Socket == nil {
 		var err error
 		transport, err = thrift.NewTServerSocket(cfg.Addr)
@@ -97,6 +97,17 @@ func NewBaseplateServer(
 	bp baseplate.Baseplate,
 	cfg ServerConfig,
 ) (baseplate.Server, error) {
+	ApplyConfigDefaults(bp, cfg)
+	srv, err := NewServer(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return ApplyBaseplate(bp, srv), nil
+}
+
+// ApplyConfigDefaults initializes the default configuration values
+// used for creating a baseplate server.
+func ApplyConfigDefaults(bp baseplate.Baseplate, cfg ServerConfig) {
 	middlewares := BaseplateDefaultProcessorMiddlewares(
 		DefaultProcessorMiddlewaresArgs{
 			EdgeContextImpl:                    bp.EdgeContextImpl(),
@@ -116,11 +127,6 @@ func NewBaseplateServer(
 	}
 	cfg.Addr = bp.GetConfig().Addr
 	cfg.Socket = nil
-	srv, err := NewServer(cfg)
-	if err != nil {
-		return nil, err
-	}
-	return ApplyBaseplate(bp, srv), nil
 }
 
 // ApplyBaseplate returns the given TSimpleServer as a baseplate Server with the
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -97,36 +97,92 @@ func NewBaseplateServer(
 	bp baseplate.Baseplate,
 	cfg ServerConfig,
 ) (baseplate.Server, error) {
-	ApplyConfigDefaults(bp, cfg)
-	srv, err := NewServer(cfg)
+
+	opts := []ServerOpt{
+		func(optCfg ServerConfig) ServerConfig {
+			return cfg
+		},
+	}
+	opts = append(opts, DefaultServerOpts(bp)...)
+
+	return NewBaseplateServerFromOpts(bp, opts...)
+}
+
+// NewBaseplateServerFromOpts returns a new Thrift implementation of a Baseplate
+// server using a config built from the supplied opts ServerOpt.
+func NewBaseplateServerFromOpts(
+	bp baseplate.Baseplate,
+	opts ...ServerOpt,
+) (baseplate.Server, error) {
+
+	srv, err := NewServer(BuildConfig(opts...))
 	if err != nil {
 		return nil, err
 	}
 	return ApplyBaseplate(bp, srv), nil
 }
 
-// ApplyConfigDefaults initializes the default configuration values
-// used for creating a baseplate server.
-func ApplyConfigDefaults(bp baseplate.Baseplate, cfg ServerConfig) {
-	middlewares := BaseplateDefaultProcessorMiddlewares(
-		DefaultProcessorMiddlewaresArgs{
-			EdgeContextImpl:                    bp.EdgeContextImpl(),
-			ErrorSpanSuppressor:                cfg.ErrorSpanSuppressor,
-			ReportPayloadSizeMetricsSampleRate: cfg.ReportPayloadSizeMetricsSampleRate,
-		},
-	)
-	middlewares = append(middlewares, cfg.Middlewares...)
-	cfg.Middlewares = middlewares
-	if cfg.Logger == nil {
+type ServerOpt func(cfg ServerConfig) ServerConfig
+
+func BuildConfig(opts ...ServerOpt) ServerConfig {
+	cfg := ServerConfig{}
+	for _, opt := range opts {
+		cfg = opt(cfg)
+	}
+
+	return cfg
+}
+
+func DefaultServerOpts(bp baseplate.Baseplate) []ServerOpt {
+	return []ServerOpt{
+		serverOptMiddleware(bp),
+		serverOptLogger(bp),
+		serverOptAddress(bp),
+		serverOptSocket(bp),
+	}
+}
+
+func serverOptMiddleware(bp baseplate.Baseplate) ServerOpt {
+	return func(cfg ServerConfig) ServerConfig {
+		middlewares := BaseplateDefaultProcessorMiddlewares(
+			DefaultProcessorMiddlewaresArgs{
+				EdgeContextImpl:                    bp.EdgeContextImpl(),
+				ErrorSpanSuppressor:                cfg.ErrorSpanSuppressor,
+				ReportPayloadSizeMetricsSampleRate: cfg.ReportPayloadSizeMetricsSampleRate,
+			},
+		)
+		cfg.Middlewares = append(middlewares, cfg.Middlewares...)
+
+		return cfg
+	}
+}
+
+func serverOptLogger(bp baseplate.Baseplate) ServerOpt {
+	return func(cfg ServerConfig) ServerConfig {
 		cfg.Logger = log.ZapWrapper(log.ZapWrapperArgs{
 			Level: bp.GetConfig().Log.Level,
 			KVPairs: map[string]interface{}{
 				"from": "thrift",
 			},
 		}).ToThriftLogger()
+
+		return cfg
 	}
-	cfg.Addr = bp.GetConfig().Addr
-	cfg.Socket = nil
+}
+
+func serverOptAddress(bp baseplate.Baseplate) ServerOpt {
+	return func(cfg ServerConfig) ServerConfig {
+		cfg.Addr = bp.GetConfig().Addr
+
+		return cfg
+	}
+}
+
+func serverOptSocket(bp baseplate.Baseplate) ServerOpt {
+	return func(cfg ServerConfig) ServerConfig {
+		cfg.Socket = nil
+		return cfg
+	}
 }
 
 // ApplyBaseplate returns the given TSimpleServer as a baseplate Server with the
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -99,7 +99,7 @@ func NewBaseplateServer(
 ) (baseplate.Server, error) {
 
 	opts := []ServerOpt{
-		func(optCfg ServerConfig) ServerConfig {
+		func(_ ServerConfig) ServerConfig {
 			return cfg
 		},
 	}
@@ -122,8 +122,12 @@ func NewBaseplateServerFromOpts(
 	return ApplyBaseplate(bp, srv), nil
 }
 
+// ServerOpt is a type used for defining configuration arguments
+// needed for creating a baseplate server.
 type ServerOpt func(cfg ServerConfig) ServerConfig
 
+// BuildConfig creates a new ServerConfig instance and applies
+// all the supplied configuration options to it.
 func BuildConfig(opts ...ServerOpt) ServerConfig {
 	cfg := ServerConfig{}
 	for _, opt := range opts {
@@ -133,6 +137,8 @@ func BuildConfig(opts ...ServerOpt) ServerConfig {
 	return cfg
 }
 
+// DefaultServerOpts builds and returns a slice of the default
+// baseplate server config options.
 func DefaultServerOpts(bp baseplate.Baseplate) []ServerOpt {
 	return []ServerOpt{
 		serverOptMiddleware(bp),
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -69,26 +69,41 @@ type ServerConfig struct {
 // and protocol to serve the given TProcessor which is wrapped with the
 // given ProcessorMiddlewares.
 func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
-	var transport thrift.TServerTransport
+
+	// Leaving this default socket initialization intact so that we
+	// don't break existing usages of NewServer. Ideally, by the time
+	// we've reached NewServer, all necessary configuration should
+	// be completed. If we are ever able to deprecate this use case
+	// we will be able to fully remove cfg.Addr and serverOptAddress
+	// as well.
 	if cfg.Socket == nil {
-		var err error
-		transport, err = thrift.NewTServerSocket(cfg.Addr)
-		if err != nil {
+		if err := ServerOptSocket(cfg.Addr)(&cfg); err != nil {
 			return nil, err
 		}
-	} else {
-		transport = cfg.Socket
 	}
 
 	server := thrift.NewTSimpleServer4(
 		thrift.WrapProcessor(cfg.Processor, cfg.Middlewares...),
-		transport,
+		cfg.Socket,
 		thrift.NewTHeaderTransportFactoryConf(nil, nil),
 		thrift.NewTHeaderProtocolFactoryConf(nil),
 	)
 	server.SetForwardHeaders(HeadersToForward)
 	server.SetLogger(cfg.Logger)
 	return server, nil
+}
+
+// NewServerFromOpts creates a new ServerConfig instance and
+// delegates to NewServer so that it can instantiate a
+// new thrift.TSimpleServer.
+func NewServerFromOpts(opts ...ServerOpt) (*thrift.TSimpleServer, error) {
+
+	cfg, err := BuildConfig(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewServer(*cfg)
 }
 
 // NewBaseplateServer returns a new Thrift implementation of a Baseplate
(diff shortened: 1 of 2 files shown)


fishy (changes requested):
While this change is fine code-wise, I have the doubt of the usefulness of this new public API.

Now the code-freeze is ended, I'd suggest you to try the new version of thrift go library (`go get github.com/apache/thrift@master`), and you likely won't need non-baseplate server after that.

marcoferrer:
> While this change is fine code-wise, I have the doubt of the usefulness of this new public API.
> 
> Now the code-freeze is ended, I'd suggest you to try the new version of thrift go library (`go get github.com/apache/thrift@master`), and you likely won't need non-baseplate server after that.

The new thrift change cause a large spike in errors until we reverted it. I think this api is necessary if it's expected that users who need additional customization of the thrift server are to rely on `NewServer`.

marcoferrer:
The scenario Im trying to avoid with this method is cases where a critical change was made in `NewBaseplateServer` but every consumer of `NewServer` is left in the dark.

konradreiche:
Yes. Also, would it be more idiomatic to have a `NewDefaultServerConfig(bp baseplate.Baseplate) ServerConfig`?

marcoferrer:
Curious as to what you have in mind? Im not sure how I would accomplish returning `ServerConfig` without passing in the user `ServerConfig` since part of configuring defaults is referencing those values. Such as prepending the default thrift middlewares.

image

kylelemons:
I tend to prefer "functional" APIs.  i.e.

```
NewBaseplateServer(foo) ~> NewServer(DefaultServerArgs(foo))
```

so then you can easily "decompose" and modify.

marcoferrer:
So I took a shot at going full functional and implemented a constructor that takes in a list of functional options `NewBaseplateServerFromOpts` with each opt being typed as `func(ServerConfig) ServerConfig`. I made the type have a return value of `ServerConfig` only to simplify backwards compatibility with the existing `NewBaseplateServer`.

I also converted the defaults to functions and adapted the original `NewBaseplateServer` to delegate to `NewBaseplateServerFromOpts` and preserve its original behavior. Looking for thoughts on the approach

thriftbp/server.go
```diff
@@ -97,30 +97,92 @@ func NewBaseplateServer(
...
-	middlewares := BaseplateDefaultProcessorMiddlewares(
-		DefaultProcessorMiddlewaresArgs{
-			EdgeContextImpl:                    bp.EdgeContextImpl(),
-			ErrorSpanSuppressor:                cfg.ErrorSpanSuppressor,
-			ReportPayloadSizeMetricsSampleRate: cfg.ReportPayloadSizeMetricsSampleRate,
+
+	opts := []ServerOpt{
+		func(optCfg ServerConfig) ServerConfig {
+			return cfg
 		},
```
marcoferrer, line 104: These lines are the reason why I opted for having the `ServerOpt` return the `ServerConfig` instance instead of unit. It was the most straight forward way I could think of maintaining backwards compatibility. Would using a pointer as an arg be the better option?

kylelemons (changes requested):
Per chat (and to appease harold):

I recommend option patterns like

```
type FooOption func(opt *FooOptions) error
```

so it can fail.  You can have an option that does `*dst = src` to override the full set of options for the fallback case.

marcoferrer:
Updated impl based on feedback.

thriftbp/server.go
```diff
@@ -69,20 +69,22 @@ type ServerConfig struct {
...
 func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
-	var transport *thrift.TServerSocket
+
+	// Leaving this default socket initialization intact so that we
+	// don't break existing usages of NewServer. Ideally, by the time
+	// we've reached NewServer, all necessary configuration should
+	// be completed. If we are ever able to deprecate this use case
+	// we will be able to fully remove cfg.Addr and serverOptAddress
+	// as well.
 	if cfg.Socket == nil {
```
marcoferrer, line 79: In sourcegraph I was only able to find two place sites where removing this would break user code.

konradreiche (approved):

fishy (changes requested):
I'm fine with the code, but I still have concerns regarding the added new public APIs.

The point is, we really want services to use `NewBaseplateServer` so their service is meeting all the expectations. `NewServer` is provided for tests and services having a really good reason to not use `NewBaseplateServer`, and they are on their own to make sure they are meeting the baseplate expectations. Even with all the new APIs, services using `NewServer` or `NewServerFromOpts` can still do things to mess up (accidentally remove some middlewares, accidentally add default middlewares twice, etc.).

thriftbp/server.go
```diff
@@ -91,36 +93,145 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+
+// ServerOpt is a type used for defining configuration arguments
+// needed for creating a baseplate server. It allows returning
+// an error to signal that a configuration value was invalid
+// or failed to be set.
+type ServerOpt func(cfg *ServerConfig) error
+
+// BuildConfig creates a new ServerConfig instance and applies
+// all the supplied configuration options to it.
+func BuildConfig(opts ...ServerOpt) (*ServerConfig, error) {
```
kylelemons, line 147: BuildConfig should return the same type that will actually be consumed.  In this case that's a struct value not a pointer.

Also, I don't really think this should be exported -- NewServerFromOpts is the API surface that a user would use I think.

thriftbp/server.go
```diff
@@ -91,36 +93,145 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+
+	srv, err := NewServerFromOpts(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return ApplyBaseplate(bp, srv), nil
+}
+
+// ServerOpt is a type used for defining configuration arguments
```
kylelemons, line 139: ServerOption or NewServerOption -- don't drop letters to save typing, especially in exported APIs, and _especially_ when it is ambiguous (Opt can mean Option, Optional, Opt-in, etc).

thriftbp/server.go
```diff
@@ -91,36 +93,145 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+
+// serverOptAddress is used to copy the address config from the
+// baseplate config to the server config. Currently, it's only
+// needed to guarantee backwards compatibility in NewServer
+func serverOptAddress(bp baseplate.Baseplate) ServerOpt {
+	return func(cfg *ServerConfig) error {
+		cfg.Addr = bp.GetConfig().Addr
+
+		return nil
+	}
```
kylelemons, line 216: return `WithListenAddr(bp.GetConfig().Addr)`?

thriftbp/server.go
```diff
@@ -91,36 +93,145 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+}
+
+func serverOptSocket(bp baseplate.Baseplate) ServerOpt {
+	return ServerOptSocket(bp.GetConfig().Addr)
+}
+
+// ServerOptSocket creates a server option that initializes a
+// thrift.TServerSocket listening on the address supplied.
+func ServerOptSocket(address string) ServerOpt {
```
kylelemons, line 225: ServerOptSocket should take a socket.  A name like `WithListenAddr` would be more appropriate for this option

thriftbp/server.go
```diff
@@ -91,36 +93,145 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+		}
+	}
+
+	return cfg, nil
+}
+
+// ServerOptFrom returns a server option that overwrites the
+// any existing values in a ServerConfig with those defined
+// in src.
+func ServerOptFrom(src ServerConfig) ServerOpt {
```
kylelemons, line 162: The convention around these names is that they tend to start `With` e.g. `WithConfigFrom(ServerConfig)`.  godoc will sort them under the appropriate type, e.g. DialOption in https://pkg.go.dev/google.golang.org/grpc

kylelemons (changes requested):

thriftbp/server.go
```diff
@@ -10,6 +12,10 @@ import (
 	"github.com/reddit/baseplate.go/log"
 )
 
+var (
+	ErrInvalidListenerAddress = errors.New("invalid listener address")
```
kylelemons, line 16: I don't think this is useful outside this package and would recommend unexporting it.  Exported error sentinels are an API burden.

thriftbp/server.go
```diff
@@ -69,20 +75,20 @@ type ServerConfig struct {
 // and protocol to serve the given TProcessor which is wrapped with the
 // given ProcessorMiddlewares.
 func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
-	var transport *thrift.TServerSocket
+
```
kylelemons, line 78: don't start indented blocks with a blank line; this has come up a few times -- don't do it :)

thriftbp/server.go
```diff
@@ -91,38 +97,201 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
 	return server, nil
 }
 
+// NewServerFromOpts creates a new ServerConfig instance and
+// delegates to NewServer so that it can instantiate a
+// new thrift.TSimpleServer.
+func NewServerFromOpts(opts ...ServerOption) (*thrift.TSimpleServer, error) {
```
kylelemons, line 103: Opts -> Options

thriftbp/server.go
```diff
@@ -91,38 +97,201 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
 	bp baseplate.Baseplate,
 	cfg ServerConfig,
 ) (baseplate.Server, error) {
-	middlewares := BaseplateDefaultProcessorMiddlewares(
-		DefaultProcessorMiddlewaresArgs{
-			EdgeContextImpl:                    bp.EdgeContextImpl(),
-			ErrorSpanSuppressor:                cfg.ErrorSpanSuppressor,
-			ReportPayloadSizeMetricsSampleRate: cfg.ReportPayloadSizeMetricsSampleRate,
-		},
+
```
kylelemons, line 118: extraneous blank line; remove

thriftbp/server.go
```diff
@@ -91,38 +97,201 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
-	cfg.Addr = bp.GetConfig().Addr
-	cfg.Socket = nil
-	srv, err := NewServer(cfg)
+
+	return NewBaseplateServerFromOpts(bp, opts...)
+}
+
+// NewBaseplateServerFromOpts returns a new Thrift implementation of a Baseplate
+// server using a config built from the supplied opts ServerOption.
+func NewBaseplateServerFromOpts(
```
kylelemons, line 129: Opts -> Options

thriftbp/server.go
```diff
@@ -91,38 +97,201 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+	bp baseplate.Baseplate,
+	opts ...ServerOption,
+) (baseplate.Server, error) {
+
```
kylelemons, line 133: extraneous blank line; remove

thriftbp/server.go
```diff
@@ -91,38 +97,201 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+	srv, err := NewServerFromOpts(opts...)
 	if err != nil {
 		return nil, err
 	}
+
```
kylelemons, line 138: nothing is changing here, leave it as-is

thriftbp/server.go
```diff
@@ -91,38 +97,201 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+func WithSocket(socket *thrift.TServerSocket) ServerOption {
+	return func(cfg *ServerConfig) error {
+		cfg.Socket = socket
+		return nil
+	}
+}
+
+// WithDefaultSocket initializes a socket listening at the
+// configured address.
+func WithDefaultSocket() ServerOption {
```
kylelemons, line 280: what's the use-case for users providing this option themselves?

Either they have a ServerConfig (in which case they should be using WithConfigFrom) or they have an address (and should use WithListenerAddress) or socket (WithSocket)

kylelemons (commented):
OK, so in general I think that configurability with variadic options is fine.  However, I feel like this is a much broader API expansion than what you currently need.  I would recommend exporting _only_ the pieces that are _strictly_ required, and providing an example that shows how they can be used.

You also don't mention in any of the documentation that these APIs are intended for advanced use-cases.  You should probably do so, and it might even be worth an adomonishment to not use them unless directed to do so by #baseplate-go

kylelemons (changes requested):
ugh, muscle memory.  clicking the right button this time...

marcoferrer:
I can update the documentation and provide examples. Is there a specific format I should for embedding code samples in go docs? Sidenote: if this is too much to introduce I can revert back to my original implementation. My goal was to surface the defaults used to configure a baseplate server. Currently the API for `NewServer` leaves it up to the user to create their own set of defaults instead of extending whats provided by baseplate. This makes updating baseplate not as straight forward and painful. I went the route of implementing a functional constructor since it seemed that was the direction that was preferred.

```go

// NewBaseplateServer returns a new Thrift implementation of a Baseplate
// server with the given config.
func NewBaseplateServer(
	bp baseplate.Baseplate,
	cfg ServerConfig,
) (baseplate.Server, error) {
	ApplyConfigDefaults(bp, cfg)
	srv, err := NewServer(cfg)
	if err != nil {
		return nil, err
	}
	return ApplyBaseplate(bp, srv), nil
}

// ApplyConfigDefaults initializes the default configuration values
// used for creating a baseplate server.
func ApplyConfigDefaults(bp baseplate.Baseplate, cfg ServerConfig) {
	middlewares := BaseplateDefaultProcessorMiddlewares(
		DefaultProcessorMiddlewaresArgs{
			EdgeContextImpl:                    bp.EdgeContextImpl(),
			ErrorSpanSuppressor:                cfg.ErrorSpanSuppressor,
			ReportPayloadSizeMetricsSampleRate: cfg.ReportPayloadSizeMetricsSampleRate,
		},
	)
	cfg.Middlewares = middlewares
	if cfg.Logger == nil {
		cfg.Logger = log.ZapWrapper(log.ZapWrapperArgs{
			Level: bp.GetConfig().Log.Level,
			KVPairs: map[string]interface{}{
				"from": "thrift",
			},
		}).ToThriftLogger()
	}
	cfg.Socket = nil
}

```

fishy:
For examples, see https://pkg.go.dev/testing#hdr-Examples

thriftbp/server.go
```diff
@@ -91,38 +96,232 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
 	return server, nil
 }
 
+// NewServerFromOptions creates a new ServerConfig instance and delegates
+// to NewServer so that it can instantiate a new thrift.TSimpleServer.
+// This is a low level api and should only be used at the recommendation
```
fishy, line 101: style nit: add a blank line (`//`) between them.

the doc comment will be rendered similar to markdown, so line break is the same as a space, and a blank line is the same as actual line break (new paragraph).

thriftbp/server.go
```diff
@@ -91,38 +96,232 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to build server config: %w", err)
+	}
+	return NewServer(cfg)
+}
+
 // NewBaseplateServer returns a new Thrift implementation of a Baseplate
-// server with the given config.
+// server with the given config. This is the primary constructor that
+// should be used when building a baseplate server.
```
fishy, line 114: similar style nit as above, in general we would prefer to limit the first paragraph of the doc comment to a single sentence (ends with the first `.`), and add a blank line to start the next one.

thriftbp/server.go
```diff
@@ -91,38 +96,232 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
-	cfg.Socket = nil
-	srv, err := NewServer(cfg)
+	return NewBaseplateServerFromOptions(bp, options...)
+}
+
+// NewBaseplateServerFromOptions returns a new Thrift implementation of a Baseplate
+// server using a config built from the supplied opts ServerOption. This is a low
+// level api and should  only be used at the recommendation of the baseplate maintainers.
+// Most application implementations should rely on NewBaseplateServer.
+func NewBaseplateServerFromOptions(
```
fishy, line 130: what's the use case scenario for this one?

thriftbp/server.go
```diff
@@ -91,38 +96,232 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+// ServerOption is a type used for defining configuration arguments
+// needed for creating a baseplate server. It allows returning
+// an error to signal that a configuration value was invalid
+// or failed to be set.
+type ServerOption func(cfg *ServerConfig) error
+
+// buildConfig creates a new ServerConfig instance and applies
+// all the supplied configuration options to it.
+func buildConfig(options ...ServerOption) (ServerConfig, error) {
+	cfg := &ServerConfig{}
```
fishy, line 150: my preference would be to use `var cfg ServerConfig` here, use `option(&cfg)` on line 152 and `return cfg, nil` on line 158.

thriftbp/server.go
```diff
@@ -91,38 +96,232 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
...
+// withDefaultListenerAddress is used to copy the address config from the
+// baseplate config to the server config. Currently, it's only
+// needed to guarantee backwards compatibility in NewServer
+func withDefaultListenerAddress(bp baseplate.Baseplate) ServerOption {
+	return WithListenerAddress(bp.GetConfig().Addr)
+}
+
+// WithMiddleware appends the arg middleware to the end of
+// the cfg.Middleware slice
+func WithMiddleware(middleware ...thrift.ProcessorMiddleware) ServerOption {
```
fishy, line 229: I can see this API being confusing and becomes a footgun. All the other `With*` APIs are to replace the corresponding config field, while this one is to append.

thriftbp/server_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"context"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/reddit/baseplate.go"
+	"github.com/reddit/baseplate.go/thriftbp"
+)
+
+func ExampleNewServerFromOptions() {
```
fishy, line 11: please add a doc comment to the example function and some inline comments to explain what you are doing in this example. see https://github.com/reddit/baseplate.go/blob/v0.9.1/thriftbp/example_client_test.go for an example (https://pkg.go.dev/github.com/reddit/baseplate.go@v0.9.1/thriftbp#example-MonitorClient is the result rendered on doc site)

fishy (changes requested):
I also don't think you addressed my concerns in https://github.com/reddit/baseplate.go/pull/463#pullrequestreview-844941016. can we hold this PR until you give the thrift fix another try? if that is successful you won't need to use `NewServer` and thus won't really need this PR.

kylelemons (changes requested):
(pending answers about thrift)

marcoferrer:
Bumping this PR.

Tested the thrift-api changes and no dice. Details can be found in the internal bp-go channel. 

Want to confirm whats needed to get this unblocked.

thriftbp/server_test.go
```diff
@@ -0,0 +1,62 @@
+package thriftbp_test
```
kylelemons, line 1: OK, so in terms of unblocking:

* If we don't need an API, or at least yet, I would prefer to not have one so that we don't complicate any future programmatic migrations to a new baseplate version.  The question of whether we need this API seems to come back to the question of whether the current thriftbp can work for your service.
* If we need an API, I think the `With`-style functional options are a good choice generally, but the details of usage matter, and I haven't seen the example PR that makes use of this so I am purely extrapolating.
* Both of these examples show using a _lot_ of options.  When a lot of callers are using a lot of options, the functional options actually aren't that great of a choice in many cases -- they're optimal for callers who need 0 or 1 option, with the rare exception of those who need more.  The 
* Most examples should be targeted at actual usage patterns.  If this is what it will actually look like for callers who need to customize one thing -- i.e. having to spell out a lot of options manually -- then I think it's going the wrong way.  If this is just to make use of all of the available options, then I wouldn't worry about that -- the options are right there in the godoc.  Make these examples focus on what actual users will need, if that makes sense.
* At the end of the day, if we don't want people to use these APIs except out of dire need, we need to discourage it.  Maybe We need to add "Dangerous" or "Unsupported" or something to the ones that we don't want people to use.

kylelemons (changes requested):

marcoferrer:
Closing in favor of https://github.com/reddit/baseplate.go/pull/470

PR #464, opened 2022-01-04
add server connection count metric

marcoferrer:
This includes the following changes:

* A baseplate wrapper for `TServerTransport` which wraps the result of `Accept` with a `MonitoredTTransport`
* A baseplate wrapper for `TTransport` which emits counter metrics for invocations of `Open` and `Close`
* A `metricsbp.Config` flag to allow users to enable recording connection count metrics from their application yaml
* A `thriftbp.ServerConfig` flag to check if reporting connection counts is enabled inside of `NewBaseplateServer`


thriftbp/server_transport.go
```diff
@@ -0,0 +1,46 @@
+package thriftbp
+
+import (
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/reddit/baseplate.go/metricsbp"
+)
+
+// TODO(marco.ferrer):1/4/22 Replace with metric name that conforms to bp conventions
+const meterNameTransportConnCounter = "thrift.conn.count"
```
marcoferrer, line 9: Not sure what the convention is for naming metrics in baseplate so I left this placeholder to update

metricsbp/config.go
```diff
@@ -53,6 +53,12 @@ type Config struct {
 	//
 	// Optional, defaults to false.
 	RunSysStats bool `yaml:"runSysStats"`
+
+	// ReportServerConnectionCount indicates that you want to publish
+	// a counter for the number of clients connected to the server
+	//
+	// Optional, defaults to false.
+	ReportServerConnectionCount bool `yaml:"ReportServerConnectionCount"`
```
kylelemons, line 61: the other yaml things are `lowerCamelCase` which while I don't love, you want to at least be consistent :D

thriftbp/server.go
```diff
@@ -80,6 +85,10 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
 		transport = cfg.Socket
 	}
 
+	transport = BaseplateWrapperTServerTransport{
+		transport, cfg.ReportServerConnectionCount,
+	}
```
kylelemons, line 90: I think `false` should mean that you don't incur any overhead at all, so maybe something more like

```suggestion
  if cfg.ReportServerConnectionCount {
	  transport = CountedTServerTransport{transport}
	}
```

thriftbp/server_transport.go
```diff
@@ -0,0 +1,49 @@
...
+
+	return transport, nil
+}
+
+type MonitoredTTransport struct {
+	thrift.TTransport
+}
+
+func (m MonitoredTTransport) Close() error {
+	defer metricsbp.M.
```
kylelemons, line 35: defer doesn't seem necessary here?

thriftbp/server_transport.go
```diff
@@ -0,0 +1,49 @@
...
+		Counter(meterNameTransportConnCounter).
+		Add(-1)
+	return m.TTransport.Close()
+}
+
+func (m MonitoredTTransport) Open() error {
+	err := m.TTransport.Open()
+	if err == nil {
+		metricsbp.M.
```
kylelemons, line 44: you might want to do the metricsbp.M.Counter bit in the constructor of the TServerTransport and store it so you don't incur that overhead every time?

kylelemons (changes requested):

thriftbp/server_transport.go
konradreiche: Baseplate specific metrics go under `baseplate` but in this case it might be similar to Thrift client pool stats which we put under `pool-active-connections` and `pool-allocated-clients`. What I would say though is to name it just `thrift.connections`. The count part is an aggregation option and would come up anyway.

thriftbp/server_transport.go
```diff
@@ -0,0 +1,49 @@
...
+func (m *CountedTTransport) Close() error {
+	m.counter.Add(-1)
+	return m.TTransport.Close()
+}
+
+func (m *CountedTTransport) Open() error {
+	err := m.TTransport.Open()
+	if err == nil {
+		m.counter.Add(1)
+	}
```
konradreiche, line 47: :no_bell: 
```go
if err := m.TTransport.Open(); err != nil {
   return err
}
m.counter.Add(1)

**Review by konradreiche — CHANGES_REQUESTED** [review]
:+1: lgtm, we just need some clarification on what metric path we want to use.

File: thriftbp/server_transport.go (hunk 6 above)
Comment (marcoferrer) [reply thriftbp/server_transport.go]: done

File: thriftbp/server_transport.go [hunk 7]
```diff
@@ -0,0 +1,48 @@
...
+
+func NewCountedTTransport(transport thrift.TTransport) thrift.TTransport {
+	return &CountedTTransport{
+		TTransport: transport,
+		counter:    metricsbp.M.Counter(meterNameTransportConnCounter),
+	}
+}
+
+func (m *CountedTTransport) Close() error {
+	m.counter.Add(-1)
```
fishy, line 38: `TTransport.Close` can be called multiple times (calling it on an already closed transport will just be no-op or return an "already closed" error, can't remember exactly which), so just blindly doing -1 will likely give you inaccurate numbers.

There's `TTransport.IsOpen` you can use, but that will likely also give you inaccurate number, because it can return false without explicitly calling `Close` (for example, the other end closed the connection).

I think the closest we can get to a reliable way to implement it is to add some atomic value in `CountedTTransport`, so that we make sure that we only do -1 once and the future `Close` calls will not try to -1 again.

fishy (changes requested):

thriftbp/server_transport.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+const meterNameTransportConnCounter = "thrift.connections"
+
+type CountedTServerTransport struct {
+	thrift.TServerTransport
+}
+
+func (m *CountedTServerTransport) Accept() (thrift.TTransport, error) {
+
```
kylelemons, line 16:
```suggestion
```

thriftbp/server_transport.go
```diff
@@ -0,0 +1,48 @@
...
+	return NewCountedTTransport(transport), nil
+}
+
+type CountedTTransport struct {
+	thrift.TTransport
+	counter metrics.Counter
+}
+
+func NewCountedTTransport(transport thrift.TTransport) thrift.TTransport {
```
kylelemons, line 30: Since this is an exported constructor you should either:
* Return `*CountedTTransport` or
* Unexport `CountedTTransport` and keep returning the interface

I would probably slightly lean toward the latter since you don't expect people outside the package to need or name the type, just the functionality.

thriftbp/server_transport.go
kylelemons: I would put a `closeOnce sync.Once` in `CountedTTransport` and do

```
m.closeOnce.Do(func() {
  m.counter.Add(-1)
})
```

kylelemons (changes requested):

thriftbp/server_transport.go
fishy: yea once should work

metricsbp/config.go
```diff
@@ -53,6 +53,12 @@ type Config struct {
 	//
 	// Optional, defaults to false.
 	RunSysStats bool `yaml:"runSysStats"`
+
+	// ReportServerConnectionCount indicates that you want to publish
+	// a counter for the number of clients connected to the server
+	//
+	// Optional, defaults to false.
+	ReportServerConnectionCount bool `yaml:"reportServerConnectionCount"`
```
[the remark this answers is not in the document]
marcoferrer: Is there a way to set the server config from the application yaml?

thriftbp/server_transport.go
marcoferrer: 👍

fishy: another concern I have is that we are not really using true counters with statsd metrics. our telegraf configuration will reset every counter every minute, which means there could be a scenario that in a minute window we report more -1s than +1s, and causing telegraf to report a negative counter to the upstream, and that could cause problems.

we won't have that problem on prometheus (as we use true counters), but for statsd it probably makes more sense to use runtime gauges, so we report the per-pod gauge, instead of a global counter.

marcoferrer: So we use a counter for the inflight metrics of our rate limiter queue and havent run into this yet. Considering that we plan to move to prometheus and this is a new API we can keep it as is and verify in the beta phase.

metricsbp/config.go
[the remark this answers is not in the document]
kylelemons: Is that desireable? Metrics are not usually the domain of configuration

kylelemons (approved):

metricsbp/config.go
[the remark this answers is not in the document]
fishy: we can also just remove this configuration and just report it always. an additional counter/gauge adds negligible overhead to a go service.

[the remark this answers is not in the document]
marcoferrer: Let me know if I should update it to always enabled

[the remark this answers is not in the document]
fishy: @kylelemons what do you think to remove this configuration and just always on?

fishy (changes requested):
Besides the configuration field, the code looks good to me.

Regarding the configuration field, I'm fine with either remove it from `metricsbp.Config` and only provide it in `thriftbp.ServerConfig`, or just remove it altogether and make it always on. But the current state is misleading and should be avoided.

marcoferrer:
> Besides the configuration field, the code looks good to me.
> 
> Regarding the configuration field, I'm fine with either remove it from `metricsbp.Config` and only provide it in `thriftbp.ServerConfig`, or just remove it altogether and make it always on. But the current state is misleading and should be avoided.

Outside the scope of this PR, but we should probably come up with a preferred convention for sourcing configuration values from the application yaml. It seems like whenever theres a new option exposed to baseplate users they end up having to hard code their settigns or go through the work of creating a new field in their configuration struct. In this scenario, the path of least resistance usually wins. ie. hardcode settings.

metricsbp/config.go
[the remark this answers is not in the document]
kylelemons: The metric is negligible but the additional thunk call might not be... though it may be dwarfed by thrift overhead.

I don't have enough experience to know, but if both of you agree it's ok to enable by default then it probably is

thriftbp/server_transport.go
```diff
@@ -0,0 +1,54 @@
...
+func newCountedTTransport(transport thrift.TTransport) thrift.TTransport {
+	return &countedTTransport{
+		TTransport: transport,
+		counter:    metricsbp.M.Counter(meterNameTransportConnCounter),
+	}
+}
+
+func (m *countedTTransport) Close() error {
+	m.closeOnce.Do(func() {
+		m.counter.Add(-1)
```
fishy, line 43: Sorry I missed this before. for statsd metrics, counter with negative diff is in general a bad idea.

we should use runtime gauge for this instead. gauge also has `Add` function in addition to `Set`: https://pkg.go.dev/github.com/go-kit/kit/metrics#Gauge.Add

metricsbp/config.go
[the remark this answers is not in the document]
fishy: which is why I prefer to just move it to `thriftbp.ServerConfig` instead of default on :)

in the current cookiecutter we always need to pass in `thriftbp.ServerConfig` to `thriftbp.NewBaseplateServer` manually, so adding a config there manually to make it opt-in is not something too out of the scope.

If you want it from yaml, you can just make `thriftbp.ServerConfig` part of your `Config` struct and override the `Processor` field (as that cannot be deserialized from yaml anyways).

thriftbp/server.go
```diff
@@ -46,6 +46,11 @@ type ServerConfig struct {
 	// If not set none of the requests will be sampled.
 	ReportPayloadSizeMetricsSampleRate float64
 
+	// Optional, used only by NewBaseplateServer.
+	//
+	// Report the number of clients connected to the server.
+	ReportServerConnectionCount bool
```
fishy, line 52: 1. since this is part of `thriftbp.ServerConfig`, the `Server` part in the name is kinda redundant, probably should rename this to `ReportConnectionCount`.
2. Please also mention the metrics path in the comments (e.g. "Report the number of clients connected to the server as RuntimeGauge of thrift.connections")

thriftbp/server.go
```diff
@@ -46,6 +46,11 @@ type ServerConfig struct {
 	// If not set none of the requests will be sampled.
 	ReportPayloadSizeMetricsSampleRate float64
 
+	// Optional, used only by NewBaseplateServer.
```
fishy, line 49: This is actually used by both `NewBaseplateServer` and `NewServer`.

fishy (approved):
Only minor comments/naming issues remaining.

fishy (approved):

konradreiche (approved):

PR #465, opened 2022-01-07
Fix "_service" vs "_slug" labels

JessicaGreben:
There is some confusion around the `*_service` label representing the name of the rpc server when really it should be the fully qualified name of the rpc service. This PR attempts to fix that confusion.  

This PR also removes the thrift service label, re: https://github.snooguts.net/reddit/baseplate.spec/pull/40.

This PR also adds a prometheus metrics spec test for grpc.

### DIFF
--- a/grpcbp/client_middlewares.go
+++ b/grpcbp/client_middlewares.go
@@ -113,7 +113,7 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
 //
 // * grpc_client_active_requests gauge with labels:
 //
-//   - grpc_service: the fully qualified name of the gRPC service, serviceSlug arg
+//   - grpc_service: the fully qualified name of the gRPC service, the serviceName arg
 //   - grpc_method: the name of the method called on the gRPC service
 //   - grpc_slug: an arbitray short string representing the backend the client is connecting to
 //
@@ -124,7 +124,7 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
 //   - grpc_success: "true" if status is OK, "false" otherwise
 //   - grpc_type: type of request, i.e unary
 //   - grpc_code: the human-readable status code, e.g. OK, Internal, etc
-func PrometheusUnaryClientInterceptor(serviceSlug, serverSlug string) grpc.UnaryClientInterceptor {
+func PrometheusUnaryClientInterceptor(serviceName, serverSlug string) grpc.UnaryClientInterceptor {
 	return func(
 		ctx context.Context,
 		method string,
@@ -137,7 +137,7 @@ func PrometheusUnaryClientInterceptor(serviceSlug, serverSlug string) grpc.Unary
 		start := time.Now()
 		m := methodSlug(method)
 		activeRequestLabels := prometheus.Labels{
-			serviceLabel:           serviceSlug,
+			serviceLabel:           serviceName,
 			methodLabel:            m,
 			remoteServiceSlugLabel: serverSlug,
 		}
@@ -148,7 +148,7 @@ func PrometheusUnaryClientInterceptor(serviceSlug, serverSlug string) grpc.Unary
 			status, _ := status.FromError(err)
 
 			latencyLabels := prometheus.Labels{
-				serviceLabel:           serviceSlug,
+				serviceLabel:           serviceName,
 				methodLabel:            m,
 				typeLabel:              unary,
 				successLabel:           success,
@@ -158,7 +158,7 @@ func PrometheusUnaryClientInterceptor(serviceSlug, serverSlug string) grpc.Unary
 			clientLatencyDistribution.With(latencyLabels).Observe(time.Since(start).Seconds())
 
 			rpcCountLabels := prometheus.Labels{
-				serviceLabel:           serviceSlug,
+				serviceLabel:           serviceName,
 				methodLabel:            m,
 				typeLabel:              unary,
 				successLabel:           success,
--- a/grpcbp/client_middlewares.go
+++ b/grpcbp/client_middlewares.go
@@ -113,9 +113,9 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
 //
 // * grpc_client_active_requests gauge with labels:
 //
-//   - grpc_service: the fully qualified name of the gRPC service, the serviceName arg
+//   - grpc_service: the fully qualified name of the gRPC service
 //   - grpc_method: the name of the method called on the gRPC service
-//   - grpc_slug: an arbitray short string representing the backend the client is connecting to
+//   - grpc_slug: an arbitray short string representing the backend the client is connecting to, the serverSlug arg
 //
 // * grpc_client_latency_seconds histogram and grpc_client_requests_total
 //   counter with labels:
@@ -124,7 +124,7 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
 //   - grpc_success: "true" if status is OK, "false" otherwise
 //   - grpc_type: type of request, i.e unary
 //   - grpc_code: the human-readable status code, e.g. OK, Internal, etc
-func PrometheusUnaryClientInterceptor(serviceName, serverSlug string) grpc.UnaryClientInterceptor {
+func PrometheusUnaryClientInterceptor(serverSlug string) grpc.UnaryClientInterceptor {
 	return func(
 		ctx context.Context,
 		method string,
@@ -135,7 +135,8 @@ func PrometheusUnaryClientInterceptor(serviceName, serverSlug string) grpc.Unary
 		opts ...grpc.CallOption,
 	) (err error) {
 		start := time.Now()
-		m := methodSlug(method)
+		serviceName, m := serviceAndMethodSlug(method)
+
 		activeRequestLabels := prometheus.Labels{
 			serviceLabel:           serviceName,
 			methodLabel:            m,
--- a/grpcbp/client_middlewares.go
+++ b/grpcbp/client_middlewares.go
@@ -138,9 +138,9 @@ func PrometheusUnaryClientInterceptor(serverSlug string) grpc.UnaryClientInterce
 		serviceName, m := serviceAndMethodSlug(method)
 
 		activeRequestLabels := prometheus.Labels{
-			serviceLabel:           serviceName,
-			methodLabel:            m,
-			remoteServiceSlugLabel: serverSlug,
+			serviceLabel:          serviceName,
+			methodLabel:           m,
+			remoteServerSlugLabel: serverSlug,
 		}
 		clientActiveRequests.With(activeRequestLabels).Inc()
 
@@ -149,22 +149,22 @@ func PrometheusUnaryClientInterceptor(serverSlug string) grpc.UnaryClientInterce
 			status, _ := status.FromError(err)
 
 			latencyLabels := prometheus.Labels{
-				serviceLabel:           serviceName,
-				methodLabel:            m,
-				typeLabel:              unary,
-				successLabel:           success,
-				remoteServiceSlugLabel: serverSlug,
+				serviceLabel:          serviceName,
+				methodLabel:           m,
+				typeLabel:             unary,
+				successLabel:          success,
+				remoteServerSlugLabel: serverSlug,
 			}
 
 			clientLatencyDistribution.With(latencyLabels).Observe(time.Since(start).Seconds())
 
 			rpcCountLabels := prometheus.Labels{
-				serviceLabel:           serviceName,
-				methodLabel:            m,
-				typeLabel:              unary,
-				successLabel:           success,
-				remoteServiceSlugLabel: serverSlug,
-				codeLabel:              status.Code().String(),
+				serviceLabel:          serviceName,
+				methodLabel:           m,
+				typeLabel:             unary,
+				successLabel:          success,
+				remoteServerSlugLabel: serverSlug,
+				codeLabel:             status.Code().String(),
 			}
 			clientRPCRequestCounter.With(rpcCountLabels).Inc()
 			clientActiveRequests.With(activeRequestLabels).Dec()
@@ -177,7 +177,7 @@ func PrometheusUnaryClientInterceptor(serverSlug string) grpc.UnaryClientInterce
 // monitoring for Streaming RPCs.
 //
 // This is not implemented yet.
-func PrometheusStreamClientInterceptor(serviceSlug, serverSlug string) grpc.StreamClientInterceptor {
+func PrometheusStreamClientInterceptor(serverSlug string) grpc.StreamClientInterceptor {
 	return func(
 		ctx context.Context,
 		desc *grpc.StreamDesc,
--- a/grpcbp/client_middlewares.go
+++ b/grpcbp/client_middlewares.go
@@ -138,9 +138,9 @@ func PrometheusUnaryClientInterceptor(serverSlug string) grpc.UnaryClientInterce
 		serviceName, m := serviceAndMethodSlug(method)
 
 		activeRequestLabels := prometheus.Labels{
-			serviceLabel:          serviceName,
-			methodLabel:           m,
-			remoteServerSlugLabel: serverSlug,
+			serviceLabel:    serviceName,
+			methodLabel:     m,
+			serverSlugLabel: serverSlug,
 		}
 		clientActiveRequests.With(activeRequestLabels).Inc()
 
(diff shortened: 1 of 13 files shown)


konradreiche (approved):

grpcbp/server_middlewares.go
```diff
@@ -164,14 +164,14 @@ func StartSpanFromGRPCContext(ctx context.Context, name string) (context.Context
 //   - grpc_success: "true" if status is OK, "false" otherwise
 //   - grpc_type: type of request, i.e unary
 //   - grpc_code: the human-readable status code, e.g. OK, Internal, etc
-func InjectPrometheusUnaryServerInterceptor(serviceSlug string) grpc.UnaryServerInterceptor {
+func InjectPrometheusUnaryServerInterceptor(serviceName string) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (_ interface{}, err error) {
 		start := time.Now()
 
 		method := methodSlug(info.FullMethod)
```
kylelemons, line 171: https://pkg.go.dev/google.golang.org/grpc#UnaryServerInfo

this gives both the serviceName and the method

thriftbp/server_middlewares.go
```diff
@@ -391,7 +391,7 @@ func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessor
 //
 // * thrift_server_active_requests gauge with labels:
 //
-//   - thrift_service: the serviceSlug arg
+//   - thrift_service: the name of the thrift service, the serviceName arg
```
kylelemons, line 394: we're going to remove this one

kylelemons (changes requested):

thriftbp/server_middlewares.go
JessicaGreben: just pushed a commit to remove this

grpcbp/server_middlewares.go
JessicaGreben: nice. let me fix. thanks.

grpcbp/tracing.go
```diff
@@ -44,3 +44,13 @@ func methodSlug(method string) string {
 	split := strings.Split(method, "/")
 	return split[len(split)-1]
 }
+
+// serviceAndMethodSlug splits the UnaryServerInfo.FullMethod and returns
+// the package.service part separate from the method part.
+// ref: https://pkg.go.dev/google.golang.org/grpc#UnaryServerInfo
```
fishy, line 50: The example given by that is `/package.service/method`, which would produce `/package.service/` and `method` as the return values, and using `/package.service/` as the service name sounds weird to me.

Also a better place for this function might be `grpcbp/prometheus.go`?

Also since both return values are strings, please use named returns as a documentation

grpcbp/tracing.go
```diff
@@ -44,3 +44,13 @@ func methodSlug(method string) string {
 	split := strings.Split(method, "/")
 	return split[len(split)-1]
 }
```
fishy, line 46: 🔕 this function is likely no longer used?

grpcbp/server_middlewares.go
```diff
@@ -154,7 +154,7 @@ func StartSpanFromGRPCContext(ctx context.Context, name string) (context.Context
 //
```
fishy, line 154: since you removed `service` suffix from `spectest`, we need to add the tests for `gprc_service` label here manually. from my quick glance of the code I don't believe we are producing the expected `grpc_service` label values.

fishy (changes requested):

grpcbp/tracing.go
kylelemons: +1 -- I would expect "package.service" and "method" when given "/package.service/method" as input

grpcbp/tracing.go
```diff
@@ -44,3 +44,13 @@ func methodSlug(method string) string {
...
+
+// serviceAndMethodSlug splits the UnaryServerInfo.FullMethod and returns
+// the package.service part separate from the method part.
+// ref: https://pkg.go.dev/google.golang.org/grpc#UnaryServerInfo
+func serviceAndMethodSlug(fullMethod string) (string, string) {
+	split := strings.Split(fullMethod, "/")
```
kylelemons, line 52: strings.SplitN(fullMethod, 3)

grpcbp/tracing.go
```diff
@@ -44,3 +44,13 @@ func methodSlug(method string) string {
...
+	method := split[len(split)-1]
```
kylelemons, line 53: check len(split) before indexing to avoid panics

thriftbp/client_middlewares.go
```diff
@@ -326,9 +326,8 @@ var (
 //
 // * thrift_client_active_requests gauge with labels:
 //
-//   - thrift_service: the serviceSlug arg
 //   - thrift_method: the method of the endpoint called
-//   - thrift_slug: the service being contacted
+//   - thrift_slug: an arbitray short string representing the backend the client is connecting to, the remoteServiceSlug arg
```
kylelemons, line 330: remoteServerSlug

thriftbp/client_middlewares.go
```diff
@@ -342,13 +341,12 @@ var (
 //     as a string if present (e.g. 404), or the empty string
 //   - thrift_baseplate_status_code: the human-readable status code, e.g.
 //     NOT_FOUND, or the empty string
-func PrometheusClientMiddleware(localServiceSlug, remoteServiceSlug string) thrift.ClientMiddleware {
+func PrometheusClientMiddleware(remoteServiceSlug string) thrift.ClientMiddleware {
```
kylelemons, line 344: server, not service

thriftbp/client_middlewares_test.go
```diff
@@ -544,7 +541,7 @@ func (srv mockBaseplateService) IsHealthy(ctx context.Context, req *baseplatethr
 func setupFake(ctx context.Context, t *testing.T, handler baseplatethrift.BaseplateServiceV2) thriftbp.ClientPool {
 	srv, err := thrifttest.NewBaseplateServer(thrifttest.ServerConfig{
 		Processor:         baseplatethrift.NewBaseplateServiceV2Processor(handler),
-		ClientMiddlewares: []thrift.ClientMiddleware{thriftbp.PrometheusClientMiddleware(localSvr, remoteSvr)},
+		ClientMiddlewares: []thrift.ClientMiddleware{thriftbp.PrometheusClientMiddleware(remoteSvr)},
```
kylelemons, line 544: svr / srv should be consistent.  I think srv is the more commently used (though I would also not mind writing it out as server or service)

kylelemons (changes requested):
just a couple precision things left!

grpcbp/tracing.go
kylelemons: just in case somehow we're give "/foo/bar/baz" -- so that we return "bar/baz"

You could also trim a / prefix from fullMethod and then SplitN with 2, so that "foo/bar" also works just as "/foo/bar" would

grpcbp/tracing.go
JessicaGreben: there is one other use of this method `methodSlug` in `MonitorInterceptorUnary`

grpcbp/server_middlewares.go
JessicaGreben: you are correct. the spec test currently only checks the thrift metrics. i need to add support for grpc and http. i think i will do that in a separate PR if thats ok.

grpcbp/prometheus.go
```diff
@@ -100,3 +102,16 @@ var (
...
+// ref: https://pkg.go.dev/google.golang.org/grpc#UnaryServerInfo
+func serviceAndMethodSlug(fullMethod string) (service string, method string) {
+	split := strings.SplitN(fullMethod, "/", 3)
+	if len(split) < 2 {
+		return "", ""
+	}
+	method = split[len(split)-1]
+	service = strings.Join(split[:len(split)-1], "")
+	return service, method
+}
```
kylelemons, line 117:
```
func serviceAndMethodSlug(fullMethod string) (service, method string) {
	split := strings.SplitN(fullMethod, "/", 3)
	if len(split) < 3 || split[0] != "" {
		return "", fullMethod
	}
	return split[1], split[2]
}
```

or

```
func serviceAndMethodSlug(fullMethod string) (service, method string) {
	fullMethod = strings.TrimPrefix(fullMethod, "/")
	split := strings.SplitN(fullMethod, "/", 2)
	if len(split) < 2 {
		return "", fullMethod
	}
	return split[0], split[1]
}
```

I kinda like the second one.

grpcbp/client_middlewares.go
```diff
@@ -124,7 +124,7 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
 //   - grpc_success: "true" if status is OK, "false" otherwise
 //   - grpc_type: type of request, i.e unary
 //   - grpc_code: the human-readable status code, e.g. OK, Internal, etc
-func PrometheusUnaryClientInterceptor(serviceSlug, serverSlug string) grpc.UnaryClientInterceptor {
+func PrometheusUnaryClientInterceptor(serverSlug string) grpc.UnaryClientInterceptor {
```
fishy, line 127: nit: now it probably makes more sense to rename this arg to just `slug`

grpcbp/prometheus_test.go
```diff
@@ -0,0 +1,57 @@
...
+		{
+			name:        "extra /",
+			fullMethod:  "/foo/bar/baz",
+			wantService: "foo",
+			wantMethod:  "bar/baz",
+		},
+		{
+			name:        "no /",
+			fullMethod:  "package.service.method",
+			wantService: "",
```
kylelemons, line 35: since this is for monitoring, you want to fail closed.  In this case, that probably means that when you don't recognize the format, you stick it all in the "method" label, since that's the most specific label and most likely matches the expected cardinality.

grpcbp/prometheus.go
```diff
@@ -100,3 +102,16 @@ var (
...
+
+// serviceAndMethodSlug splits the UnaryServerInfo.FullMethod and returns
+// the package.service part separate from the method part.
+// ref: https://pkg.go.dev/google.golang.org/grpc#UnaryServerInfo
+func serviceAndMethodSlug(fullMethod string) (service string, method string) {
+	split := strings.SplitN(fullMethod, "/", 3)
+	if len(split) < 2 {
+		return "", ""
+	}
+	method = split[len(split)-1]
```
fishy, line 114: since you already checked length, at this point the only possible `len(split)` is 3, so you can just use `method = split[2]` here.

grpcbp/prometheus_test.go
```diff
@@ -0,0 +1,57 @@
...
+			wantMethod:  "",
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			gotService, gotMethod := serviceAndMethodSlug(tt.fullMethod)
+			if got, want := gotService, tt.wantService; got != want {
+				t.Errorf("got %v, want %v", got, want)
```
kylelemons, line 50: if the output was
```
got "a", want "b"
got "c", want "d"
```

you wouldn't know which was which.  So we prefix them:

```suggestion
				t.Errorf("serviceAndMethodSlug(%q).service = %v, want %v",  tt.fullMethod, got, want)
```

Thus, if you see this output

```
serviceAndMethodSlug("/package.service/method").service = "", want "package.service"
serviceAndMethodSlug("/package.service/method").method = "package.service", want "method"
```

you actally don't need to look at the test at all, it includes all the info you need to understand the failure.

grpcbp/prometheus.go
```diff
@@ -100,3 +102,16 @@ var (
...
+	service = strings.Join(split[:len(split)-1], "")
```
fishy, line 115: I... really don't know what I should expect from the full method in a grpc call. the example they gave will have `split[0]` being the empty string and `split[1]` being the service fqn, so this join works. but is there a scenario that `split[0]` will be non-empty? and if that happens, do we want to just join that against `split[1]`?

can someone link to some grpc doc to explain what to expect here? the ref linked above only provided one example.

kylelemons (changes requested):

grpcbp/prometheus.go
kylelemons: Put differently, I think the static indices are easier to reason about than len-1 / len-2 -- and checking `len(split) < 2` when you have `SplitN(..., 3)` looks like an off-by-one error, even though it's not.  (it took me awhile to recognize this)

grpcbp/server_middlewares.go
fishy: that means this PR can produce totally wrong labels and that will only get fixed by the next PR that adds the tests, which doesn't sound ideal to me.

grpcbp/prometheus.go
kylelemons: 😅 actually len can be 2 or 3

fishy: you are right

grpcbp/prometheus.go
kylelemons: [gRPC's http/2 interface uses /service/method as the path -- see PROTOCOL-HTTP2 and search for :path](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)

grpcbp/server_middlewares.go
JessicaGreben: ok i will add the spec test changes here to include grpc

grpcbp/prometheus.go
JessicaGreben: agree, i like the second one. will add.

internal/prometheusbp/spectest/spec.go
```diff
@@ -142,22 +145,42 @@ func validateLabels(name, prefix, clientOrServer string, gotLabels map[string]st
 
 // buildLables returns a set of expected labels for the metric name provided.
 // prefix is either thrift, http, or grpc.
+func buildLables(name, prefix, clientOrServer string) map[string]struct{} {
```
kylelemons, line 148: `Lables` -> `Labels`

internal/prometheusbp/spectest/spec.go
```diff
@@ -142,22 +145,42 @@ func validateLabels(name, prefix, clientOrServer string, gotLabels map[string]st
...
+	var labelSuffixes = []string{}
```
kylelemons, line 149:
```suggestion
	var labelSuffixes []string
```

As written this conceptually (though not in practice) "allocates" a new slice, when appenad and such all work just fine on the zero value (nil) as above.

internal/prometheusbp/spectest/spec.go
```diff
@@ -142,22 +145,42 @@ func validateLabels(name, prefix, clientOrServer string, gotLabels map[string]st
...
+	switch prefix {
+	case thriftPrefix:
+		labelSuffixes = thriftSpecificLabels(name)
+	case grpcPrefix:
+		labelSuffixes = grpcSpecificLabels(name)
+	default:
+		labelSuffixes = []string{}
```
kylelemons, line 156: this is a no-op, and it should be `= nil` since "allocating" a zero slice is unnecessary.

Probably just drop the default

internal/prometheusbp/spectest/spec.go
```diff
@@ -142,22 +145,42 @@ func validateLabels(name, prefix, clientOrServer string, gotLabels map[string]st
...
+	}
+
+	if clientOrServer == client {
+		labelSuffixes = append(labelSuffixes, "slug")
+	}
+
+	var wantLabels = map[string]struct{}{}
```
kylelemons, line 163: Minor memory optimization:

```suggestion
	wantLabels := make(map[string]struct{}, len(labelSuffixes))
```

Not necessary in this specific case, since it's a test, but a good habit

internal/prometheusbp/spectest/spec.go
```diff
@@ -168,16 +191,37 @@ func buildLables(name, prefix, clientOrServer string) map[string]struct{} {
...
+	labelSuffixes := []string{"service", "method"}
+	switch {
+	case strings.HasSuffix(name, "_latency_seconds"):
+		labelSuffixes = append(labelSuffixes, "type", "success")
+	case strings.HasSuffix(name, "_requests_total"):
+		labelSuffixes = append(labelSuffixes, "type", "success", "code")
+	case strings.HasSuffix(name, "_active_requests"):
+		// no op
+	default:
+		labelSuffixes = []string{}
```
kylelemons, line 222: I would probably say "return nil" if you want to emphasize that you are "ignoring" labels for things that don't match

kylelemons (changes requested):
Thanks so much for getting this done! Just a few minor points

grpcbp/prometheus.go
```diff
@@ -1,19 +1,21 @@
...
-	typeLabel              = "grpc_type"
-	successLabel           = "grpc_success"
-	codeLabel              = "grpc_code"
-	remoteServiceSlugLabel = "grpc_slug"
+	serviceLabel          = "grpc_service"
+	methodLabel           = "grpc_method"
+	typeLabel             = "grpc_type"
+	successLabel          = "grpc_success"
+	codeLabel             = "grpc_code"
+	remoteServerSlugLabel = "grpc_slug"
```
fishy, line 18: we originally added `remote-` prefix to this name because we thought the other label (`grpc_service`) is for local. but now we know both are for the remote, the `remote-` prefix in the name doesn't make much sense any more, and I'd suggest just rename this const to `serverSlugLabel` or even just `slugLabel`.

fishy (approved):

kylelemons (approved):

grpcbp/prometheus.go
JessicaGreben: i have a slight preference to `serverSlugLabel`. Will make that change.

PR #466, opened 2022-01-10
Add Prometheus metrics to http, grpc, and thrift.

JessicaGreben:
This PR adds the Prometheus branch to master which contains all default Prometheus metrics for HTTP, GRPC, and Thrift client and servers based on the [baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#http).

Background:


JessicaGreben:
made a new branch and copied over commits to fix merge conflicts: https://github.com/reddit/baseplate.go/pull/467

PR #467, opened 2022-01-10
Add Prometheus metrics to http, grpc, and thrift. 

JessicaGreben:
This PR adds the Prometheus branch to master which contains all default Prometheus metrics for HTTP, GRPC, and Thrift client and servers based on the [baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#http).

Background:


grpcbp/client_middlewares.go
```diff
@@ -101,3 +105,87 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+		ctx context.Context,
+		method string,
+		req interface{},
+		reply interface{},
+		cc *grpc.ClientConn,
+		invoker grpc.UnaryInvoker,
+		opts ...grpc.CallOption,
+	) (err error) {
+		start := time.Now()
+		serviceName, m := serviceAndMethodSlug(method)
```
kylelemons, line 138: sorry for missing this; I don't really like one-letter variable names unless I could tell you what it meant with no context (i, w, r, etc.)... so maybe name this "method" and call method "serviceMethod" or "wireMethod"?

grpcbp/client_middlewares.go
```diff
@@ -101,3 +105,87 @@ func ForwardEdgeContextStreaming(ecImpl ecinterface.Interface) grpc.StreamClient
...
+//   - grpc_success: "true" if status is OK, "false" otherwise
+//   - grpc_type: type of request, i.e unary
+//   - grpc_code: the human-readable status code, e.g. OK, Internal, etc
+func PrometheusUnaryClientInterceptor(serverSlug string) grpc.UnaryClientInterceptor {
+	return func(
+		ctx context.Context,
+		method string,
+		req interface{},
+		reply interface{},
+		cc *grpc.ClientConn,
```
kylelemons, line 133: conn or clientConn?

grpcbp/server_middlewares_test.go
```diff
@@ -181,3 +184,123 @@ func (t *mockService) PingList(req *pb.PingRequest, c pb.TestService_PingListSer
...
+			}
+
+			clientActiveRequestLabels := prometheus.Labels{
+				serviceLabel:    serviceName,
+				methodLabel:     tt.method,
+				serverSlugLabel: serverSlug,
+			}
+
+			defer promtest.NewPrometheusMetricTest(t, "server latency", serverLatencyDistribution, serverLatencyLabels).CheckExists()
+			defer promtest.NewPrometheusMetricTest(t, "server rpc count", serverRPCRequestCounter, serverTotalRequestLabels).CheckDelta(1)
```
kylelemons, line 283: Some of these are named almost identically, e.g. "clientActiveRequests" <-> "clientActiveRequestLabels" -- maybe the rest should be unified too?  This (RPC vs Total) stands out the most

kylelemons (changes requested):
Definitely interested in getting this into preference-go and seeing what it says!

konradreiche (approved):

kylelemons (approved):

fishy (approved):
Since we added fixup commits to this branch, we can no longer _just_ rebase merge this PR (which will make the fixup commits land on master branch).

Do we want to squash everything into a single commit? Or if we still want individual commits to be rebased onto master branch in the end, then we need to figure out how to fix every individual commits.

JessicaGreben:
@fishy I was assuming we would squash everything. But if we prefer individual commits, then I need to fix this. Let me know @fishy/@kylelemons .

fishy:
We went thought the rebase before so that we could still keep the individual commits in the end 😅 

But I'm also not against just squash a single commit in the end ¯\\\_(ツ)\_/¯

kylelemons:
Recommended merge strategy:
* Combine the last ones (that don't ahve a PR number) into one commit, say titled "Final polish for merge into master (#467)"
* and then rebase-and-merge.

This keeps the work units nicely separated, keeps links to the original PRs, etc

PR #470, opened 2022-01-13
add thrift socket timeout impl

marcoferrer:
This PR implements the experimental server configuration `ThriftSocketTimeout` as detailed in internal discussions.

### DIFF
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -1,13 +1,19 @@
 package thriftbp
 
 import (
+	"strings"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/reddit/baseplate.go/metricsbp"
 
 	"github.com/reddit/baseplate.go"
 	"github.com/reddit/baseplate.go/errorsbp"
 	"github.com/reddit/baseplate.go/log"
+)
+
+const (
+	meterNameThriftSocketErrorCounter = "thrift.socket.timeout"
 )
 
 // ServerConfig is the arg struct for both NewServer and NewBaseplateServer.
@@ -57,6 +63,16 @@ type ServerConfig struct {
 	// Deprecated: No-op for now, will be removed in a future release.
 	Timeout time.Duration
 
+	// Optional, This duration is used to set both the read and write idle timeouts
+	// for the thrift.TServerSocket used by the baseplate server.
+	//
+	// This is an experimental configuration and is subject to change or deprecation
+	// without notice. Setting a socket timeout will also override the default thrift
+	// server logger to one that emits metrics instead of logs in the event of a
+	// socket disconnect. A zero value means I/O read or write operations
+	// will not time out.
+	ThriftSocketTimeout time.Duration
+
 	// Optional, used only by NewServer.
 	// In NewBaseplateServer the address and timeout set in bp.Config() will be
 	// used instead.
@@ -72,7 +88,7 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
 	var transport *thrift.TServerSocket
 	if cfg.Socket == nil {
 		var err error
-		transport, err = thrift.NewTServerSocket(cfg.Addr)
+		transport, err = thrift.NewTServerSocketTimeout(cfg.Addr, cfg.ThriftSocketTimeout)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


thriftbp/server.go
```diff
@@ -57,6 +63,16 @@ type ServerConfig struct {
...
 
+	// Optional, This duration is used to set both the read and write idle timeouts
+	// for the thrift.TServerSocket used by the baseplate server.
+	//
+	// This is an experimental configuration and is subject to change or deprecation
+	// without notice. Setting a socket timeout will also override the default thrift
+	// server logger to one that emits metrics instead of logs in the event of a
+	// socket disconnect. A zero value means I/O read or write operations
+	// will not time out.
+	ThriftSocketTimeout time.Duration
```
fishy, line 74: this is already inside `thriftbp.ServerConfig` so `Thrift-` prefix in the name is kinda redundant.

I would also suggest to just keep using `Timeout` above, and keep the "Deprecated:" comment. so it's still deprecated (if people try to use it, staticcheck will complain, and they can add a `//go:lint` line to suppress the complaint when they know they do need it). we can remove the deprecated comment if we decide that we want to keep it long term.

we also need to add some clarifications to how this is used in `NewBaseplaseServer` vs. `NewServer`:

1. The timeout itself is used by both.
2. The logger behavior currently mentioned here only applies to `NewBaseplateServer`.

fishy (changes requested):

thriftbp/server.go
```diff
@@ -72,7 +88,7 @@ func NewServer(cfg ServerConfig) (*thrift.TSimpleServer, error) {
 	var transport *thrift.TServerSocket
 	if cfg.Socket == nil {
 		var err error
-		transport, err = thrift.NewTServerSocket(cfg.Addr)
+		transport, err = thrift.NewTServerSocketTimeout(cfg.Addr, cfg.ThriftSocketTimeout)
```
kylelemons, line 91: I assume this doesn't set a timeout if it's zero?

I _might_ do an explicit

```
if cfg.ThriftSocketTimeout > 0 && cfg.Socket == nil { ...
  // call NewTServerSocketTimeout
  cfg.Logger = ...
} else if cfg.Socket == nil {
  // original code
}
```

This keeps the special / experimental codepath easy to see and keeps all of the repercussions of setting it in a single narrow branch

kylelemons (changes requested):

thriftbp/server.go
fishy: calling `NewTServerSocketTimeout` with 0 timeout is the same as no timeout. but calling `NewTServerSocketTimeout` with negative timeout could have undefined behavior. so +1 to add an explicit check.

thriftbp/server.go
marcoferrer: Based on the discussion I think we're intentionally using a new configuration field so that existing usages of `Timeout` can continue to function as a noop. Re-using the field would undo that.

fishy: That's fair, but my points about marking it as deprecated and the `Thrift-` prefix in the name is redundant still stand.

thriftbp/server.go
```diff
@@ -106,12 +124,7 @@ func NewBaseplateServer(
 	)
 	middlewares = append(middlewares, cfg.Middlewares...)
 	cfg.Middlewares = middlewares
-	cfg.Logger = log.ZapWrapper(log.ZapWrapperArgs{
-		Level: bp.GetConfig().Log.Level,
-		KVPairs: map[string]interface{}{
-			"from": "thrift",
-		},
-	}).ToThriftLogger()
+	cfg.Logger = buildTServerLogger(bp.GetConfig().Log.Level, cfg.ThriftSocketTimeout > 0)
```
kylelemons, line 127: 🔕 Can this be rewritten to be like

```
cfg.Logger = ...
if cfg.ThriftSocketTimeout > 0 {
  cfg.Logger = suppressTimeoutLogger(cfg.Logger)
}
```

?

kylelemons (approved):

thriftbp/server.go
marcoferrer: fixed

konradreiche (changes requested):

konradreiche (approved):

fishy (approved):

PR #471, opened 2022-01-18
Add prometheus middleware to default middleware

JessicaGreben:
While @kylelemons and I were working on making a new [upgrader](https://github.snooguts.net/reddit-go/baseplate-go-updater) for baseplate.go v0.9.2, we realized it would be better to add the Prometheus metric middleware by default so we don't have to add it for each repo. That means we would like to cut another tag v0.9.3 after this change is merged.

### DIFF
--- a/httpbp/client_middlewares.go
+++ b/httpbp/client_middlewares.go
@@ -38,8 +38,8 @@ func (f roundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
 
 // NewClient returns a standard HTTP client wrapped with the default middleware
 // plus any additional client middleware passed into this function. Default
-// middlewares are: MonitorClient and Retries. ClientErrorWrapper is included
-// as transitive middleware through Retries.
+// middlewares are: MonitorClient, Retries, and PrometheusClientMetrics.
+// ClientErrorWrapper is included as transitive middleware through Retries.
 func NewClient(config ClientConfig, middleware ...ClientMiddleware) (*http.Client, error) {
 	if err := config.Validate(); err != nil {
 		return nil, err
--- a/httpbp/client_middlewares.go
+++ b/httpbp/client_middlewares.go
@@ -64,6 +64,7 @@ func NewClient(config ClientConfig, middleware ...ClientMiddleware) (*http.Clien
 	defaults := []ClientMiddleware{
 		MonitorClient(config.Slug),
 		Retries(config.MaxErrorReadAhead, config.RetryOptions...),
+		PrometheusClientMetrics(config.Slug),
 	}
 
 	// prepend middleware to ensure Retires with ClientErrorWrapper is still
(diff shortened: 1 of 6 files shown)


httpbp/client_middlewares.go
```diff
@@ -64,6 +64,7 @@ func NewClient(config ClientConfig, middleware ...ClientMiddleware) (*http.Clien
 	defaults := []ClientMiddleware{
 		MonitorClient(config.Slug),
 		Retries(config.MaxErrorReadAhead, config.RetryOptions...),
+		PrometheusClientMetrics(config.Slug),
```
fishy, line 67: please update the doc comment of this function (which described what are the default middlewares added). same for other functions affected.

httpbp/middlewares.go
```diff
@@ -433,7 +434,7 @@ func RecordStatusCode() Middleware {
 // * http_server_requests_total counter with all labels above plus:
 //
 //   - http_response_code: numeric status code as a string, e.g. 200
-func PrometheusServerMetrics(serverSlug string) Middleware {
+func PrometheusServerMetrics() Middleware {
```
fishy, line 437: uh this is a breaking change. we can't cut this as `v0.9.3`.

fishy (changes requested):

httpbp/middlewares.go
fishy: it looks like we accidentally added an arg that's not really used. so in this version we should just use `""` in the default middleware function, and add a TODO to remove this arg from this function in a future, breaking change.

fishy: we can also rename this arg into `unused` and add some comment in the doc comment of this function to explain that this arg is unused and will be removed in a future version.

JessicaGreben: :(

kylelemons: My policy on breaking changes is that we can make them as long as it doesn't break anyone, _especially_ since this is still technically v0.  I am not aware of this API being used outside of baseplate.go, so we should fix it and fix it before it does become used.

If you really want to, we can retract v0.9.2

See the baseplate 2022 compatibility promise:

https://docs.google.com/document/d/1RD5bDhmnQir-gNyEMFWiZjtyc6c8_s51smHkPNlbr4w/edit#heading=h.tsja10upcmb0

kylelemons (approved):

httpbp/middlewares.go
fishy: A breaking change is a breaking change. no one uses it doesn't make it non-breaking.

If someone is already using `v0.9.2` and using this function, breaking it in `v0.9.3` will mean that they need to make a breaking change when they upgrade to `v0.9.3`. retracting `v0.9.2` will help preventing new services from using it (so we probably should still do), but it does not help people already using it.

if no one is using this function from `v0.9.2`, then after `v0.9.3` is cut no one needs to call it any more anyways (because it's already in the default middlewares), and when we make the breaking change in `v0.10.0` they don't need to fix it because they are not using it directly (but via the default middlewares that will also be fixed in `v0.10.0`).

fishy (approved):

konradreiche (approved):

fishy (approved):

PR #472, opened 2022-01-21
feat(filewatcher,secrets): Support for  parsing  Vault CSI Directories

Tediferous:
Did a rework of this PR. instead of adding a new object ive just allowed filewatcher to handle directories as well

This feat gives baseplate-go a feature that has been in baseplate-py for a while


directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+	Stop()
+}
+
+// AddFile is a type of function that should be ran to handle
+// adding data after a file has been parsed
+type AddFile func(d interface{}, file interface{}) (data interface{}, err error)
+
+// RemoveFile is a type of function that should be ran to handle removing
+// data after a file has been removed from the watcher
+type RemoveFile func(d interface{}, path string) (data interface{}, err error)
```
fishy, line 41: I assume an user case would be that we split a huge json file into smaller ones, and the `AddFile` callback is supposed to stitch the json data together into a single object. assuming we have:

```json
// foo.json
{
  "foo": {
    "a": 1,
    "b": 2,
  }
}

// bar.json
{
  "bar": {
    "a": 3,
    "b": 4,
  }
}
```

so we want the stitched together object be something like:
```go
map[string]object{
  "foo": object{
    A: 1,
    B: 2,
  },
  "bar": object{
    A: 3,
    B: 4,
  },
}
```

Now `bar.json` is removed, `RemoveFile` is called with `d` being the whole object and path being `bar.json`. But how does the implementation know that it should remove the `bar` key from the object? The args of `RemoveFile` and `AddFile` are not symmetry, so it has no knowledge on the association between the path and the actual data to remove.

I'm also questioning the decision to make it recursive. From the currently defined interface we are treating all the files equally (e.g. there's no difference between "<root>/foo.json" and "<root>/foo/bar.json" if they have the same content), but if we treat all files equally what's the point of having directory hierarchy?

fishy (changes requested):
duplicate a lot of the code from filewatcher to directorywatcher is also something we should avoid. it likely makes much more sense to consolidate everything into `filewatcher`, we just auto detect that if the path is a file it's a single file, if the path is a directory we watch the directory instead.

also I would really love to see a design doc of this first.

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+	"github.com/reddit/baseplate.go/filewatcher"
+	"github.com/reddit/baseplate.go/internal/limitopen"
+	"github.com/reddit/baseplate.go/log"
+)
+
+// DirectoryWatcher loads and parses data from a file and watches for changes to that
+// file in order to refresh it's stored data.
+type DirectoryWatcher interface {
+	// Get returns the latest, parsed data from the DirectoryWatcher.
+	Get() interface{}
```
kylelemons, line 23: I strongly prefer APIs structured as

```
Parse(dstPtr interface{}) error
```

so that the caller doesn't have to deal with type assertions

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// Stop stops the DirectoryWatcher.
+	//
+	// After Stop is called you won't get any updates on the file content,
+	// but you can still call Get to get the last content before stopping.
+	//
+	// It's OK to call Stop multiple times.
+	// Calls after the first one are essentially no-op.
+	Stop()
```
kylelemons, line 32: This is typically spelled `Close() error` -- types that have cleanup operations typically implement the `io.Closer` interface.

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+	"gopkg.in/fsnotify.v1"
+
+	"github.com/reddit/baseplate.go/filewatcher"
+	"github.com/reddit/baseplate.go/internal/limitopen"
+	"github.com/reddit/baseplate.go/log"
+)
+
+// DirectoryWatcher loads and parses data from a file and watches for changes to that
+// file in order to refresh it's stored data.
+type DirectoryWatcher interface {
```
kylelemons, line 21: "Provide concrete types, accept interfaces."

In this case, you are providing `*Result` and thus don't need to export the interface that it implements.  I don't think you use this interface at all, so it can probably be removed.  Callers can define their own interface if they need to abstract over a subset of its behavior.

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+				// do nothing.
+			case fsnotify.Create: // add to watcher, parse if file
+				// Wrap with an anonymous function to make sure that defer works.
+				func() {
+					if isDir {
+						watcher.Add(ev.Name)
+					} else {
+						f, err := limitopen.OpenWithLimit(ev.Name, softLimit, hardLimit)
+						if err != nil {
+							logger.Log(context.Background(), "directorywatcher: I/O error: "+err.Error())
```
kylelemons, line 137: you are passing context.Background, which should only ever be used in main and test functions.  Always propagate your context.

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+					// remove data related to file
+					data := r.data.Load()
+					data, err := remove(data, ev.Name)
+					if err != nil {
+						logger.Log(context.Background(), "directorywatcher: remove file error: "+err.Error())
+						return
+					}
+					r.data.Store(data)
+				}()
+			case fsnotify.Write, fsnotify.Chmod: //parse
```
kylelemons, line 167: spaces after `//`

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+
+		case ev := <-watcher.Events:
+
+			isDir, _ := isDirectory(ev.Name)
+
+			switch ev.Op {
+			default:
+				// do nothing.
+			case fsnotify.Create: // add to watcher, parse if file
+				// Wrap with an anonymous function to make sure that defer works.
```
kylelemons, line 130: If you need this, refactor it out into a method instead

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+				func() {
+					if isDir {
+						watcher.Add(ev.Name)
```
kylelemons, line 133: return and outdent the else block

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+func (r *Result) watcherLoop(
+	watcher *fsnotify.Watcher,
+	parser filewatcher.Parser,
+	add AddFile,
+	remove RemoveFile,
+	softLimit, hardLimit int64,
+	logger log.Wrapper,
+) {
+	for {
+		select {
```
kylelemons, line 114: In my opinion, trying to deal with the specific operations is too complex to be justified.

I would recommend instead using a debounce.

Whenever you get any notification, start a timer.  Say, 100ms, 500ms, or 1s depending on how fast you need to react.  If you get another notification before it fires, reset the timer.  When the timer fires, do a complete pass over all files and load them.

This way you can write one set of logic that handles the files the way they appear on the filesystem, you don't need to worry about delta-encoding and the _many_ issues inherent therein.

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+
+	var watcher *fsnotify.Watcher
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+
+	var d interface{}
+	res := &Result{}
+	// res.data.Store(Folder{
```
kylelemons, line 219: don't comment out code; either delete it, or `if false {}` it or something if you will need it in the future.

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+	dirPath := filepath.Clean(cfg.Path)
+
+	isDir, err := isDirectory(dirPath)
+	if !isDir {
+		return nil, fmt.Errorf("directorywatcher: %q is not a directory", dirPath)
+	} else if err != nil {
+		return nil, err
+	}
+
+	err = filepath.WalkDir(dirPath, func(path string, info fs.DirEntry, err error) error {
```
kylelemons, line 233: There are enough nested functions here that you probably want to factor at least some of this out into helper methods.  The amount of possible local state and closure capturing is really high, which makes it hard to read and reason about.

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,288 @@
...
+		if info.IsDir() {
+			return watcher.Add(path)
```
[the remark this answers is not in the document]
kylelemons: I'm pretty sure replacing a directory will trigger the inotify on that directlry, right?

You might need to rewatch the subdirectories though I am not sure.

directorywatcher/directorywatcher_test.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer data.Stop()
+
+	path1 := filepath.Join(dir, "foo")
+	if _, err = os.Create(path1); err != nil {
+		t.Fatal(err)
+	}
+	if err = os.WriteFile(path1, []byte("{\"number\":17}"), 0644); err != nil {
```
kylelemons, line 73: `if err =` is a red flag for me.  Any `=` in an if clause, actually.  You probably want `:=`.  I would look around for other instances of this too.

directorywatcher/directorywatcher_test.go
```diff
@@ -0,0 +1,147 @@
...
+		t.Fatal(err)
+	}
+
+	time.Sleep(1 * time.Second) // give the watcher time to parse
```
kylelemons, line 77: Create a synchronous way to detect that it has parsed, rather than using a sleep.  It can be unexported if you want.

For example, a channel that, if non-nil, has a boolean pushed onto it whenever parsing completes.  Now you can have a `select` on that channel along with a timeout, so you can say `parse failed to execute within 1 second` if it times out but don't need to wait the full second if it happens fast.

kylelemons (changes requested):

directorywatcher/directorywatcher.go
Tediferous: So from the perspective of the Vault CSI plugin, a files position in a directory would imply a difference in data since it mirrors how secrets are stored in vault. Each secret will be its own file.

I do agree with the assymetry of the add and remove functions being weird so I may take Kyle's advice and ditch the remove function entirely

directorywatcher/directorywatcher.go
Tediferous: I do like this but I still think extra logic for merging data between files will be needed somehow

directorywatcher/directorywatcher.go
```diff
@@ -0,0 +1,268 @@
...
+	"gopkg.in/fsnotify.v1"
+
+	"github.com/reddit/baseplate.go/filewatcher"
+	"github.com/reddit/baseplate.go/internal/limitopen"
+	"github.com/reddit/baseplate.go/log"
+)
+
+// AddFile is a type of function that should be ran to handle
+// adding data after a file has been parsed
+type AddFile func(d interface{}, file interface{}) (data interface{}, err error)
```
konradreiche, line 21: I have a hard time wrapping my head around this function without reading further down how it is being utilized, part of this is the constraint of the empty interface being unconcrete here but I also wonder if we can improve it with either different parameter names or more document.

Is `d` the existing data so far and file is the content of a parsed file that will be merged into the existing data `?` Maybe we can clarify this.

directorywatcher/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package directorywatcher provides a go implementation of baseplate's directorywatcher:
+// https://baseplate.readthedocs.io/en/stable/api/baseplate/lib/dir_watcher.html
```
konradreiche, line 2: I get a 404 for this link. It could be useful to document the purpose of the package here separately.

konradreiche (changes requested):
I would also love to see a consolidation with the filewatcher package too, especially coming from the Unix philosophy of treating directories as files.

filewatcher/filewatcher.go
```diff
@@ -308,6 +305,26 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
...
+	isDir, err := isDirectory(filepath.Clean(cfg.Path))
+	if err != nil {
+		return nil, err
+	}
+	if isDir {
+		// Need to walk recursively because the watcher
+		// doesnt support recursion by itself
+		dirPath := filepath.Clean(cfg.Path)
+
+		err = filepath.WalkDir(dirPath, func(path string, info fs.DirEntry, err error) error {
```
kylelemons, line 317: if err :=

secrets/secrets.go
```diff
@@ -298,3 +324,45 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+
+			for k, v := range parsedSecrets.Secrets {
+				inputSecrets.Secrets[k] = v
+			}
+		}
+		secretsDocument = inputSecrets
+	} else {
+		// parse file
+		var secretFile CSIFile
+		err = json.NewDecoder(file).Decode(&secretFile)
```
kylelemons, line 360: if err :=

filewatcher/filewatcher.go
```diff
@@ -308,6 +305,26 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
 	if err != nil {
 		return nil, err
 	}
+	isDir, err := isDirectory(filepath.Clean(cfg.Path))
```
kylelemons, line 308: What is the reason for the `Clean`s? Is this necessary, or is it to make errors better, or what?

secrets/secrets.go
```diff
@@ -298,3 +324,45 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+	}
+	file, err := os.Open(path)
+	if err != nil {
+		return secretsDocument, err
+	}
+	fileInfo, err := file.Stat()
+	if err != nil {
+		return secretsDocument, err
+	}
+	if fileInfo.IsDir() {
```
kylelemons, line 340: you have a helper for this right?

secrets/secrets.go
```diff
@@ -298,3 +324,45 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+		// parse file
+		var secretFile CSIFile
+		err = json.NewDecoder(file).Decode(&secretFile)
+		if err != nil {
+			return secretsDocument, err
+		}
+		secretsDocument.Secrets[path] = secretFile.Secret
+	}
+
+	return
```
kylelemons, line 367: don't use bare returns

secrets/secrets.go
```diff
@@ -298,3 +324,45 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+			parsedSecrets, err := csiPathParser(filepath.Join(path, f.Name()), inputSecrets)
+			if err != nil {
+				return secretsDocument, err
+			}
+
+			for k, v := range parsedSecrets.Secrets {
+				inputSecrets.Secrets[k] = v
+			}
+		}
+		secretsDocument = inputSecrets
```
kylelemons, line 356: return and outdent the else case

secrets/secrets.go
```diff
@@ -252,10 +262,26 @@ type Vault struct {
...
+				secretsDocument, err = csiPathParser(e.Path, secretsDocument)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				return nil, err
+			}
+		default:
+			return nil, err
+		}
```
kylelemons, line 284:
```suggestion
		}
		return nil, err
```

secrets/secrets.go
```diff
@@ -252,10 +262,26 @@ type Vault struct {
...
 func NewSecrets(r io.Reader) (*Secrets, error) {
-	var secretsDocument Document
+	secretsDocument := Document{
+		Secrets: make(map[string]GenericSecret),
+	}
 	err := json.NewDecoder(r).Decode(&secretsDocument)
 	if err != nil {
-		return nil, err
+		switch e := err.(type) {
+		case *fs.PathError:
```
kylelemons, line 271: probably just do an `if e, ok := err.(*fs.PathError)` instead of a type switch

secrets/secrets.go
```diff
@@ -252,10 +262,26 @@ type Vault struct {
...
+			// check if the path is a directory, then assume
+			// the secret provider is Vault CSI
+			if strings.Contains(e.Error(), "is a directory") {
```
kylelemons, line 274: why are you worrying about directory errors in an `io.Reader`-targeted API?

kylelemons (changes requested):

filewatcher/filewatcher.go
```diff
@@ -308,6 +305,26 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
...
+		// Need to walk recursively because the watcher
+		// doesnt support recursion by itself
+		dirPath := filepath.Clean(cfg.Path)
+
+		err = filepath.WalkDir(dirPath, func(path string, info fs.DirEntry, err error) error {
+			if info.IsDir() {
+				return watcher.Add(path)
+			}
+			return nil
+
```
fishy, line 322: nit: excessive blank line

secrets/secrets.go
```diff
@@ -4,6 +4,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/fs"
+	"io/ioutil"
```
fishy, line 8: `io/ioutil` is already deprecated, new code should use the replacements (usually from either `io` or `os`) instead.

secrets/secrets.go
```diff
@@ -298,3 +324,45 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+	file, err := os.Open(path)
+	if err != nil {
+		return secretsDocument, err
+	}
+	fileInfo, err := file.Stat()
+	if err != nil {
+		return secretsDocument, err
+	}
+	if fileInfo.IsDir() {
+		files, err := ioutil.ReadDir(path)
```
fishy, line 341: oh so here is where `ioutil` is used. for this you want to use [`os.ReadDir`](https://pkg.go.dev/os#ReadDir) instead.

secrets/secrets.go
fishy: that helper function is in another package and is unexported :) and we likely don't want to export it (e.g. having `IsDirectory` being part of `filewatcher` api doesn't really make sense)

but again, since `NewSecrets` takes an `io.Reader` as an arg, I'm not really sure how we could get the error to say it's actually a directory not a file (you can't get an `io.Reader` out of a directory can you?) and get into here.

filewatcher/filewatcher.go
```diff
@@ -308,6 +305,26 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
...
+		return nil, err
+	}
+	if isDir {
+		// Need to walk recursively because the watcher
+		// doesnt support recursion by itself
+		dirPath := filepath.Clean(cfg.Path)
+
+		err = filepath.WalkDir(dirPath, func(path string, info fs.DirEntry, err error) error {
+			if info.IsDir() {
+				return watcher.Add(path)
```
fishy, line 319: I'm not sure that this works as you intended it to be.

so between filewatcher and directory watcher, the actual parser api would be different. for filewatcher since it's just a file, you can just pass the `io.Reader` to the parser to let it do the parsing. but for directory watcher you can't really get an `io.Reader` out of a directory, so what you want to do is just notify the caller that "file(s) under this directory changed, you want to reload everything", with potentially the path to the directory, or maybe `[]fs.DirEntry` (this is what's returned by `os.ReadDir`), passed into the parser callback.

also I think kyle's earlier suggestion of wait for an interval without any notifications before doing that is important, otherwise we could be calling the directory callback in very quick succession.

fishy (changes requested):

konradreiche (changes requested):

filewatcher/filewatcher.go
Tediferous: Yeah no real reason other than that. If you feel its extraneous I can take them out

secrets/secrets.go
Tediferous: When using vault csi the secret path can be a directory so I have to check for this case and handle the directory differently. Ive moved this into its own helper

secrets/secrets.go
Tediferous: This specific dir check is used to determine which whether the files within the parent dir are files during the recursion. And like Yuxuan said, that helper I made is in the other package

filewatcher/filewatcher.go
Tediferous: So I need to check watch all of the underlying paths and act accordingly when events happen. And in this case the parser being used is `csiPathParser` so I do think it works similarly enough to the regular filewatcher.

How would you expect the timer to be implemented? I wanted to minimize changes to the behavior of the current watcherloop so I did it this way

secrets/secrets.go
```diff
@@ -298,3 +312,56 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+	secretsDocument.Secrets[path] = secretFile.Secret
+
+	return secretsDocument, err
+}
+
+func checkForVaultCSI(secretsDocument Document, err error) (Document, error) {
+	if e, ok := err.(*fs.PathError); ok {
+		// check if the path is a directory, then assume
+		// the secret provider is Vault CSI
+		if strings.Contains(e.Error(), "is a directory") {
```
fishy, line 362: what you want here is `errors.Is(err, syscall.EISDIR)`, example: https://go.dev/play/p/VkiHFhVeQhI

but I still think that using the error to pass the path to the directory is a very anti-pattern way to do that and we should have made the distinction long before getting the error.

this is basically the last resort if we do not want to change the parser api in anyway (for a directory we still just pass in an `io.Reader` to the parser). I think we should make a different api for directory parser instead. but if that turns out to be infeasible then this is... probably fine.

filewatcher/filewatcher.go
fishy: so if you look at the doc of [`time.AfterFunc`](https://pkg.go.dev/time#AfterFunc), it returns you a `*time.Timer` that you can `Reset` or `Stop`.

basically you want to store the `*time.Timer` returned by it in the `*Result` struct, along with a lock, so basically in `watcherLoop`, every time a file under a directory changes, you do something like this:

```go
if isDirectory(...) {
  // use a lambda to make sure `defer` works as expected. you can also just refactor this part into a function
  func() {
    r.dirTimerLock.Lock()
    defer r.dirTimerLock.Unlock()
    if r.dirTimer != nil {
      // this means this is not the first event in thischange, just reset the timer
      r.dirTimer.Reset(configuredDuration)
      return
    }
    // otherwise, this is the first event in this change and we need to create a new timer
    r.dirTimer = time.AfterFunc(configuredDuration, func() {
      // this is the function fired to call the parser and remove the stored timer
      r.dirTimerLock.Lock()
      r.dirTimerLock = nil
      r.dirTimerLock.Unlock()
      dirParser(...)
    })
  }()
}
```

basically, for every fsnotify event, we push the timer on `configuredDuration`, and if no new fsnotify event happens during this duration we actually call the parser to reload everything.

fishy (changes requested):

kylelemons (changes requested):

filewatcher/filewatcher.go
```diff
@@ -379,3 +393,12 @@ func (fw *MockFileWatcher) Get() interface{} {
...
 
 var _ FileWatcher = (*MockFileWatcher)(nil)
+
+func isDirectory(path string) (bool, error) {
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return false, err
+	}
+
+	return fileInfo.IsDir(), err
```
konradreiche, line 403:
```suggestion
	return fileInfo.IsDir(), nil
```
Return nil error explicitly here.

secrets/secrets.go
konradreiche: +1 on using `errors.Is` API. Forcing an error to handle this does not sound great. If we look here the path is being passed in before delegating it to the parser method (which is not exported): https://github.com/reddit/baseplate.go/blob/master/secrets/store.go#L48-L62

Couldn't we extend the unexported `parse` method here and add a new `newSecrets` constructor which also accepts a path?

konradreiche (changes requested):

filewatcher/filewatcher.go
```diff
@@ -308,6 +305,26 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
...
+		// Need to walk recursively because the watcher
+		// doesnt support recursion by itself
+		err := filepath.WalkDir(cfg.Path, func(path string, info fs.DirEntry, err error) error {
+			if info.IsDir() {
+				return watcher.Add(path)
+			}
+			return nil
+		})
+		if err != nil {
+			return nil, err
```
kylelemons, line 325: add context to the error

filewatcher/filewatcher.go
```diff
@@ -379,3 +396,19 @@ func (fw *MockFileWatcher) Get() interface{} {
 func (fw *MockFileWatcher) Stop() {}
 
 var _ FileWatcher = (*MockFileWatcher)(nil)
+
+// DummyReadCloser is a mock struct used to hold the path for directory watchers
+type DummyReadCloser struct {
```
kylelemons, line 401: don't export this

filewatcher/filewatcher.go
```diff
@@ -308,6 +305,26 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
 	if err != nil {
 		return nil, err
 	}
+	isDir, err := isDirectory(cfg.Path)
+	if err != nil {
+		return nil, err
+	}
+	if isDir {
+		f = DummyReadCloser{
+			Path: cfg.Path,
```
kylelemons, line 314: if Read or Close gets called on this, it will panic... so this doesn't feel right to me.  What is this trying to do?  `f` doesn't seem to be used after this

kylelemons (changes requested):

secrets/store.go
```diff
@@ -70,6 +80,18 @@ func (s *Store) parser(r io.Reader) (interface{}, error) {
 	return secrets, nil
 }
 
+func (s *Store) dirParser(r io.Reader) (interface{}, error) {
+	drc := r.(filewatcher.DummyReadCloser)
```
konradreiche, line 84:
```suggestion
	drc, ok := r.(filewatcher.DummyReadCloser)
	if !ok {
	  return nil, fmt.Errorf("secrets.Store.dirParser: invalid type: %T", r)
	}
```

I think we should check if the type assertion was successful. We might be able to guarantee the type but if the code ever changes it could change into a panic.

filewatcher/filewatcher.go
konradreiche: If I understand correctly, this dummy read closer was added, so that we can pass through the path using an `io.Reader` interface due to the limitation of the existing API.

`f` is being used down here https://github.com/reddit/baseplate.go/pull/472/files#diff-eb266ac0bad4d099e00f9c659d61679d726029ef44c28bb064647bf4e9d1dc7dR330

If I read this correctly, the watcher loop will then call it and pass in the dummy again where it is being type asserted: https://github.com/reddit/baseplate.go/pull/472/files#diff-c6bb5cecf85ab156298ad011c2a0fd07ecb8bc37ec2f3dc06923703b44f6cbd2R83

konradreiche (changes requested):
Can we also add a test to cover some of these new code paths, especially because of the different types that we add to make this work :slightly_smiling_face:

secrets/store.go
Tediferous: good call, ty

filewatcher/filewatcher.go
Tediferous: Yes thats what its for. Because you cant make io.Readers from directories I needed to have this extra object to support it. The entire filewatcher is built around io.readers and Im not sure how to support directories without a rework of the filewatcher or a new class altogether.

filewatcher/filewatcher.go
Tediferous: I export this so I can use it to type assert in secrets/store.go

secrets/secrets.go
Tediferous: ive dropped this pattern of error checking and handle dir paths before trying to read

filewatcher/filewatcher.go
```diff
@@ -64,6 +66,8 @@ type Result struct {
 
 	ctx    context.Context
 	cancel context.CancelFunc
+	lock   sync.Mutex
```
fishy: style nit: group (by adding blank lines) lock and the fields it guards together.

filewatcher/filewatcher.go
```diff
@@ -104,20 +108,47 @@ func (r *Result) watcherLoop(
 	pollingInterval time.Duration,
 ) {
 	forceReload := func(mtime time.Time) {
+		isDir, err := isDirectory(path)
+		if err != nil {
+			logger.Log(context.Background(), "filewatcher: isDirectory error: "+err.Error())
+			return
+		}
 		f, err := limitopen.OpenWithLimit(path, softLimit, hardLimit)
```
fishy, line 113: shouldn't you check `isDir` before doing this, and only do this if it's not a directory?

filewatcher/filewatcher.go
```diff
@@ -379,3 +426,19 @@ func (fw *MockFileWatcher) Get() interface{} {
 func (fw *MockFileWatcher) Stop() {}
 
 var _ FileWatcher = (*MockFileWatcher)(nil)
+
+// DummyReadCloser is a mock struct used to hold the path for directory watchers
+type DummyReadCloser struct {
+	io.Reader
+	io.Closer
```
fishy, line 413: [the `Parser` interface only requires `io.Reader`](https://pkg.go.dev/github.com/reddit/baseplate.go/filewatcher#Parser), so there's no need to add dummy closer here.

Also please add blank lines between embedded and non-embedded fields.

filewatcher/filewatcher_test.go
```diff
@@ -20,6 +21,27 @@ func parser(f io.Reader) (interface{}, error) {
 	return io.ReadAll(f)
 }
 
+func dirParser(f io.Reader) (interface{}, error) {
+	data := []byte{}
```
fishy, line 25: style nit: prefer `var data []byte`

filewatcher/filewatcher_test.go
```diff
@@ -20,6 +21,27 @@ func parser(f io.Reader) (interface{}, error) {
...
+	drc := f.(filewatcher.DummyReadCloser)
+	err := filepath.Walk(drc.Path,
```
fishy, line 27: style nit: one-per-line: https://github.com/reddit/baseplate.go/blob/master/Style.md#one-per-line-style

filewatcher/filewatcher_test.go
```diff
@@ -20,6 +21,27 @@ func parser(f io.Reader) (interface{}, error) {
...
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.Mode().IsDir() {
+				return nil
+			}
+			// parse file
+			data, err = ioutil.ReadFile(path)
```
fishy, line 36: so you are overriding `data` with every file walked, in the end you are returning the `data` from the last file walked and throw away all the previous reads. that certainly doesn't sound right to me.

fishy: also `ioutil` is already deprecated. use `os.ReadFile` instead.

secrets/secrets.go
```diff
@@ -298,3 +320,31 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
 	}
 	return secrets, nil
 }
+
+func csiPathParser(path string) (Document, error) {
+	secretsDocument := Document{
+		Secrets: make(map[string]GenericSecret),
+	}
+	err := filepath.Walk(path,
```
fishy, line 328: same one-per-line comment here.

fishy (changes requested):

filewatcher/filewatcher.go
fishy: I think what we need here is to:
1. Define a type, `DirParser`, for use cases with a directory
2. Provide a wrapper to wrap `DirParser` into `Parser` so they can be used in the current API

and secrets should use their own `DirParser`, which should be roughly the same as the callback you are currently using in `Walk`.

fishy (changes requested):

filewatcher/filewatcher.go
Tediferous: that is true about the parser but the limitopen module returns a ReadCloser, so many parts of the code treat it as such, take the watcherloop for example.

I will add the space for the embedded field.

filewatcher/filewatcher_test.go
Tediferous: Yes that is what Im doing here, but during tests only one file is written so the data turns out to be the same. In the tests I show that the data returned is the same wether you watch the file or the dir that holds it.

Also, I will remove ioutil thanks for keeping me honest

filewatcher/filewatcher.go
kylelemons: What is not OK is leaving this around to panic.  You should _implement_ Read and Close.  They should return an error that says "this operation is not supported on a directory" rather than panic'ing.

(but see the comment below about an alternate API proposal)

kylelemons (changes requested):

filewatcher/filewatcher.go
fishy: actually the `DirParser` api should be something like this:

```go
type DirParer func(path string) (data interface{}, err error)
```

then we provide another api to convert that into a `Parser`:

```go
func DirParserToParser(dp DirParser) Parser {
  return func(f io.Reader) (interface{}, error) {
    return dp(f.(dummyReader).Path)
  }
}
```

and we no longer need to export `DummyReaderCloser`. `Walk` should be implemented inside `DirParser` instead.

filewatcher/filewatcher.go
konradreiche: +1 on that :slightly_smiling_face:

konradreiche (changes requested):

filewatcher/filewatcher.go
Tediferous: added that check

filewatcher/filewatcher.go
Tediferous: Im on it

filewatcher/filewatcher.go
Tediferous: ah nice, thank you

kylelemons: I'm still interested in seeing this approach considered, as I still consider mis-implementing an interface to be a loaded footgun we don't want to saddle ourselves with

kylelemons (changes requested):

filewatcher/filewatcher.go
```diff
@@ -20,6 +22,10 @@ import (
 // new file watcher.
 const DefaultPollingInterval = 30 * time.Second
 
+// DefaultParseDelay is the default time needed without an event to parse again. Used in
+// directory watchers when many events can happen at once
```
fishy, line 26: nit: s/at once/in a short time window/

filewatcher/filewatcher.go
```diff
@@ -58,12 +64,28 @@ const (
 // Inconsistent type will cause panic, as does returning nil data and nil error.
 type Parser func(f io.Reader) (data interface{}, err error)
 
+// DirParser is a callback function to be called when a file in a watched directory
```
fishy, line 67: nit: s/a file/file(s)/

filewatcher/filewatcher.go
```diff
@@ -102,22 +124,50 @@ func (r *Result) watcherLoop(
 	softLimit, hardLimit int64,
 	logger log.Wrapper,
 	pollingInterval time.Duration,
+	parseDelay time.Duration,
 ) {
 	forceReload := func(mtime time.Time) {
-		f, err := limitopen.OpenWithLimit(path, softLimit, hardLimit)
+		var f io.ReadCloser
```
fishy, line 130: you can change this part to make `dummyReadCloser` no longer need to implement closer as that's not really used:
```go
var reader io.Reader
...
if isDir {
  reader = dummyReader{
    Path: path, // actually Path field no longer need to be exported either.
  }
} else {
  f, err := limitopen.OpenWithLimit(...)
  if err != nil {
    ...
  }
  defer f.Close()
  reader = f
}
// now you can use reader instead of f
```

filewatcher/filewatcher_test.go
```diff
@@ -20,6 +20,27 @@ func parser(f io.Reader) (interface{}, error) {
...
+			if info.Mode().IsDir() {
+				return nil
+			}
+			// parse file
+			data, err = os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+			return nil
+		})
```
fishy, line 40: style nit: break `}` and `)` into 2 lines.

secrets/secrets.go
```diff
@@ -298,3 +320,32 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+				return err
+			}
+			var secretFile CSIFile
+			err = json.NewDecoder(file).Decode(&secretFile)
+			if err != nil {
+				return err
+			}
+			secretsDocument.Secrets[path] = secretFile.Secret
+			return nil
+		})
```
fishy, line 349: same style nit here

secrets/secrets.go
```diff
@@ -298,3 +320,32 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+			file, err := os.Open(path)
+			if err != nil {
+				return err
+			}
+			var secretFile CSIFile
+			err = json.NewDecoder(file).Decode(&secretFile)
+			if err != nil {
+				return err
+			}
+			secretsDocument.Secrets[path] = secretFile.Secret
```
fishy, line 347: 🔕 I wonder if we want to guard against the case that the same `path` appearing in different files. e.g. check that `path` is not already in the map and return an error if that's the case:

```go
if _, ok := secretsDocument.Secrets[path]; ok {
  return someError
}
secretsDocument.Secrets[path] = secretFile.Secret
```

filewatcher/filewatcher.go
```diff
@@ -58,12 +64,28 @@ const (
...
 type Parser func(f io.Reader) (data interface{}, err error)
 
+// DirParser is a callback function to be called when a file in a watched directory
+// has been touched, or read for the first time.
+// Should always return a constant type or will cause panic
+type DirParser func(path string) (data interface{}, err error)
+
+// DirParserToParser wraps a DirParser so that it may be used as a Parser in a
+// filewatcher
+func DirParserToParser(dp DirParser) Parser {
```
fishy, line 74: I'm bad at names and I was just using `DirParserToParser` as a placeholder in my previous comment, not that it's a good name 😅 . maybe `DirParserWrapper` or something is better?

fishy (approved):
a few nits left but I'm happy with the current state.

filewatcher/filewatcher.go
```diff
@@ -379,3 +455,23 @@ func (fw *MockFileWatcher) Get() interface{} {
 func (fw *MockFileWatcher) Stop() {}
 
 var _ FileWatcher = (*MockFileWatcher)(nil)
+
+// dummyReader is a mock struct used to hold the path for directory watchers
+type dummyReader struct {
+	io.Reader
```
fishy, line 461: nit: now that you implemented the `Read` function, this embedding is no longer needed.

filewatcher/filewatcher.go
```diff
@@ -379,3 +455,23 @@ func (fw *MockFileWatcher) Get() interface{} {
...
+
+	Path string
```
fishy, line 463: nit: this field also no longer needs to be exported.

filewatcher/filewatcher.go
```diff
@@ -308,9 +360,29 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
...
+			Path: cfg.Path,
+		}
+		// Need to walk recursively because the watcher
+		// doesnt support recursion by itself
+		err := filepath.WalkDir(cfg.Path, func(path string, info fs.DirEntry, err error) error {
+			if info.IsDir() {
+				return watcher.Add(path)
+			}
+			return nil
+		})
```
fishy, line 378: I just remembered, there's still a corner case we haven't handled.

basically we only add the directory and all its subdirectories to watch when filewatcher itself is created, and the watch list is never updated later. here's a scenario:

1. initially filewatcher is created for `/foo/`, which contains `/foo/bar/` and `/foo/baz/` and other files
2. later, a new subdirectory, `/foo/bar3/` was added, we do not add it to the watch list, so a new file `/foo/bar3/file` being created will not trigger anything
3. `/foo/bar/` or `/foo/baz/` can also be removed later (but that likely won't cause issues in the current implementation)

I think what we need to do in `parse` callback is that if it's a directory, remove all subdirectories we added before (we likely need to record that info in the struct), and re-add all subdirectories again.

fishy (changes requested):

secrets/secrets.go
Tediferous: The Vault CSI provider should ensure that two files refer to different secrets.  It creates one file per secret path. So in this example the secrets
`secret/creds/user` and `secret/creds/pass`
should appear in a directory like
```
secret
└── creds
    ├── user
    └── pass
```

Duplicate paths shouldnt show up

filewatcher/filewatcher.go
konradreiche: Is this expected to happen for Vault CSI or to cover the case of directory watching in general?

konradreiche (approved):
No concerns on my end, though curious about the edge case Yuxuan mentioned.

filewatcher/filewatcher.go
Tediferous: good call out
Ive added and extra block in the watcherloop to add any new sub directotories to the watcher. Ive also added a test for this case

filewatcher/filewatcher.go
```diff
@@ -161,14 +212,28 @@ func (r *Result) watcherLoop(
...
+				}
+				if isDir {
+					eventIsDir, err := isDirectory(ev.Name)
+					if err != nil {
+						logger.Log(context.Background(), "filewatcher: isDirectory error: "+err.Error())
+						return
+					}
+					if eventIsDir && ev.Op == fsnotify.Create {
+						watcher.Add(ev.Name)
+					}
```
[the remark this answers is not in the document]
fishy: actually I think we need to watch all files (not only directories).

filewatcher/filewatcher.go
```diff
@@ -97,27 +119,57 @@ func getMtime(path string) (time.Time, error) {
 
 func (r *Result) watcherLoop(
 	watcher *fsnotify.Watcher,
+	watched map[string]bool,
```
fishy, line 122: nit: since the list is generated by `filepath.WalkDir`, there shouldn't be any concerns of duplications, so just a `[]string` should be enough.

filewatcher/filewatcher.go
```diff
@@ -161,14 +227,21 @@ func (r *Result) watcherLoop(
...
+				// If a directory is created in a directory watcher we want
+				// to watch that as well
+				isDir, err := isDirectory(path)
+				if err != nil {
+					logger.Log(context.Background(), "filewatcher: isDirectory error: "+err.Error())
+					continue
+				}
+				if isDir {
+					refreshWatched()
+				}
```
fishy, line 243: this should be moved into `parse` instead, before calling the parser callback (so we only refresh it once per parser call, not for every fsnotify event)

filewatcher/filewatcher.go
```diff
@@ -135,14 +187,28 @@ func (r *Result) watcherLoop(
...
 		}
 	}
+	refreshWatched := func() {
+		for path := range watched {
+			watcher.Remove(path)
+			delete(watched, path)
+		}
+		err := filepath.WalkDir(path, func(p string, info fs.DirEntry, err error) error {
+			watched[p] = true
+			return watcher.Add(p)
```
fishy, line 197: we should ignore the return of `watcher.Add` and always return `true` to the walkdir callback. the reason is for whatever reason fsnotify failed to watch a particular file/dir, that should not stop us from adding the rest of files/dirs to fsnotify.

fishy: also we need to always add the parent directory of `path` additionally (likely before the `filepath.WalkDir` call), like we did in the initial parse.

filewatcher/filewatcher.go
```diff
@@ -161,14 +227,21 @@ func (r *Result) watcherLoop(
...
 			}
 
-			if filepath.Base(ev.Name) != file {
-				continue
-			}
-
 			switch ev.Op {
 			default:
 				// Ignore uninterested events.
 			case fsnotify.Create, fsnotify.Write:
```
fishy, line 233: with the add of directory support, I think we also need to add `fsnotify.Remove` to this `case`.

kylelemons (changes requested):
As this gets more and more complex, I have to wonder... is it worth pushing for this in v0 when we have CSI support natively in v2?  In the v2 world, it's also watching on a per-file level with a TTL fallback, so it is much easier to deal with a lot of these complexities and corner-cases.

Given the difficulty of testing / validating this behavior, and the widespread adoption of baseplate.go v0, I'm thinking we may want to hold off on this.

fishy (changes requested):
now that [0b74864](https://github.com/reddit/baseplate.go/commit/0b748640e206524bb61ddcc8e75ad2972c50d710) is merged, please rebase this PR on top of latest master and reimplement the secrets part of the change with the new DirParser API provided.

Tediferous:
closing in favor of https://github.com/reddit/baseplate.go/pull/565

PR #473, opened 2022-01-24
metricsbp: Default to use blackhole sink without Endpoint

fishy:
Add BufferInMemoryForTesting to metricsbp.Config, with yaml tag
disabled, to be used by test code that reads from the metricsbp.Statsd
buffer explicitly.

### DIFF
--- a/log/wrapper_example_test.go
+++ b/log/wrapper_example_test.go
@@ -18,6 +18,19 @@ type ExtendedLogWrapper struct {
 	log.Wrapper
 }
 
+var st = metricsbp.NewStatsd(
+	context.Background(),
+	metricsbp.Config{
+		// This is to make sure that when we read the metricsbp statsd buffer out
+		// manually later in the example they are not already sent to a blackhole
+		// sink.
+		// This is NOT needed in production code as in production code the sink
+		// shall be configured as the actual statsd collector and not read out
+		// manually like in this example.
+		BufferInMemoryForTesting: true,
+	},
+)
+
 // UnmarshalText implements encoding.TextUnmarshaler.
 //
 // In addition to the implementations log.Wrapper.UnmarshalText supports, it
@@ -39,6 +52,7 @@ func (e *ExtendedLogWrapper) UnmarshalText(text []byte) error {
 		e.Wrapper = metricsbp.LogWrapper(metricsbp.LogWrapperArgs{
 			Counter: parts[1],
 			Wrapper: w,
+			Statsd:  st,
 		})
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


konradreiche (approved):

log/wrapper_example_test.go
```diff
@@ -18,6 +18,19 @@ type ExtendedLogWrapper struct {
...
+var st = metricsbp.NewStatsd(
+	context.Background(),
+	metricsbp.Config{
+		// This is to make sure that when we read the metricsbp statsd buffer out
+		// manually later in the example they are not already sent to a blackhole
+		// sink.
+		// This is NOT needed in production code as in production code the sink
+		// shall be configured as the actual statsd collector and not read out
+		// manually like in this example.
+		NoBlackhole: true,
```
kylelemons, line 30: what about `ForceInMemoryBuffer` or `BufferInMemoryForTesting`?

Can this be an unexported struct field?  (I assume it's exported since you think other packages need this too)

metricsbp/statsd.go
```diff
@@ -124,14 +98,17 @@ func NewStatsd(ctx context.Context, cfg Config) *Statsd {
 	}
 	st.ctx, st.cancel = context.WithCancel(ctx)
 
+	var sink io.Writer
```
kylelemons, line 101: you can do

```
sink := io.Discard
if cfg.Endpoint != "" {
  sink = ...
}
```

kylelemons (changes requested):

log/wrapper_example_test.go
fishy: This must be exported, for example the line you commented on is outside of the original package.

Will rename it to `BufferInMemoryForTesting`.

metricsbp/statsd.go
fishy: this is 3-state, not 2-state. I can do that but I would still need:

```go
sink := io.Discard
if cfg.Endpoint != "" {
  sink = ...
} else if cfg.NoBlackhole {
  sink = nil
}
```
otherwise the `if` on line 107 will stop working.

metricsbp/statsd.go
```diff
@@ -354,6 +331,9 @@ func (st *Statsd) Close() error {
 // so in most cases you shouldn't be using it in production code.
 // But it's useful in unit tests to verify that you have the correct metrics you
 // want to report.
+//
+// When you use this in test code you also want to set BufferInMemoryForTesting
+// to true in the statsd Config, otherwise your test could become flaky.
```
kylelemons, line 336: Wouldn't it just fail? since it's dropping data, why would it sometimes work and sometimes fail?

kylelemons (approved):

metricsbp/statsd.go
fishy: the dropping of the data happens on a ticker. it's possible/likely when the test runs the ticker hasn't fired yet.

PR #474, opened 2022-01-26
Adjust default client middleware order to handle retries

fishy:
For thriftbp, we used to use two MonitorClient middlewares, one before
and one after Retry, to record both raw and -with-retry metrics, so do
the same thing for PrometheusClientMiddleware.

For httpbp, move PrometheusClientMetrics to before Retries so it records
the metrics with retries, same as how MonitorClient middleware is doing
today.

### DIFF
--- a/httpbp/client_middlewares.go
+++ b/httpbp/client_middlewares.go
@@ -16,6 +16,7 @@ import (
 	"github.com/reddit/baseplate.go/breakerbp"
 	"github.com/reddit/baseplate.go/retrybp"
 	"github.com/reddit/baseplate.go/tracing"
+	"github.com/reddit/baseplate.go/transport"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


kylelemons (changes requested):
hmm.  Are we sure we want this? This means that I can't tell how many _actual_ requests are being sent (QPS wise)

fishy:
> hmm. Are we sure we want this? This means that I can't tell how many _actual_ requests are being sent (QPS wise)

For thrift we report both with different slug tags (slug and slug-with-retry), with the one without retry is the raw QPS.

For http, I'm just matching the current behavior of `MonitorClient` in the default middlewares. I'm not sure what was the reason to only report with-retry metrics but not the raw ones. cc @konradreiche who originally added the default http client middlewares.

JessicaGreben (approved):
I have a slight preference of httpbp and thriftbp Prom Middleware both tracking before and after retries. But if this PR matches past behavior, then I can be convinced this is fine the way it is.

konradreiche (approved):
This looks correct to me, before the addition of Prometheus client metrics retries was executed last too.

fishy:
:haircut: Added both -with-retry raw metrics for httpbp clients.

cc @konradreiche this will change the behavior of your dashboards, basically what client side metrics you get now for `client="foo"`, you need to use `client="foo-with-retry"` instead to get the same metrics (the `client="foo"` metrics are now raw metrics).

konradreiche (approved):

kylelemons (approved):

PR #477, opened 2022-02-02
Fix connect leak from clientpool

fishy:
This seems to be triggered by envoy closing the connections differently
from thrift servers.

### DIFF
--- a/clientpool/channel.go
+++ b/clientpool/channel.go
@@ -60,6 +60,14 @@ func (cp *channelPool) Get() (client Client, err error) {
 		if c.IsOpen() {
 			return c, nil
 		}
+		// For thrift connections, IsOpen could return false in both explicit and
+		// implicit closed situations.
+		// In implicit closed situation, IsOpen does a connectivity check and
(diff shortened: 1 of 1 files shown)


kylelemons (approved):

konradreiche (approved):

fishy:
Alternatively we can also do it [here](https://github.com/reddit/baseplate.go/blob/49770a8f3a30516f816825a2e8b07d5a6afc18d7/clientpool/channel.go#L62), @kylelemons & @konradreiche do you have a strong preference either way?

alienth (approved):

fishy:
:haircut: moved the fix to clientpool so it's more foolproof (for example, in the future we no longer wrap clients with ttlClient).

konradreiche (approved):

kylelemons (approved):

fishy:
💇 added the fix to another place, which is the actual one triggered by envoy.

konradreiche (approved):

PR #478, opened 2022-02-04
Add prometheus metrics to all the existing statsd metrics

fishy:
Except the ones in example code, and the one emitted from the
experimental thrift server config.

### DIFF
--- a/breakerbp/failure_ratio.go
+++ b/breakerbp/failure_ratio.go
@@ -6,10 +6,29 @@ import (
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/sony/gobreaker"
 
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/metricsbp"
+)
+
+const (
+	promNamespace = "breakerbp"
+	nameLabel     = "breaker"
+)
+
+var (
+	breakerLabels = []string{
+		nameLabel,
+	}
+
+	breakerClosed = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: promNamespace,
+		Name:      "closed",
+		Help:      "0 means the breaker is currently tripped, 1 otherwise (closed)",
+	}, breakerLabels)
 )
 
 // FailureRatioBreaker is a circuit breaker based on gobreaker that uses a low-water-mark and
@@ -36,8 +55,10 @@ type Config struct {
 	Name string `yaml:"name"`
 
 	// EmitStatusMetrics sets whether the failure breaker will regularly update a gauge on the breakers state (closed or open/halfopen).
-	// When enabled, it emits metrics using the interval defined by metricsbp.SysStatsTickerInterval.
-	EmitStatusMetrics bool `yaml:"emitStatusMetrics"`
+	// When enabled, it emits metrics using the interval of EmitStatusMetricsInterval.
+	// If EmitStatusMetricsInterval <=0, metricsbp.SysStatsTickerInterval will be used as the fallback.
+	EmitStatusMetrics         bool          `yaml:"emitStatusMetrics"`
+	EmitStatusMetricsInterval time.Duration `yaml:"emitStatusMetricsInterval"`
 
 	// Logger is the logger to be called when the breaker changes states.
 	Logger log.Wrapper `yaml:"logger"`
@@ -75,15 +96,18 @@ func NewFailureRatioBreaker(config Config) FailureRatioBreaker {
 
 	failureBreaker.goBreaker = gobreaker.NewCircuitBreaker(settings)
 	if config.EmitStatusMetrics {
-		go failureBreaker.runStatsProducer()
+		go failureBreaker.runStatsProducer(config.EmitStatusMetricsInterval)
 	}
 	return failureBreaker
 }
 
-func (cb FailureRatioBreaker) runStatsProducer() {
+func (cb FailureRatioBreaker) runStatsProducer(interval time.Duration) {
+	if interval <= 0 {
+		interval = metricsbp.SysStatsTickerInterval
+	}
 	circuitBreakerGauge := metricsbp.M.RuntimeGauge(cb.name + "-circuit-breaker-closed")
 
-	tick := time.NewTicker(metricsbp.SysStatsTickerInterval)
+	tick := time.NewTicker(interval)
 	defer tick.Stop()
 	for {
 		select {
@@ -137,6 +161,14 @@ func (cb FailureRatioBreaker) shouldTrip(counts gobreaker.Counts) bool {
 }
 
 func (cb FailureRatioBreaker) stateChanged(name string, from gobreaker.State, to gobreaker.State) {
+	var value float64
+	if to != gobreaker.StateOpen {
+		value = 1
+	}
+	breakerClosed.With(prometheus.Labels{
+		nameLabel: cb.name,
+	}).Set(value)
+
 	message := fmt.Sprintf("circuit breaker %v state changed from %v to %v", name, from, to)
 	cb.logger.Log(context.Background(), message)
 }
--- a/httpbp/prometheus.go
+++ b/httpbp/prometheus.go
@@ -107,6 +107,25 @@ var (
 	}, clientActiveRequestsLabels)
 )
 
+const (
+	// Note that this is not used by prometheus metrics defined in Baseplate spec.
+	promNamespace   = "httpbp"
+	subsystemServer = "server"
+)
+
+var (
+	panicRecoverLabels = []string{
+		methodLabel,
+	}
+
+	panicRecoverCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: promNamespace,
+		Subsystem: subsystemServer,
+		Name:      "panic_recover_total",
+		Help:      "The number of panics recovered from http server handlers",
+	}, panicRecoverLabels)
+)
+
 // PerformanceMonitoringMiddleware returns optional Prometheus historgram metrics for monitoring the following:
 //    1) http server time to write header in seconds
 //    2) http server time to write header in seconds
--- a/internal/limitopen/limitopen.go
+++ b/internal/limitopen/limitopen.go
@@ -9,8 +9,28 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/metricsbp"
+)
+
+const (
+	promNamespace = "limitopen"
+	pathLabel     = "path"
+)
+
+var (
+	sizeLabels = []string{
+		pathLabel,
+	}
+
+	sizeGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: promNamespace,
+		Name:      "file_size_bytes",
+		Help:      "The size of the file opened by limitopen.Open",
+	}, sizeLabels)
 )
 
 // Open opens a path for read.
@@ -63,7 +83,8 @@ type readCloser struct {
 // OpenWithLimit calls Open with limit checks.
 //
 // It always reports the size of the path as a runtime gauge
-// (with "limitopen.size" as the metrics path and path label).
+// (with "limitopen.size" as the metrics path and path label for statsd,
+// "limitopen_file_size_bytes" for prometheus).
 // When softLimit > 0 and the size of the path as reported by the os is larger,
 // it will also use log.ErrorWithSentry to report it.
 // When hardLimit > 0 and the size of the path as reported by the os is larger,
@@ -74,9 +95,13 @@ func OpenWithLimit(path string, softLimit, hardLimit int64) (io.ReadCloser, erro
 		return nil, err
 	}
 
+	pathValue := filepath.Base(path)
 	metricsbp.M.RuntimeGauge("limitopen.size").With(
-		"path", filepath.Base(path),
+		"path", pathValue,
 	).Set(float64(size))
+	sizeGauge.With(prometheus.Labels{
+		pathLabel: pathValue,
+	}).Set(float64(size))
 
 	if softLimit > 0 && size > softLimit {
 		const msg = "limitopen.OpenWithLimit: file size > soft limit"
--- a/kafkabp/consumer.go
+++ b/kafkabp/consumer.go
@@ -7,9 +7,31 @@ import (
 	"sync/atomic"
 
 	"github.com/Shopify/sarama"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/reddit/baseplate.go/metricsbp"
 	"github.com/reddit/baseplate.go/tracing"
+)
+
+const (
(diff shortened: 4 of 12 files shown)


breakerbp/failure_ratio.go
```diff
@@ -6,12 +6,29 @@ import (
...
+	nameLabel = "name"
+)
+
+var (
+	circuitBreakerLabels = []string{
+		nameLabel,
+	}
+
+	circuitBreakerClosed = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "circuit_breaker_closed",
```
kylelemons, line 27: wdyt about

```suggestion
		Name: "breakerbp_breaker_tripped",
```

the package name will avoid conflicts if/when we migrate to a different breaker package

https://prometheus.io/docs/practices/naming/

kafkabp/consumer.go
```diff
@@ -7,11 +7,28 @@ import (
...
+	successLabel = "success"
+)
+
+var (
+	rebalanceLabels = []string{
+		successLabel,
+	}
+
+	rebalanceCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "kafka_consumer_rebalance_total",
```
kylelemons, line 27: kafkabp?

redis/db/redisbp/monitored_client.go
```diff
@@ -7,9 +7,52 @@ import (
...
+	stateIdle  = "idle"
+	stateStale = "stale"
+)
+
+var (
+	gaugeLabels = []string{
+		nameLabel,
+	}
+
+	hitsGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
kylelemons, line 30: these seem like all the same metric

```
redisbp_pool_gets_total{name="$name",result="hit|miss|timeout"}
```

This lets you get an overall QPS while still breaking it down

https://prometheus.io/docs/practices/naming/#labels

redis/db/redisbp/monitored_client.go
```diff
@@ -7,9 +7,52 @@ import (
...
+	}, gaugeLabels)
+
+	connectionGaugeLabels = []string{
+		nameLabel,
+		stateLabel,
+	}
+
+	connectionsGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "redisbp_pool_connections",
+		Help: "Number of connections from this redisbp pool",
```
kylelemons, line 52: number of connections currently in the connection pool?

thriftbp/prometheus.go
```diff
@@ -88,3 +88,93 @@ var (
...
+		protoLabel,
+	}
+
+	// 8 bytes to 8 megabytes
+	// some endpoints can have very small, almost zero payloads (for example,
+	// is_healthy), but we do have some endpoints with very large payloads
+	// (up to ~400KB).
+	payloadSizeBuckets = prometheus.ExponentialBuckets(8, 2, 20)
+
+	payloadSizeRequestHisto = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
kylelemons, line 127: `s/Histo/Bytes/`

It'll be clear in context that it is a histogram since you add a sample

thriftbp/prometheus.go
```diff
@@ -88,3 +88,93 @@ var (
 		Help: "The number of in-flight requests",
 	}, clientActiveRequestsLabels)
 )
+
+var (
+	serverConnectionsGauge = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "thrift_server_connections",
```
kylelemons, line 94: thriftbp?

thriftbp/prometheus.go
```diff
@@ -88,3 +88,93 @@ var (
...
+
+var (
+	ttlClientReplaceLabels = []string{
+		serverSlugLabel,
+		successLabel,
+	}
+
+	ttlClientReplaceCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "thrift_ttl_client_connection_houskeeping_total",
+		Help: "Total connection housekeeping done in thrift ttlClient",
```
kylelemons, line 107: "successLabel" doesn't make sense to me in this context; what is a successful vs a failed "housekeeping"?

kylelemons (changes requested):
nice

thriftbp/prometheus.go
fishy: I try to keep consistent with other `thrift_server_` metrics, I think using `thriftbp_` here might cause confusions.

But if you think `thrift_server_` is only for things defined in baseplate spec and all other metrics should be namespaced by the package then that's fine by me as well.

thriftbp/prometheus.go
fishy: whether we successfully replaced the connection in the background.

redis/db/redisbp/monitored_client.go
fishy: I'm not 100% sure that's true semantically, as it does have breakdown of total/idle/stale.

thriftbp/prometheus.go
kylelemons: My thought is that `thriftbp` is for things that are specific to this thrift "framework" but might change if we make a different framework for thrift servers.  The `thrift_` ones must remain consistent across such changes, but that's a bit "easier" since there's a spec saying what they should be.

I don't really have a strong opinion though.

thriftbp/prometheus.go
kylelemons: 🔕 so maybe a name like `thriftbp_ttlclient_connection_recycle_total` and a name like `replacement_success`?

or maybe just update the doc a little?

kylelemons (approved):

thriftbp/prometheus.go
fishy: that makes sense. changed to `thriftbp_`

thriftbp/prometheus.go
fishy: updated the doc.

konradreiche (approved):

httpbp/prometheus.go
```diff
@@ -107,6 +107,17 @@ var (
 	}, clientActiveRequestsLabels)
 )
 
+var (
+	panicRecoverLabels = []string{
+		methodLabel,
+	}
+
+	panicRecoverCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "httpbp_server_panic_recover_total",
```
bjk-reddit, line 116: Hmm, `bp` seems a bit uncessary in this naming. If this is an `http` server, it should just be under the `http` namespace. Also we tend to imply `server` for most metrics that are not explicitly `client`.

```suggestion
		Namespace: "http",
		Name: "panics_recover_total",
```

internal/limitopen/limitopen.go
```diff
@@ -9,10 +9,28 @@ import (
...
+	pathLabel = "path"
+)
+
+var (
+	sizeLabels = []string{
+		pathLabel,
+	}
+
+	sizeGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "limitopen_file_size_bytes",
```
bjk-reddit, line 29: Let's use `Namespace` here. It's also good to make this a package const.

```suggestion
const (
	metricsNamespace = "limitopen"
)
		Namespace: metricsNamespace,
		Name: "limitopen_file_size_bytes",
```

kafkabp/consumer.go
```diff
@@ -7,11 +7,28 @@ import (
...
+	successLabel = "success"
+)
+
+var (
+	rebalanceLabels = []string{
+		successLabel,
+	}
+
+	rebalanceCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "kafkabp_consumer_rebalance_total",
```
bjk-reddit, line 27: Same thing here, the namespace is just `kafka`.

redis/db/redisbp/monitored_client.go
```diff
@@ -7,9 +7,50 @@ import (
...
 
 	"github.com/go-redis/redis/v8"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
 	"github.com/reddit/baseplate.go/metricsbp"
 )
 
+const (
+	nameLabel = "name"
```
bjk-reddit, line 17: Probably `pool`?

```suggestion
	nameLabel = "pool"
```

redis/db/redisbp/monitored_client.go
```diff
@@ -109,11 +150,35 @@ func MonitorPoolStats(ctx context.Context, client PoolStatser, name string, tags
 		case <-ticker.C:
 			stats := client.PoolStats()
 			hitsGauge.Set(float64(stats.Hits))
+			statsGauge.With(prometheus.Labels{
+				nameLabel:   name,
+				resultLabel: resultHits,
+			}).Set(float64(stats.Hits))
```
bjk-reddit, line 156: So, I'm not sure we should hook into the metrics the same way with this one.

This is going to have a couple problems, first the interval/cadence of updates is going to have a missmatch between this and the scape interval.

I would recommend, if we can, to insert our instrumentation at the actual point of change. For example, for a `hit`, we put a counter there and `Inc()`. This more direct instrumentation is less prone to problems.

There are also a bunch of different metrics here that shouldn't be labels, but rather separate metric names.

For example, if you `sum(redisbp_pool_stats)`, its going to include hits+miss+total, which is probably not what the user wants.

Typically in Prometheus we either have a metric for `status="hit"` + `status="miss"`, this way `sum()` works.

Or we have separate `hits_total` and `requests_total` metrics so to get misses you do `requests_total - hits_total`.

bjk-reddit (changes requested):
Some naming convention ideas, and questions about how to better hook into some of these things.

We should avoid using `gauge.Set()` where we know things are actually counters.

If we have counters we want to expose from existing metrics vars, we can use "Const" metrics and register a handler function, similar to how stand-alone exporters like node_exporter work.

breakerbp/failure_ratio.go
bjk-reddit: I actually liked this the other way.

In Prometheus I like to make boolean values 1 for the positive thing. Like `up` is.

kafkabp/consumer.go
bjk-reddit: I think `kafka` is better in this case. The `bp` thing is a bit weird to me.

breakerbp/failure_ratio.go
```diff
@@ -6,12 +6,29 @@ import (
...
+	nameLabel = "name"
+)
+
+var (
+	breakerLabels = []string{
+		nameLabel,
+	}
+
+	breakerTripped = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "breakerbp_tripped",
```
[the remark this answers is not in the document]
fishy: no it's not specific for thrift.

internal/limitopen/limitopen.go
fishy: is namespace similar to a meta label? also do we want to repeat namespace and the name? (here the namespace is also the prefix of the name)

thriftbp/server_middlewares.go
```diff
@@ -299,16 +299,22 @@ func ReportPayloadSizeMetrics(rate float64) thrift.ProcessorMiddleware {
...
+							labels := prometheus.Labels{
+								methodLabel: name,
+								protoLabel:  proto,
+							}
 							metricsbp.M.HistogramWithRate(metricsbp.RateArgs{
 								Name:             "payload.size." + name + ".request",
 								Rate:             1,
 								AlreadySampledAt: metricsbp.Float64Ptr(rate),
 							}).With("proto", proto).Observe(float64(itrans.Size()))
+							payloadSizeRequestBytes.With(labels).Observe(float64(itrans.Size()))
```
bjk-reddit, line 311: This section needs to be restructured a bit.

`if randbp.ShouldSampleWithRate(rate) {` above needs to be moved to wrap around the `metricsbp.M.HistogramWithRate()` calls so that we can use `prometheus.Observe()` on every call, but only sample for statsd calls.

bjk-reddit (changes requested):

internal/limitopen/limitopen.go
fishy: how does namespace show up in grafana? we did not use namespaces in any prometheus metrics defined in the spec.

redis/db/redisbp/monitored_client.go
fishy: the upstream redis library does not provide hooks for us to actually emit counters, so I separated the gauges.

thriftbp/server_middlewares.go
fishy: that means we need to apply the overhead of counting the bytes to all requests. we don't want to do that (for now). we might revisit that after we removed statsd metrics.

bjk-reddit: Let's leave this out of the change for now. We do not want to introduce pre-sampled histograms into Prometheus metrics. It's just not handled by the ecosystem.

fishy: this is an important metrics that we don't want to lose when we migrate to prometheus. also I don't think it will break the system in any way? we just have randomly sampled reporting instead of reporting all, which does not really have any real impact on p99/p50 in statistical ways?

fishy:
@bjk-reddit I removed redisbp's hits/misses/timeouts from this PR, PTAL.

thriftbp/server_middlewares.go
kylelemons: We need to defer to the expert here, and the clear message is that we do not sample prometheus metrics, full stop.

If you are concerned about performance, I recommend making that decision based on a benchmark or a load test.  If the data says that it's expensive, it should be opt-in (or maybe opt-out), but still recording samples at 100% when enabled.  Either collecting the data is worth the cost or it's not.

kylelemons (changes requested):

breakerbp/failure_ratio.go
```diff
@@ -92,8 +119,10 @@ func (cb FailureRatioBreaker) runStatsProducer() {
 		case <-tick.C:
 			if cb.State() == gobreaker.StateOpen {
 				circuitBreakerGauge.Set(0)
+				breakerClosed.With(labels).Set(0)
 			} else {
 				circuitBreakerGauge.Set(1)
+				breakerClosed.With(labels).Set(1)
```
bjk-reddit, line 125: These should be part of the `stateChanged()` method, so that the values are updated any time the `gobreaker.State` changes.

bjk-reddit (changes requested):

thriftbp/client_pool.go
```diff
@@ -451,6 +452,22 @@ func newClientPool(
 			cfg.PoolGaugeInterval,
 			tags,
 		)
+
+		if err := prometheus.Register(clientPoolGaugeExporter{
```
SuperQ, line 456: While I like the use of `Register()` over `MustRegister()`, this isn't going to be very effective in catching errors since `func (e clientPoolGaugeExporter) Describe()` is a noop.

I guess this is fine, since we can improve upon things later. But it's going to provide a misleading amount of safety as-is.

fishy: I disagree. If the only choices we have is between this and panic (via `MustRegister`), I don't see how this is worse than panic.

SuperQ: It's not worse, it's just not doing what you think it's doing.

fishy: I can add a comment to say "this should not happen because desc is no-op but just in case" if that makes you happy.

SuperQ (approved):
LGTM, Nice!

kylelemons (approved):

PR #479, opened 2022-02-08
Add http_endpoint label for Prometheus metrics

konradreiche:
This adds a separate label for HTTP Prometheus server metrics to differentiate individual endpoints. For HTTP servers we have the notion of endpoint names, depending on whether we want to move away from defining them as part of a HTTP server router we could also use a different naming schema.


kylelemons (approved):

fishy (approved):

JessicaGreben (approved):
thanks for adding this

PR #480, opened 2022-02-08
Add Prometheus metrics to redisbp

SuperQ:
Add a Prometheus exporter interface implementation to the redisbp
`NewMonitoredClient()`.

### DIFF
--- a/redis/db/redisbp/monitored_client.go
+++ b/redis/db/redisbp/monitored_client.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/go-redis/redis/v8"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/reddit/baseplate.go/metricsbp"
 )
 
@@ -25,6 +26,13 @@ type PoolStatser interface {
 func NewMonitoredClient(name string, opt *redis.Options) *redis.Client {
 	client := redis.NewClient(opt)
 	client.AddHook(SpanHook{ClientName: name})
+
+	if err := prometheus.Register(newExporter(client, name)); err != nil {
+		// prometheus.Register should never fail because
+		// exporter.Describe is a no-op, but just in case.
+		return nil
+	}
+
 	return client
 }
 
@@ -33,6 +41,13 @@ func NewMonitoredClient(name string, opt *redis.Options) *redis.Client {
 func NewMonitoredFailoverClient(name string, opt *redis.FailoverOptions) *redis.Client {
 	client := redis.NewFailoverClient(opt)
 	client.AddHook(SpanHook{ClientName: name})
+
+	if err := prometheus.Register(newExporter(client, name)); err != nil {
+		// prometheus.Register should never fail because
+		// exporter.Describe is a no-op, but just in case.
+		return nil
+	}
+
 	return client
 }
 
@@ -77,6 +92,12 @@ func (cc *ClusterClient) Wait(ctx context.Context, args WaitArgs) (replicas int6
 func NewMonitoredClusterClient(name string, opt *redis.ClusterOptions) *ClusterClient {
 	client := redis.NewClusterClient(opt)
 	client.AddHook(SpanHook{ClientName: name})
+
+	if err := prometheus.Register(newExporter(client, name)); err != nil {
+		// prometheus.Register should never fail because
+		// exporter.Describe is a no-op, but just in case.
+		return nil
+	}
 
 	return &ClusterClient{client}
 }
--- a/redis/db/redisbp/prometheus.go
+++ b/redis/db/redisbp/prometheus.go
@@ -1,0 +1,128 @@
+package redisbp
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	promNamespace = "redisbp"
+	subsystemPool = "pool"
+
+	nameLabel = "pool"
+)
+
+// exporter provides an interface for Prometheus metrics.
+type exporter struct {
+	client PoolStatser
+	name   string
+
+	poolHitsCounterDesc     *prometheus.Desc
+	poolMissesCounterDesc   *prometheus.Desc
+	poolTimeoutsCounterDesc *prometheus.Desc
+	totalConnectionsDesc    *prometheus.Desc
+	idleConnectionsDesc     *prometheus.Desc
+	staleConnectionsDesc    *prometheus.Desc
+}
+
+func newExporter(client PoolStatser, name string) *exporter {
+	labels := []string{
+		nameLabel,
+	}
+
+	return &exporter{
+		client: client,
+		name:   name,
+
+		// Upstream docs: https://pkg.go.dev/github.com/go-redis/redis/v8/internal/pool#Stats
+
+		// Counters.
+		poolHitsCounterDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "hits_total"),
+			"Number of times free connection was found in the pool",
+			labels,
+			nil,
+		),
+		poolMissesCounterDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "misses_total"),
+			"Number of times free connection was NOT found in the pool",
+			labels,
+			nil,
+		),
+		poolTimeoutsCounterDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "timeouts_total"),
+			"Number of times a wait timeout occurred",
+			labels,
+			nil,
+		),
+
+		// Gauges.
+		totalConnectionsDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "connections"),
+			"Number of connections in this redisbp pool",
+			labels,
+			nil,
+		),
+		idleConnectionsDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "idle_connections"),
+			"Number of idle connections in this redisbp pool",
+			labels,
+			nil,
+		),
+		staleConnectionsDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "stale_connections"),
+			"Number of stale connections in this redisbp pool",
+			labels,
+			nil,
+		),
+	}
+}
+
+// Describe implements the prometheus.Collector interface.
+func (e *exporter) Describe(ch chan<- *prometheus.Desc) {
+	// All metrics are described dynamically.
+}
+
+// Collect implements prometheus.Collector.
+func (e *exporter) Collect(ch chan<- prometheus.Metric) {
+	stats := e.client.PoolStats()
+
+	// Counters.
+	ch <- prometheus.MustNewConstMetric(
+		e.poolHitsCounterDesc,
+		prometheus.CounterValue,
+		float64(stats.Hits),
+		e.name,
+	)
+	ch <- prometheus.MustNewConstMetric(
+		e.poolMissesCounterDesc,
+		prometheus.CounterValue,
+		float64(stats.Misses),
+		e.name,
+	)
+	ch <- prometheus.MustNewConstMetric(
+		e.poolTimeoutsCounterDesc,
+		prometheus.CounterValue,
+		float64(stats.Timeouts),
+		e.name,
+	)
+
+	// Gauges.
+	ch <- prometheus.MustNewConstMetric(
+		e.totalConnectionsDesc,
+		prometheus.GaugeValue,
+		float64(stats.TotalConns),
+		e.name,
+	)
+	ch <- prometheus.MustNewConstMetric(
+		e.idleConnectionsDesc,
+		prometheus.GaugeValue,
+		float64(stats.IdleConns),
+		e.name,
+	)
+	ch <- prometheus.MustNewConstMetric(
+		e.staleConnectionsDesc,
+		prometheus.GaugeValue,
+		float64(stats.StaleConns),
+		e.name,
+	)
+}
--- a/redis/db/redisbp/prometheus_test.go
+++ b/redis/db/redisbp/prometheus_test.go
@@ -1,0 +1,46 @@
+package redisbp
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/go-redis/redis/v8"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type fakeClient redis.Client
+
+// PoolStats returns connection pool stats.
+func (c fakeClient) PoolStats() *redis.PoolStats {
+	return &redis.PoolStats{
+		Hits:     1,
+		Misses:   2,
+		Timeouts: 3,
+
+		TotalConns: 4,
+		IdleConns:  5,
+		StaleConns: 6,
+	}
+}
+
+func TestRedisPoolExporter(t *testing.T) {
+	client := &fakeClient{}
+
+	exporter := newExporter(
+		client,
+		"test",
+	)
+	// No real test here, we just want to make sure that Collect call will not
+	// panic, which would happen if we have a label mismatch.
+	ch := make(chan prometheus.Metric)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for range ch {
+		}
+	}()
+	exporter.Collect(ch)
+	close(ch)
+	wg.Wait()
+}


SuperQ:
I still need to implement a couple more stats, but they should follow this pattern. I also need to read the docs to make sure I get the types correct.

redis/db/redisbp/monitored_client.go
```diff
@@ -72,6 +83,75 @@ func (cc *ClusterClient) Wait(ctx context.Context, args WaitArgs) (replicas int6
...
+	client PoolStatser
+	name   string
+}
+
+func NewExporter(client PoolStatser, name string) *Exporter {
+	return &Exporter{
+		client: client,
+		name:   name,
+	}
+}
```
fishy, line 124: I don't think we need to export either `Exporter` or `NewExporter`? Or if you expect people to construct the client themselves without using `NewMonitoredClient`, then `NewExporter` might be useful but `Exporter` still does not need to be exported (or just remove `NewExporter` and keep `Exporter`, as you didn't actually do anything extra inside `NewExporter`).

redis/db/redisbp/monitored_client.go
```diff
@@ -72,6 +83,75 @@ func (cc *ClusterClient) Wait(ctx context.Context, args WaitArgs) (replicas int6
...
+
+	staleConnectionsDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(promNamespace, subsystemPool, "stale_connections"),
+		"Number of stale connections in this redisbp pool",
+		gaugeLabels,
+		nil,
+	)
+)
+
+// Exporter provides an interface for Prometheus merics.
```
fishy, line 113: typo: `s/merics/metrics/`

redis/db/redisbp/monitored_client.go
```diff
@@ -72,6 +83,75 @@ func (cc *ClusterClient) Wait(ctx context.Context, args WaitArgs) (replicas int6
...
+// Describe implements the prometheus.Collector interface.
+func (e *Exporter) Describe(ch chan<- *prometheus.Desc) {
+	// All metrics are described dynamically.
+}
+
+// Collect implements prometheus.Collector.
+func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
+	stats := e.client.PoolStats()
+
+	ch <- prometheus.MustNewConstMetric(
```
fishy, line 135: is there a reason we have to use the `Must` version, instead of using `promauto` to declare global variables? I really don't like the idea of using `Must*` outside of `init` or global variable declarations.

redis/db/redisbp/monitored_client.go
```diff
@@ -25,6 +33,9 @@ type PoolStatser interface {
 func NewMonitoredClient(name string, opt *redis.Options) *redis.Client {
 	client := redis.NewClient(opt)
 	client.AddHook(SpanHook{ClientName: name})
+
+	prometheus.MustRegister(NewExporter(client, name))
```
fishy, line 37: same comment about `Must*` here.

looking at the doc, it says:

```
	// If the provided Collector is equal to a Collector already registered
	// (which includes the case of re-registering the same Collector), the
	// returned error is an instance of AlreadyRegisteredError, which
	// contains the previously registered Collector.
	//
	// A Collector whose Describe method does not yield any Desc is treated
	// as unchecked. Registration will always succeed. No check for
	// re-registering (see previous paragraph) is performed. Thus, the
	// caller is responsible for not double-registering the same unchecked
	// Collector, and for providing a Collector that will not cause
	// inconsistent metrics on collection. (This would lead to scrape
	// errors.)
```

so my understanding is that this will always success and never panic. which makes using `MustRegister` fine in this case, but we should add a comment here to explain that.

fishy (changes requested):

redis/db/redisbp/monitored_client.go
SuperQ: This is the standard for metrics collection when the values you're gathering are external to your application.

SuperQ: https://prometheus.io/docs/instrumenting/writing_exporters/#collectors

redis/db/redisbp/monitored_client.go
SuperQ: Sure, those can also be private. I will make this change.

redis/db/redisbp/monitored_client.go
fishy: ok looking at https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewConstMetric, the only way it would cause panic is label mismatch, which I guess is fine.

but in that case, we should add a test to call this `Collect` function, so that if we actually have a label mismatch in the code, the test would panic (and thus fail).

redis/db/redisbp/monitored_client.go
```diff
@@ -72,6 +83,75 @@ func (cc *ClusterClient) Wait(ctx context.Context, args WaitArgs) (replicas int6
...
+func NewExporter(client PoolStatser, name string) *Exporter {
+	return &Exporter{
+		client: client,
+		name:   name,
+	}
+}
+
+// Describe implements the prometheus.Collector interface.
+func (e *Exporter) Describe(ch chan<- *prometheus.Desc) {
+	// All metrics are described dynamically.
```
konradreiche, line 128: Is this necessary though, wouldn't we be able to send all descriptors in this instance here?
```
  // The sent descriptors fulfill the
	// consistency and uniqueness requirements described in the Desc
	// documentation.
```

konradreiche (changes requested):

redis/db/redisbp/monitored_client.go
SuperQ: We need at least a noop as the Prometheus interface requires it at runtime.

redis/db/redisbp/monitored_client.go
kylelemons: registering an unchecked collector sounds like we're gonna trip over this, can we make it a collector that Describes itself so we _do_ get this checking?

redis/db/redisbp/monitored_client.go
kylelemons: +1 to making this fail at unit test time

kylelemons (changes requested):
+1 to the comment

redis/db/redisbp/monitored_client.go
```diff
@@ -25,6 +26,13 @@ type PoolStatser interface {
 func NewMonitoredClient(name string, opt *redis.Options) *redis.Client {
 	client := redis.NewClient(opt)
 	client.AddHook(SpanHook{ClientName: name})
+
+	if err := prometheus.Register(newExporter(client, name)); err != nil {
+		// prometheus.Register should never fail because
+		// exporter.Describe is a no-op, but just in case.
+		return nil
```
kylelemons, line 33: 🔕 Log it? Export a counter :trollface: ?

redis/db/redisbp/prometheus.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return &exporter{
+		client: client,
+		name:   name,
+
+		// Upstream docs: https://pkg.go.dev/github.com/go-redis/redis/v8/internal/pool#Stats
+
+		// Counters.
+		poolHitsCounterDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "hits_total"),
```
kylelemons, line 40: is there a reason not to combine these as `cache{result="hit|miss"}` so that sums can be used for totaling?

redis/db/redisbp/prometheus.go
```diff
@@ -0,0 +1,128 @@
...
+		poolTimeoutsCounterDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "timeouts_total"),
+			"Number of times a wait timeout occurred",
+			labels,
+			nil,
+		),
+
+		// Gauges.
+		totalConnectionsDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(promNamespace, subsystemPool, "connections"),
```
kylelemons, line 60: same question but with `connection{state="active|idle|stale"}`

kylelemons (approved):

redis/db/redisbp/monitored_client.go
fishy, line 33: 🔕 I personally would slightly prefer to use `MustRegiester` over returning an `nil` client (but still add a comment to explain that this is not supposed to happen).

pros of `MustRegister`: when it actually happens, the panic message will be the error returned by `Register`, so it's more obvious that it's the prometheus exporter causing trouble. vs. returning `nil`, which likely will also cause panic, but the panic message would be about nil dereference, and it's much harder to figure out why.

pros of using `nil`: caller can actually check for `nil` and fail more gracefully than just panic.

redis/db/redisbp/prometheus.go
fishy: I _think_ the reason is that hits+misses may or may not be the total.

redis/db/redisbp/prometheus.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+// Describe implements the prometheus.Collector interface.
+func (e *exporter) Describe(ch chan<- *prometheus.Desc) {
+	// All metrics are described dynamically.
+}
+
+// Collect implements prometheus.Collector.
+func (e *exporter) Collect(ch chan<- prometheus.Metric) {
```
fishy, line 86: please add an unit test to this function, similar to https://github.com/reddit/baseplate.go/blob/70ef150cf4a06005b64806bcf3c62bc73b128710/thriftbp/prometheus_test.go#L156

fishy (changes requested):

redis/db/redisbp/monitored_client.go
SuperQ: Yes, that's why I chose to just return nil, it's easy to check.

redis/db/redisbp/prometheus.go
SuperQ: When dealing with external values like this where you don't have good control of the underlying implementation I prefer to not guess and expose them as separate metrics.

I'd prefer "cache requests" and "cache misses", but we only have hits and misses.

redis/db/redisbp/prometheus.go
```diff
@@ -0,0 +1,128 @@
...
+			"Number of stale connections in this redisbp pool",
+			labels,
+			nil,
+		),
+	}
+}
+
+// Describe implements the prometheus.Collector interface.
+func (e *exporter) Describe(ch chan<- *prometheus.Desc) {
+	// All metrics are described dynamically.
```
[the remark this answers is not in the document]
SuperQ: Yes, it's better to include it. It avoids the situation where someone tries to use both this method and `promauto` for the same metric family.

redis/db/redisbp/prometheus.go
SuperQ: Done, but I'm having some issues with the results stalling.

redis/db/redisbp/prometheus_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	exporter := newExporter(
+		&client,
+		"test",
+	)
+	// No real test here, we just want to make sure that Collect call will not
+	// panic, which would happen if we have a label mismatch.
+	//
+	// TODO: This hangs.
+	// exporter.Collect(make(chan prometheus.Metric, 2))
```
fishy, line 37: this hangs because you are pushing more than 2 metrics to the channel in the `Collect` function. I used a fixed 2 in the other test because that one is simple and only writes 2 metrics. since this one is more complicated, I'd suggest something like this:

```go
ch := make(chan prometheus.Metric)
var wg sync.WaitGroup
wg.Add(1)
go func() {
  defer wg.Done()
  for range ch {}
}()

exporter.Collect(ch)
close(ch)
wg.Wait()
```

fishy: we probably should change the other test to this as well (instead of using magic number of 2) 😅

redis/db/redisbp/prometheus_test.go
```diff
@@ -0,0 +1,38 @@
...
+		IdleConns:  5,
+		StaleConns: 6,
+	}
+}
+
+func TestRedisPoolExporter(t *testing.T) {
+	client := fakeClient{}
+
+	exporter := newExporter(
+		&client,
```
fishy, line 30: nit: since only `*fakeClient` implements it, using the non-pointer as `client` is misleading, we should just do `client := &fakeClient{}` (or `client := new(fakeClient)`) on line 27 and just use `client` here, or just inline it here.

(also it's really unnecessary to make it a pointer receiver on line 14)

fishy (changes requested):

redis/db/redisbp/prometheus_test.go
SuperQ: Ohh, that makes sense. Thanks!

redis/db/redisbp/prometheus_test.go
SuperQ: Yes, I was going to do that, but forgot to fix it. :facepalm:

redis/db/redisbp/prometheus_test.go
```diff
@@ -0,0 +1,47 @@
...
+func TestRedisPoolExporter(t *testing.T) {
+	client := &fakeClient{}
+
+	exporter := newExporter(
+		client,
+		"test",
+	)
+	// No real test here, we just want to make sure that Collect call will not
+	// panic, which would happen if we have a label mismatch.
+	//
```
fishy, line 35: nit: no need for the extra blank comment line :)

fishy (approved):

redis/db/redisbp/prometheus.go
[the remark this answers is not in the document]
SuperQ: Ok, I've implemented the Desc functionality.

redis/db/redisbp/prometheus.go
```diff
@@ -0,0 +1,133 @@
...
+			prometheus.BuildFQName(promNamespace, subsystemPool, "stale_connections"),
+			"Number of stale connections in this redisbp pool",
+			labels,
+			nil,
+		),
+	}
+}
+
+// Describe implements the prometheus.Collector interface.
+func (e *exporter) Describe(ch chan<- *prometheus.Desc) {
```
fishy, line 81: 🔕 In this approach, will it cause problems if someone call `NewMonitoredClient` twice (and thus create two `exporter`s to be registered)? maybe we should add a test to make sure that is fine?

fishy: actually it will cause the second `Register` call to return an error, as prometheus consider them duplicated: https://github.com/prometheus/client_golang/blob/v1.12.1/prometheus/registry.go#L295-L299

so we need to go back to the no-op `Describe` and dynamically register it way.

fishy (changes requested):

redis/db/redisbp/prometheus.go
```diff
@@ -0,0 +1,128 @@
...
+type exporter struct {
+	client PoolStatser
+	name   string
+
+	poolHitsCounterDesc     *prometheus.Desc
+	poolMissesCounterDesc   *prometheus.Desc
+	poolTimeoutsCounterDesc *prometheus.Desc
+	totalConnectionsDesc    *prometheus.Desc
+	idleConnectionsDesc     *prometheus.Desc
+	staleConnectionsDesc    *prometheus.Desc
```
fishy, line 24: in the case of a service needs more than one redis clients (and thus they will implicitly create more than one `exporter`s), since all those `Desc`s don't really change between the `exporter` instances, I think it's better to just define them as global variables (what you originally did and what I did in the thrift pool one), over making them part of the `exporter` and create during the constructor.

redis/db/redisbp/prometheus.go
```diff
@@ -0,0 +1,128 @@
...
+			prometheus.BuildFQName(promNamespace, subsystemPool, "stale_connections"),
+			"Number of stale connections in this redisbp pool",
+			labels,
+			nil,
+		),
+	}
+}
+
+// Describe implements the prometheus.Collector interface.
+func (e *exporter) Describe(ch chan<- *prometheus.Desc) {
```
fishy, line 81: can you also add an unit test to make sure that multiple `prometheus.Register(exporter)` will not cause issues, like I'm doing in https://github.com/reddit/baseplate.go/pull/494/files?

fishy (approved):
the 2 comments are non-blocking, I guess

konradreiche (approved):

fishy:
I'll merge this in and address the 2 comments I left in a follow up PR.

PR #481, opened 2022-02-16
Add support for *string in redisx

domenipavec:
Add support for *string data receiver in redisx
Using `*string` receiver for data from redis enables distinguishing between empty string values and `nil` (e.g. stored empty string vs. cache miss).

### DIFF
--- a/redis/cache/redisx/request.go
+++ b/redis/cache/redisx/request.go
@@ -250,7 +250,8 @@ func (r Request) setStructValue(dst reflect.Value, src reflect.Value) error {
 func (r Request) convertAndSetByteSlice(dst reflect.Value, src reflect.Value) error {
 	asBytes, _ := src.Interface().([]byte)
 	asStr := string(asBytes)
-	if dst.Kind() == reflect.Int64 {
+	switch dst.Kind() {
+	case reflect.Int64:
 		asInt, err := strconv.ParseInt(asStr, 10, 64)
 		if err != nil {
 			return &InvalidInputError{
@@ -258,14 +259,16 @@ func (r Request) convertAndSetByteSlice(dst reflect.Value, src reflect.Value) er
 			}
 		}
 		dst.Set(reflect.ValueOf(asInt))
-	} else if dst.Kind() == reflect.String {
+	case reflect.String:
 		dst.Set(reflect.ValueOf(asStr))
-	} else if dst.Type() == stringPtrT {
-		dst.Set(reflect.ValueOf(&asStr))
-	} else {
-		return &ResponseInputTypeError{
-			Cmd:               r.Cmd,
-			ResponseInputType: dst.Type(),
+	default:
+		if dst.Type() == stringPtrT {
+			dst.Set(reflect.ValueOf(&asStr))
+		} else {
+			return &ResponseInputTypeError{
+				Cmd:               r.Cmd,
+				ResponseInputType: dst.Type(),
+			}
 		}
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


redis/cache/redisx/request.go
```diff
@@ -249,18 +250,19 @@ func (r Request) setStructValue(dst reflect.Value, src reflect.Value) error {
 func (r Request) convertAndSetByteSlice(dst reflect.Value, src reflect.Value) error {
 	asBytes, _ := src.Interface().([]byte)
 	asStr := string(asBytes)
-	switch dst.Kind() {
-	case reflect.Int64:
```
rbietin, line 253: Why did you change this switch?

domenipavec: @rbietin Because in my case I have to check `dst.Type()` and not `dst.Kind()` that is checked in other cases.

rbietin (approved):

kylelemons (changes requested):

redis/cache/redisx/request.go
domenipavec: @fishy @kylelemons Ok, changed back to using `switch`.

redis/cache/redisx/request_benchmark_test.go
```diff
@@ -11,6 +11,8 @@ import (
 // BenchmarkSetValue
 // BenchmarkSetValue/string
 // BenchmarkSetValue/string-12         	14946445	        67.1 ns/op
+// BenchmarkSetValue/*string
+// BenchmarkSetValue/*string-12	         6187052	        169 ns/op
```
fishy, line 15: 🔕 this looks a lot slower than others (for example, `string`). it's probably due to your machine being slower than the original source of the benchmark results, you should replace the whole result, not only add the new one you added.

fishy (approved):

konradreiche (approved):

redis/cache/redisx/request_benchmark_test.go
domenipavec: The values for others are comparable on my machine. This is slower than the `string` benchmark, because it's going from `[]byte` to `*string` so it's doing type conversion, while `string` benchmark is just doing the assignment from string to string. If I change the `string` benchmark to go from `[]byte` it's just as slow.

redis/cache/redisx/request.go
```diff
@@ -261,9 +262,13 @@ func (r Request) convertAndSetByteSlice(dst reflect.Value, src reflect.Value) er
...
-			Cmd:               r.Cmd,
-			ResponseInputType: dst.Type(),
+		if dst.Type() == stringPtrT {
+			dst.Set(reflect.ValueOf(&asStr))
+		} else {
+			return &ResponseInputTypeError{
+				Cmd:               r.Cmd,
+				ResponseInputType: dst.Type(),
+			}
 		}
```
kylelemons, line 272: 🔕 this can be a switch too

```suggestion
		switch dst.Type() {
		case stringPtrT:
			dst.Set(reflect.ValueOf(&asStr))
		default:
			return &ResponseInputTypeError{
				Cmd:               r.Cmd,
				ResponseInputType: dst.Type(),
			}
		}
```

kylelemons (approved):

pacejackson (changes requested):
The change seems fine but it should probably also support *int64, it seems like it would be confusing to support this for the string conversion but not the integer one. 

There’s also documentation here that needs to be updated to reflect the new types you can use. https://github.com/reddit/baseplate.go/blob/70ef150cf4a06005b64806bcf3c62bc73b128710/redis/cache/redisx/syncx.go#L27

redis/cache/redisx/request_benchmark_test.go
domenipavec: I've modified the benchmarks to include source and destination types so the results are easier to interpret.

domenipavec:
@pacejackson Added support for `*int64` and updated documentation.

PR #482, opened 2022-02-23
thriftbp: Handle wrapped TException in prometheus middlewares

fishy:
For thrift's server middleware, the error type is of TException instead
of error, so usually freeform errors need to be wrapped by
thrift.WrapTException, and just try to get the type via %T will hide the
actual type returned by the server handler.

Add a helper function stringifyErrorType to try to unwrap such cases.

### DIFF
--- a/thriftbp/client_middlewares.go
+++ b/thriftbp/client_middlewares.go
@@ -3,9 +3,7 @@ package thriftbp
 import (
 	"context"
 	"errors"
-	"fmt"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
@@ -361,11 +359,10 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
 				clientActiveRequests.With(activeRequestLabels).Inc()
 
 				defer func() {
-					var exceptionTypeLabel, baseplateStatusCode, baseplateStatus string
+					var baseplateStatusCode, baseplateStatus string
+					exceptionTypeLabel := stringifyErrorType(err)
 					success := strconv.FormatBool(err == nil)
 					if err != nil {
-						exceptionTypeLabel = strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
-
 						var bpErr baseplateError
 						if errors.As(err, &bpErr) {
 							code := bpErr.GetCode()
--- a/thriftbp/prometheus.go
+++ b/thriftbp/prometheus.go
@@ -1,6 +1,11 @@
 package thriftbp
 
 import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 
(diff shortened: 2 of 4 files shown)


thriftbp/prometheus.go
```diff
@@ -235,3 +240,18 @@ func (e clientPoolGaugeExporter) Collect(ch chan<- prometheus.Metric) {
 		e.slug,
 	)
 }
+
+func stringifyErrorType(err error) string {
+	if err == nil {
+		return ""
+	}
+	var te thrift.TException
+	if errors.As(err, &te) && te.TExceptionType() == thrift.TExceptionTypeUnknown {
```
nanassito, line 249: niiice

nanassito (approved):

JessicaGreben (commented):
does this change create unbounded cardinality?

nanassito:
It does increase cardinality a bit since the correct exception type is now being put in the label. But that is only exception *types* we are talking about here, not exception messages.

fishy:
> does this change create unbounded cardinality?

Yes, but that's a calculated risk and the intended behavior (from the baseplate spec discussion).

kylelemons (approved):

konradreiche (approved):
:clap:

thriftbp/client_middlewares.go
```diff
@@ -361,11 +359,10 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
...
 				defer func() {
-					var exceptionTypeLabel, baseplateStatusCode, baseplateStatus string
+					var baseplateStatusCode, baseplateStatus string
+					exceptionTypeLabel := stringifyErrorType(err)
 					success := strconv.FormatBool(err == nil)
 					if err != nil {
-						exceptionTypeLabel = strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
-
 						var bpErr baseplateError
 						if errors.As(err, &bpErr) {
```
nanassito, line 367: ah oops. This will never match because `err` is a `thrift.TException` not a `baseplateError`

nanassito: sorry I'm discovering this after you merged

fishy: no an concrete error type can implement 2 interfaces.

nanassito: well I just ran it through the debugger and it doesn't enter the `if errors.As(err, &bpErr) {` block

nanassito: I just created a repro in auth-go which uses `thrift_version=9bee877` and crafted a request so that the response is a `Baseplate.Error`.

adding a break point here, I can see that `err` is `github.com/apache/thrift/lib/go/thrift.TException(*github.snooguts.net/reddit/reddit-service-authentication-go/thrift/reddit/baseplate.Error) *{Code: *401, Message: *"Unauthorized", Details: map[string]string ["reason": "could not find user Identity. NotFoundError({})", ]}` and then stepping through execution I can see it pass over the `if errors.As(err, &bpErr) {...}` block.

nanassito: [T](https://github.com/reddit/baseplate.go/pull/485) is what I mean

PR #483, opened 2022-02-24
Add baseplate admin package

JessicaGreben:
baseplate-cookiecutter adds this boilerplate to new baseplate services ([code](https://github.snooguts.net/reddit/baseplate-cookiecutter/blob/master/baseplate_cookiecutter/make_project/go/%7B%7Bcookiecutter.project_slug%7D%7D/main.go#L86)):
```
func serveAdmin() {
    // This goroutine opens http://localhost:6060/ for internal functions:
    //    pprof   - https://blog.golang.org/pprof
    //    metrics - will serve /metrics for prometheus (TBD)
    //    status  - will serve /health for health checking
    go func() {
        http.HandleFunc("/health", service.IsHealthy)
        http.Handle("/metrics", promhttp.Handler())
        log.Infof("Serving admin on %s", *adminAddr)
        log.Warnw("admin http serve exited", "err", http.ListenAndServe(*adminAddr, nil))
    }()
}
```

This PR adds that boilerplate into baseplate.go to create an opinionated `baseplate.go/admin` package that reduces this boilerplate (per Ben's rec [here](https://reddit.slack.com/archives/CLMUD9GTE/p1645719276430109)).

### DIFF
--- a/admin/server.go
+++ b/admin/server.go
@@ -1,0 +1,55 @@
+package admin
+
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/reddit/baseplate.go/log"
+)
+
+const (
+	DefaultAdminAddr = ":6060"
+)
+
+// ServerArgs contain optional configuration for the admin server.
+type ServerArgs struct {
+	// AdminAddr is a custom address for the admin server. Defaults to port 6060 on localhost if not set.
+	AdminAddr string
+	// HealthCheckFn is the HTTP handler for health check.
+	HealthCheckFn func(w http.ResponseWriter, r *http.Request)
+}
+
+// NewServer returns a new admin server for internal functionality.
+func NewServer(args *ServerArgs) *Server {
+	adminAddr := args.AdminAddr
+	if args.AdminAddr == "" {
+		adminAddr = DefaultAdminAddr
+	}
+
+	return &Server{
+		adminAddr:     adminAddr,
+		healthCheckFn: args.HealthCheckFn,
+	}
+}
+
+type Server struct {
+	adminAddr     string
+	healthCheckFn func(w http.ResponseWriter, r *http.Request)
+}
+
+// Serve starts an HTTP server for internal functions:
+//    metrics       - serve /metrics for prometheus
+//    health check  - serve /health for health checking
+//    pprof         - https://blog.golang.org/pprof
+// Default server address is http://localhost:6060.
+func (s *Server) Serve() {
+	go func() {
+		if s.healthCheckFn != nil {
+			http.HandleFunc("/health", s.healthCheckFn)
+		}
+		http.Handle("/metrics", promhttp.Handler())
+		log.Infof("Serving admin on %s", s.adminAddr)
+		log.Warnw("admin http serve exited", "err", http.ListenAndServe(s.adminAddr, nil))
+	}()
+}
--- a/admin/doc.go
+++ b/admin/doc.go
@@ -1,0 +1,2 @@
+// Package admin is baseplate boilerplate for admin features.
+package admin
--- a/httpbp/example_server_test.go
+++ b/httpbp/example_server_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/reddit/baseplate.go"
+	"github.com/reddit/baseplate.go/admin"
 	"github.com/reddit/baseplate.go/ecinterface"
 	"github.com/reddit/baseplate.go/httpbp"
 	"github.com/reddit/baseplate.go/log"
@@ -25,27 +26,35 @@ type body struct {
 	Y int `json:"y"`
 }
 
-type Handlers struct {
+type isHealthyResponse struct {
+	Status string `json:"status,omitempty"`
+}
+
+type TestService struct {
 	secrets    *secrets.Store
 	redisAddrs []string
 }
 
-func (h Handlers) Home(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+func (s *TestService) IsHealthy(w http.ResponseWriter, r *http.Request) {
+	httpbp.NewResponse(isHealthyResponse{Status: "healthy"})
+}
+
+func (s *TestService) Home(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	return httpbp.WriteJSON(w, httpbp.NewResponse(body{X: 1, Y: 2}))
 }
 
-func (h Handlers) ServerErr(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+func (s *TestService) ServerErr(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	return httpbp.JSONError(httpbp.InternalServerError(), errors.New("example"))
 }
 
-func (h Handlers) Ratelimit(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+func (s *TestService) Ratelimit(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	return httpbp.JSONError(
 		httpbp.TooManyRequests().Retryable(w, time.Minute),
 		errors.New("rate-limit"),
 	)
 }
 
-func (h Handlers) InvalidInput(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+func (s *TestService) InvalidInput(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	return httpbp.JSONError(
 		httpbp.BadRequest().WithDetails(map[string]string{
 			"foo": "must be >= 0",
@@ -55,26 +64,26 @@ func (h Handlers) InvalidInput(ctx context.Context, w http.ResponseWriter, r *ht
 	)
 }
 
-func (h Handlers) Endpoints() map[httpbp.Pattern]httpbp.Endpoint {
+func (s *TestService) Endpoints() map[httpbp.Pattern]httpbp.Endpoint {
 	return map[httpbp.Pattern]httpbp.Endpoint{
 		"/": {
 			Name:    "home",
-			Handle:  h.Home,
+			Handle:  s.Home,
 			Methods: []string{http.MethodGet},
 		},
 		"/err": {
 			Name:    "err",
-			Handle:  h.ServerErr,
+			Handle:  s.ServerErr,
 			Methods: []string{http.MethodGet, http.MethodPost},
 		},
 		"/ratelimit": {
 			Name:    "ratelimit",
-			Handle:  h.Ratelimit,
+			Handle:  s.Ratelimit,
 			Methods: []string{http.MethodGet},
 		},
 		"/invalid-input": {
 			Name:    "invalid-input",
-			Handle:  h.InvalidInput,
+			Handle:  s.InvalidInput,
 			Methods: []string{http.MethodPost},
 		},
 	}
@@ -110,17 +119,19 @@ func ExampleNewBaseplateServer() {
 	}
 	defer bp.Close()
 
-	handlers := Handlers{
+	svc := TestService{
 		secrets:    bp.Secrets(),
 		redisAddrs: cfg.Redis.Addrs,
 	}
 	server, err := httpbp.NewBaseplateServer(httpbp.ServerArgs{
 		Baseplate:   bp,
-		Endpoints:   handlers.Endpoints(),
+		Endpoints:   svc.Endpoints(),
 		Middlewares: []httpbp.Middleware{loggingMiddleware},
 	})
 	if err != nil {
 		panic(err)
 	}
+	adminServer := admin.NewServer(&admin.ServerArgs{HealthCheckFn: svc.IsHealthy})
+	adminServer.Serve()
 	log.Info(baseplate.Serve(ctx, baseplate.ServeArgs{Server: server}))
 }
--- a/thriftbp/example_server_test.go
+++ b/thriftbp/example_server_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/reddit/baseplate.go"
+	"github.com/reddit/baseplate.go/admin"
 	"github.com/reddit/baseplate.go/ecinterface"
 	bpgen "github.com/reddit/baseplate.go/internal/gen-go/reddit/baseplate"
 	"github.com/reddit/baseplate.go/log"
@@ -41,5 +42,7 @@ func ExampleNewBaseplateServer() {
 		log.Fatal(err)
 	}
 
+	adminServer := admin.NewServer(&admin.ServerArgs{})
+	adminServer.Serve()
 	log.Info(baseplate.Serve(ctx, baseplate.ServeArgs{Server: server}))
 }


admin/server.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Serve starts an HTTP server for internal functions:
+//    metrics       - serve /metrics for prometheus
+//    health check  - serve /health for health checking
+//    pprof         - https://blog.golang.org/pprof
+// Default server address is http://localhost:6060.
+func (s *Server) Serve() {
+	go func() {
+		if s.healthCheckFn != nil {
+			http.HandleFunc("/health", s.healthCheckFn)
```
marcoferrer, line 49: Is it worth having a default for the health check fn as well?

admin/server.go
```diff
@@ -0,0 +1,55 @@
...
+	healthCheckFn func(w http.ResponseWriter, r *http.Request)
+}
+
+// Serve starts an HTTP server for internal functions:
+//    metrics       - serve /metrics for prometheus
+//    health check  - serve /health for health checking
+//    pprof         - https://blog.golang.org/pprof
+// Default server address is http://localhost:6060.
+func (s *Server) Serve() {
+	go func() {
```
fishy, line 47: I would prefer to document that this is a block function and the caller should spawn a goroutine, over doing this in a public api.

admin/server.go
```diff
@@ -0,0 +1,55 @@
...
+const (
+	DefaultAdminAddr = ":6060"
+)
+
+// ServerArgs contain optional configuration for the admin server.
+type ServerArgs struct {
+	// AdminAddr is a custom address for the admin server. Defaults to port 6060 on localhost if not set.
+	AdminAddr string
+	// HealthCheckFn is the HTTP handler for health check.
+	HealthCheckFn func(w http.ResponseWriter, r *http.Request)
```
fishy, line 20: nit: this is the same as [`http.HandlerFunc`](https://pkg.go.dev/net/http#HandlerFunc) so I'd rather just use that:

```suggestion
	HealthCheckFn http.HandlerFunc
```

admin/server.go
```diff
@@ -0,0 +1,55 @@
...
+
+type Server struct {
+	adminAddr     string
+	healthCheckFn func(w http.ResponseWriter, r *http.Request)
+}
+
+// Serve starts an HTTP server for internal functions:
+//    metrics       - serve /metrics for prometheus
+//    health check  - serve /health for health checking
+//    pprof         - https://blog.golang.org/pprof
```
fishy, line 44: so this does not happen with this function. the service still needs to do the blank import in their main package, and we can not do that for them here (because blank imports shall only happen in main package), which means this is techincally... a lie.

I'm not sure what's a good way to handle this, though.

fishy (changes requested):

admin/server.go
fishy: this is only meant for http services. for thrift services providing another http health check endpoint could gets confusing.

SuperQ: I think a default handler would be useful. It would allow users to always include a service health check in their supervisor (i.e. Kubernetes pod health checks).

Even if it's an always-ok stub, it allows for consistency of deployment configuration.

admin/server.go
SuperQ: Instead of doing a blank import, we could explicitly setup pprof by explicitly setting up the http handlers in `Serve()` rather than depend on the main `init()`.

https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/net/http/pprof/pprof.go;l=80-86

admin/server.go
SuperQ: I was thinking we should drop the `go` in this method and prefer to do `go admin.Serve()` in `main()`.

admin/server.go
fishy: I considered that, but that means if in the future a new pprof endpoint is added to go, people won't get them automatically by updating go, and we have to add them here.

and I'm not sure what would happen if they also blank import it in their main (e.g. the endpoints are double registered)

SuperQ: Yea, it's a bit more work for us, but I don't think it's that bad. That `init()` section hasn't changed since 2017, so not much churn. 

Yea, I thought about the double registering. I guess we could do some tests.

JessicaGreben: I have a slight preference for copying the pprof `init` func into this PR.  But I don't know what double registering would do. I will test it out.

Otherwise, we can make sure the blank import is happening in baseplate cookiecutter.

admin/server.go
JessicaGreben: I'm fine with either adding a default health check or not. @fishy do you have a preference?

fishy: We might want to do that eventually, but now is not the time yet (we need more discussion, etc.. there's already a discussion on how the v2 health check can be provided by the framework/library), so we should leave that out for a future PR.

admin/server.go
fishy: I would prefer to change the documentation to something like "this also provides pprof if you also have blank import in your main".

alternatively we can use a non-default mux and copy pprof's `init` here, which avoids the double registration problem. but we still have the "if upstream adds a new useful pprof endpoint we have to add the support manually here" problem.

JessicaGreben: double registration results in this panic `panic: http: multiple registrations for /debug/pprof/`.

We could use the non-default mux and copy pprof's init, but I agree @fishy I prefer to add documentation in the comment that says to "add pprof if you also have blank import in your main".  Will do that.

admin/server.go
```diff
@@ -0,0 +1,57 @@
...
+	return &Server{
+		adminAddr:     adminAddr,
+		healthCheckFn: args.HealthCheckFn,
+	}
+}
+
+type Server struct {
+	adminAddr     string
+	healthCheckFn http.HandlerFunc
+}
```
fishy, line 39: I find the names `NewServer` and `Server` too broad, like they can only handle admin requests but the name seem to imply they are the main server. I'd rather just have `func ServeAdmin(cfg AdminConfig)` and call it a day.

admin/server.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Serve starts a blocking HTTP server for internal functions:
+//    metrics       - serve /metrics for prometheus
+//    health check  - serve /health for health checking
+// Default server address is http://localhost:6060.
```
fishy, line 44: default address is actually `http://:6060`.

also, please add blank lines between paragraphs.

fishy (changes requested):

admin/server.go
kylelemons: I am in agreement about the top-level function part, but I have experience with this kind of API and the struct options ended up in pain.  We had to pivot to variadic options to stay sane.  That said, for now I think being strongly opinionated will end up better for us, since we have a small number of entrypoints.

I think we should go for
```
package httpbp
func ServeAdmin(isHealthy http.Handler)

// and

package thriftbp
func ServeAdmin()
```

to enforce the strong opinions we want.

admin/server.go
kylelemons: Not serving /debug/pprof unless you have a magic import in your main is unacceptable for this kind of "fire and forget" API.

We have a few options:
* Document it as a side-effect of importing the package
* Serve them on a different path
* Use our own servemux

I would go with the last one: make our own top-level AdminMux and init()-register the /debug/ handlers.  The init changes VERY infrequently because only special-handling profiles need to be added to it, Index dynamically lists and dispatches most new profiles that would be added.

admin/server.go
JessicaGreben: @fishy re: `NewServer` being too broad ... the entire call with package name will be `admin.NewServer()` so that is why I originally removed the `admin` from `ServeAdmin` because it seemed redundant like `admin.ServeAdmin`. I'm fine to do it either way, I don't feel strongly one way or the other.

@kylelemons  I'm interested to understand how this "kind of API and the struct opts end up a pain"?  And Im interested to understand why variadic options was an improvement?  Re: "be strongly opinionated" ... I was trying to avoid being strongly opinionated (i.e. passing `isHealthy` as an arg) since it seems it will make backwards compatibility harder in the future having a strict API.

admin/server.go
JessicaGreben: Sounds good. I will add our own AdminMux.

SuperQ: Yes, having a separate AdminMux is probably the best option. It makes sure we explicitly separate traffic to the admin port.

admin/server.go
kylelemons: Struct options are one way to handle optional parameters, but they are optimized for cases where _every_ call site has to pass some parameters because of the added syntax and cognitive overhead.  Compared to the alternatives, they're best when every call site needs to pass 2+ options.  If most call-sites need zero or one option, if option values require validation, if you have multiple use-cases, if options themselves have options, or if you want the flexibility to adjust internals, the variadic option pattern is a good choice.  If you only have a few use-cases, having specific top-level functions with non-optional parameters is even better still.

For the hole we fell into with an API like this specifically: we started out with just a "Port" field, but that turned out to be a pain to test because in tests you want to use a random port.  So we allowed passing in zero, but then we needed a way to exfiltrate the port that was chosen.  So we ended up also having a "Listener" field, but now you have this weird "what if both the listener AND the port are provided?" question.  I think similar things also happened around logging and error handling and the mux and some other configuration.  Using variadic options simplified these a lot, for example `admin.Serve(admin.WithPort(123))` vs `admin.Serve(admin.WithListener(lis))` and now you can also have `admin.Serve(admin.WithAddr("localhost:1234")` without having to have _three_ possibly-conflicting fields.

As I mentioned elsewhere though, having top-level APIs that are opinionated and purpose-built is often the. best overall.  `thrfitbp.ServeAdmin()` + `httpbp.ServeAdmin(isHealthy)` for example -- with this arrangement, there is no "optional" arguments to worry about.  We can still expose the "low level" `admin` library, but caution that most people should not use it directly.

kylelemons (changes requested):

admin/server.go
JessicaGreben: Interesting. Thanks for the extra info @kylelemons.  
So if we do `thrfitbp.ServeAdmin()` + `httpbp.ServeAdmin(isHealthy)` which then removes the need for the `AdminServerArgs struct`, then we should not allow an adminAddr as an option? Or do we want `addr` to still be configurable and in that case have:
`thrfitbp.ServeAdmin(addr)` + `httpbp.ServeAdmin(addr, isHealthy)`
?

Trying to remove the `AdminServerArgs struct` for now seems ok, but what about down the road if we end up wanting to add more options... following this pattern we could end up with more servers like ... 
```
thrfitbp.ServeAdmin(addr) 
thrfitbp.ServeAdmin(addr, newNeededOpt) 
httpbp.ServeAdmin(addr, isHealthy)
httpbp.ServeAdmin(addr, isHealthy, newNeededOpt)
httpbp.ServeAdmin(addr, newNeededOpt)
```
Is that right? Or am I misunderstanding something?

kylelemons: Listening on any address other than 0.0.0.0 is basically incorrect for kubernetes, so no I don't think providing the address is something we need.

I can't really come up with any additional configuration or customization that I think we would want here.  We have the option of exposing an admin package in the future to allow for deeper customization but I'm not really sure what we would want to let people tweak.  If we did have them though, we could add `http.ServeAdminOpts` or we could invoke the clause in the baseplate 2022 compatibility promise that "if it doesn't break any visible code, we can do it" and switch from a bare function to a func with variadic parameters, because we can relatively quickly find out if that breaks anyone (and work with them to deal with the breakage).

internal/admin/server.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const DefaultAdminAddr = ":6060"
+
+// DefaultAdminServeMux is the default ServeMux to be used for admin servers in packages like httpbp, thriftbp, etc.
+// DefaultAdminServeMux configures the following routes:
+//    metrics       - serve /metrics for prometheus
+//    profiling     - serve /debug/pprof for profiling, ref: https://pkg.go.dev/net/http/pprof
+var DefaultAdminServeMux = http.NewServeMux()
```
kylelemons, line 16: honestly just `Mux` is probably fine, though `ServeMux` would be fine by me as well.  HTTP calls it `DefaultServeMux` but in our case it's not the default, it's the only, and we don't have other kinds of muxes.

thriftbp/admin.go
```diff
@@ -0,0 +1,43 @@
...
+//    profiling     - serve /debug/pprof for profiling, ref: https://pkg.go.dev/net/http/pprof
+//
+// Default server address is admin.DefaultAdminAddr.
+//
+// This method is blocking, to prevent blocking run it as a goroutine.
+func (s *AdminServer) Serve() {
+	log.Infof("Serving admin on %s", s.addr)
+	log.Warnw("admin http serve exited", "err", http.ListenAndServe(s.addr, admin.DefaultAdminServeMux))
+	log.Info("admin returnings")
+}
```
kylelemons, line 43:
```suggestion
// ServeAdmin starts a blocking HTTP server for internal functions:
//    metrics       - serve /metrics for prometheus
//    profiling     - serve /debug/pprof for profiling, ref: https://pkg.go.dev/net/http/pprof
//
// Default server address is admin.DefaultAdminAddr.
//
// This function blocks, so it should be run as its own goroutine.
func ServeAdmin() {
	log.Warnw("thriftbp: admin serving exited", "err", admin.Serve())
}
```

httpbp/admin.go
```diff
@@ -0,0 +1,51 @@
...
+//
+// This method is blocking, to prevent blocking run it as a goroutine.
+func (s *AdminServer) Serve() {
+	if s.healthCheckFn != nil {
+		admin.DefaultAdminServeMux.HandleFunc("/health", s.healthCheckFn)
+	}
+	log.Infof("Serving admin on %s", s.addr)
+	log.Warnw("admin http serve exited", "err", http.ListenAndServe(s.addr, admin.DefaultAdminServeMux))
+	log.Info("admin returnings")
+}
```
kylelemons, line 51:
```suggestion
// ServeAdmin starts a blocking HTTP server for internal functions:
//    health        - serve /health for health checking
//    metrics       - serve /metrics for prometheus
//    profiling     - serve /debug/pprof for profiling, ref: https://pkg.go.dev/net/http/pprof
//
// Default server address is admin.DefaultAdminAddr.
//
// This function blocks, so it should be run as its own goroutine.
func ServeAdmin(healthCheck http.HandlerFunc) {
  admin.ServeMux.HandleFunc("/health", healthCheck)
	log.Warnw("thriftbp: admin serving exited", "err", admin.Serve())
}
```

internal/admin/server.go
```diff
@@ -0,0 +1,27 @@
+package admin
+
+import (
+	"net/http"
+	"net/http/pprof"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const DefaultAdminAddr = ":6060"
```
kylelemons, line 10: DefaultAddr

internal/admin/server.go
```diff
@@ -0,0 +1,27 @@
...
+
+// DefaultAdminServeMux is the default ServeMux to be used for admin servers in packages like httpbp, thriftbp, etc.
+// DefaultAdminServeMux configures the following routes:
+//    metrics       - serve /metrics for prometheus
+//    profiling     - serve /debug/pprof for profiling, ref: https://pkg.go.dev/net/http/pprof
+var DefaultAdminServeMux = http.NewServeMux()
+
+func init() {
+	// The debug/pprof endpoints follow the pattern from the init function in net/http/pprof package.
```
kylelemons, line 19: the description is good to have, but I would also (for convenience) provide a link

kylelemons (changes requested):

internal/admin/server.go
JessicaGreben: we don't have any other kinds of muxes now, but could in the future.  I don't feel strongly either way tho.

thriftbp/admin.go
JessicaGreben: @kylelemons  I'm fine with making this change, but my one concern is that people won't be able to pass in a different port to run the admin server on. This code will either make people forced to use port 6060 or modify it like this:
```
import "github.com/reddit/baseplate.go/internal/admin"
admin.DefaultAddr = ":8081"
go.ServeAdmin()
```
Modifying it like this seems fine, but if this is how we expect people to modify the port value, then I want to change the name to not say "Default".

admin/server.go
JessicaGreben: @kylelemons I'm more concerned about people wanting to specify the port, not the host.

If you really don't think we will want to add any other configuration or customization, then I can implement your suggestions to remove the `ServerArgs`, but I feel its likely people will want to add more endpoints in the future. And the endpoints could need a custom handler.

SuperQ (approved):
LGTM, Nice

httpbp/admin.go
```diff
@@ -0,0 +1,21 @@
...
+
+// ServeAdmin starts a blocking HTTP server for internal functions:
+//    health        - serve /health for health checking
+//    metrics       - serve /metrics for prometheus
+//    profiling     - serve /debug/pprof for profiling, ref: https://pkg.go.dev/net/http/pprof
+//
+// Default server address is admin.Addr.
+//
+// This function blocks, so it should be run as its own goroutine.
+func ServeAdmin(healthCheck http.HandlerFunc) {
```
fishy, line 18: if we move this into `httpbp`, things start to get trickier: we actually have [`httpbp.HandlerFunc`](https://pkg.go.dev/github.com/reddit/baseplate.go/httpbp#HandlerFunc) that's different from `http.HandlerFunc`.

I'm not saying we should just use `httpbp.HandlerFunc` here, but mixing `http.HandlerFunc` inside `httpbp` package can get really confusing.

internal/admin/server.go
```diff
@@ -0,0 +1,35 @@
...
+	// ref: https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/net/http/pprof/pprof.go;l=80
+	Mux.HandleFunc("/debug/pprof/", pprof.Index)
+	Mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	Mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	Mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	Mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+	Mux.Handle("/metrics", promhttp.Handler())
+}
```
fishy, line 30: 🔕 now that the `admin` package is under `internal`, I would slightly prefer to make `Mux` a function that returns `*http.ServeMux` and do the registrations inside the `Mux` call (over using `init`). But I don't feel strongly.

kylelemons: the point of the Mux global var is that multiple packages can share it, so we could make a package to add additional admin pages that would share it, etc.

httpbp/admin.go
kylelemons: make it func() error?

kylelemons (approved):

httpbp/admin.go
JessicaGreben: @fishy I don't have a preference what we use. You tell me whats best... either:
`func ServeAdmin(healthCheck http.HandlerFunc) {`, or
`func ServeAdmin(healthCheck httpbp.HandlerFunc) {`, or
what kyle is saying?

fishy: I don't think `func() error` works for the case to support different health check probes. I would prefer `httpbp.HandlerFunc` to reduce confusion, but we need a good way to handle the error return (e.g. we need to write the error return to `ResponseWriter` in some way, which is not something straightforward, cc @pacejackson ).

but maybe for health check purpose we can just log the error and write a standard 500 error to `ResponseWriter` should be good enough?

fishy: actually https://github.com/reddit/baseplate.go/blob/b1ba649a8929c56f48d57ed43074854082ce2c25/httpbp/handler.go#L34 is a good example on how to handle the error from `httpbp.HandlerFunc`

internal/admin/server.go
fishy: while theoretically that is true, in reality if a service provides both http and thrift endpoints, it won't actually call both `thriftbp.ServeAdmin` and `httpbp.ServeAdmin` in its main (they try to bind on the same port, for one), so I'm not sure how this sharing via global variable will actually happen in reality. (the `admin` package is under internal so services cannot use it directly)

but again, I don't feel strongly either way ¯\\\_(ツ)\_/¯

fishy (approved):

PR #484, opened 2022-02-28
Add Redistest

zack-keim:
Adds redistest implementation


konradreiche (changes requested):
Could we apply this to existing test code in this repository while we are at it? I noticed that https://github.com/reddit/baseplate.go/blob/master/redis/cache/redispipebp/init_test.go uses miniredis

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/reddit/baseplate.go/redis/cache/redisx"
+)
+
+// NewMockRedisClient sets up a mock redis cluster, client and sender
+// This should be called from TestMain since the miniredis instance created is locked
+func NewMockRedisClient(ctx context.Context, timeout time.Duration) (client redisx.Syncx, teardown func(), err error) {
+	redisCluster, err := miniredis.Run()
+	if err != nil {
+		return redisx.Syncx{}, func() {}, err
```
fishy, line 19: nit: in an error return, we would want to return `nil` for non-error returns whenever possible. you cannot return `nil` for the `client`, but you certainly can return `nil` for the `teardown` (we don't want people using it to blindly `defer teardown()` without checking for errors first).

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/alicebob/miniredis/v2"
+	"github.com/joomcode/redispipe/redis"
+	"github.com/joomcode/redispipe/redisconn"
+
+	"github.com/reddit/baseplate.go/redis/cache/redisx"
+)
+
+// NewMockRedisClient sets up a mock redis cluster, client and sender
+// This should be called from TestMain since the miniredis instance created is locked
+func NewMockRedisClient(ctx context.Context, timeout time.Duration) (client redisx.Syncx, teardown func(), err error) {
```
fishy, line 16: if this is only intended to be used by unit tests, we do not really need to return `teardown` and `err`. just pass in`tb testing.TB` as an arg (so it can be used by both tests and benchmarks), then when an error happens we can just `tb.Fatalf("Failed to ...: %v", err)`, and at the end, before returning, we can just call `tb.Cleanup(teardown)`.

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,43 @@
...
+	// Teardown closure
+	teardown = func() {
+		redisCluster.Close()
+		conn.Close()
+	}
+
+	return client, teardown, nil
+}
+
+func FlushRedis(ctx context.Context, client redisx.Syncx) error {
```
fishy, line 41: I don't see this used by `NewMockRedisClient`, and it's not explained by the PR description either.

fishy (changes requested):

redis/cache/redisx/redisxtest/redisxtest.go
zack-keim: Removing. This was a func we were using locally that got dropped in here. Agree it is probably not necessary.

redis/cache/redisx/redisxtest/redisxtest.go
zack-keim: @konradreiche that is correct. We have been following the pattern of setup in `TestMain` since places where the mock redis cluster is used tend to be places where it made sense to init there. This approach also gives the flexibility to use this func in both tests directly but also in `TestMain`. However, if we feel strongly that supporting `TestMain` doesn't have much value, then I am happy to change this.

redis/cache/redispipebp/init_test.go
```diff
@@ -53,7 +60,9 @@ func TestMain(m *testing.M) {
 	client = redisx.BaseSync{
```
fishy, line 60: this value of `client` is no longer used by anything?

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Close shuts down the MockRedisCluster
+func (mrc *MockRedisCluster) Close() {
+	mrc.redisCluster.Close()
+}
+
+// NewMockRedisClient sets up a client and sender to a mock redis cluster
+func NewMockRedisClient(
+	ctx context.Context,
+	redisCluster MockRedisCluster,
```
fishy, line 48: this feels like an over-design to me.

if in most tests we only need a single, global, mocked cluster and client, then we don't really need separated `NewMockRedisCluster` and `NewMockRedisClient`, a single `NewMockRedisClient` with its returned `teardown` also tears down the cluster should be sufficient (the cluster does not need to be exposed to the caller at all).

but if that's not the case, then I think only the cluster itself should be global (that's handled inside `TestMain`), the clients should be created per test, so my previous comment regarding passing in `testing.TB` and no need to return `error` nor `teardown` applies.

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,70 @@
...
+	opts redisconn.Opts,
+) (client redisx.BaseSync, teardown func(), err error) {
+
+	// Create connection
+	conn, err := redisconn.Connect(ctx, redisCluster.Addr(), opts)
+	if err != nil {
+		redisCluster.Close()
```
fishy, line 55: this is a very unexpected side effect. if a client cannot connect to the server, it will also kill the server, causing other, clients to stop working.

fishy (changes requested):

redis/cache/redispipebp/init_test.go
zack-keim: Removed.

redis/cache/redisx/redisxtest/redisxtest.go
zack-keim: Removed.

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,69 @@
...
+	return mockRedisCluster, teardown, nil
+}
+
+// Addr returns address of mock redis cluster e.g. '127.0.0.1:12345'.
+func (mrc *MockRedisCluster) Addr() string {
+	return mrc.redisCluster.Addr()
+}
+
+// Close shuts down the MockRedisCluster
+func (mrc *MockRedisCluster) Close() {
```
fishy, line 41: nit: make this function to return an error, so it implements [`io.Closer`](https://pkg.go.dev/io#Closer) interface

since `miniredis.Miniredis.Close` does not return an error, you can just always `return nil` inside this function.

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,69 @@
...
+		return MockRedisCluster{}, nil, err
+	}
+
+	mockRedisCluster = MockRedisCluster{
+		redisCluster: redisCluster,
+	}
+
+	teardown = func() {
+		mockRedisCluster.Close()
+	}
```
fishy, line 30: this is a very standard closing pattern in go (especially when you make `MockRedisCluster` to implement `io.Closer`), so I would say returning a `teardown` is unnecessary, caller could just do:
```go
cluster, err := NewMockRedisCluster()
if err != nil { ... }
defer cluster.Close()
```

redis/cache/redisx/redisxtest/redisxtest.go
fishy: and looking at this function again I don't see it doing anything special. all it does are pretty standard stuff about create a redis connection (you actually only need the address passed in, not the whole cluster object) and create a redis client out of this connection. the `teardown` is also standard stuff about closing the client, so I don't really see much value of providing (and maintaining) this as part of the public API.

If we think a `func NewClient(ctx, addr, opts) (redisx.BaseSync, error)` API is useful, then that should be part of `redisx`, not `redisxtest`.

fishy (changes requested):

redis/cache/redisx/redisxtest/redisxtest.go
zack-keim: Removed

redis/cache/redisx/redisxtest/redisxtest.go
zack-keim: Reason for the change to have separated `NewMockRedisCluster` and `NewMockRedisClient` was to support the cases like that in `TestMain` in `init_test`, where it looks like we need a global cluster. Perhaps an alternative solution to this would be to have the following API for redisxtest:

```
func NewMockRedisClient(tb, ctx) (client, error)
func NewMockRedisClientWithCluster(tb, ctx, clusterAddr) (client, error)
```
These functions would handle cleanup and client creation via tb as you have suggested. 

I am really not opposed at all to going down the path updating the redisx API to have a `func NewClient()`. This is actually our preferred route. We saw adding `redisxtest` as being minimally invasive, which was the reason for going this path. If you are ok with us doing that, happy to open another PR for that change and close potentially close this one if we see no value in what I have proposed.

konradreiche (approved):

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,64 @@
...
+	mrc.redisCluster.Close()
+	return nil
+}
+
+// NewMockRedisClient sets up a client and sender to a mock redis cluster
+func NewMockRedisClient(
+	ctx context.Context,
+	address string,
+	opts redisconn.Opts,
+) (client redisx.BaseSync, teardown func(), err error) {
```
fishy, line 45: I would still recommend change this function signature into:
```
func NewMockRedisClient(ctx context.Context, tb testing.TB, address string, opts redisconn.Opts) redisx.BaseSync
```

it's OK for the mocked cluster to be a global singleton that's initialized in `TestMain` and shared between tests, because a mocked cluster is probably using much more resource and stand it up and tear it down in every test case is probably too slow/wasteful, but for clients, the pattern we should promote should be that every unit test create it and clean it up, instead of doing it in `TestMain`.

fishy (changes requested):

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// Create client
+	client := redisx.BaseSync{
+		SyncCtx: redis.SyncCtx{S: conn},
+	}
+
+	// Cleanup
+	tb.Cleanup(func() {
+		conn.Close()
+	})
```
fishy, line 36: nit: I would move this up to right after the `err` check (to line 27).

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+
+	"github.com/joomcode/redispipe/redis"
+	"github.com/joomcode/redispipe/redisconn"
+
+	"github.com/reddit/baseplate.go/redis/cache/redisx"
+)
+
+// NewMockRedisClient sets up a redis client for use in testing
+func NewMockRedisClient(
```
fishy, line 14: now there's nothing really "mocking" here (the mocking part is done via miniredis, which is no longer done here).

I would rename this function to `NewTestRedisClient`, and document that:

1. the `address` arg should be coming from miniredis (we should also provide an example on how to it, see https://pkg.go.dev/testing#hdr-Examples, we also have prior arts in baseplate.go, for example https://github.com/reddit/baseplate.go/blob/master/errorsbp/batch_size_example_test.go)
2. we should also document that the returned client will be auto released at the end of test and there's no need to do that explicitly (because the caller actually still have access to that via `client.SyncCtx.S.Close()`

redis/cache/redisx/redisxtest/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package redisxtest provides utilities for baseplate.go/redids/cache/redisx testing.
```
fishy, line 1: no need to spell out the full `baseplate.go/redids/cache/redisx`, just say "... for redisx testing" is enough.

redis/cache/redisx/redisxtest/doc.go
```diff
@@ -0,0 +1,4 @@
...
+//
+// This includes functions for mocking redis instances.
```
fishy, line 3: same comment here regarding "mocking". I think with the current state of this change the first paragraph is enough for the package doc and we can just remove this paragraph.

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,39 @@
...
+	ctx context.Context,
+	tb testing.TB,
+	address string,
+	opts redisconn.Opts,
+) (*redisx.BaseSync, error) {
```
fishy, line 19: looking at how existing tests work, we do need the returned client to be of type `redisx.BaseSync` instead of the pointer, for example https://github.com/reddit/baseplate.go/blob/a6f0c8aa515567b05b5da3820b3aa879d33a1ce4/redis/cache/redispipebp/errors_test.go#L57

fishy (changes requested):
we would still want to change the existing tests to use this, that includes:

1. instead of store the client globally (https://github.com/reddit/baseplate.go/blob/a6f0c8aa515567b05b5da3820b3aa879d33a1ce4/redis/cache/redispipebp/init_test.go#L19), we store the miniredis address globally
2. in tests currently using the global client, use this function to get the client instead
3. change [`flushRedis`](https://github.com/reddit/baseplate.go/blob/a6f0c8aa515567b05b5da3820b3aa879d33a1ce4/redis/cache/redispipebp/init_test.go#L61) to take tb and client as args.

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,39 @@
...
+	tb.Helper()
+
+	// Create connection to redis
+	conn, err := redisconn.Connect(ctx, address, opts)
+	if err != nil {
+		return nil, err
+	}
```
fishy, line 26: oh and this function no longer need to return an error. this can be just:
```suggestion
	if err != nil {
		tb.Fatalf("Failed to create redis client: %v", err)
	}
```

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,39 @@
...
+	tb testing.TB,
+	address string,
+	opts redisconn.Opts,
+) (*redisx.BaseSync, error) {
+	tb.Helper()
+
+	// Create connection to redis
+	conn, err := redisconn.Connect(ctx, address, opts)
+	if err != nil {
+		return nil, err
```
kylelemons, line 25: If this is a testing helper, in my opinion it's better to call t.Fatalf and document that it must be called directly in a test / subtest, rather than do what is inevitable: have every user of this make their own _local_ helper that wraps _this_ helper that does the transition.  I prefix my t.Fatalf errors with "SETUP:" to make it clear that it's not a failure of the code under test, and I would name the function "Setup" not "New" to encourage that interpretation as well.

redis/cache/redisx/redisxtest/redisxtest.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	// Create client
+	client := redisx.BaseSync{
+		SyncCtx: redis.SyncCtx{S: conn},
+	}
+
+	// Cleanup
+	tb.Cleanup(func() {
+		conn.Close()
```
kylelemons, line 35: check the error and log it if it fails?

kylelemons (changes requested):

PR #486, opened 2022-03-02
Relax interface for bpError in the Prometheus middlewares

nanassito:
Right now we expect the user to have the same definition of `baseplate.thrift` than we do but nothing enforces it. In the case of the Prometheus middleware this leads to us not recognizing an error as being a Baseplate.Error and therefore not adding the labels for the error code.

This PR introduces a relaxed interface only to be used by the prometheus middleware to ensure that they will identity a Baseplate.Error as such regardless of past or future changes to the `baseplate.thrift` file.

### DIFF
--- a/thriftbp/errors.go
+++ b/thriftbp/errors.go
@@ -32,6 +32,15 @@ type baseplateError interface {
 	GetDetails() map[string]string
 }
 
+// We create this interface instead of using baseplateError because we want it to always match
+// regardless of the version of the thrift file the user has put in their repository.
+type baseplateErrorCoder interface {
+	thrift.TException
+
+	IsSetCode() bool
+	GetCode() int32
+}
+
 var (
 	_ baseplateError = (*baseplatethrift.Error)(nil)
 )
@@ -69,7 +78,7 @@ func BaseplateErrorFilter(codes ...int32) retrybp.Filter {
 		codeMap[code] = true
 	}
 	return func(err error, next retry.RetryIfFunc) bool {
-		var bpErr baseplateError
+		var bpErr baseplateErrorCoder
 		if errors.As(err, &bpErr) {
 			return codeMap[bpErr.GetCode()]
 		}
@@ -87,7 +96,7 @@ func IDLExceptionSuppressor(err error) bool {
 	if !errors.As(err, &te) {
 		return false
 	}
-	var bpErr baseplateError
+	var bpErr baseplateErrorCoder
 	if errors.As(err, &bpErr) {
 		// If this is also baseplate.Error,
 		// only suppress it if the error code is outside of [500, 600).
--- a/thriftbp/prometheus.go
+++ b/thriftbp/prometheus.go
@@ -255,3 +255,12 @@ func stringifyErrorType(err error) string {
 	}
 	return strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
 }
+
+// We create this interface instead of using baseplateError because we want it to always match
+// regardless of the version of the thrift file the user has put in their repository.
+type baseplateErrorCode interface {
+	thrift.TException
+
+	IsSetCode() bool
+	GetCode() int32
+}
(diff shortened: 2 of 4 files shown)


fishy (approved):

thriftbp/prometheus.go
```diff
@@ -255,3 +255,12 @@ func stringifyErrorType(err error) string {
 	}
 	return strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
 }
+
+// We create this interface instead of using errors.go/Error because we want it
+//to always match the Baseplate.Error regardless of the version the user has put in their repository.
+type baseplateErrorMinimalIface interface {
```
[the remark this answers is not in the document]
nanassito: `minimalBaseplateError` then in case we later add the retriable.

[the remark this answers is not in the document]
fishy: I actually slightly prefer `baseplateErrorCode` between the 2 names I suggested, because this is more specific (we only need code for prometheus purpose, so the name reflects that, while the term "minimal" is quite ambiguous).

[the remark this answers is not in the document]
nanassito: agreed, updating again

thriftbp/prometheus.go
```diff
@@ -255,3 +255,12 @@ func stringifyErrorType(err error) string {
 	}
 	return strings.TrimPrefix(fmt.Sprintf("%T", err), "*")
 }
+
+// We create this interface instead of using errors.go/Error because we want it
```
fishy, line 259: I think you meant to say "baseplateError" instead of "errors.go/Error"? I'm not sure what "errors.go/Error" is.

thriftbp/prometheus.go
```diff
@@ -255,3 +255,12 @@ func stringifyErrorType(err error) string {
...
+//to always match the Baseplate.Error regardless of the version the user has put in their repository.
```
fishy, line 260: nits:
1. add a space after `//`
2. don't capitalize `Baseplate` here because `baseplate.Error` means type `Error` from package `baseplate` (and we don't capitalize package names in go) :)

fishy (approved):

thriftbp/prometheus.go
nanassito: updated

thriftbp/prometheus.go
nanassito: updated

nanassito:
flaky test ? It passes locally and appears unrelated to the change here

fishy (approved):

fishy:
> flaky test ? It passes locally and appears unrelated to the change here

looks so. https://github.com/reddit/baseplate.go/pull/487

fishy:
Could you please also change [`IDLExceptionSuppressor`](https://pkg.go.dev/github.com/reddit/baseplate.go/thriftbp#IDLExceptionSuppressor) and [`BaseplateErrorFilter`](https://pkg.go.dev/github.com/reddit/baseplate.go/thriftbp#BaseplateErrorFilter) to use `baseplateErrorCode` instead of `baseplateError`.

fishy (approved):

thriftbp/errors.go
```diff
@@ -32,6 +32,15 @@ type baseplateError interface {
 	GetDetails() map[string]string
 }
 
+// We create this interface instead of using baseplateError because we want it to always match
+// regardless of the version of the thrift file the user has put in their repository.
+type baseplateErrorCode interface {
```
kylelemons, line 37: It's not an error code, it's a thing that gives you codes, so `baseplateErrorCoder` (reader, writer, etc)

kylelemons (approved):
Ah, indeed! I think this worked before because I actually wrote it inside the service.

I'm hopeful we can avoid these issues structurally in gRPC land.

thriftbp/errors.go
nanassito: renamed (again)

nanassito:
I don't have merge permission in here so I'll need one of you to merge it for me if that's ok

konradreiche (approved):

PR #487, opened 2022-03-02
httpbp: Fix flaky test

fishy:
https://github.com/reddit/baseplate.go/runs/5384488436?check_suite_focus=true

This is likely caused by the previous timeout was too short and on some
slow runs it did not have enough time to react.


konradreiche (approved):

kylelemons (approved):
aw, I was expecting an actual flakiness fix, not a 1% to .001% flakiness change ;-)

PR #488, opened 2022-03-02
Add gRPC baseplate server implementation

konradreiche:
This adds the neccessary code to have a baseplate service backed by a
gRPC server.

ServerConfig can be used to configure specific keepalive settings that
are used for point-to-point healthchecks, otherwise defaults are used.
It is also used to pass in the concrete gRPC service implementation
since gRPC service implementations do not share a common interface.

The baseplate server Close method is implemented using GracefulStop,
this may need to change to grpc.Server#Stop for an immediate shutdown.

### DIFF
--- a/grpcbp/server.go
+++ b/grpcbp/server.go
@@ -34,60 +34,80 @@ type ServerConfig struct {
 	// pings the client to see if the transport is still alive. If set below 1s,
 	// a minimum value of 1s will be used instead. The current default value is 2
 	// hours.
-	Time time.Duration `yaml:"time"`
+	PingInactiveClientsAfter time.Duration `yaml:"pingInactiveClientsAfter"`
 
 	// After having pinged for keepalive check, the server waits for a duration
-	// of Timeout and if no activity is seen even after that the connection is
+	// of KeepAliveTimeout and if no activity is seen even after that the connection is
 	// closed. The current default value is 20 seconds.
-	Timeout time.Duration `yaml:"timeout"`
+	KeepAliveTimeout time.Duration `yaml:"keepAliveTimeout"`
 
-	// MinTime is the minimum amount of time a client should wait before sending
+	// KeepAliveMinTime is the minimum amount of time a client should wait before sending
 	// a keepalive ping.
-	MinTime time.Duration `yaml:"minTime"`
+	KeepAliveMinTime time.Duration `yaml:"keepAliveMinTime"`
 
 	// If true, server allows keepalive pings even when there are no active
 	// streams(RPCs). If false, and client sends ping when there are no active
 	// streams, server will send GOAWAY and close the connection.
-	PermitWithoutStream bool `yaml:"permitWithoutStream"`
+	KeepAlivePermitWithoutStream bool `yaml:"keepAlivePermitWithoutStream"`
+}
 
-	// RegisterServerFunc is used to pass in a generated gRPC service
-	// implementation and register it on the created gRPC.Server.
-	RegisterServerFunc func(*grpc.Server) `yaml:"-"`
+// Validate ensures all set values are within bound.
+func (s ServerConfig) Validate() error {
+	return nil
+}
+
+// GRPCServer provides passing in the generated gRPC service implementation and
+// reigster it on the created gRPC.Server.
+type GRPCServer interface {
+	RegisterOn(*grpc.Server)
 }
 
 // NewBaseplateServer returns a new gRPC implementation of a Baseplate server
 // with the given config.
-func NewBaseplateServer(bp baseplate.Baseplate, cfg ServerConfig) (baseplate.Server, error) {
+func NewBaseplateServer(bp baseplate.Baseplate, server GRPCServer, cfg ServerConfig) (baseplate.Server, error) {
 	lis, err := net.Listen("tcp", bp.GetConfig().Addr)
 	if err != nil {
 		return nil, err
 	}
+	grpcServer, err := NewServer(bp, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	server.RegisterOn(grpcServer)
+
+	return ApplyBaseplate(bp, grpcServer, lis), nil
+}
+
+// NewServer returns a new instance of grpc.Server with any baseplate-related
+// server options.
+func NewServer(bp baseplate.Baseplate, cfg ServerConfig) (*grpc.Server, error) {
+	if err := cfg.Validate(); err != nil {
+		return nil, err
+	}
 
 	kaep := keepalive.EnforcementPolicy{
-		MinTime:             cfg.MinTime,
-		PermitWithoutStream: cfg.PermitWithoutStream,
+		MinTime:             cfg.KeepAliveMinTime,
+		PermitWithoutStream: cfg.KeepAlivePermitWithoutStream,
 	}
 
 	kasp := keepalive.ServerParameters{
 		MaxConnectionIdle:     cfg.MaxConnectionIdle,
 		MaxConnectionAge:      cfg.MaxConnectionAge,
 		MaxConnectionAgeGrace: cfg.MaxConnectionAgeGrace,
-		Time:                  cfg.Time,
-		Timeout:               cfg.Timeout,
+		Time:                  cfg.PingInactiveClientsAfter,
+		Timeout:               cfg.KeepAliveTimeout,
 	}
 
 	middlewares := BaseplateDefaultMiddlewares(DefaultMiddlewaresArgs{
 		EdgeContextImpl: bp.EdgeContextImpl(),
 	})
 
-	server := grpc.NewServer(
+	return grpc.NewServer(
 		grpc.KeepaliveEnforcementPolicy(kaep),
 		grpc.KeepaliveParams(kasp),
 		middlewares,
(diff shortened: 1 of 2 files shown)


grpcbp/server.go
```diff
@@ -0,0 +1,128 @@
...
+	// MaxConnectionIdle is a duration for the amount of time after which an idle
+	// connection would be closed by sending a GoAway. Idleness duration is
+	// defined since the most recent time the number of outstanding RPCs became
+	// zero or the connection establishment. The current default value is
+	// infinity.
+	MaxConnectionIdle time.Duration `yaml:"maxConnectionIdle"`
+
+	// MaxConnectionAge is a duration for the maximum amount of time a
+	// connection may exist before it will be closed by sending a GoAway. A
+	// random jitter of +/-10% will be added to MaxConnectionAge to spread out
```
fishy, line 24: I'm not sure adding the jitter on the server side is useful.. unless the client will auto reconnect once it lose the connection.

grpcbp/server.go
```diff
@@ -0,0 +1,128 @@
...
+	// MaxConnectionAgeGrace is an additive period after MaxConnectionAge after
+	// which the connection will be forcibly closed. The current defualt value is
+	// infinity.
+	MaxConnectionAgeGrace time.Duration `yaml:"maxConnectionAgeGrace"`
+
+	// After a duration of this time if the server doesn't see any activity it
+	// pings the client to see if the transport is still alive. If set below 1s,
+	// a minimum value of 1s will be used instead. The current default value is 2
+	// hours.
+	Time time.Duration `yaml:"time"`
```
fishy, line 37: `Time` feels like a very confusing/non-descriptive name for it. Maybe `KeepaliveTime`?

grpcbp/server.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// After having pinged for keepalive check, the server waits for a duration
+	// of Timeout and if no activity is seen even after that the connection is
+	// closed. The current default value is 20 seconds.
+	Timeout time.Duration `yaml:"timeout"`
+
+	// MinTime is the minimum amount of time a client should wait before sending
+	// a keepalive ping.
+	MinTime time.Duration `yaml:"minTime"`
```
fishy, line 46: same here, maybe `MinKeepaliveTime`?

grpcbp/server.go
kylelemons: From my understanding, it will.

grpcbp/server.go
kylelemons: I agree the name is confusing.  KeepaliveTime is fine with me.  Also could go with a more descritive name like PingQuietClientsAfter.

grpcbp/server.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// If true, server allows keepalive pings even when there are no active
+	// streams(RPCs). If false, and client sends ping when there are no active
+	// streams, server will send GOAWAY and close the connection.
+	PermitWithoutStream bool `yaml:"permitWithoutStream"`
+
+	// RegisterServerFunc is used to pass in a generated gRPC service
+	// implementation and register it on the created gRPC.Server.
+	RegisterServerFunc func(*grpc.Server) `yaml:"-"`
```
kylelemons, line 55: I recommend defining an interface, called `GRPCService` with `RegisterOn(*grpc.Server)` as its only method and using that.

We can make an HandlerFunc adapter for it, but it actually turns out to be super nice to just pass the object itself as the GRPCService

grpcbp/server.go
```diff
@@ -0,0 +1,128 @@
...
+	middlewares := BaseplateDefaultMiddlewares(DefaultMiddlewaresArgs{
+		EdgeContextImpl: bp.EdgeContextImpl(),
+	})
+
+	server := grpc.NewServer(
+		grpc.KeepaliveEnforcementPolicy(kaep),
+		grpc.KeepaliveParams(kasp),
+		middlewares,
+	)
+	cfg.RegisterServerFunc(server)
```
kylelemons, line 88: Hiding all of this away makes it really hard to use in tests.

I would have a top-level function that takes in your config and returns a grpc.Server

grpcbp/server.go
```diff
@@ -0,0 +1,128 @@
...
+// will take care of this for you.
+func ApplyBaseplate(bp baseplate.Baseplate, server *grpc.Server, lis net.Listener) baseplate.Server {
+	return impl{
+		bp:  bp,
+		lis: lis,
+		srv: server,
+	}
+}
+
+type impl struct {
```
kylelemons, line 106: "impl" is a no-name name.  "server" or "grpcServer" would be better names for this, since it's implementing baseplate.Server

kylelemons (changes requested):

grpcbp/server.go
konradreiche: Would this change the signature of the constructor though? Trying to figure out what type of testing we want to make more accessible. I'd think generally we want to be able to have users define their own server options altogether.

grpcbp/server.go
konradreiche: Maybe `PingInactiveClientsAfter`? The official names are all very misleading but I prefixed any keepalive related configuration with `KeepAlive` hoping that it makes it more clear.

grpcbp/server.go
```diff
@@ -0,0 +1,148 @@
...
+	// If true, server allows keepalive pings even when there are no active
+	// streams(RPCs). If false, and client sends ping when there are no active
+	// streams, server will send GOAWAY and close the connection.
+	KeepAlivePermitWithoutStream bool `yaml:"keepAlivePermitWithoutStream"`
+}
+
+// Validate ensures all set values are within bound.
+func (s ServerConfig) Validate() error {
+	return nil
+}
```
konradreiche, line 57: I wanted to make sure there is a validate method from the start but I believe the only meaningful validation is to check if any durations are negative.

grpcbp/server.go
```diff
@@ -0,0 +1,148 @@
+package grpcbp
+
+import (
+	"net"
+	"time"
+
+	"github.com/reddit/baseplate.go"
```
fishy, line 7: nit: this should be in the 3rd import group.

grpcbp/server.go
fishy: then maybe only add this api when it's meaningful?

adding an api in the future is not a breaking change, while removing an api in the future is.

grpcbp/server.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// Validate ensures all set values are within bound.
+func (s ServerConfig) Validate() error {
+	return nil
+}
+
+// GRPCServer provides passing in the generated gRPC service implementation and
+// reigster it on the created gRPC.Server.
+type GRPCServer interface {
```
fishy, line 61: nit: since the name of this package is already `grpcbp`, `grpcbp.GRPCServer` feels redundant, we should rename this to just `Server` so it's `grpcbp.Server`.

grpcbp/server.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	server.RegisterOn(grpcServer)
+
+	return ApplyBaseplate(bp, grpcServer, lis), nil
+}
+
+// NewServer returns a new instance of grpc.Server with any baseplate-related
+// server options.
+func NewServer(bp baseplate.Baseplate, cfg ServerConfig) (*grpc.Server, error) {
```
fishy, line 84: maybe it's my lack of context around grpc, but `NewBaseplateServer` vs. `NewServer` confuses me here, even with the function doc.

is it that we are supposed to first call `grpcbp.NewServer` then pass that into `grpcbp.NewBaseplateServer`?

fishy (changes requested):

grpcbp/server.go
konradreiche: I'll have to defer to what @kylelemons had in mind. The way I changed it made it more confusing. The goal here is to offer a way for the service-specific code to register their methods. For example for this here https://grpc.io/docs/languages/go/basics/#starting-the-server it would be:

```
pb.RegisterRouteGuideServer(grpcServer, newServer())
```
and since baseplate.go takes care of instantiating `*grpc.Server` we need to expose this instantiated type somehow so that the baseplate user can register their concrete types. Previously I had this in the server config, another way could look like this:

```go
func NewBaseplateServer(bp baseplate.Baseplate, registerFunc func(*grpc.Server), cfg ServerConfig) (baseplate.Server, error) {
	lis, err := net.Listen("tcp", bp.GetConfig().Addr)
	if err != nil {
		return nil, err
	}
	srv, err := NewServer(bp, cfg)
	if err != nil {
		return nil, err
	}

        registerFunc(srv)

	return ApplyBaseplate(bp, srv, lis), nil
}
```
But I believe Kyle was thinking about some specific using an interface type :slightly_smiling_face:

grpcbp/server.go
kylelemons: PingInactiveClientsAfter sounds fine, though I went with "Quiet" since it seems a bit clearer than "Inactive" (what counts as active?) but this is just bikeshedding.

konradreiche:
Closing in favor of a different API approach :clap:

PR #489, opened 2022-03-03
🐛 | Swap Prometheus and recoverPanic middlewares

nanassito:
Right now PrometheusMiddleware is the leaf middleware. This means that its defer function doesn't see panics as being a wrapped error. So it logs them as successes. Making the RecoverPanic middleware the leaf appears to solve this problem

### DIFF
--- a/thriftbp/server_middlewares.go
+++ b/thriftbp/server_middlewares.go
@@ -69,9 +69,9 @@ type DefaultProcessorMiddlewaresArgs struct {
 //
 // 5. ReportPayloadSizeMetrics
 //
-// 6. RecoverPanic
+// 6. PrometheusServerMiddleware
 //
-// 7. PrometheusServerMiddleware
+// 7. RecoverPanic
 func BaseplateDefaultProcessorMiddlewares(args DefaultProcessorMiddlewaresArgs) []thrift.ProcessorMiddleware {
 	return []thrift.ProcessorMiddleware{
 		ExtractDeadlineBudget,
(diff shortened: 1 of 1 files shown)


kylelemons (approved):

kylelemons:
This makes sense to me, though the panic middleware also sets the `panicRecoverCounter` metric.

I woner if we should have `success: true/false/panic` in the prometheus middleware? (which would mean that it should go back after recoverpanic).  probably not...

nanassito:
I would prefer `panic` be treated as the error class.

That said I would prefer the success label to not exist and be a result label instead that contains "success" or the error class.

nanassito:
but I think this change makes it so that in case of panic we will label as `thrift.wrappedTException` which isn't what we want here

nanassito:
as for the `panicRecoverCounter` I can see it for WF but not Prometheus

fishy:
> This makes sense to me, though the panic middleware also sets the `panicRecoverCounter` metric.
> 
> I woner if we should have `success: true/false/panic` in the prometheus middleware? (which would mean that it should go back after recoverpanic). probably not...

I actually think this is an opportunity in the prometheus world: since we also label error types in prometheus, instead of having a separated counter, we can just give it an unique error type and we get counters and timers out of the standard prometheus metrics for free

but that certainly are out of the scope of this pr :)

fishy (approved):

nanassito:
I think we should at least label as `panic` as part of this PR

nanassito:
ah it will be logged as the type that was passed in the panic

fishy:
> but I think this change makes it so that in case of panic we will label as `thrift.wrappedTException` which isn't what we want here

no it should never be labeled as that one.

[inside `RecoverPanic`](https://github.com/reddit/baseplate.go/blob/9770a2ef0bc4f95e16a771d6beabfd3b62c7fa3b/thriftbp/server_middlewares.go#L374-L378), we either return the original error as-is (if `panic()` was called on an error), or use `fmt.Errorf`.

nanassito:
yeah my tests didn't have your fix on that applied yet that's why I didn't see it.  Should we still wrap in a new `Panic` error type ?

fishy:
> yeah my tests didn't have your fix on that applied yet that's why I didn't see it. Should we still wrap in a new `Panic` error type ?

that should be done in a separated PR (if we decided to do it, that is)

nanassito:
Okay in that case can one of you merge this ?

PR #490, opened 2022-03-03
Bump thrift dependency to v0.16.0

fishy:
Also rerun thrift compiler v0.16.0 against baseplate.thrift.

Also while I'm here, update CI to use go 1.17.7.


kylelemons (approved):

fishy:
actually go 1.17.8 was just released today so bumping that to 1.17.8 again 😅

PR #491, opened 2022-03-04
runtimebp: Clearly log the `GOMAXPROCS` configuration on startup

jstrate-reddit:
Logs like `GOMAXPROCS: 48 4` aren't very clear.  Identifying the configuration values in the default
and updated states is useful to the operator.

### DIFF
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -37,5 +37,5 @@ func InitFromConfig(cfg Config) {
 		min = cfg.NumProcesses.Min
 	}
 	prev, current := GOMAXPROCS(min, max)
-	fmt.Fprintf(os.Stderr, "GOMAXPROCS: %d %d\n", prev, current)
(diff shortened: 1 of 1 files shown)


fishy (approved):

runtimebp/config.go
```diff
@@ -37,5 +37,5 @@ func InitFromConfig(cfg Config) {
 		min = cfg.NumProcesses.Min
 	}
 	prev, current := GOMAXPROCS(min, max)
-	fmt.Fprintf(os.Stderr, "GOMAXPROCS: %d %d\n", prev, current)
+	fmt.Fprintf(os.Stderr, "GOMAXPROCS: previous: %d current: %d\n", prev, current)
```
fishy, line 40: 🔕 or simply old/new (we actually wanted to use `old`/`new` as the variable names instead of `prev`/`current`, but `new` is a keyword in go so we have to pick our second choices 😅)

kylelemons (approved):

PR #492, opened 2022-03-04
metricsbp|prometheusbp: Add BoolString

fishy:
Currently in Baseplate spec, we want different string values for bool
values in statsd vs. prometheus metric tags/labels: In statsd we want
capitalized "True"/"False" while in prometheus we want "true"/"false".
This can be confusing and lead to inconsistency, especially when
services are transitioning from statsd to prometheus metrics.

Provide metricsbp.BoolString and prometheusbp.BoolString in hope to help
reduce the confusion. For devs it should be easier to:

* If I'm using metricsbp to emit a metric, use metricsbp.BoolString
* If I'm using prometheus to emit a metric, use prometheusbp.BoolString

### DIFF
--- a/grpcbp/client_middlewares.go
+++ b/grpcbp/client_middlewares.go
@@ -3,7 +3,6 @@ package grpcbp
 import (
 	"context"
 	"errors"
-	"strconv"
 	"time"
 
 	"github.com/opentracing/opentracing-go"
@@ -12,6 +11,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/reddit/baseplate.go/ecinterface"
+	"github.com/reddit/baseplate.go/prometheusbp"
 	"github.com/reddit/baseplate.go/tracing"
 )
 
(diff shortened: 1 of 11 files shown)


JessicaGreben (commented):
why is the baseplate spec different? Can we not standardize on either all being "True" or "true" instead of needing to support both?

fishy:
> why is the baseplate spec different? Can we not standardize on either all being "True" or "true" instead of needing to support both?

we had the spec for statsd metrics first, at the time we picked capitalized "True"/"False" likely because that's how python generates strings from booleans.

I don't know exactly why did we pick `true`/`false` for prometheus metrics spec, but I guess would be that's from already established prometheus convention, or something like that.

kylelemons:
> I don't know exactly why did we pick true/false for prometheus metrics spec, but I guess would be that's from already established prometheus convention, or something like that.

Yep.  lower-case true/false is what existing prom metrics will almost universally use.

kylelemons (approved):

PR #494, opened 2022-03-07
thriftbp: Minor unit test improvement on clientPoolGaugeExporter

fishy:
Instead of using a fixed buffer size of 2 for the channel, spawn a
goroutine to drain the channel. This way the unit test is more robust
against future changes to the Collect implementation.

### DIFF
--- a/thriftbp/prometheus_test.go
+++ b/thriftbp/prometheus_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/apache/thrift/lib/go/thrift"
@@ -153,14 +154,47 @@ func (fakePool) NumAllocated() int32 {
 	return 2
 }
 
-func TestClientPoolGaugeExporter(t *testing.T) {
+func TestClientPoolGaugeExporterRegister(t *testing.T) {
+	// This test is to make sure that a service creating more than one thrift
+	// client pool will not cause issues in prometheus metrics.
+	exporters := []clientPoolGaugeExporter{
+		{
+			slug: "foo",
+			pool: fakePool{},
+		},
+		{
+			slug: "bar",
+			pool: fakePool{},
+		},
+	}
+	for i, exporter := range exporters {
+		if err := prometheus.Register(exporter); err != nil {
+			t.Errorf("Register #%d failed: %v", i, err)
+		}
+	}
+}
+
+func TestClientPoolGaugeExporterCollect(t *testing.T) {
 	exporter := clientPoolGaugeExporter{
 		slug: "slug",
 		pool: fakePool{},
 	}
+	ch := make(chan prometheus.Metric)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		// Drain the channel
+		for range ch {
+		}
+	}()
+	t.Cleanup(func() {
+		close(ch)
+		wg.Wait()
+	})
 	// No real test here, we just want to make sure that Collect call will not
 	// panic, which would happen if we have a label mismatch.
-	exporter.Collect(make(chan prometheus.Metric, 2))
+	exporter.Collect(ch)
 }
 
 // This is an thrift.TException implementation that does not wrap any error.


thriftbp/prometheus_test.go
```diff
@@ -158,9 +159,22 @@ func TestClientPoolGaugeExporter(t *testing.T) {
 		slug: "slug",
 		pool: fakePool{},
 	}
+	ch := make(chan prometheus.Metric)
```
kylelemons, line 162: I would invert this.  It lets the main test goroutine handle all of the coordination and validation and thus you don't need a waitgroup.

```
done := make(chan interface{}) // capture panic value or nil if no panic
go func() {
  defer func() {
    done <- recover()
  }()
  defer close(ch)
  exporter.Collect(ch)
}
for _, def := range ch {
  t.Logf(...) // so humans can sanity check
}
if r := <-done; r != nil {
  t.Errorf("Collect panic: %v", r)
}
```

kylelemons (changes requested):

thriftbp/prometheus_test.go
fishy: also skip go testing's auto panic handling means that we are actually losing the line number where the panic happens. this is the current test code when I make the labels inconsistent:
```
 go test
--- FAIL: TestClientPoolGaugeExporter (0.00s)
panic: inconsistent label cardinality: expected 1 label values but got 0 in []string(nil) [recovered]
        panic: inconsistent label cardinality: expected 1 label values but got 0 in []string(nil)

goroutine 197 [running]:
testing.tRunner.func1.2({0x8eee60, 0xc000516890})
        /gnu/store/ayhq6nzj64sp0rh8xqrfjjis88zpz393-go-1.17.7/lib/go/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
        /gnu/store/ayhq6nzj64sp0rh8xqrfjjis88zpz393-go-1.17.7/lib/go/src/testing/testing.go:1212 +0x218
panic({0x8eee60, 0xc000516890})
        /gnu/store/ayhq6nzj64sp0rh8xqrfjjis88zpz393-go-1.17.7/lib/go/src/runtime/panic.go:1038 +0x215
github.com/prometheus/client_golang/prometheus.MustNewConstMetric(...)
        /home/fishy/.gopath/pkg/mod/github.com/prometheus/client_golang@v1.11.0/prometheus/value.go:107
github.com/reddit/baseplate.go/thriftbp.clientPoolGaugeExporter.Collect({{0x98a087, 0xc00031c660}, {0xa44a50, 0xc000516850}}, 0xc0002f6f50)
        /home/fishy/work/reddit/baseplate.go/thriftbp/prometheus.go:236 +0x173
github.com/reddit/baseplate.go/thriftbp.TestClientPoolGaugeExporter(0xc0004ea680)
        /home/fishy/work/reddit/baseplate.go/thriftbp/prometheus_test.go:177 +0x146
testing.tRunner(0xc0004ea680, 0x9b1280)
        /gnu/store/ayhq6nzj64sp0rh8xqrfjjis88zpz393-go-1.17.7/lib/go/src/testing/testing.go:1259 +0x102
created by testing.(*T).Run
        /gnu/store/ayhq6nzj64sp0rh8xqrfjjis88zpz393-go-1.17.7/lib/go/src/testing/testing.go:1306 +0x35a
exit status 2
FAIL    github.com/reddit/baseplate.go/thriftbp 0.015s
```

this is when I added `recover` and use `t.Errorf` explicitly (line 179 is the `t.Errorf` call, not the actual place panicked):
```
$ go test
--- FAIL: TestClientPoolGaugeExporter (0.00s)
    prometheus_test.go:179: panic: inconsistent label cardinality: expected 1 label values but got 0 in []string(nil)
...
```

fishy: the `t.Logf` part will also not be not very useful, this is even if we used `%#v`:
```
prometheus_test.go:169: metric: &prometheus.constMetric{desc:(*prometheus.Desc)(0xc000388070), valType:2, val:1, labelPairs:[]*io_prometheus_client.LabelPair{(*io_prometheus_client.LabelPair)(0xc00059d1a0)}}
```

kylelemons: Boo, their error messages are bad.  You could grab the stack trace and send it back over the channel too, but that starts to make it cumbersome again.

The benefit is that this is how I would require the test to be written in production: explicit communication from the goroutine back to the synchronous API, rather than waitgroup which is easy to screw up.

Don't use `%#v` pick some fields to log (probably the name and fields if you can find them)

kylelemons (approved):

konradreiche (approved):

fishy:
💇 I added another test to make sure that multiple `Register` call will not cause an issue.

konradreiche (approved):

thriftbp/prometheus_test.go
kylelemons: `.Desc()` is a Stringer that has a human-readable representation of it

kylelemons (approved):
see note about .Desc

thriftbp/prometheus_test.go
fishy: but `.Desc()` is only the spec part, not the value part, so it's still not very helpful:
```
Desc{fqName: "thriftbp_client_pool_active_connections", help: "The number of active (in-use) connections of a thrift client pool", constLabels: {}, variableLabels: [thrift_slug]}
```

also if we do have a label mismatch and the panic happens, the panic happens before it's written to the channel, so whatever we log from reading from the channel will not include the one that causes problems, so it's still... not helpful.

PR #497, opened 2022-03-09
redisbp: Minor improvements on the prometheus exporter

fishy:
1. Move the Desc from the exporter to global variable as they really
   don't change between different exporter instances.
2. Add a test to make sure that a service creating multiple redisbp
   clients won't cause prometheus exporter registration issues.

cc @SuperQ


konradreiche (approved):

kylelemons (approved):

SuperQ:
:+1: Thanks!

PR #499, opened 2022-03-17
thriftbp: Deprecate ClientPoolConfig.InitialConnectionsFallback

fishy:
Currently ClientPoolConfig.InitialConnectionsFallback is default to
false, due to how zero values in go work. Deprecate it to make it
default to true, because cascading failures during an outage is in
general much worse.

Add InitialConnectionsNoFallback to allow the current default behavior
(that could cause cascading failures) for services that absolutely need
it.

Also instead of fallback to 0 connections, return whatever connections
we already established with the pool. This requires a small behavior
change inside clientpool package, that clientpool.NewChannelPool now
could return both non-nil Pool and error. An internal code search shows
that there are 2 services using that directly (we probably should have
made that an internal package to avoid that, but that ship has already
sailed). I'll send out separated PRs to handle the both non-nil Pool and
error case for the 2 services to avoid connection leaks when they
upgrade Baseplate.go to this version.

### DIFF
--- a/clientpool/channel.go
+++ b/clientpool/channel.go
@@ -17,6 +17,10 @@ type channelPool struct {
 var _ Pool = (*channelPool)(nil)
 
 // NewChannelPool creates a new client pool implemented via channel.
+//
+// Note that this function could return both non-nil Pool and error,
+// when we failed to create all asked initialClients.
+// In such case the returned Pool would have the clients we already established.
 func NewChannelPool(initialClients, maxClients int, opener ClientOpener) (Pool, error) {
 	if initialClients > maxClients {
 		return nil, &ConfigError{
@@ -25,16 +29,18 @@ func NewChannelPool(initialClients, maxClients int, opener ClientOpener) (Pool, 
 		}
 	}
 
+	var finalErr error
 	pool := make(chan Client, maxClients)
 	for i := 0; i < initialClients; i++ {
 		c, err := opener()
 		if err != nil {
-			return nil, fmt.Errorf(
-				"error creating client #%d/%d: %w",
+			finalErr = fmt.Errorf(
+				"clientpool: error creating client #%d/%d: %w",
 				i,
 				initialClients,
 				err,
 			)
+			break
 		}
 		pool <- c
 	}
@@ -44,7 +50,7 @@ func NewChannelPool(initialClients, maxClients int, opener ClientOpener) (Pool, 
 		opener:         opener,
 		initialClients: initialClients,
 		maxClients:     maxClients,
-	}, nil
+	}, finalErr
 }
 
 // Get returns a client from the pool.
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -62,9 +62,23 @@ type ClientPoolConfig struct {
 	// "${host}:${port}"
 	Addr string `yaml:"addr"`
 
-	// InitialConnections is the inital number of thrift connections created by
-	// the client pool.
-	InitialConnections int `yaml:"initialConnections"`
+	// InitialConnections is the desired inital number of thrift connections
+	// created by the client pool.
+	//
+	// If an error occurred when we try to establish the initial N connections for
+	// the pool, we log the error with InitialConnectionsFallbackLogger,
+	// then return the pool with the <N connections we already established.
+	//
+	// If that's unacceptable, then RequiredInitialConnections can be used to set
+	// the hard requirement of minimal initial connections to be established.
+	// Note that enabling this can cause cascading failures during an outage,
+	// so it shall only be used for extreme circumstances.
+	InitialConnections               int         `yaml:"initialConnections"`
+	InitialConnectionsFallbackLogger log.Wrapper `yaml:"initialConnectionsFallbackLogger"`
+	RequiredInitialConnections       int         `yaml:"requiredInitialConnections"`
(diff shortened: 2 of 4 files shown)


konradreiche (approved):

clientpool/channel.go
```diff
@@ -17,6 +17,10 @@ type channelPool struct {
 var _ Pool = (*channelPool)(nil)
 
 // NewChannelPool creates a new client pool implemented via channel.
+//
+// Note that this function could return both non-nil Pool and error,
+// When we failed to create all asked initialClients.
+// In such case the returned Pool would have the clients we already established.
```
kylelemons, line 23: This seems like it's making the API harder to use correctly -- this is likely to lead to un-Close-d pools when we return an error.  You will have to say
```go
pool, err := NewChannelPool(...)
if pool != nil {
  defer pool.Close()
}
if err != nil {
  ...
}
```

which is not very natural.

If we're going to return a Pool, I think we need to _not_ return an error.

thriftbp/client_pool.go
```diff
@@ -62,9 +62,29 @@ type ClientPoolConfig struct {
...
+	//
+	// If <N initial connections is not acceptable, you can set
+	// InitialConnectionsNoFallback to true. In that case we return an error if
+	// any error occurred while trying to establish the initial connections.
+	// But please note that enabling this can cause cascading failures during an
+	// outage, as when the upstream is in an outage, you won't be able to do any
+	// deploys because new pods will fail creating the client pool.
+	InitialConnections               int         `yaml:"initialConnections"`
+	InitialConnectionsFallbackLogger log.Wrapper `yaml:"initialConnectionsFallbackLogger"`
+	InitialConnectionsNoFallback     bool        `yaml:"initialConnectionsNoFallback"`
```
kylelemons, line 84: Maybe `RequireInitialConnections int` to set how many connections the pool must have in order to be considered successfully constructed?  (the default being zero, i.e. don't fail if you can't establish any connections)

kylelemons (changes requested):

clientpool/channel.go
fishy: there are a few instances in the stdlib that the function can return both non-nil result and error, although the only one non-nil pattern is far more common.

but I like your suggestion of changing the config to a required int, so I'll do that instead.

fishy: but on second thought, that means we are losing the info about how the M-th connection was failed to be established (unless we make a breaking change to also pass in a logger to log that, or something like that).

`clientpool` is considered a low-level api that most services will not use it directly (internal code search shows only 2 repo uses it, see my PR description/commit message), how strongly do you feel that we should just swallow the error and don't return an error?

kylelemons: doesn't the logger satisfy that need of wanting to control where and how errors are handled?

kylelemons (approved):
🔕 If we don't want people to use this API, we should probably mark it as deprecated too, since when we make baseplate v2 we will probably remove such "internal" APIs.

clientpool/channel.go
fishy: we don't have logger in this API now. adding it will be a breaking change.

in my experience that error usually can provide info to shed light on whether the error is on dns or on the remote service.

fishy:
> 🔕 If we don't want people to use this API, we should probably mark it as deprecated too, since when we make baseplate v2 we will probably remove such "internal" APIs.

that's basically what I said in the commit message, that we should have made it `internal` but didn't. making it internal now will be a breaking change :)

clientpool/channel.go
kylelemons: I meant `InitialConnectionsFallbackLogger`?  Since you're going to be _trying_ to stand up the connections, some will just fail, and they'll be logged there.

The workflow you're describing is primarily human driven, so the `error` isn't necessary as long as it can be logged.

kylelemons (approved):

thriftbp/client_pool.go
```diff
@@ -62,9 +62,23 @@ type ClientPoolConfig struct {
...
+	// created by the client pool.
+	//
+	// If an error occurred when we try to establish the initial N connections for
+	// the pool, we log the error with InitialConnectionsFallbackLogger,
+	// then return the pool with the <N connections we already established.
+	//
+	// If that's unacceptable, then RequiredInitialConnections can be used to set
+	// the hard requirement of minimal initial connections to be established.
+	// Note that enabling this can cause cascading failures during an outage,
+	// so it shall only be used for extreme circumstances.
```
kylelemons, line 75: ... or special cases like integration or load tests?

clientpool/channel.go
fishy: that's part of the thriftbp api, not part of the clientpool api (clientpool is the more low-level api called by thriftbp)

thriftbp/client_pool.go
fishy: I don't think it should even be used in these 2 cases. Say you set `RequiredInitialConnections` to 1 in a (integration|load) test and the first client failed, that does not necessarily mean the remote server is just unavailable. it could well be just a dns blip, or any network blip. The test shall only fail in cases of, say, the failure rate is higher than some threshold. The first client connection attempt failed does not really map to the failure rate threshold in any way.

konradreiche (approved):

PR #500, opened 2022-03-30
Workaround a bug in envoy

marcoferrer:
This is a fix to bring baseplate go in sync with baseplate py.

### DIFF
--- a/thriftbp/headers.go
+++ b/thriftbp/headers.go
@@ -2,6 +2,7 @@ package thriftbp
 
 import (
 	"context"
+	"strings"
 
 	"github.com/apache/thrift/lib/go/thrift"
 
@@ -43,4 +44,28 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
 	ctx = thrift.SetHeader(ctx, key, value)
 	headers = append(headers, key)
 	return thrift.SetWriteHeaderList(ctx, headers)
+}
+
+// Header gets the value of a thrift header by key
+//
+// If the value is not present we fall back to a case-insensitive check
+func Header(ctx context.Context, key string) (v string, ok bool) {
+
+	// Option 1.
+	v, ok = thrift.GetHeader(ctx, key)
+	if !ok {
+		v, ok = thrift.GetHeader(ctx, strings.ToLower(key))
+	}
+
+	// Option 2.
+	// We can potentially recreate the map in the context with lower-case keys
+	// then we wouldn't need to iterate over the map.
+	for _, header := range thrift.GetReadHeaderList(ctx) {
+		if strings.EqualFold(header, key) {
+			v, ok = thrift.GetHeader(ctx, header)
+			break
+		}
+	}
+
+	return
 }
--- a/thriftbp/headers.go
+++ b/thriftbp/headers.go
@@ -46,26 +46,16 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
 	return thrift.SetWriteHeaderList(ctx, headers)
 }
 
-// Header gets the value of a thrift header by key
+// header gets the value of a thrift header by key
 //
 // If the value is not present we fall back to a case-insensitive check
-func Header(ctx context.Context, key string) (v string, ok bool) {
-
-	// Option 1.
+func header(ctx context.Context, key string) (v string, ok bool) {
 	v, ok = thrift.GetHeader(ctx, key)
 	if !ok {
+		// We fall back to a case-insensitive check so that we can have
+		// functional parity with baseplate-py
 		v, ok = thrift.GetHeader(ctx, strings.ToLower(key))
 	}
-
-	// Option 2.
-	// We can potentially recreate the map in the context with lower-case keys
-	// then we wouldn't need to iterate over the map.
-	//for _, header := range thrift.GetReadHeaderList(ctx) {
-	//	if strings.EqualFold(header, key) {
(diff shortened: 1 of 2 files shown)


thriftbp/headers.go
```diff
@@ -44,3 +45,27 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
 	headers = append(headers, key)
 	return thrift.SetWriteHeaderList(ctx, headers)
 }
+
+// Header gets the value of a thrift header by key
+//
+// If the value is not present we fall back to a case-insensitive check
+func Header(ctx context.Context, key string) (v string, ok bool) {
```
fishy, line 52: I don't think we want to export this. if a service code need to do the same thing they can just duplicate this function.

thriftbp/headers.go
```diff
@@ -44,3 +45,27 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
...
+
+	// Option 1.
+	v, ok = thrift.GetHeader(ctx, key)
+	if !ok {
+		v, ok = thrift.GetHeader(ctx, strings.ToLower(key))
```
fishy, line 57: if you are certain that what envoy does is just lowercase the key, then this is enough and we won't need option 2.

fishy (changes requested):

thriftbp/headers.go
```diff
@@ -44,3 +45,17 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
 	headers = append(headers, key)
 	return thrift.SetWriteHeaderList(ctx, headers)
 }
+
+// header gets the value of a thrift header by key
+//
+// If the value is not present we fall back to a case-insensitive check
```
fishy, line 51: s/fall back to a case-insensitive check/fall back to all lowercase check to workaround a bug in envoy/

fishy (commented):
the code lgtm, but before we can merge it:

1. The PR title should be something like "Workaround a bug in envoy" as this is not a bug in baseplate.go
2. We need to file a bug on envoy, and link the bug here, so we could revert this change in the future when the bug is fixed in envoy.

fishy:
https://github.com/envoyproxy/envoy/issues/20595

thriftbp/headers.go
```diff
@@ -44,3 +45,17 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
...
+func header(ctx context.Context, key string) (v string, ok bool) {
+	v, ok = thrift.GetHeader(ctx, key)
+	if !ok {
+		// We fall back to a case-insensitive check so that we can have
```
kylelemons, line 55: This is incorrect -- it's not a case-insensitive check, it's a lower-case check.

I think you should say
```
We fall back to checking the key that an envoy proxy will send, which is a lower-case
version of the thrift key.

For details on this workaround, see:
* <link to this PR or whatever>
* <link to the envoyproxy/envoy bug>
```

kylelemons (changes requested):

thriftbp/headers.go
```diff
@@ -44,3 +45,17 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
...
+		// functional parity with baseplate-py
```
marcoferrer, line 56:
```suggestion
		// We fall back to checking the key that an envoy proxy will send, which is a lower-case
		// version of the thrift key.
		// 
		// 
		// For details on this workaround, see:
		// 	* https://github.com/reddit/baseplate.py/blob/0d189bae9bc15459b8d9d62856f31e45cbfcdb1a/baseplate/frameworks/thrift/__init__.py#L111	
		// 	* https://github.com/envoyproxy/envoy/issues/20595
		// 	* https://github.com/reddit/baseplate.go/pull/500
```

thriftbp/headers.go
```diff
@@ -44,3 +45,23 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
...
+// If the value is not present we fall back to a case-insensitive check
+func header(ctx context.Context, key string) (v string, ok bool) {
+	v, ok = thrift.GetHeader(ctx, key)
+	if !ok {
+		// We fall back to checking the key that an envoy proxy will send, which is a lower-case
+		// version of the thrift key.
+		// 
+		// 
+		// For details on this workaround, see:
+		// 	* https://github.com/reddit/baseplate.py/blob/0d189bae9bc15459b8d9d62856f31e45cbfcdb1a/baseplate/frameworks/thrift/__init__.py#L111
```
marcoferrer, line 60:
```suggestion
		// 	* https://github.com/reddit/baseplate.py/commit/daf1c5e32dcc019470d75f3940006bb2f7b9855e	
```

thriftbp/headers.go
```diff
@@ -44,3 +45,23 @@ func AddClientHeader(ctx context.Context, key, value string) context.Context {
 	headers = append(headers, key)
 	return thrift.SetWriteHeaderList(ctx, headers)
 }
+
+// header gets the value of a thrift header by key
+//
+// If the value is not present we fall back to a case-insensitive check
```
marcoferrer, line 51:
```suggestion
// If the value is not present we fall back to all lowercase check to workaround a bug in envoy
```

fishy (approved):

kylelemons (approved):

pacejackson (approved):

fishy:
you have extra trailing spaces: https://github.com/reddit/baseplate.go/runs/5775116141?check_suite_focus=true

PR #502, opened 2022-03-30
log: Add PrometheusCounterWrapper

fishy:
This is the prometheus version of metricsbp.LogWrapper.

### DIFF
--- a/log/wrapper.go
+++ b/log/wrapper.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/getsentry/sentry-go"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 // DefaultWrapper defines the default one to be used in log.Wrapper.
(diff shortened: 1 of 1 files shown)


log/wrapper.go
```diff
@@ -272,6 +273,27 @@ func ErrorWithSentryWrapper() Wrapper {
...
+// Please note that it's not possible to deserialize this Wrapper directly from
+// yaml, so you usually need to override it in your main function, after
+// baseplate.ParseConfigYAML call, for example:
+//
+//     if err := baseplate.ParseConfigYAML(&cfg); err != nil {
+//       log.Fatal(err)
+//     }
+//     cfg.Config.Tracing.Logger = log.PrometheusCounterWrapper(
+//       cfg.Config.Tracing.Logger,       // delegate
+//       tracingFailureCounter.With(nil), // counter
```
pacejackson, line 288: why the `.With(nil)`?

pacejackson (approved):

log/wrapper.go
fishy: In a typical prometheus counter, the global variable you defined is a [`CounterVec`](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec), and [`With`](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec.With) gives you an actual [`Counter`](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Counter).

The args to `With` is the label values. For things used globally as a log wrapper, you usually don't really have any labels defined for the counter, thus the `nil`.

pacejackson: would it make sense to just do `counter = counter.With(nil)` in PrometheusCounterWrapper before returning the function?

fishy: actually let me expand the example a little bit.

pacejackson (approved):

log/wrapper.go
```diff
@@ -272,6 +273,36 @@ func ErrorWithSentryWrapper() Wrapper {
...
+// yaml, so you usually need to override it in your main function, after
+// baseplate.ParseConfigYAML call, for example:
+//
+//     // a global variable
+//     var tracingFailures = promauto.NewCounterVec(prometheus.CounterOpts{
+//       Namespace: "myservice",
+//       Subsystem: "tracing",
+//       Name:      "failures_total",
+//       Help:      "Total number of failures when sending tracing spans to the sidecar",
+//     }, nil /* no labels */).With(nil /* no labels */)
```
kylelemons, line 289: Don't use a Vec if you don't have labels.

If you're going to have an example with labels, the With call should happen at the PrometheusCounterWrapper call-site.

kylelemons (changes requested):

kylelemons (approved):

konradreiche (approved):

PR #503, opened 2022-03-30
filewatcher: Change to pull+push mode

fishy:
Currently filewatcher only work on push mode, by using fsnotify to
monitor the parent directory. This might break when the whole parent
directory is remounted. Add pull mode, that checks the file mtime when
Get is called, and force reload the file if the mtime is after the last
parsed mtime.

### DIFF
--- a/filewatcher/filewatcher.go
+++ b/filewatcher/filewatcher.go
@@ -16,6 +16,10 @@ import (
 	"github.com/reddit/baseplate.go/log"
 )
 
+// DefaultPollingInterval is the default PollingInterval used when creating a
+// new file watcher.
+const DefaultPollingInterval = 30 * time.Second
+
 // FileWatcher loads and parses data from a file and watches for changes to that
 // file in order to refresh it's stored data.
 type FileWatcher interface {
@@ -43,7 +47,7 @@ var InitialReadInterval = time.Second / 2
 //
 // It's 1MiB, which is following the size limit of Apache ZooKeeper nodes.
 const (
-	DefaultMaxFileSize  = 1024 * 1024
+	DefaultMaxFileSize  = 1 << 20
 	HardLimitMultiplier = 10
 )
 
@@ -64,10 +68,12 @@ type Result struct {
 
 // Get returns the latest parsed data from the file watcher.
 //
+// It might also call Parser if the file is changed from the latest parsed data.
+//
 // Although the type is interface{},
 // it's guaranteed to be whatever actual type is implemented inside Parser.
 func (r *Result) Get() interface{} {
-	return r.data.Load()
+	return r.data.Load().(*atomicData).data
 }
 
 // Stop stops the file watcher.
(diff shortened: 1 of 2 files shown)


pacejackson (commented):
Why do both the fsnotify and stat check on get? It makes this more complex and it's not clear what the value is.

filewatcher/filewatcher.go
```diff
@@ -107,26 +114,64 @@ func (r *Result) watcherLoop(
...
+		return
+	}
+	defer f.Close()
+	d, err := r.parser(f)
+	if err != nil {
+		r.logger.Log(context.Background(), "filewatcher: parser error: "+err.Error())
+	} else {
+		r.data.Store(&atomicData{
+			data:  d,
+			mtime: mtime,
```
pacejackson, line 161: I think we probably want to re-grab the mtime from the file rather than using the one that's passed in case the file was updated between when we checked the mtime and when we opened the file?

fishy: in that scenario we just do an extra `forceReload` and load in the same data, which is not a big deal.

filewatcher/filewatcher.go
```diff
@@ -107,26 +114,64 @@ func (r *Result) watcherLoop(
...
+		})
+	}
+}
```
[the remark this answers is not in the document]
fishy: in that race condition we just have 2 competing goroutines trying to load the same data and save the same result to the atomic value. it's not ideal but not a big deal either.

fishy (commented):
>Why do both the fsnotify and stat check on get? It makes this more complex and it's not clear what the value is.

The value of push is to move the cost of parser out of the critical path. Usually `result.Get()` is called on critical path (e.g. during your endpoint handler), so if you have to call parser there, the latency added by parser is added to your critical path.

fishy:
so we can pump the mtime from `limitopen.Open`, but that requires some breaking changes (also `limitopen` is under internal so breaking changes should be OK?). I feel that it's probably not worth all the trouble as the consequences are very minor, but if people feel strongly I can do that.

filewatcher/filewatcher.go
```diff
@@ -60,14 +60,27 @@ type Result struct {
...
 // Get returns the latest parsed data from the file watcher.
 //
+// It might also call Parser if the file is changed from the latest parsed data.
+//
 // Although the type is interface{},
 // it's guaranteed to be whatever actual type is implemented inside Parser.
 func (r *Result) Get() interface{} {
-	return r.data.Load()
+	r.reload()
+	d, _ := r.data.Load().(*atomicData)
```
kylelemons, line 79: if this type assertion cannot fail, don't make it a checked type asertion and _definitely_ don't silently ignore it

filewatcher/filewatcher.go
```diff
@@ -81,22 +94,16 @@ func (r *Result) Stop() {
...
 		select {
 		case <-r.ctx.Done():
 			watcher.Close()
 			return
 
 		case err := <-watcher.Errors:
-			logger.Log(context.Background(), "filewatcher: watcher error: "+err.Error())
+			r.logger.Log(context.Background(), "filewatcher: watcher error: "+err.Error())
 
 		case ev := <-watcher.Events:
```
kylelemons, line 108: When I was suggesting polling, I was imagining a new constructor that specified an additional fixed polling period, plus some sensible default (e.g. 30 seconds) for the existing ones, not checking on every Get call.

WDYT?

kylelemons (changes requested):

filewatcher/filewatcher.go
fishy: I certainly like that better as it makes things much simpler, let me ask internally on what's an appropriate default.

filewatcher/filewatcher.go
kylelemons: returning nil is silently ignoring it -- it's pushing the failure into user code, where we have lost context around what the actual failure was.  What we need to know is what the type was if it wasn't *atomicData

kylelemons (changes requested):

pacejackson (approved):

PR #505, opened 2022-04-04
log: PrometheusCounterWrapper -> CounterWrapper

fishy:
Relax the interface to a minimal Counter and drop the prometheus
dependency.

### DIFF
--- a/log/wrapper.go
+++ b/log/wrapper.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/getsentry/sentry-go"
-	"github.com/prometheus/client_golang/prometheus"
 )
 
 // DefaultWrapper defines the default one to be used in log.Wrapper.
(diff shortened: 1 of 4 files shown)


kylelemons (approved):
Slightly annoying that the common method is Add and not Inc, but... 🤷

log/wrapper.go
```diff
@@ -273,7 +272,15 @@ func ErrorWithSentryWrapper() Wrapper {
 	}
 }
 
-// PrometheusCounterWrapper returns a Wrapper implementation that increases
+// Counter is a minimal interface for a counter.
+//
+// This is implemented by both prometheus counter and statsd counter from
+// metricsbp.
+type Counter interface {
```
konradreiche, line 279: Does this need to be exported though?

fishy: it's required by exported function `CounterWrapper` so yes it needs to be exported.

konradreiche (approved):

PR #508, opened 2022-04-08
Adds ReplicaPolicy Config when creating a clustered redis baseplate client

loftiskg:
This PR adds the ability to specify the ReplicaPolicy when creating a clustered Redis client.

### DIFF
--- a/redis/cache/redispipebp/baseplate.go
+++ b/redis/cache/redispipebp/baseplate.go
@@ -37,6 +37,11 @@ type BaseplateRedisClientArgs struct {
 	// Retry is the retry options for the redis client.
 	// This is optional, if it is empty or nil, then we will not wrap the client in a RetrySync
 	Retry []retry.Option
+	// ReplicaPolicy is the config enumeration of policies of redis replica hosts usage.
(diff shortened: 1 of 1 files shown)


konradreiche (approved):
I wanted to suggest to expose this as a configuration here 
https://github.com/reddit/baseplate.go/blob/624989c06094d040289f803277d6204f6a0e74c7/redis/cache/redispipebp/config.go#L83-L92

But it looks like there would be no way with the existing constructor to apply it since `rediscluster.Opts` does not let you define the policy, it only works by wrapping it manually :shrug:

fishy (approved):
I don't really love the idea of passing through those configs but in the current api I can't think of a better way to do that so 🤷

loftiskg:
Yeah, I am in the same boat.

kylelemons (approved):
Yeah, it's a bit unfortunate that it's tied to the same API.

 Can you add a TODO inside NewBaseplateRedisCluster that we should make it a separate config object when we next make breaking API changes?

loftiskg:
@kylelemons sure, added a TODO comment.

PR #509, opened 2022-04-15
Register important prometheus error counters

fishy:
For important prometheus error counters with labels, register them
(With the labels but without Inc/Add), so that we are always emitting 0
even if those error counters never fire. This way they can be monitored
correctly (with rate query producing expected results).

### DIFF
--- a/httpbp/middlewares.go
+++ b/httpbp/middlewares.go
@@ -287,6 +287,9 @@ func SupportedMethods(method string, additional ...string) Middleware {
 // bubble up into other middlewares. Since it is always added to the middleware
 // chain is a specific position, it is not exported.
 func recoverPanic(name string, next HandlerFunc) HandlerFunc {
+	counter := panicRecoverCounter.With(prometheus.Labels{
+		methodLabel: name,
+	})
 	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
 		defer func() {
 			if r := recover(); r != nil {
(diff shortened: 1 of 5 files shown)


fishy:
also cc @SuperQ

pacejackson (approved):

konradreiche (approved):

httpbp/middlewares.go
```diff
@@ -287,6 +287,9 @@ func SupportedMethods(method string, additional ...string) Middleware {
 // bubble up into other middlewares. Since it is always added to the middleware
 // chain is a specific position, it is not exported.
 func recoverPanic(name string, next HandlerFunc) HandlerFunc {
+	promCounter := panicRecoverCounter.With(prometheus.Labels{
```
kylelemons, line 290: 🔕 since 'prom' is the way of the future, can we just call it "counter" or "metric"?

kafkabp/consumer.go
```diff
@@ -33,6 +33,11 @@ var (
 		Name:      "rebalance_total",
 		Help:      "The number of times consumer rebalance happened",
 	}, rebalanceLabels)
+
+	// Register the error counter so that it can be monitored.
+	_ = rebalanceCounter.With(prometheus.Labels{
+		successLabel: prometheusbp.BoolString(false),
+	})
```
kylelemons, line 40: I think I would prefer this to be in an `init()`.

I think you should also register both `true` and `false`, no?

kylelemons (approved):

kafkabp/consumer.go
fishy: we can register both true and false, it's just register true will not be very useful. `reset` is called in the constructor, so as soon as someone construct a consumer the true counter should emit 1.

kylelemons (approved):

kafkabp/consumer.go
```diff
@@ -35,6 +35,13 @@ var (
 	}, rebalanceLabels)
 )
 
+func init() {
+	// Register the error counter so that it can be monitored.
+	rebalanceCounter.With(prometheus.Labels{
+		successLabel: prometheusbp.BoolString(false),
```
SuperQ, line 41: I think we want to avoid success boolean labels. Prometheus best practices prefers separate "all FOO" and "error FOO" counters. This would avoid the need to do this init here.

PR #511, opened 2022-04-18
redispipebp: Add prom timer for MonitoredClient

fishy:


pacejackson (approved):

redis/cache/redispipebp/monitored.go
```diff
@@ -3,34 +3,73 @@ package redispipebp
...
 }
 
+func normalizeCommand(cmd string) string {
+	cmd = strings.ToLower(cmd)
+	if index := strings.IndexByte(cmd, ' '); index >= 0 {
+		cmd = cmd[:index]
+	}
+	return cmd
+}
+
```
kylelemons, line 54: this is going to be doing _a lot_ of allocating, because of the strings.ToLower.  Are we sure we want that?

kylelemons (approved):

redis/cache/redispipebp/monitored.go
fishy: I think redis is case-insensitive so you can use `SET foo`/`SeT foo`/`set foo` and they all work, but not doing `ToLower` will greatly increase prometheus cardinality (with `ToLower` the actual number of redis commands you are use is limited).

I can move `ToLower` to after the space cut, so that `ToLower` will only be applied to a much shorter string, wdyt?

konradreiche: The question to me is, should we even take care of this? If anyone types them out as lowercase, ideally this is being enforced through guidelines and not code. Alternatively: settle on upper case and check `IsUpper` and normalize otherwise / warn?

konradreiche (changes requested):
Left a question regarding responsibility of normalizing commands in middleware.

redis/cache/redispipebp/monitored.go
fishy: We should. Redis commands are case insensitive: https://redis.io/commands/command/#:~:text=Note%3A%20Redis%20command%20names%20are%20case%2Dinsensitive.

So it could totally happen that in one place someone wrote:
```go
client.Do(ctx, "SET", ...)
```

and in another place of the code, or even in another service (prometheus metrics are global so all services emit the same metrics):
```go
client.Do(ctx, "set", ...)
```

they work the same, so they should emit the same metrics. so the case normalization is necessary.

fishy: I don't have a strong opinion between lower and upper though.

konradreiche (approved):
I think lowercase is more in line with Prometheus conventions but I could be wrong. I don't feel strongly about the performance implications, it will likely depend a lot on what we see in profilings.

redis/cache/redispipebp/monitored.go
kylelemons: ToUpper / ToLower has a fast path that doesn't allocate if it is already upper- or lower-case (respectively), so _if_ we normalize, we should strongly prefer to select the one that is most-used.

In some ways, however, I feel like we should just not normalize.  If people are annoyed by having "SET" and "set" they should unify their code to use "SET" or maybe someone should build an opt-in middleware that normalizes the command.  Regardless, I am not sure it's the place of _monitoring_ to make such editorial decisions, especially at the cost of performance.

fishy: re performance: I moved it to *after* the space cut, so `ToLower`/`ToUpper` would only work on the actual redis command (before the first space). looks like the longest length of possible redis commands is something around 10?

kylelemons: unfortunately it doesn't really matter how long the commands are -- the performance cost is purely due to the number of allocations this is going to cause.  That's why using a map to intern the strings is beneficial if this is more than a few hundred QPS -- it spends a bit of CPU in order to avoid calling into the allocator.  (and if it's less than a few hundred QPS, the overhead of spending the CPU is negligible)

fishy: ok removed `ToLower` and renamed this function to `extractCommand`.

kylelemons (approved):

PR #512, opened 2022-04-20
Use strconv.FormatBool

SuperQ:
Remove reimplementaiton of stdlib function strconv.FormatBool().

### DIFF
--- a/grpcbp/client_middlewares.go
+++ b/grpcbp/client_middlewares.go
@@ -3,6 +3,7 @@ package grpcbp
 import (
 	"context"
 	"errors"
+	"strconv"
 	"time"
 
 	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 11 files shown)


fishy (changes requested):

konradreiche (changes requested):
Can we instead mark it as deprecated?

konradreiche (approved):

fishy (changes requested):
The commit message of 9ad9f79d34867828c720a48b6c8a01aa283867b7 explained why we added this, marking it as deprecated does not resolve the problem.

We can probably mark it as deprecated when we mark the whole `metricsbp` package as deprecated, and even that is still debatable. But we certainly should not mark it as deprecated now.

kylelemons (changes requested):
I'm with @fishy I think.

We can deprecate this function when we deprecate metricsbp, and we can use a source rewrite to remove it from call-sites.  (I would replace it with a string, not the call to strconv, when it's a literal `true` or `false` as well.)

PR #513, opened 2022-04-21
redispipebp: Add name to prometheus metric labels

fishy:
In af1f91b0c661b89078a42cab5ccbe36c2b9e1d44 I totally forgot to add this
label, so when a service has 2 redispipebp.MonitoredSync the prometheus
metrics won't be able to distinguish between them.

### DIFF
--- a/redis/cache/redispipebp/monitored.go
+++ b/redis/cache/redispipebp/monitored.go
@@ -20,12 +20,14 @@ import (
 const (
 	promNamespace = "redispipebp"
 
+	labelSlug    = "redis_slug"
 	labelCommand = "redis_command"
 	labelSuccess = "redis_success"
 )
 
 var (
 	labels = []string{
+		labelSlug,
 		labelCommand,
 		labelSuccess,
 	}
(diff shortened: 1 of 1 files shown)


redis/cache/redispipebp/monitored.go
```diff
@@ -20,12 +20,14 @@ import (
 const (
 	promNamespace = "redispipebp"
 
+	labelName    = "redis_name"
```
konradreiche, line 23: Slug vs. name?

konradreiche (approved):

redis/cache/redispipebp/monitored.go
fishy: I don't feel strongly either way. In `MonitoredSync` api it's called `Name` so I kept it consistent with that, but I can rename the prometheus label to slug if that's preferred.

kylelemons: I don't have a good model for when something is a name vs a slug, so... 🤷 

Other names that might work: name, tag, target, group, remote, id

kylelemons (approved):

redis/cache/redispipebp/monitored.go
fishy: we need andrew's tie breaking vote here but he's OOO today 😅

fishy: ok renamed to `redis_slug`. currently across baseplate.go code we have several usages of _slug prometheus label and zero usage of _name in prometheus labels.

pacejackson: Here's what I said to @fishy when he asked about this:

> When I was making the some other tooling, I think there was both a name and slug, name was the “human readable” one and slug was for the “machine readable” one that was actually in the metrics

PR #514, opened 2022-04-25
log: Add fallback to debug.BuildInfo to version stamping

fishy:
Taking advantage of go 1.18's new -buildvcs flag, read the version info
from runtime/debug.BuildInfo when no version is stamped in log package.

Although we don't have -buildvcs in our production builds yet, this
makes us ready to take advantage of that whenever it's ready.

This also requires us to drop support to go 1.17, as
runtime/debug.BuildSetting is added in go 1.18.

### DIFF
--- a/log/version.go
+++ b/log/version.go
@@ -1,0 +1,83 @@
+package log
+
+import (
+	"runtime/debug"
+)
+
+// Version is the version tag value to be added to the global logger.
+//
+// If it's changed to non-empty value before the calling of Init* functions
+// (InitFromConfig/InitLogger/InitLoggerJSON/InitLoggerWithConfig/InitSentry),
+// the global logger initialized will come with a tag of VersionKey("v"),
+// added to every line of logs.
+// For InitSentry the global sentry will also be initialized with a tag of
+// "version".
+//
+// In order to use it, either set it in your main function early,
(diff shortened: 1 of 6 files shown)


konradreiche (approved):

log/version.go
```diff
@@ -0,0 +1,86 @@
...
+func getVersion() string {
+	if Version != "" {
+		return Version
+	}
+	info, ok := debug.ReadBuildInfo()
+	if !ok || info == nil {
+		return ""
+	}
+	return getVersionFromBuildInfo(info)
+}
```
kylelemons, line 54: I haven't cehcked the final implementation, but I think initially it seemed like ReadBuildInfo wasn't something we would want to call often.

How about populating Version in init() based on ReadBuildInfo if it's not already populated?

kylelemons (changes requested):

log/version.go
kylelemons: `init()` happens after execve and before calling main(), so it's at runtime -- ldflags happens at compile time.

So we don't need sync.once

fishy: ok. changed to `init()`.

kylelemons (approved):

mterwill (approved):

PR #517, opened 2022-04-27
Inject k8s metadata in response headers

clayton-gonsalves:
## 💸 TL;DR
Consume downward facing k8s info.
enrich http response headers with above metadata for debugging. 

The last commit is a potential implementation with the k8s client added to the metadatabp pkg. It would not be needed immediately but I can see it being required to fetch more fine grained data from k8s. 

Tested locally. 

## ✅ Checks


fishy (changes requested):

pacejackson (changes requested):
It's not clear to me what the benefit of this is and, like yuxuan, I would worry about it leaking.

httpbp/middlewares.go
```diff
@@ -510,3 +520,31 @@ func (rr *responseRecorder) WriteHeader(code int) {
...
+	redditK8sMetaFmt   = "%s-%s-%s"
+)
+
+// InjectDebugHeaders injects debug headers into the http response.
+// headers are only injected if X-Reddit-Debug is set in the corresponding request.
+func InjectDebugHeaders(metadata *metadatabp.Config) Middleware {
+	return func(name string, next HandlerFunc) HandlerFunc {
+		return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
+			wrapped := &statusCodeRecorder{ResponseWriter: w}
+			if value := r.Header.Get(RedditDebug); value == RedditDebugEnabled && metadata != nil {
```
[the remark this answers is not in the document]
pacejackson: If what you want to do is be able to track down the logs for a request, I think a safer way to do that is to instead add a log tag into your context rather than returning k8s metadata to the client. Then you can search the logs for just that tag.

httpbp/middlewares.go
```diff
@@ -76,11 +77,20 @@ func DefaultMiddleware(args DefaultMiddlewareArgs) []Middleware {
 	if args.TrustHandler == nil {
 		args.TrustHandler = NeverTrustHeaders{}
 	}
+
+	// create metadata config to pass to InjectDebugHeaders middleware func
+	// since this isnt part of critical functionality, log error and move on.
+	config, err := metadatabp.New()
+	if err != nil {
+		log.Warn("could not set up func InjectDebugHeaders()", err)
```
kylelemons, line 85: metadatabp.New needs to always return a valid config, because if it returns a nil pointer and this warning prints out, any usage of the struct below will panic.

If there's an error, you should set a valid config, or you should always return a valid one and not return an error, or you should not add a middleware in the case of failure.

httpbp/middlewares.go
```diff
@@ -510,3 +520,31 @@ func (rr *responseRecorder) WriteHeader(code int) {
 	rr.ResponseWriter.WriteHeader(code)
 	rr.responseCode = code
 }
+
+const (
+	RedditDebug        = "X-Reddit-Debug"
+	RedditDebugEnabled = "enabled"
+	RedditK8sMeta      = "X-Reddit-K8S-Meta"
+	redditK8sMetaFmt   = "%s-%s-%s"
```
kylelemons, line 528: Don't move format strings to constants.  This makes it much harder to understand at the use-site.

httpbp/middlewares_test.go
```diff
@@ -323,3 +324,94 @@ func TestSupportedMethods(t *testing.T) {
 		)
 	}
 }
+
+func TestInjectDebugHeaders(t *testing.T) {
```
kylelemons, line 328: (sorry I didn't make it all the way through this, gotta get back to our team summit, will try to get further next round)

kylelemons (changes requested):

konradreiche (changes requested):
Is this something that could also be solved with tracing?

PR #518, opened 2022-04-27
thriftbp: Add prometheus histogram from client set timeout

fishy:
This can be useful for the service owners to get a sense of what their
clients are setting for the request timeout, and for example if their
SLO is 100ms but a lot of the client requests have a timeout <100ms,
they can talk to the client owners to fix the timeouts.

Note that due to the unusual nature of this histogram, we are usually
looking at the "reverted percentile" of it, for example instead of
looking at 99th percentile, we should be looking at the 1st percentile
for "the 1% of the lowest client set timeouts". As a result this
histogram will not work in a statsd metric, but it should work in a
prometheus metric.

### DIFF
--- a/thriftbp/server_middlewares.go
+++ b/thriftbp/server_middlewares.go
@@ -216,11 +216,19 @@ func ExtractDeadlineBudget(name string, next thrift.TProcessorFunction) thrift.T
 	return thrift.WrappedTProcessorFunction{
 		Wrapped: func(ctx context.Context, seqID int32, in, out thrift.TProtocol) (bool, thrift.TException) {
 			if s, ok := header(ctx, transport.HeaderDeadlineBudget); ok {
-				v, err := strconv.ParseInt(s, 10, 64)
-				if err == nil && v >= 1 {
+				if v, err := strconv.ParseInt(s, 10, 64); err == nil && v >= 1 {
+					timeout := time.Duration(v) * time.Millisecond
+
 					var cancel context.CancelFunc
-					ctx, cancel = context.WithTimeout(ctx, time.Millisecond*time.Duration(v))
+					ctx, cancel = context.WithTimeout(ctx, timeout)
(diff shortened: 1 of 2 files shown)


konradreiche (approved):

pacejackson (approved):

thriftbp/server_middlewares.go
```diff
@@ -219,8 +219,15 @@ func ExtractDeadlineBudget(name string, next thrift.TProcessorFunction) thrift.T
 				v, err := strconv.ParseInt(s, 10, 64)
 				if err == nil && v >= 1 {
 					var cancel context.CancelFunc
-					ctx, cancel = context.WithTimeout(ctx, time.Millisecond*time.Duration(v))
+					timeout := time.Duration(v) * time.Millisecond
+					ctx, cancel = context.WithTimeout(ctx, timeout)
 					defer cancel()
+
+					client, _ := header(ctx, transport.HeaderUserAgent)
```
kylelemons, line 226: Comment why it's ok to ignore the error?

thriftbp/server_middlewares.go
```diff
@@ -219,8 +219,15 @@ func ExtractDeadlineBudget(name string, next thrift.TProcessorFunction) thrift.T
 				v, err := strconv.ParseInt(s, 10, 64)
 				if err == nil && v >= 1 {
 					var cancel context.CancelFunc
```
kylelemons, line 221: keep this right above the `ctx, cancel` line since it's being added so that line can avoid shadowing ctx

bonus points, `// avoid shadowing ctx`

kylelemons (approved):

thriftbp/server_middlewares.go
fishy: the dropped return is not an error, it's an `ok bool`

PR #520, opened 2022-05-02
Update NumCPU for unlimited quota

SuperQ:
## 💸 TL;DR

Simplify runtimebp.NumCPU() by removing defer function.

Explicitly return runtime.NumCPU() when `cpu.cfs_quota_us` returns `-1`,
which is usef for no quota. This matches normal Go behavior.

[0]: https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt

## ✅ Checks

### DIFF
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -12,46 +12,53 @@ import (
 // NumCPU returns the number of CPUs assigned to this running container.
 //
 // This is the container aware version of runtime.NumCPU.
-// It reads from the cgroup sysfs values.
+// It reads from the cgroup cpu.cfs_quota_us and cpu.cfs_period_us values
+// to determine the hard CPU limit of the container.
 //
 // If the current process is not running inside a container,
 // or if there's no limit set in cgroup,
 // it will fallback to runtime.NumCPU() instead.
-//
-// When fallback happens, it also prints the reason to stderr.
-func NumCPU() (n float64) {
+func NumCPU() float64 {
 	const (
 		quotaPath  = "/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
 		periodPath = "/sys/fs/cgroup/cpu/cpu.cfs_period_us"
 	)
 
-	var err error
-	defer func() {
-		if err != nil || n <= 0 {
-			// Fallback and log to stderr.
-			fmt.Fprintf(
-				os.Stderr,
-				"NumCPU: falling back to use runtime.NumCPU(): %v, %v\n",
-				n,
-				err,
-			)
-			// fallback to physical cpus
-			n = float64(runtime.NumCPU())
-		}
-	}()
+	// Default to the standard runtime package value.
+	defaultCPUs := float64(runtime.NumCPU())
 
 	// Big enough buffer to read the number in the file wholly into memory.
 	buf := make([]byte, 1024)
-	var quota, period float64
 
-	quota, err = readNumberFromFile(quotaPath, buf)
+	quota, err := readNumberFromFile(quotaPath, buf)
 	if err != nil {
-		return
+		fmt.Fprintf(
+			os.Stderr,
+			"NumCPU: Failed to read quota file, falling back to use runtime.NumCPU(): %v\n",
+			err,
+		)
+		return defaultCPUs
 	}
 
-	period, err = readNumberFromFile(periodPath, buf)
+	// CFS quota returns -1 if there is no limit, return the default.
+	if quota < 0 {
+		fmt.Fprintf(
+			os.Stderr,
+			"NumCPU: Quota file returned %f, falling back to use runtime.NumCPU(): %v\n",
+			quota,
+			err,
+		)
+		return defaultCPUs
+	}
+
+	period, err := readNumberFromFile(periodPath, buf)
 	if err != nil {
-		return
+		fmt.Fprintf(
+			os.Stderr,
+			"NumCPU: Failed to read period file, falling back to use runtime.NumCPU(): %v\n",
+			err,
+		)
+		return defaultCPUs
 	}
 
 	return quota / period
--- a/runtimebp/cpu_test.go
+++ b/runtimebp/cpu_test.go
@@ -59,6 +59,10 @@ func TestReadNumberFromFile(t *testing.T) {
 			Content: "123.456",
 			Error:   true,
 		},
+		"negative": {
+			Content:  "-1",
+			Expected: -1,
+		},
 	}
 
 	for label, data := range cases {


konradreiche (approved):

runtimebp/cpu.go
```diff
@@ -48,6 +48,12 @@ func NumCPU() (n float64) {
 		return
 	}
 
+	// CFS quota returns -1 if there is no limit.
+	// Return the default NumCPU for this case to avoid squashing CPU on unlimited quota.
+	if quota < 0 {
+		return float64(runtime.NumCPU())
+	}
```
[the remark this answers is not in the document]
SuperQ: Wow, that's convoluted.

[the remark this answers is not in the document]
SuperQ: Using CPU shares is not a good idea either, if there's no quota, you don't want to blindly rely on shares.

There are good use cases where you would not set shares either, to take advantage of CFS shares in order to provide a scaleable workload, but have low priority by having a very small or no share request.

This code makes too many false assumptions about how CPU limits and shares work.

runtimebp/cpu.go
```diff
@@ -48,52 +36,18 @@ func NumCPU() (n float64) {
...
-	}()
-
-	// Big enough buffer to read the number in the file wholly into memory.
-	buf := make([]byte, 1024)
-	var shares float64
-
-	shares, err = readNumberFromFile(sharesPath, buf)
+	period, err = readNumberFromFile(periodPath, buf)
 	if err != nil {
 		return
```
fishy, line 47: with the removal of `defer` this will return 0. if you only want to remove the fallback to shares you should just change line 38 in the original code.

as to whether we want to remove the fallback to shares, that's still up to debate.

fishy (changes requested):

runtimebp/cpu.go
SuperQ: Good catch, I'll fix it.

runtimebp/cpu.go
```diff
@@ -48,52 +39,17 @@ func NumCPU() (n float64) {
...
 	}
 
-	period, err = readNumberFromFile(periodPath, buf)
-	if err != nil {
+	// CFS quota returns -1 if there is no limit, return the default.
+	if quota < 0 {
 		return
```
mterwill, line 44: Are named return values common elsewhere in baseplate.go? Personally favor explicit returns but not sure if we have an "official" style recommendation here 🤷‍♂️

runtimebp/cpu.go
```diff
@@ -48,52 +39,17 @@ func NumCPU() (n float64) {
...
-
-	shares, err = readNumberFromFile(sharesPath, buf)
+	period, err = readNumberFromFile(periodPath, buf)
 	if err != nil {
 		return
 	}
 
-	return float64(shares) / denominator
+	return quota / period
 }
```
mterwill, line 53: 💭 While we're refactoring this, worth making it testable as well? Could use an [`fs.FS`](https://pkg.go.dev/io/fs#FS) + pull the `runtime.NumCPU` dependency up into a `func() int` variable that could be overwritten in tests.

runtimebp/cpu.go
pacejackson: I wouldn't say they are super common, we mostly use them in monitoring contexts.  That said, I think we've been trying to encourage everyone to still have the return be explicit when we have named returns but this predates us being more deliberate about that 😅

fishy: there are two different things here: named return and naked return (while naked return is only possible with named return).

named return is necessary when you need to access (and optionally, modify) the return value, and we use that pattern a lot in middlewares. named return is also useful for documentation purpose (when the return type is not enough to infer the meaning of the return value).

naked return is in general not recommended, this code was just written early enough to predate that style guide 😅

SuperQ: Done, removed named return / naked returns.

runtimebp/cpu.go
```diff
@@ -19,81 +19,37 @@ import (
...
 //
 // When fallback happens, it also prints the reason to stderr.
-func NumCPU() (n float64) {
+func NumCPU() float64 {
 	const (
 		quotaPath  = "/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
 		periodPath = "/sys/fs/cgroup/cpu/cpu.cfs_period_us"
 	)
 
 	var err error
```
kylelemons, line 28: there is no longer a defer, so the predeclared `err` is no longer justified.

runtimebp/cpu.go
```diff
@@ -19,81 +19,37 @@ import (
...
-	}()
+
+	// Default to the standard runtime package value.
+	defaultCPUs := float64(runtime.NumCPU())
 
 	// Big enough buffer to read the number in the file wholly into memory.
 	buf := make([]byte, 1024)
 	var quota, period float64
 
 	quota, err = readNumberFromFile(quotaPath, buf)
```
kylelemons, line 37: `quota, err :=`

runtimebp/cpu.go
```diff
@@ -19,81 +19,37 @@ import (
...
-			n = float64(runtime.NumCPU())
-		}
-	}()
-
-	// Big enough buffer to read the number in the file wholly into memory.
-	buf := make([]byte, 1024)
-	var shares float64
-
-	shares, err = readNumberFromFile(sharesPath, buf)
+	period, err = readNumberFromFile(periodPath, buf)
```
kylelemons, line 47: `period, err :=`

kylelemons (changes requested):
With no "cpu shares" fallback, doesn't this code no longer handle sub-1000 millicpu allocations in kubernetes?

SuperQ:
@kylelemons The min should still be 1, so anything less than 1000m should get 1 GOMAXPROCS. That's the best we can do with POSIX thread granularity.

runtimebp/cpu.go
```diff
@@ -19,81 +19,34 @@ import (
 // it will fallback to runtime.NumCPU() instead.
 //
 // When fallback happens, it also prints the reason to stderr.
```
fishy, line 21: you changed the behavior so the comments should also be updated. you also removed the logging when fallback happens, which I think can still be useful.

also this function, as a public API from `runtimebp`, is supposed to provide container-aware version of `runtime.NumCPU`. if a container requested 2 cpu running on a 48 cpu physical machine, is 48 the expected/correct return for this api? maybe we just want to change the behavior of `runtimebp.GOMAXPROCS`, but keep `runtimebp.NumCPU` intact (we can have some refactoring to reduce code duplication, if we do believe GOMAXPROCS should not fallback to shares).

fishy (changes requested):
since this is something totally override-able by services, I would prefer to actually have some experiments, and probably an internal one-pager, before we conclude that this is the correct default behavior we want.

for example, one issue I have with this approach is that if we have a k8s service that requested 2 cpu (but no cpu limit is set), pods deployed on different physical machines will perform differently based on how many actual cpu the physical node has, which sounds wrong to me as I would expect the pods from the same deploy to perform similarly.

runtimebp/cpu.go
SuperQ: Yes, "request" is a Kubernetes-specific concept. In Kubernetes, Request affects scheduling, and CFS shares as a side effect. It's intended that containers are allowed to use any CPU more than their "request" in the Kubernetes context. If there are competing containers, CFS will proportionally share the "extra" CPU time between containers based on those requests.

Since we don't know what the intended use the application is, we should not second guess and set limits that would be a major change to typical Go runtime.

Keeping the reading of quota, which is "limit" in a Kubernetes context is what we want to keep. With hard quota / limit, we have a much better idea on how many CPUs are intended to be given to the app, therefore overriding the usual number of runtime threads is a good idea.

SuperQ: Updated the docs.

runtimebp/cpu_test.go
```diff
@@ -59,6 +59,10 @@ func TestReadNumberFromFile(t *testing.T) {
 			Content: "123.456",
 			Error:   true,
 		},
+		"negative": {
+			Content:  "-1",
+			Expected: -1,
+		},
 	}
```
kylelemons, line 66: > @kylelemons The min should still be 1, so anything less than 1000m should get 1 GOMAXPROCS. That's the best we can do with POSIX thread granularity.

@superq Can you add a test case that covers this case?  It's not clear to me how we will end up setting GOMAXPROCS to 1 if we have a limit of, say, 500 milli cpu.  Does that show up as 5000 per 10 or something?  If so, then we'll return 0, not 1.

kylelemons (changes requested):

runtimebp/cpu.go
```diff
@@ -12,88 +12,40 @@ import (
...
-
-	shares, err = readNumberFromFile(sharesPath, buf)
+	period, err := readNumberFromFile(periodPath, buf)
 	if err != nil {
-		return
+		return defaultCPUs
 	}
 
-	return float64(shares) / denominator
+	return quota / period
```
kylelemons, line 48: 🔕 If this is the code path we hit for .5 cpu limit, then it'll round down to 0, so this would be `(quota+quota-1) / period` if we want to round up.

fishy: this function returns a float number (so it returns 0.5 instead of rounding), [`runtimebp.GOMAXPROCS`](https://pkg.go.dev/github.com/reddit/baseplate.go/runtimebp#GOMAXPROCS) is the one actually using it and do the rounding, and it rounds up.

runtimebp/cpu_test.go
SuperQ: As @fishy said, this function just returns the cpus/mllicpu as a float. The `runtimebp.defaultMaxProcsFormula()` function does the default rounding up.

runtimebp/cpu.go
```diff
@@ -12,88 +12,40 @@ import (
...
-	var err error
-	defer func() {
-		if err != nil || n <= 0 {
-			// Fallback and log to stderr.
-			fmt.Fprintf(
-				os.Stderr,
-				"NumCPU: falling back to use shares: %v, %v\n",
-				n,
-				err,
-			)
```
fishy, line 37: I still think the logging about why the fallback happened is useful information and should not be removed. I'd suggest just remove the `numCPUSharesFallback` fallback and just do `n = float64(runtime.NumCPU())` on the next line instead.

fishy (changes requested):
Removing the shares fallback makes sense to me now, but I don't think we should remove the logging when fallbacks happen.

runtimebp/cpu.go
fishy: https://github.com/reddit/baseplate.go/pull/530 did the same thing so I don't think we need this PR any more?

kylelemons (approved):

runtimebp/cpu.go
```diff
@@ -12,46 +12,55 @@ import (
...
-	var quota, period float64
 
-	quota, err = readNumberFromFile(quotaPath, buf)
+	quota, err := readNumberFromFile(quotaPath, buf)
 	if err != nil {
-		return
+		fmt.Fprintf(
+			os.Stderr,
+			"NumCPU: Failed to read quota file, falling back to use runtime.NumCPU(): %v, %v\n",
+			n,
```
kylelemons, line 38: n doesn't exist anymore

SuperQ: Fixed

runtimebp/cpu.go
SuperQ: It's a good refactor, but this finishes the change by simplifying the code to be more explicit about the behavior.

runtimebp/cpu.go
```diff
@@ -12,46 +12,52 @@ import (
...
+			err,
+		)
+		return defaultCPUs
+	}
+
+	// CFS quota returns -1 if there is no limit, return the default.
+	if quota < 0 {
+		fmt.Fprintf(
+			os.Stderr,
+			"NumCPU: Quota file returned -1, falling back to use runtime.NumCPU(): %v\n",
```
fishy, line 47: nit: change `-1` to `%d` with `quota`. the `if` is only checking <0, so you can't be sure it's -1 here.

fishy (approved):
:fish: after https://github.com/reddit/baseplate.go/pull/520#discussion_r891704750 is addressed :)

runtimebp/cpu.go
fishy:
```suggestion
			"NumCPU: Quota file returned %d, falling back to use runtime.NumCPU(): %v\n",
			quota,
```

SuperQ: Agreed, good call.

fishy: oh wait it's float64, so you have to use `%v` instead of `%d`

fishy (approved):

PR #522, opened 2022-05-06
add graceful drain thrift middleware

marcoferrer:
## 💸 TL;DR
This PR introduces a thrift middleware that will attach a "drain" header to a response if the application is currently terminating

## 📜 Details
The specific header is `:drain` which is explicitly checked by envoy while handling an upstream response. This envoy feature was introduced in the following PR:
https://github.com/envoyproxy/envoy/pull/20236

[Jira]()

## ✅ Checks

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync/atomic"
 	"time"
 
 	"github.com/reddit/baseplate.go/batchcloser"
@@ -67,6 +68,8 @@ type Baseplate interface {
 
 	EdgeContextImpl() ecinterface.Interface
 	Secrets() *secrets.Store
+
+	State() runtimebp.State
 }
 
 // Server is the primary interface for baseplate servers.
@@ -128,6 +131,11 @@ type ServeArgs struct {
 func Serve(ctx context.Context, args ServeArgs) error {
 	server := args.Server
 
+	setRuntimeState := func(runtimebp.State) { /*noop*/ }
+	if v, ok := server.Baseplate().(interface{ setState(runtimebp.State) }); ok {
+		setRuntimeState = v.setState
+	}
+
 	// Initialize a channel to return the response from server.Close() as our
 	// return value.
 	shutdownChannel := make(chan error)
@@ -140,6 +148,8 @@ func Serve(ctx context.Context, args ServeArgs) error {
 	go runtimebp.HandleShutdown(
 		ctx,
 		func(signal os.Signal) {
+			setRuntimeState(runtimebp.StateDraining)
+
 			// Check if the server has a StopTimeout configured.
 			//
 			// If one is set, we will only wait for that duration for the server to
@@ -202,6 +212,8 @@ func Serve(ctx context.Context, args ServeArgs) error {
 			shutdownChannel <- err
 		},
 	)
+
+	setRuntimeState(runtimebp.StateServing)
 
 	// Start the server.
 	//
@@ -281,7 +293,8 @@ type NewArgs struct {
 // The returned context will be cancelled when the Baseplate is closed.
 func New(ctx context.Context, args NewArgs) (context.Context, Baseplate, error) {
 	cfg := args.Config.GetConfig()
-	bp := impl{cfg: cfg, closers: batchcloser.New()}
+	state := int64(runtimebp.StateStarting)
+	bp := impl{cfg: cfg, closers: batchcloser.New(), state: &state}
 
 	runtimebp.InitFromConfig(cfg.Runtime)
 
@@ -343,6 +356,15 @@ type impl struct {
 	cfg     Config
 	ecImpl  ecinterface.Interface
 	secrets *secrets.Store
+	state   *int64
+}
+
+func (bp impl) State() runtimebp.State {
+	return runtimebp.State(int(atomic.LoadInt64(bp.state)))
+}
+
+func (bp impl) setState(state runtimebp.State) {
+	atomic.StoreInt64(bp.state, int64(state))
 }
 
 func (bp impl) GetConfig() Config {
--- a/runtimebp/signal.go
+++ b/runtimebp/signal.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+
+	"github.com/reddit/baseplate.go"
 )
 
 // ShutdownHandler is the callback type used in HandleSignals.
@@ -15,6 +17,30 @@ var defaultSignals = []os.Signal{
 	os.Interrupt,
 	// Ref: https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods
 	syscall.SIGTERM,
+}
+
+// State is an enum representing the runtime state of the baseplate server
+type State int
+
+const (
+	StateUnknown State = iota
+	StateRunning
+	StateShuttingDown
+)
+
+// drainer is used to track the server state
+//
+// This drainer is updated by HandleShutdown before invoking the
+// user supplied ShutdownHandler
+var drainer = baseplate.Drainer()
+
+// ServerState returns the current runtimebp.State of the application
+func ServerState() State {
+	if drainer.IsHealthy(context.Background()) {
+		return StateRunning
+	}
+
+	return StateShuttingDown
 }
 
 // HandleShutdown register a handler to do cleanups for a graceful shutdown.
@@ -43,8 +69,10 @@ func HandleShutdown(ctx context.Context, handler ShutdownHandler, signals ...os.
 	)
 	select {
 	case signal := <-c:
+		drainer.Close()
 		handler(signal)
 	case <-ctx.Done():
+		drainer.Close()
 		// do nothing, just unblock the select block so it will return after it.
 	}
 }
--- a/runtimebp/signal.go
+++ b/runtimebp/signal.go
@@ -5,8 +5,6 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
-
-	"github.com/reddit/baseplate.go"
 )
 
 // ShutdownHandler is the callback type used in HandleSignals.
@@ -23,25 +21,12 @@ var defaultSignals = []os.Signal{
 type State int
 
 const (
+	//TODO docs
 	StateUnknown State = iota
-	StateRunning
-	StateShuttingDown
+	StateStarting
+	StateServing
+	StateDraining
 )
-
-// drainer is used to track the server state
-//
-// This drainer is updated by HandleShutdown before invoking the
-// user supplied ShutdownHandler
-var drainer = baseplate.Drainer()
-
-// ServerState returns the current runtimebp.State of the application
-func ServerState() State {
-	if drainer.IsHealthy(context.Background()) {
-		return StateRunning
-	}
-
-	return StateShuttingDown
-}
 
 // HandleShutdown register a handler to do cleanups for a graceful shutdown.
 //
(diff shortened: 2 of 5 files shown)


thriftbp/server_middlewares.go
```diff
@@ -477,3 +484,32 @@ func PrometheusServerMiddleware(method string, next thrift.TProcessorFunction) t
...
+
+	const envoyDrainHeader = ":drain"
+
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
+	var shouldDrain int32
+	go func() {
+		<-sig
+		atomic.StoreInt32(&shouldDrain, 1)
+	}()
```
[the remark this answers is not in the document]
fishy: I would slightly prefer to make a change to `runtimebp` package to provide this global singleton instead of reimplement it here.

my other concern is that, can you elaborate on what real difference this makes on the envoy level? is it to make sure that envoy does not send more requests to this server pod? the current implementation (the one not using global singleton) can totally be implemented in service code instead, can you implement this in your service first, test it in staging/run it in production, fix bugs/verify that it does what we expect it does, then upstream it to baseplate?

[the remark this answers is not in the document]
pacejackson: You can wire function calls into both the pre and post stop steps in baseplate.go (I think this would need to be in pre-stop). https://pkg.go.dev/github.com/reddit/baseplate.go#ServeArgs

Rather than using a global state, you could make a struct that stores the state and has methods for the shutdown handler and the middleware.

```go
const envoyDrainHeader = ":drain"

type EnvoyGracefulDrainHeader struct {
	shouldDrain *int32
}

func (e *EnvoyGracefulDrainHeader) Close() error {
	atomic.StoreInt32(e.shouldDrain, 1)	
}

func (e *EnvoyGracefulDrainHeader) ThriftProcessorMiddleware(name string, next thrift.TProcessorFunction) thrift.TProcessorFunction {
	return thrift.WrappedTProcessorFunction{
		Wrapped: func(ctx context.Context, seqID int32, in, out thrift.TProtocol) (ok bool, err thrift.TException) {
			if atomic.LoadInt32(e.shouldDrain) > 0 {
				if t, ok := out.(*thrift.THeaderProtocol); ok {
					t.SetWriteHeader(envoyDrainHeader, "true")
				}
			}

			return next.Process(ctx, seqID, in, out)
		},
	}
}
```

pacejackson (changes requested):
Could you also add a test for this and explain in the code comment what this header actually does.

thriftbp/server_middlewares.go
```diff
@@ -477,3 +484,32 @@ func PrometheusServerMiddleware(method string, next thrift.TProcessorFunction) t
 	}
 	return thrift.WrappedTProcessorFunction{Wrapped: process}
 }
+
+// EnvoyGracefulDrainHeader returns middleware which sets a drain signal header on server responses.
+func EnvoyGracefulDrainHeader() thrift.ProcessorMiddleware {
+
```
kylelemons, line 490:
```suggestion
```

thriftbp/server_middlewares.go
```diff
@@ -477,3 +484,32 @@ func PrometheusServerMiddleware(method string, next thrift.TProcessorFunction) t
...
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
+	var shouldDrain int32
+	go func() {
+		<-sig
+		atomic.StoreInt32(&shouldDrain, 1)
+	}()
+
+	return func(name string, next thrift.TProcessorFunction) thrift.TProcessorFunction {
+
```
kylelemons, line 502:
```suggestion
```

thriftbp/server_middlewares.go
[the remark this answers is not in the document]
pacejackson: > I would slightly prefer to make a change to runtimebp package to provide this global singleton instead of reimplement it here.

This also sounds like a reasonable approach, but maybe it should be on the `Baseplate` object rather than a global? Reason being is the function in runtimebp is really just a helper, while I think this notion of the server running belongs in baseplate.Serve. Maybe add a `ServerIsRunning() bool` function to the Baseplate?

kylelemons (changes requested):
Personally I'm happy to keep the API surfaces small while we're still in v0 of baseplate.go, and consider moving to runtimebp when / if we need to share this elsewhere.  But I don't have a strong opinion either way.

The main thing I think that should be clarified is why we should do this rather than through the Kubernetes readiness checks.  The contract for kubernetes services is that once they start draining they should stop marking themselves as ready, and we should make sure our envoys handle that gracefully.  We can do both, of course, since this is likely to be a more responsive indicator than the polling-based readiness probe, but at its core we should ensure the kubernetes contract is respected.

marcoferrer:
So a couple of things.
* Is safe to say we couldnt make this middleware a default with this pattern since theres no way to pass it to the baseplate shutdown args. Just clarifying to be sure Im not missing anything.
* What happens if one of the other shutdown handlers is slow to execute? Are they ran in parallel? We would need this to be invoked immediately for it to have the intended effect

pacejackson:
what about this? https://github.com/reddit/baseplate.go/blob/master/drainer.go

marcoferrer:
@fishy Thanks! I didnt see that. But would be happy to use that in the impl instead.

@kylelemons We're relying on this in thing service + k8s readiness checks + k8s endpoint conditions through xds

pacejackson:
> Is safe to say we couldnt make this middleware a default with this pattern since theres no way to pass it to the baseplate shutdown args. Just clarifying to be sure Im not missing anything.

The way it currently is, that's true. It's probably a good option for testing it in your own service though (or if we just don't want to bake it into the default).

> What happens if one of the other shutdown handlers is slow to execute? Are they ran in parallel? We would need this to be invoked immediately for it to have the intended effect

No, they run in serial, so any shutdown handler being slow blocks the following ones. I think this would be ok though? It really only needs to run before you actually call the "Close" method on the server, not *first* right? Adding it into the PreStop would accomplish that (and if you add it to the beginning of that list it should run first anyways).

I think having the server state in `baseplate.Baseplate` might be the right way forward? We can explicitly change that at the beginning of the shutdown handler in `baseplate.Serve` and can share that with any components that care about it. @fishy @kylelemons does that sound reasonable if we decide we want to bake this into baseplate itself?

fishy:
My understanding is that this is a limitation that when we use envoy as the lb over k8s' built-in lb, it loses the access k8s' lb has regarding health checks, so it does not know a pod is shutting down and should not receive any further requests like k8s' lb does, and rely on this sidechannel to have that knowledge.

If we plan to deploy envoys to more services (we probably should eventually, we've seen a lot of issues with k8s' builtin lb), then this does make sense to be a default middleware as envoy will be part of our default configuration.

marcoferrer:
So its less of a limitation of more of an optimization / safety measure. If there is a lot of pod churn there is slight delay of x ms before envoy has a new a endpoint list pushed to it. Also if there is a xds service discovery outage this would prevent envoy from routing to terminating pods.

fishy:
if this is not a limitation per se then I'd slightly prefer andrew's proposed solution to just provide a pre-shutdown closer services can choose to add it themselves as that's the cleaner way to implement it.

or just implement it in your service first, see how much difference it actually makes, then we would have some data to decide whether this needs to be a default middleware or just a nice-to-have one.

fishy (changes requested):
make harold happy. I think the next step should be that you work on this on your own service, gather some real world data, then we can come back to this PR to decide whether we want to make this default and what's the approach we should use?

runtimebp/signal.go
```diff
@@ -17,6 +19,30 @@ var defaultSignals = []os.Signal{
...
+var drainer = baseplate.Drainer()
+
+// ServerState returns the current runtimebp.State of the application
+func ServerState() State {
+	if drainer.IsHealthy(context.Background()) {
+		return StateRunning
+	}
+
+	return StateShuttingDown
+}
```
pacejackson, line 44: I don't really think this should be here like this, to me `HandleShutdown` should just be helper and this state should be managed in `baseplate.Serve` on the passed in `Baseplate` object rather than as a global.

pacejackson (changes requested):

marcoferrer:
> make harold happy. I think the next step should be that you work on this on your own service, gather some real world data, then we can come back to this PR to decide whether we want to make this default and what's the approach we should use?

So this is something we're already using in `thing`. We're going to be testing this go impl next week in our services but I can confirm that it was valuable for mitigating any errors during scale down events.

runtimebp/signal.go
```diff
@@ -17,6 +17,17 @@ var defaultSignals = []os.Signal{
...
+// State is an enum representing the runtime state of the baseplate server
+type State int
+
+const (
+	//TODO docs
+	StateUnknown State = iota
+	StateStarting
+	StateServing
+	StateDraining
+)
```
pacejackson, line 29: 🔕 I'm not sure that it makes sense to have this in runtimebp while the rest of it is in baseplate? It should probably all just be in there IMO.

thriftbp/server_middlewares.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/reddit/baseplate.go/runtimebp"
```
pacejackson, line 12: nit, this should be in the block below

thriftbp/server_middlewares.go
```diff
@@ -53,6 +54,8 @@ type DefaultProcessorMiddlewaresArgs struct {
 	//
 	// If it's not set, the global one from ecinterface.Get will be used instead.
 	EdgeContextImpl ecinterface.Interface
+
+	RuntimeStateReader func() runtimebp.State
```
pacejackson, line 58: missing doc string

thriftbp/server_middlewares.go
```diff
@@ -477,3 +483,20 @@ func PrometheusServerMiddleware(method string, next thrift.TProcessorFunction) t
 	}
 	return thrift.WrappedTProcessorFunction{Wrapped: process}
 }
+
+// EnvoyGracefulDrainHeader returns middleware which sets a drain signal header on server responses.
```
pacejackson, line 487:
```suggestion
// EnvoyGracefulDrainHeader returns middleware which sets the Envoy drain signal header on server responses.
```

runtimebp/signal.go
```diff
@@ -17,6 +17,17 @@ var defaultSignals = []os.Signal{
 	syscall.SIGTERM,
 }
 
+// State is an enum representing the runtime state of the baseplate server
+type State int
+
+const (
+	//TODO docs
```
pacejackson, line 24: `TODO` 😅

baseplate.go
```diff
@@ -128,6 +131,11 @@ type ServeArgs struct {
 func Serve(ctx context.Context, args ServeArgs) error {
 	server := args.Server
 
+	setRuntimeState := func(runtimebp.State) { /*noop*/ }
+	if v, ok := server.Baseplate().(interface{ setState(runtimebp.State) }); ok {
```
pacejackson, line 135: 🔕 
I'm not 100% sure about this, but I feel like `setState` should just be a hidden part of the `Baseplate` interface, especially since `State` won't work properly without it. Looking back, we probably should have just made `Baseplate` a struct with non-public fields rather than an interface with a hidden implementation where we would have likely just done it this way to begin with 🤷

pacejackson (changes requested):

marcoferrer:
Its low value for us to add this to baseplate only to have to manually instrument it in every service that will have to be fronted by envoy (not limiting to core services). If we're sticking to manual instrumentation we can add it the internal core platform library and source for there. Closing out as a noop

fishy:
> Its low value for us to add this to baseplate only to have to manually instrument it in every service that will have to be fronted by envoy (not limiting to core services). If we're sticking to manual instrumentation we can add it the internal core platform library and source for there. Closing out as a noop

I think if the data shows that this is valueable with envoy, we'll likely want it to be a default middleware (go back to one that's more similar to the original approach). but that is probably still a few months away (we probably want to enable envoy for the majority of services first).

PR #528, opened 2022-06-03
redisbp: Add prometheus metrics for latency

fishy:


prometheusbp/promtest/testutil.go
```diff
@@ -10,24 +10,36 @@ import (
...
-	got := p.getValue()
+	got, _ := p.getValues()
 	got -= float64(p.initValue)
 	if got != delta {
 		p.tb.Errorf("%s metric delta: wanted %v, got %v", p.name, delta, got)
 	}
 }
 
+// CheckHistoCountDelta checks that the number of samples of histogram is the exactly delta from when Helper was called.
+func (p *PrometheusMetricTest) CheckHistoCountDelta(delta int) {
```
kylelemons, line 33: I don't like abbreviating words in exported API names.

CheckSampleDelta?

prometheusbp/promtest/testutil.go
```diff
@@ -10,24 +10,36 @@ import (
...
+	p.tb.Helper()
+
+	_, got := p.getValues()
```
kylelemons, line 36: I don't really like conflating getValues to return a second argument all the time...

it makes it look like errors are being disccarded.

Maybe make a new getSampleCount helper?

If not, `getValues -> valueAndSamples` maybe to make it clear at the call site?

kylelemons (approved):
Just some naming suggestions

prometheusbp/promtest/testutil.go
fishy: I actually think the current implementation of `CheckDelta` for histogram is a mistake, what does a delta value for a histogram mean, exactly? If we just change histogram to store the sample count in its `initValue` and change its `CheckDelta` to check the sample count delta, that addresses both of your concerns, what do you think?

technically that is a breaking change (if someone has some code actually using `CheckDelta` on histograms it will break), but I really don't know how that is usable right now.

cc @JessicaGreben

fishy: oh actually someone are actually using it to check the observed latency seconds. ok I'll just spell `Histogram` fully instead. I don't think the name `CheckSampleDelta` emphasize enough that it's for histograms only.

redis/db/redisbp/hooks_test.go
```diff
@@ -1,27 +1,34 @@
...
 )
 
 func TestSpanHook(t *testing.T) {
 	ctx, _ := thriftbp.StartSpanFromThriftContext(context.Background(), "foo")
-	hooks := redisbp.SpanHook{ClientName: "redis"}
+	hooks := SpanHook{ClientName: "redis"}
 	statusCmd := redis.NewStatusCmd(ctx, "ping")
 	stringCmd := redis.NewStringCmd(ctx, "get", "1")
-	stringCmd.SetErr(redis.Nil)
+	stringCmd.SetErr(nil)
```
fishy, line 21: @pacejackson I changed this because the [doc](https://pkg.go.dev/github.com/go-redis/redis/v8#Nil) says `redis.Nil` is what redis returns when the key does not exist, which makes me think it's an error but it's actually not and the error we get in the hook is nil, which is kinda confusing so I think just use `nil` here is more clear.

konradreiche (approved):

pacejackson (approved):

prometheusbp/promtest/testutil.go
kylelemons: Both histograms and summaries have samples, so it makes sense for more than just histograms.

kylelemons (approved):

mterwill (approved):

prometheusbp/promtest/testutil.go
fishy: oh shit I didn't see this comment before merging it. will revert and redo it.

PR #533, opened 2022-06-08
thrifttest: fix bug where Server.Close hangs

pacejackson:
## 💸 TL;DR
Update thrifttest.Server.Close to close the ClientPool before
trying to close the server. If you don't do this, then the
underlying server.Close will hang while it waits for the client
to close (which won't happen until the server closes) unless
you explicitly set thift.ServerStopTimeout

## ✅ Checks

### DIFF
--- a/thriftbp/thrifttest/server.go
+++ b/thriftbp/thrifttest/server.go
@@ -2,7 +2,6 @@ package thrifttest
 
 import (
 	"context"
-	"io"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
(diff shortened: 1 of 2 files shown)


fishy (approved):
fyi the server close will also be "fixed" in thrift v0.17.0 via https://github.com/apache/thrift/commit/2fa907e84b5bf29c263c4cde657e99c2e894562f, but to take advantage of that we would still need to set `thrift.ServerStopTimeout` to a small value in the test.

pacejackson:
@fishy ya I ran into this in one of my projects and was able to fix it with `thrift.ServerStopTimeout` so it's not blocking me atm.

konradreiche (approved):

thriftbp/thrifttest/server.go
```diff
@@ -113,11 +114,14 @@ func (s *Server) Start(ctx context.Context) {
...
+	closers := make([]io.Closer, 0, 3)
+	// close the ClientPool first so the server doesn't hang waiting for it to
+	// close while trying to close itself.
 	if s.ClientPool != nil {
-		closers.Add(s.ClientPool)
+		closers = append(closers, s.ClientPool)
 	}
-	return closers.Close()
+	closers = append(closers, s.Server, s.Baseplate())
+	return batchcloser.New(closers...).Close()
```
kylelemons, line 124: 🔕 so the only change is the order of closing?  Why is this switching away from using batchcloser.New at the beginning?

kylelemons (approved):

thriftbp/thrifttest/server.go
pacejackson: changed

PR #534, opened 2022-06-15
configbp: Add Int64String type

fishy:
This is a helper type used in yaml configs, mainly to deal with a helm
bug [1] that large int64 values in yaml config could have precision loss
when going through helm, so we would want to put the string versions of
them in yaml to preserve the precision.

[1]: https://github.com/helm/helm/issues/11045

### DIFF
--- a/configbp/yaml.go
+++ b/configbp/yaml.go
@@ -1,0 +1,33 @@
+package configbp
+
+import (
+	"fmt"
+	"strconv"
+
+	"gopkg.in/yaml.v2"
+)
+
+// Int64String is an int64 type that can be yaml deserialized from strings.
+//
+// It's useful when the yaml config goes through helm,
+// which could cause precision loss on large int64 numbers:
+// https://github.com/helm/helm/issues/11045
+type Int64String int64
+
+var (
+	_ yaml.Unmarshaler = (*Int64String)(nil)
+)
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (i *Int64String) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var s string
+	if err := unmarshal(&s); err != nil {
+		return err
+	}
+	i64, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		return fmt.Errorf("cannot parse %q as int64: %v", s, err)
+	}
+	*i = Int64String(i64)
+	return nil
+}
--- a/configbp/yaml_test.go
+++ b/configbp/yaml_test.go
@@ -1,0 +1,108 @@
+package configbp_test
+
+import (
+	"fmt"
+	"math"
+	"strconv"
+	"strings"
+	"testing"
+	"testing/quick"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/reddit/baseplate.go/configbp"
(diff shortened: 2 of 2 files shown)


configbp/yaml.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"gopkg.in/yaml.v2"
+)
+
+// Int64 is an int64 type that can be yaml deserialized from strings.
+//
+// It's useful when the yaml config goes through helm,
+// which could cause precision loss on large int64 numbers:
+// https://github.com/helm/helm/issues/11045
+type Int64 int64
```
kylelemons, line 15: HelmInt64 or Int64String?

kylelemons (changes requested):
Is this really worth checking into baseplate.go to work around helm, which is not officially supported except for ads, which has its own helper libraries?

configbp/yaml.go
fishy: Renamed to `YAMLInt64`.

while the bug is on helm, the main difference between this type and vanilla `int64` is on how it handles yaml deserialization, and also it's quite possible that tools other than helm having a similar bug.

fishy (commented):
While it's true that ads is one of the bigger internal users of helm, I don't think we are the only one.

configbp/yaml.go
kylelemons: This is not the "right" way to store int64s in YAML, so I don't like the name YAMLInt64.  `HelmInt64` tells when to use this, and `Int64String` explains that there are strings involved.

Saying "it's also quite possible that other tools that's not helm could have this bug" is a bit specious.  Sure, _all_ code could have a bug, but unless we have evidence that other tools that we use share this specific bug, it seems premature to assume they would.

kylelemons (changes requested):
> While it's true that ads is one of the bigger internal users of helm, I don't think we are the only one.

Helm is not a "supported" method for deploying services anymore, though.  There are a few lingering ones around, but _very_ few of them are likely to be in Go outside of ads because of the timing of the starlark rollout and Go appearing in the cookie-cutter.

configbp/yaml.go
fishy: fair, `Int64String` it is.

our log ingestor has this exact same bug, because json is from js, and js only defined a single number type that's float64. and yaml, in some sense, is just an extended json type.

configbp/yaml.go
```diff
@@ -0,0 +1,33 @@
+package configbp
+
+import (
+	"fmt"
+	"strconv"
+
+	"gopkg.in/yaml.v2"
+)
+
+// Int64String is an int64 type that can be yaml deserialized from strings.
```
kylelemons, line 10: 🔕 (nit)

```suggestion
// Int64String is an int64 type that can be deserialized from yaml strings.
```

kylelemons (approved):

configbp/yaml.go
fishy: I feel that "yaml string" doesn't really mean anything (or it means a string that's the yaml content, which is not what we meant here), while "yaml deserialize" could be used to mean the action of deserialize yaml content?

pacejackson (approved):

PR #537, opened 2022-06-22
[BASE-29] Create internalv2compat library

kylelemons:
## 💸 TL;DR
This PR introduces the hooks necessary for the upcoming v2 compatibility module.
This does not introduce any new user-facing APIs, and it is very clearly indicated that this is not to be used by users.

## 📜 Details
See BASE-29.

## ✅ Checks

### DIFF
--- a/internalv2compat/compat.go
+++ b/internalv2compat/compat.go
@@ -1,7 +1,7 @@
 // Package internalv2compat is an INTERNAL ONLY library provided for transitional projects that need
 // to use baseplate v2 and v0 in the same module.
 //
-// DO NOT USE THIS LIBRARY DIRECTLY.
+// DO NOT USE THIS LIBRARY DIRECTLY.  Breaking changes may be made to this package at any time.
 //
 // Deprecated: This is an internal library and should not be used directly.
 package internalv2compat
--- a/internalv2compat/compat.go
+++ b/internalv2compat/compat.go
@@ -41,7 +41,7 @@ var v0logDisabled uint64
 // SetGlobalLogger allows baseplate v0 to set the global logger.
 func SetGlobalLogger(logger *zap.SugaredLogger) {
 	if atomic.LoadUint64(&v0logDisabled) > 0 {
-		globalLogger.Warnf("Suppressing global log override; use v2 loging instead")
+		globalLogger.Warn("ineffectual call to SetGlobalLogger; baseplate.go v2 compatibility mode is active")
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


internalv2compat/compat.go
```diff
@@ -0,0 +1,55 @@
...
+func GlobalLogger() *zap.SugaredLogger {
+	return globalLogger
+}
+
+var v0logDisabled uint64
+
+// SetGlobalLogger allows baseplate v0 to set the global logger.
+func SetGlobalLogger(logger *zap.SugaredLogger) {
+	if atomic.LoadUint64(&v0logDisabled) > 0 {
+		globalLogger.Warnf("Suppressing global log override; use v2 loging instead")
```
konradreiche, line 44:
```suggestion
		globalLogger.Warnf("Suppressing global log override; use v2 logging instead")
```
or possibly a rewording: `ineffectual call to SetGlobalLogger, use OverrideLogger`. V2 logging also has a double connotation at Reddit :slightly_smiling_face:

konradreiche (changes requested):

konradreiche (approved):

mterwill (approved):

internalv2compat/compat.go
```diff
@@ -0,0 +1,55 @@
...
+func GlobalLogger() *zap.SugaredLogger {
+	return globalLogger
+}
+
+var v0logDisabled uint64
+
+// SetGlobalLogger allows baseplate v0 to set the global logger.
+func SetGlobalLogger(logger *zap.SugaredLogger) {
+	if atomic.LoadUint64(&v0logDisabled) > 0 {
+		globalLogger.Warn("ineffectual call to SetGlobalLogger; baseplate.go v2 compatibility mode is active")
```
fishy, line 44: we probably want to emit a prometheus counter for this so it can be monitored/alerted?

fishy (approved):

pacejackson (approved):

internalv2compat/compat.go
kylelemons: My plan is to emit the prometheus metric from the compat library itself, WDYT?

fishy: isn't this "the compat library"? do you have a slightly more concrete example of what you have in mind?

fishy: :fish:

kylelemons: Discussed offline

PR #540, opened 2022-06-27
thriftbp: Fix IDL exceptions in client middlewares

fishy:
In client middlewares, any error defined in thrift IDL is actually not
returned via the error return, but written as part of the result
TStruct. As a result, none of the current client middlewares actually
get the exceptions defined in IDL correctly (this includes
baseplate.Error).

Add a helper function to get those errors out of the result TStruct via
reflection.

### DIFF
--- a/thriftbp/client_middlewares.go
+++ b/thriftbp/client_middlewares.go
@@ -3,6 +3,7 @@ package thriftbp
 import (
 	"context"
 	"errors"
+	"reflect"
 	"strconv"
 	"time"
 
@@ -208,7 +209,7 @@ func MonitorClient(args MonitorClientArgs) thrift.ClientMiddleware {
 				defer func() {
 					span.FinishWithOptions(tracing.FinishOptions{
 						Ctx: ctx,
-						Err: s.Wrap(err),
+						Err: s.Wrap(getClientError(result, err)),
 					}.Convert())
 				}()
 
@@ -281,7 +282,7 @@ func Retry(defaults ...retry.Option) thrift.ClientMiddleware {
 					func() error {
 						var err error
 						lastMeta, err = next.Call(ctx, method, args, result)
-						return err
+						return getClientError(result, err)
 					},
 					defaults...,
 				)
@@ -361,11 +362,12 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
 
 				defer func() {
 					var baseplateStatusCode, baseplateStatus string
-					exceptionTypeLabel := stringifyErrorType(err)
-					success := prometheusbp.BoolString(err == nil)
-					if err != nil {
+					finalErr := getClientError(result, err)
+					exceptionTypeLabel := stringifyErrorType(finalErr)
+					success := prometheusbp.BoolString(finalErr == nil)
+					if finalErr != nil {
 						var bpErr baseplateErrorCoder
-						if errors.As(err, &bpErr) {
+						if errors.As(finalErr, &bpErr) {
 							code := bpErr.GetCode()
 							baseplateStatusCode = strconv.FormatInt(int64(code), 10)
 							if status := baseplate.ErrorCode(code).String(); status != "<UNSET>" {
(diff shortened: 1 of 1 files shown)


thriftbp/client_middlewares.go
```diff
@@ -398,3 +400,28 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
...
+	}
+	v := reflect.ValueOf(result)
+	v = reflect.Indirect(v)
+	if v.Kind() != reflect.Struct {
+		return nil
+	}
+	resultField := v.FieldByName("Success").Interface()
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i).Interface()
+		if field == resultField {
```
kylelemons, line 416: It looks like this code is trying to skip any struct fields named "Success".  Typically I would do that by `typ := v.Type(); ... if typ.Field(i).Name == "Success" { continue }`.

Right now this skips any field whose value matches the Success field, so if the success field is a boolean true, it'll skip any fields that are true.  This might function as desired but it's pretty weird, and may have unexpected panics on comparing types that can't be compared with `==`

thriftbp/client_middlewares.go
```diff
@@ -398,3 +400,28 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
...
+			continue
+		}
+		if tExc, ok := field.(thrift.TException); ok {
+			if tExc == nil || tExc.TExceptionType() != thrift.TExceptionTypeCompiled {
+				continue
+			}
+			return tExc
+		}
```
kylelemons, line 424: 🔕 I think type switches are cleaner, but they take up a bit more vertical space.

```suggestion
		switch t := field.(type) {
			case thrift.TException:
				if tExc == nil || tExc.TExceptionType() != thrift.TExceptionTypeCompiled {
					continue
				}
				return tExc
		}
```

kylelemons (changes requested):
(I think the only blocking comment is using reflect.Type to skip the Success field not the value)

fishy:
> I think this only gets us observability right? Why not parse and return the error in an additional bottom-layer middleware so that BaseplateErrorWrapper & Retry etc. can take effect as well?

I just forgot about retry. added.

adding a bottom-layer middleware is non-trivial (it requires thrift compiler change as well)

thriftbp/client_middlewares.go
```diff
@@ -398,3 +400,28 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
 		}
 	}
 }
+
+func getClientError(result thrift.TStruct, err error) error {
```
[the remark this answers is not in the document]
kylelemons: I meant more "point to in a comment" but thanks for the out of band links 👍 

This makes me wonder how many folks like me misunderstood how to "return thrift errors" and just returned them like in Go.

thriftbp/client_middlewares.go
```diff
@@ -398,3 +400,25 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
...
+	if err != nil {
+		return err
+	}
+	v := reflect.Indirect(reflect.ValueOf(result))
+	if v.Kind() != reflect.Struct {
+		return nil
+	}
+	typ := v.Type()
+	for i := 0; i < v.NumField(); i++ {
+		if typ.Field(i).Name == "Success" {
```
kylelemons, line 414: Is this just an optimization? I feel like line 419 will also skip all of the success fields since they won't have any methods.

kylelemons (approved):

thriftbp/client_middlewares.go
fishy: this is more of "in theory people can define an exception type as their rpc return type and we could accidentally use that as error" (e.g. currently in baseplate.thrift's `is_healthy` endpoint, the return type is `bool`, but in theory someone can define an endpoint with a return type being also an exception)

konradreiche (approved):

pacejackson (approved):

thriftbp/client_middlewares.go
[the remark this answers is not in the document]
fishy: Added comment. Also as said in slack for handler implementations just return the error is fine (and that's the only way to do it :) )

PR #541, opened 2022-06-29
thriftbp: Fix panic caused by typed-nil

fishy:
cccd1f1 would cause panic because of typed nil:

    panic: value method path/to/baseplate.Error.TExceptionType called using nil *Error pointer

Check reflect value's IsZero instead.

### DIFF
--- a/thriftbp/client_middlewares.go
+++ b/thriftbp/client_middlewares.go
@@ -430,8 +430,11 @@ func getClientError(result thrift.TStruct, err error) error {
 		if typ.Field(i).Name == "Success" {
 			continue
 		}
-		field := v.Field(i).Interface()
-		tExc, ok := field.(thrift.TException)
+		field := v.Field(i)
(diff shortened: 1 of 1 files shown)


thriftbp/client_middlewares.go
```diff
@@ -430,6 +430,9 @@ func getClientError(result thrift.TStruct, err error) error {
 		if typ.Field(i).Name == "Success" {
 			continue
 		}
+		if v.Field(i).IsZero() {
```
fishy, line 433: note: there are both `IsZero` and `IsNil`, reading the doc I _think_ `IsZero` is more robust in this case.

pacejackson (approved):
Someone who knows more than me would have to verify your `IsZero` vs `IsNil` question 😅

konradreiche (approved):

pacejackson (approved):

PR #542, opened 2022-07-08
upgrade log to panic

migleeson:
## 💸 TL;DR
Currently, if the `admin.Serve()` crashes, the resulting error is logged with a warning and the process keeps chugging along without everything it provides (like metrics). Upgrading the log to a Panicw will cause the go routine to panic and crash the program meaning the software can never be in a state where the metics server is not also running successfully.

Test suite ran with `make test`

### DIFF
--- a/httpbp/admin.go
+++ b/httpbp/admin.go
@@ -21,6 +21,6 @@ func ServeAdmin(healthCheck HandlerFunc) {
 	if err := admin.Serve(); errors.Is(err, http.ErrServerClosed) {
 		log.Info("httpbp: server closed")
 	} else {
-		log.Panicw("httpbp: admin serving exited", "err", err)
+		log.Panicw("httpbp: admin serving failed", "err", err)
 	}
 }
--- a/httpbp/admin.go
+++ b/httpbp/admin.go
@@ -1,6 +1,9 @@
 package httpbp
 
 import (
+	"errors"
+	"net/http"
+
 	"github.com/reddit/baseplate.go/internal/admin"
 	"github.com/reddit/baseplate.go/log"
 )
@@ -15,5 +18,9 @@ import (
 // This function blocks, so it should be run as its own goroutine.
 func ServeAdmin(healthCheck HandlerFunc) {
 	admin.Mux.Handle("/health", handler{handle: healthCheck})
-	log.Warnw("httpbp: admin serving exited", "err", admin.Serve())
+	if err := admin.Serve(); errors.Is(err, http.ErrServerClosed) {
+		log.Info("httpbp: server closed")
+	} else {
+		log.Panicw("httpbp: admin serving exited", "err", err)
+	}
 }
--- a/httpbp/admin.go
+++ b/httpbp/admin.go
@@ -19,7 +19,7 @@ import (
 func ServeAdmin(healthCheck HandlerFunc) {
 	admin.Mux.Handle("/health", handler{handle: healthCheck})
 	if err := admin.Serve(); errors.Is(err, http.ErrServerClosed) {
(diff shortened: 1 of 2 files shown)


thriftbp/admin.go
```diff
@@ -13,5 +13,5 @@ import (
 //
 // This function blocks, so it should be run as its own goroutine.
 func ServeAdmin() {
-	log.Warnw("thriftbp: admin serving exited", "err", admin.Serve())
+	log.Panicw("thriftbp: admin serving exited", "err", admin.Serve())
```
fishy, line 16: I don't believe this is the best way to do it.

There are historical reasons that we don't make this log level higher: the http server's `Serve` function [always returns non-nil error eventually](https://pkg.go.dev/net/http#Server.Serve). Currently we do not actually ever shutdown the admin http server so it won't happen, but if we ever change the `internal/admin`'s api to shut it down during graceful shutdown, this panic will cause the graceful shutdown process to panic and interrupt other cleanups.

The thing you want to guard against is that the 6060 port is unavailable so `admin.Serve` fails early (vs. normal shutdown). To achieve that, we should actually check the error returned by `admin.Serve` and only panic if it's not [`http.ErrServerClosed`](https://pkg.go.dev/net/http#ErrServerClosed). Something like this:

```go
if err := admin.Serve(); errors.Is(err, http.ErrServerClosed) {
  log.Infow(...)
} else {
  log.Panicw(...)
}
```

Also you want to do the same in `httpbp.ServeAdmin` as well.

(another historical reason of not make the log level higher is that admin port used to only serve pprof, which is really not critical enough to bring the whole pod down. now with prometheus I think it's justified)

konradreiche: +1 on that

konradreiche (changes requested):

pacejackson (changes requested):
💅

httpbp/admin.go
```diff
@@ -15,5 +18,9 @@ import (
 // This function blocks, so it should be run as its own goroutine.
 func ServeAdmin(healthCheck HandlerFunc) {
 	admin.Mux.Handle("/health", handler{handle: healthCheck})
-	log.Warnw("httpbp: admin serving exited", "err", admin.Serve())
+	if err := admin.Serve(); errors.Is(err, http.ErrServerClosed) {
+		log.Info("httpbp: server closed")
+	} else {
+		log.Panicw("httpbp: admin serving exited", "err", err)
```
fishy, line 24: maybe you want a word stronger than "exited"? maybe "failed"?

fishy (approved):

httpbp/admin.go
migleeson: I kept the language from the initial log message. Agreed that a stronger message is better. I'll update

pacejackson (approved):

konradreiche (approved):

httpbp/admin.go
```diff
@@ -15,5 +18,9 @@ import (
 // This function blocks, so it should be run as its own goroutine.
 func ServeAdmin(healthCheck HandlerFunc) {
 	admin.Mux.Handle("/health", handler{handle: healthCheck})
-	log.Warnw("httpbp: admin serving exited", "err", admin.Serve())
+	if err := admin.Serve(); errors.Is(err, http.ErrServerClosed) {
+		log.Info("httpbp: server closed")
```
fishy, line 22: actually, can we change this log (and the thriftbp one) to "*admin* server closed"?

the current log sounds like the main httpbp/thriftbp server was closed.

migleeson: done

fishy (approved):

PR #543, opened 2022-07-08
runtimebp: Honor GOMAXPROCS environment variable if set

fishy:
This makes overriding GOMAXPROCS easier, and be more consistent with the
expected behavior.

### DIFF
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -21,7 +21,20 @@ type Config struct {
 }
 
 // InitFromConfig sets GOMAXPROCS using the given config.
+//
+// NOTE: If the GOMAXPROCS environment variable is set,
+// this function will skip setting GOMAXPROCS,
+// even if the environment variable is set to some bogus value
+// (in that case it will be set to number of physical CPUs).
 func InitFromConfig(cfg Config) {
-	prev, current := GOMAXPROCS(1, math.MaxInt)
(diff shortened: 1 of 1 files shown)


nanassito (approved):
Thank you !

konradreiche (approved):

runtimebp/cpu.go
```diff
@@ -131,7 +131,19 @@ func defaultMaxProcsFormula(n float64) int {
 // in bound of [min, max].
 //
 // Currently the default formula is NumCPU() rounding up.
+//
+// NOTE: If the GOMAXPROCS environment variable is set,
+// this function does nothing besides printing out a log,
+// even if the environment variable is set to some bogus value
+// (in that case it will be set to number of physical CPUs).
 func GOMAXPROCS(min, max int) (oldVal, newVal int) {
+	if v, ok := os.LookupEnv("GOMAXPROCS"); ok {
```
pacejackson, line 140: 🔕 I could be off base, but I feel like _this_ specific function should still override GOMAXPROCS and this check should maybe be in baseplate.New before calling this?

pacejackson (changes requested):

runtimebp/cpu.go
fishy: Good point. moved to `InitFromConfig` (that's the function called by `baseplate.New`)

pacejackson (approved):

PR #544, opened 2022-07-11
Update kafka offset commit frequency to 5s

sydjryan:
The kafka library we use [defaults](https://github.com/Shopify/sarama/blob/main/config.go#L370-L387) to committing offsets every 1 second. Frequent offset commits can be hard on kafka clusters if there are many consumers, so lets change the default to commit a bit less frequently (Kafka's default is every 5s). This was successfully tested on one baseplate service already, so I'm confident it works.

### DIFF
--- a/kafkabp/config.go
+++ b/kafkabp/config.go
@@ -100,6 +100,12 @@ type ConsumerConfig struct {
 	// or it might make things worse.
 	// You are advised to test before using non-empty rack id in production.
 	RackID RackIDFunc `yaml:"rackID"`
+
+	// Optional. Defaults to 5s. Values are in seconds.
+	//
+	// This affects how often the kafka consumer writes its offsets back to
+	// the cluster.
+	AutoCommitInterval int `yaml:"autoCommitInterval"`
 }
 
 // Since not all sarama's default config are zero values,
@@ -139,9 +145,13 @@ func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
 	// Return any errors that occurred to the Errors channel.
 	c.Consumer.Return.Errors = true
 
-	// Set the commit frequency to 5s
+	// Set the commit frequency to 5s or other config value if provided
 	c.Consumer.Offsets.AutoCommit.Enable = true
-	c.Consumer.Offsets.AutoCommit.Interval = 5 * time.Second
+	if cfg.AutoCommitInterval != 0 {
+		c.Consumer.Offsets.AutoCommit.Interval = time.Duration(cfg.AutoCommitInterval) * time.Second
+	} else {
+		c.Consumer.Offsets.AutoCommit.Interval = 5 * time.Second
+	}
 
 	c.ClientID = cfg.ClientID
 	c.Version = version
--- a/kafkabp/config.go
+++ b/kafkabp/config.go
@@ -101,11 +101,11 @@ type ConsumerConfig struct {
 	// You are advised to test before using non-empty rack id in production.
 	RackID RackIDFunc `yaml:"rackID"`
 
-	// Optional. Defaults to 5. Values are in seconds.
+	// Optional. Defaults to 5 seconds.
 	//
 	// This affects how often the kafka consumer writes its offsets back to
 	// the cluster.
-	AutoCommitInterval time.Duration `yaml:"autoCommitInterval"`
+	AutoCommitIntervalSeconds int `yaml:"autoCommitInterval"`
 }
 
 // Since not all sarama's default config are zero values,
@@ -147,8 +147,8 @@ func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
 
 	// Set the commit frequency to 5 seconds or other config value if provided
 	c.Consumer.Offsets.AutoCommit.Enable = true
-	if cfg.AutoCommitInterval != 0 {
-		c.Consumer.Offsets.AutoCommit.Interval = cfg.AutoCommitInterval * time.Second
+	if cfg.AutoCommitIntervalSeconds != 0 {
+		c.Consumer.Offsets.AutoCommit.Interval = time.Duration(cfg.AutoCommitIntervalSeconds) * time.Second
 	} else {
 		c.Consumer.Offsets.AutoCommit.Interval = 5 * time.Second
 	}
--- a/kafkabp/config.go
+++ b/kafkabp/config.go
@@ -3,6 +3,7 @@ package kafkabp
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/Shopify/sarama"
 
--- a/kafkabp/config.go
+++ b/kafkabp/config.go
@@ -101,7 +101,7 @@ type ConsumerConfig struct {
 	// You are advised to test before using non-empty rack id in production.
 	RackID RackIDFunc `yaml:"rackID"`
 
-	// Optional. Defaults to 5s. Values are in seconds.
+	// Optional. Defaults to 5. Values are in seconds.
 	//
 	// This affects how often the kafka consumer writes its offsets back to
 	// the cluster.
@@ -145,7 +145,7 @@ func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
 	// Return any errors that occurred to the Errors channel.
 	c.Consumer.Return.Errors = true
 
-	// Set the commit frequency to 5s or other config value if provided
+	// Set the commit frequency to 5 seconds or other config value if provided
 	c.Consumer.Offsets.AutoCommit.Enable = true
 	if cfg.AutoCommitInterval != 0 {
 		c.Consumer.Offsets.AutoCommit.Interval = time.Duration(cfg.AutoCommitInterval) * time.Second
--- a/kafkabp/config.go
+++ b/kafkabp/config.go
@@ -138,6 +138,10 @@ func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
 	// Return any errors that occurred to the Errors channel.
 	c.Consumer.Return.Errors = true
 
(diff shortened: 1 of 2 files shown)


kafkabp/config.go
```diff
@@ -138,6 +139,10 @@ func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
 	// Return any errors that occurred to the Errors channel.
 	c.Consumer.Return.Errors = true
 
+	// Set the commit frequency to 5s
+	c.Consumer.Offsets.AutoCommit.Enable = true
+	c.Consumer.Offsets.AutoCommit.Interval = 5 * time.Second
```
fishy, line 144: :no_bell: one way to make it more flexible is to define it as a global, exported variable (e.g.`var CommitInterval = 5 * time.Second`), so if 5s doesn't work for some services they an override it by changing the value of the global variable before creating their consumers. but we don't necessarily need to do that now and can wait until such need actually arise.

fishy (approved):

konradreiche (approved):

kafkabp/config.go
mterwill: Agree it might be nice to have this configurable, especially since it appears most code calls `NewConsumer` rather than this directly so (from a cursory glance) I'm not sure if it's possible to modify as-is.

https://github.com/reddit/baseplate.go/blob/088a0beef0815a5961b5b6e0efccc025ae318c0d/kafkabp/consumer.go#L97-L101

mterwill (approved):
Seems reasonable to me. I believe this would exacerbate double-processing of messages for services that don't properly close their consumers, though that's a bug they should fix regardless.

kafkabp/config.go
sydjryan: I added it to the ConsumerConfig, let me know what you think - yall are right that some consumers might want to change this value (though as Matt says below, imo that's indicative of other problems with the consumer). let me know what you think.

kafkabp/config.go
```diff
@@ -99,6 +100,12 @@ type ConsumerConfig struct {
 	// or it might make things worse.
 	// You are advised to test before using non-empty rack id in production.
 	RackID RackIDFunc `yaml:"rackID"`
+
+	// Optional. Defaults to 5. Values are in seconds.
+	//
+	// This affects how often the kafka consumer writes its offsets back to
+	// the cluster.
+	AutoCommitInterval int `yaml:"autoCommitInterval"`
```
mterwill, line 108: Can we use `time.Duration` directly here?

mterwill (changes requested):
Small comment otherwise LGTM! Thanks for adding tests too 🙂

kafkabp/config.go
sydjryan: great call, done. I barely know go, so the tests are for my own confidence more than anything, haha!

kafkabp/config.go
```diff
@@ -99,6 +100,12 @@ type ConsumerConfig struct {
 	// or it might make things worse.
 	// You are advised to test before using non-empty rack id in production.
 	RackID RackIDFunc `yaml:"rackID"`
+
+	// Optional. Defaults to 5. Values are in seconds.
```
mterwill, line 104:
```suggestion
	// Optional. Defaults to 5 seconds.
```

kafkabp/config.go
```diff
@@ -138,6 +145,14 @@ func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
 	// Return any errors that occurred to the Errors channel.
 	c.Consumer.Return.Errors = true
 
+	// Set the commit frequency to 5 seconds or other config value if provided
+	c.Consumer.Offsets.AutoCommit.Enable = true
+	if cfg.AutoCommitInterval != 0 {
+		c.Consumer.Offsets.AutoCommit.Interval = cfg.AutoCommitInterval * time.Second
+	} else {
+		c.Consumer.Offsets.AutoCommit.Interval = 5 * time.Second
+	}
```
mterwill, line 154: This does some subtly funky stuff as "bare" `time.Duration` values are in nanoseconds.

We want to use the value the user provides to us directly, without multiplying by `time.Second` (which [is actually the integer 1000000000](https://go.dev/play/p/HJ2PVk9zhnY)).

We also probably want to set a lower bound of 1 second so that users don't inadvertently set their commit value in nanoseconds 😅 

And as I'm writing this out I'm convincing myself we should just go back to the way that you had it originally where `AutoCommitInterval` is just a regular old `int`... (and maybe we call it `AutoCommitIntervalSeconds` for clarity)

🤷‍♂️

mterwill (changes requested):

kafkabp/config.go
mterwill: (The nice thing about `time.Duration` is that it allows users to specify the value in the unit that makes sense in the context. In this case I think your original approach is better though. Sorry for the waffling – I didn't consider the implication of "bare" `time.Duration` values)

kafkabp/config.go
sydjryan: Providing the value in nanoseconds seems potentially confusing to me, especially because the way you'd set this in kafka would be in milliseconds so folks familiar with that might mix them up. Adding a lower bound would help, but I think going back to an int makes the most sense and is the least error prone. The renaming you suggested sounds good, I'll do that

kafkabp/config.go
```diff
@@ -138,6 +145,14 @@ func (cfg *ConsumerConfig) NewSaramaConfig() (*sarama.Config, error) {
 	// Return any errors that occurred to the Errors channel.
 	c.Consumer.Return.Errors = true
 
+	// Set the commit frequency to 5 seconds or other config value if provided
+	c.Consumer.Offsets.AutoCommit.Enable = true
+	if cfg.AutoCommitInterval != 0 {
+		c.Consumer.Offsets.AutoCommit.Interval = cfg.AutoCommitInterval * time.Second
```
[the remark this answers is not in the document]
sydjryan: Good catch, thank you. I am actually surprised that the test didnt catch this 🤔

kafkabp/config.go
mterwill: fwiw you don't _have_ to

mterwill: fwiw you don't _have_ to provide the value in nanoseconds (see: https://pkg.go.dev/time#ParseDuration), that's just the default

kafkabp/config.go
sydjryan: It's all good, I'm learning :)

kafkabp/config.go
fishy: depending on how often/rarely services actually want to change it, we don't necessarily want it to be part of config.

part of the config is suitable for cases that this is more likely to be changed by services (say, ~40% of the services would want to use a non-default value), while defining a global variable is more like "in most cases you don't want to change it but if you really do and know what you are doing you can change it" case.

kafkabp/config.go
```diff
@@ -99,6 +100,12 @@ type ConsumerConfig struct {
 	// or it might make things worse.
 	// You are advised to test before using non-empty rack id in production.
 	RackID RackIDFunc `yaml:"rackID"`
+
+	// Optional. Defaults to 5 seconds.
+	//
+	// This affects how often the kafka consumer writes its offsets back to
+	// the cluster.
+	AutoCommitIntervalSeconds int `yaml:"autoCommitInterval"`
```
mterwill, line 108: Should we call the yaml value `autoCommitIntervalSeconds` as well so the unit is clear to someone just looking at the config without having to trace through the code?

pacejackson, line 108:
```suggestion
	AutoCommitIntervalSeconds int `yaml:"autoCommitIntervalSeconds"`
```

That said... I would keep this as `time.Duration` 😅 

I can see the argument for doing it this way as a way to enforce not setting a time below one second

fishy (changes requested):
go idiom is to use `time.Duration` instead of `int`.

with `time.Duration` you actually specify the unit in your yaml config and makes it much more readable:
```yaml
  autoCommitInterval: 5s
```

kafkabp/config.go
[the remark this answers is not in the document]
fishy: if we really want to enforce a mininal interval, then we can just document it as-is (so use a different minimal instead of 0)

kafkabp/config_test.go
```diff
@@ -108,4 +109,32 @@ func TestConsumerConfig(t *testing.T) {
...
+
+	cfg.AutoCommitIntervalSeconds = 8
+	t.Run("auto-commit-interval", func(t *testing.T) {
+		sc, err := cfg.NewSaramaConfig()
+		if err != nil {
+			t.Errorf("expected no error, got %v", err)
+		}
+		if sc == nil {
+			t.Fatal("expected config to be non-nil, got nil")
+		}
```
mterwill, line 121: could probably simplify this:

```suggestion
		sc, err := cfg.NewSaramaConfig()
		if err != nil {
			t.Fatalf("expected no error parsing config, got %v", err)
		}
```

You can assume (it's idiomatic Go) if no error is returned that the other return value is non-nil.

kafkabp/config_test.go
```diff
@@ -108,4 +109,32 @@ func TestConsumerConfig(t *testing.T) {
...
+		if sc.Consumer.Offsets.AutoCommit.Interval != 8*time.Second {
+			t.Errorf("expected auto-commit interval to be %q, got %q", 8, sc.Consumer.Offsets.AutoCommit.Interval)
+		}
```
mterwill, line 124:
```suggestion
		if got := sc.Consumer.Offsets.AutoCommit.Interval, want := 8*time.Second; got != want {
			t.Errorf("got auto-commit interval = %q, want %q", got, want)
		}
```

avoids some potentially subtle copy pasta errors for future developers where they update the conditional but not the error message

PR #546, opened 2022-07-15
runtimebp: Extend GOMAXPROCS logic to account for additional scenarios

isugimpy:
On discussion with kevlar, we came to the conclusion that there should be
4 possible settings for GOMAXPROCS which should cover all needs and
ensure adequate automation.  These are:
1. GOMAXPROCS if provided,
2. limit if known (with configurable multiplier),
3. fallback to a known environment variable Infrared can provide,
4. exactly 2 (for minimum parallelism)

By providing this hierarchy, we give maximum flexibility for devs,
while also providing safe and sane guardrails to work with, and
never end up in a scenario where bp.go tries to use physical core count.

### DIFF
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 )
 
+var millicoreRegexp = regexp.MustCompile(`^(P<millis>[1-9][0-9]*)m$`)
+
 // NumCPU returns the number of CPUs assigned to this running container.
 //
 // This is the container aware version of runtime.NumCPU.
@@ -167,7 +169,6 @@ func fetchCPURequest() (int, error) {
 			// k8s automatically rounds it up to the nearest integer unit.  That's how
 			// we plan to set this variable, but just in case it might be worth doing this
 			// check in case they ever break that.  There's no official contract for that.
-			millicoreRegexp := regexp.MustCompile(`^(P<millis>[1-9][0-9]*)m$`)
 			match := millicoreRegexp.FindStringSubmatch(v)
 			if match == nil {
 				fmt.Fprintf(
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -42,13 +42,14 @@ func NumCPU() float64 {
 		return n
 	}
 
-	// Fallback to the standard runtime package value.
+	// If nothing is set, falling back to the core count is probably dangerous.
+	// Instead, fall back to the lowest value that still allows parallelism.
 	fmt.Fprintf(
 		os.Stderr,
-		"runtimebp.NumCPU: Failed to read cgroup v1, fallback to NumCPU on the physical machine: %v\n",
+		"runtimebp.NumCPU: Failed to read cgroup v1, fallback to 2 to allow parallelism: %v\n",
 		err,
 	)
-	return float64(runtime.NumCPU())
+	return float64(2)
 }
 
 func numCPUCgroupsV2(buf []byte) (float64, error) {
@@ -127,11 +128,59 @@ func defaultMaxProcsFormula(n float64) int {
 	return int(math.Ceil(n))
 }
 
-// GOMAXPROCS sets runtime.GOMAXPROCS with the default formula,
+func scaledMaxProcsFormula(n float64) int {
+	i := fetchMaxProcsMult()
+	return int(math.Ceil(n) * float64(i))
+}
+
+func fetchMaxProcsMult() int {
+	// Allow using a multiplier for number of processes relative to limit.
+	// Not catching the ok here because this function will never be called unless
+	// the environment variable is set
+	v, _ := os.LookupEnv("GOMAXPROCSMULT")
+	i, err := strconv.Atoi(v)
+	if err != nil {
+		fmt.Fprintf(
+			os.Stderr,
+			"runtimebp.fetchMaxProcsMult: GOMAXPROCSMULT is set to %q, which is invalid, falling back to 2 to ensure parallelism.",
+			v,
+		)
+		i = 2
+	}
+	return i
+}
+
+// fetchCpuRequest checks for the magic CPUREQUEST variable set as a fallback
+// by Infrared.  If this is set, we use it and the multiplier to set our
+// GOMAXPROCS as a fallback.
+func fetchCPURequest() (int, error) {
+	if v, ok := os.LookupEnv("CPUREQUEST"); ok {
+		req, err := strconv.Atoi(v)
+		if err != nil {
+			fmt.Fprintf(
+				os.Stderr,
+				"runtimebp.GOMAXPROCS: CPUREQUEST is set to %q, which is invalid, ignoring.",
+				v,
+			)
+			req = 1
+		}
+		mult := fetchMaxProcsMult()
+		return int(req * mult), nil
+	}
+	return 0, fmt.Errorf("CPUREQUEST unset")
+}
+
+// GOMAXPROCS sets runtime.GOMAXPROCS with the formula,
 // in bound of [min, max].
 //
 // Currently the default formula is NumCPU() rounding up.
 func GOMAXPROCS(min, max int) (oldVal, newVal int) {
+	if req, err := fetchCPURequest(); err == nil {
+		return GOMAXPROCSwithFormula(req, req, defaultMaxProcsFormula)
+	}
+	if _, ok := os.LookupEnv("GOMAXPROCSMULT"); ok {
+		return GOMAXPROCSwithFormula(min, max, scaledMaxProcsFormula)
+	}
 	return GOMAXPROCSwithFormula(min, max, defaultMaxProcsFormula)
 }
 
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -133,7 +133,7 @@ func defaultMaxProcsFormula(n float64) int {
 
 func scaledMaxProcsFormula(n float64) int {
 	i := fetchMaxProcsMult()
-	return int(math.Ceil(n) * float64(i))
+	return int(math.Ceil(n * float64(i)))
 }
 
 func fetchMaxProcsMult() int {
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -179,7 +179,7 @@ func fetchCPURequest() (int, error) {
 			} else {
 				m, err := strconv.Atoi(match[0])
 				if err != nil {
-					req = int(math.Ceil(float64(m / 1000.0)))
+					req = int(math.Ceil(float64(m) / 1000.0))
 				} else {
 					fmt.Fprintf(
 						os.Stderr,
@@ -187,8 +187,8 @@ func fetchCPURequest() (int, error) {
 						req,
 						err,
 					)
+					req = 1
 				}
-				req = 1
 			}
 		}
 		return scaledMaxProcsFormula(float64(req)), nil
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -11,7 +11,9 @@ import (
 	"strings"
 )
 
-var millicoreRegexp = regexp.MustCompile(`^(P<millis>[1-9][0-9]*)m$`)
+const defaultMultiplier = 2
+
+var millicoreRegexp = regexp.MustCompile(`^([1-9][0-9]*)m$`)
 
 // NumCPU returns the number of CPUs assigned to this running container.
 //
@@ -49,10 +51,11 @@ func NumCPU() float64 {
 	// Instead, fall back to the lowest value that still allows parallelism.
 	fmt.Fprintf(
 		os.Stderr,
-		"runtimebp.NumCPU: Failed to read cgroup v1, fallback to 2 to allow parallelism: %v\n",
+		"runtimebp.NumCPU: Failed to read cgroup v1, fallback to %d to allow parallelism: %v\n",
+		defaultMultiplier,
 		err,
 	)
-	return 2
+	return defaultMultiplier
 }
 
 func numCPUCgroupsV2(buf []byte) (float64, error) {
@@ -132,6 +135,10 @@ func defaultMaxProcsFormula(n float64) int {
 }
 
 func scaledMaxProcsFormula(n float64) int {
+	// If we have a request below 1, round up so we don't end up single-threaded
+	if n < 1 {
+		n = 1
+	}
 	i := fetchMaxProcsMult()
 	return int(math.Ceil(n * float64(i)))
 }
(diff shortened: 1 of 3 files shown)


fishy (changes requested):

isugimpy:
Going to be doing some additional tinkering here, but I wanted it in front of eyes.  I think I've got a sane way to do it off the request now that's going to require a bit more parsing.  Good feedback there, Yuxuan, I'll address.

runtimebp/cpu.go
```diff
@@ -127,11 +128,59 @@ func defaultMaxProcsFormula(n float64) int {
 	return int(math.Ceil(n))
 }
 
-// GOMAXPROCS sets runtime.GOMAXPROCS with the default formula,
+func scaledMaxProcsFormula(n float64) int {
+	i := fetchMaxProcsMult()
+	return int(math.Ceil(n) * float64(i))
```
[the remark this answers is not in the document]
isugimpy: `i` in this case is always going to be an integer, so I think we're still alright here.

fishy:
We probably also want to hardcode the min arg passed in from [`InitFromConfig`](https://github.com/reddit/baseplate.go/blob/e34b108af09beab61f33126dfacf075ff58f5116/runtimebp/config.go#L37) to be 2 to enforce the minimal of 2.

runtimebp/cpu.go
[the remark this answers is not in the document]
fishy: even in that case ceil before multiplication and ceil after multiplication can still give you different results (example: n is 1.4 and i is 2)

runtimebp/cpu.go
```diff
@@ -127,11 +129,84 @@ func defaultMaxProcsFormula(n float64) int {
...
+func fetchCPURequest() (int, error) {
+	if v, ok := os.LookupEnv("CPUREQUEST"); ok {
+		req, err := strconv.Atoi(v)
+		if err != nil {
+			// This case will probably never be hit, but I'm keeping it just in case.
+			// In k8s 1.21 and newer, when you present the CPU request via the Downward API
+			// k8s automatically rounds it up to the nearest integer unit.  That's how
+			// we plan to set this variable, but just in case it might be worth doing this
+			// check in case they ever break that.  There's no official contract for that.
+			millicoreRegexp := regexp.MustCompile(`^(P<millis>[1-9][0-9]*)m$`)
```
fishy, line 170: `MustCompile` should only be used in global variable definition, and this regexp should be moved to be a global variable definition (as compiling is expensive and we only need to do it once)

runtimebp/cpu.go
```diff
@@ -42,13 +45,14 @@ func NumCPU() float64 {
 		return n
 	}
 
-	// Fallback to the standard runtime package value.
+	// If nothing is set, falling back to the core count is probably dangerous.
+	// Instead, fall back to the lowest value that still allows parallelism.
 	fmt.Fprintf(
 		os.Stderr,
-		"runtimebp.NumCPU: Failed to read cgroup v1, fallback to NumCPU on the physical machine: %v\n",
+		"runtimebp.NumCPU: Failed to read cgroup v1, fallback to 2 to allow parallelism: %v\n",
```
fishy, line 52: and if we change `2` to a constant that should be a `%d` instead of hardcoding in the log message

runtimebp/cpu.go
```diff
@@ -127,11 +131,83 @@ func defaultMaxProcsFormula(n float64) int {
...
+				fmt.Fprintf(
+					os.Stderr,
+					"runtimebp.fetchCPURequest: CPUREQUEST is set to %q, which is invalid, ignoring.",
+					v,
+				)
+				req = 1
+			} else {
+				m, err := strconv.Atoi(match[0])
+				if err != nil {
+					req = int(math.Ceil(float64(m) / 1000.0))
```
fishy, line 183: :no_bell:/super nit: in go you don't have to add `.0`.

runtimebp/cpu.go
```diff
@@ -127,11 +131,83 @@ func defaultMaxProcsFormula(n float64) int {
...
+			if match == nil {
+				fmt.Fprintf(
+					os.Stderr,
+					"runtimebp.fetchCPURequest: CPUREQUEST is set to %q, which is invalid, ignoring.",
+					v,
+				)
+				req = 1
+			} else {
+				m, err := strconv.Atoi(match[0])
+				if err != nil {
```
fishy, line 182: you have the `if` condition reversed?

runtimebp/cpu.go
```diff
@@ -127,11 +131,83 @@ func defaultMaxProcsFormula(n float64) int {
...
+			match := millicoreRegexp.FindStringSubmatch(v)
+			if match == nil {
+				fmt.Fprintf(
+					os.Stderr,
+					"runtimebp.fetchCPURequest: CPUREQUEST is set to %q, which is invalid, ignoring.",
+					v,
+				)
+				req = 1
+			} else {
+				m, err := strconv.Atoi(match[0])
```
fishy, line 181: if you only use numeric match groups, not the named ones, then there's probably no need to use named match groups in the regexp itself.

also I'm pretty sure the match group 0 is the wrong string and 1 is the first subgroup. you probably want to write some unit tests for this regexp and its related handling logic :)

runtimebp/cpu.go
```diff
@@ -127,11 +131,83 @@ func defaultMaxProcsFormula(n float64) int {
...
+				if err != nil {
+					req = int(math.Ceil(float64(m) / 1000.0))
+				} else {
+					fmt.Fprintf(
```
fishy, line 185: :no_bell: based on the regexp I don't think this case is possible but ¯\\\_(ツ)\_/¯

runtimebp/cpu.go
```diff
@@ -127,11 +131,83 @@ func defaultMaxProcsFormula(n float64) int {
...
+						req,
+						err,
+					)
+					req = 1
+				}
+			}
+		}
+		return scaledMaxProcsFormula(float64(req)), nil
+	}
+	return 0, fmt.Errorf("CPUREQUEST unset")
```
fishy, line 197: it looks like it would be better to change the error return to a boolean return (e.g. `ok bool`), as the only non-nil error we could return from this function is when the env var was not set.

runtimebp/cpu.go
```diff
@@ -127,11 +131,83 @@ func defaultMaxProcsFormula(n float64) int {
...
+}
+
+// GOMAXPROCS sets runtime.GOMAXPROCS with the formula,
 // in bound of [min, max].
 //
 // Currently the default formula is NumCPU() rounding up.
```
fishy, line 203: the comments here need to be updated to reflect the new logic.

looking at the new logic, the priority is that if we have `CPUREQUEST` env var set, we always use that and ignore whatever cpu limits set. is that intentional? I think giving limits the highest priority and fallback to requests if limits are not set would make more sense?

runtimebp/cpu.go
fishy: :no_bell: also I'm wondering if it would make more sense to keep `req` as a float64 when passing it into `scaledMaxProcsFormula`, vs. ceil it here.

runtimebp/cpu.go
```diff
@@ -127,11 +131,83 @@ func defaultMaxProcsFormula(n float64) int {
...
+	v, _ := os.LookupEnv("GOMAXPROCSMULT")
+	i, err := strconv.Atoi(v)
+	if err != nil {
+		fmt.Fprintf(
+			os.Stderr,
+			"runtimebp.fetchMaxProcsMult: GOMAXPROCSMULT is set to %q, which is invalid, falling back to 2 to ensure parallelism: %v",
+			v,
+			err,
+		)
+		i = 2
```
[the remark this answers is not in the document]
isugimpy: So, the reason I defaulted the multiplier to 2 is that if we have it set to 1 we can literally never burst above request, and that seems like it doesn't really align with the desired state for default behavior.  It'd function identically to having a limit in that case.  Does that make sense?

runtimebp/cpu.go
isugimpy: I can drop the `.0` no prob, but the value generally comes in as an int, so otherwise I think I'd have to do a bunch of hoop jumping in reverse.

runtimebp/cpu.go
isugimpy: I think you're actually right, and if you want I can remove it.

runtimebp/cpu.go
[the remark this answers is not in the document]
fishy: there are two different "bursting" here.

one is os/container level cpu bursting, that is controlled by cpu limits, and has nothing to do with GOMAXPROCS

the other is the actual "parallelism" of the go program, which is controlled by GOMAXPROCS, but since we have an overall minimal of 2 (that's after all the multiplications we did), I don't think having a default multiplier set to 2 is meaningful.

runtimebp/cpu.go
fishy: isn't the whole point of this part to handle non-int cpu requests?

so what I'm suggesting is:
1. define `var req float64` at the beginning of this function
2. in the `ok` branch, use a local variable to get the converted int value from `strconv.Atoi`, and cast+assign it to `req` if the conversion was successful
3. here we just do `req = float64(m) / 1000` without ceil nor converting to int.

runtimebp/cpu.go
fishy: we probably still want the error handling branch, even though the error is theoretically impossible to happen. we have some "prior arts" of that by just adding a comment of `// should not happen, but just in case` and still do error logging/whatever is appropriate

runtimebp/cpu.go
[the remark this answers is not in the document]
isugimpy: So, the thing is, if GOMAXPROCS is set to 2, you can use at most 2 CPU cores.  We actually had an incident yesterday where this came up.  It's functionally the same as having the CPU limit actually set to 2.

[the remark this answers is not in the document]
fishy: ok then :)

runtimebp/cpu.go
isugimpy: Right, but it still would be coming in as an int in the string.  Like, if it came in as `500m`, after the regex capture we'd have `500`, which is an int, same as the whole number version we'd get.  I'll do what you're describing though.

fishy: no here you converted `500m` to `0.5`, then ceil it and convert it to `1` (all in one line). what I said is to keep `0.5` instead.

runtimebp/cpu.go
```diff
@@ -127,11 +128,59 @@ func defaultMaxProcsFormula(n float64) int {
...
+	return i
+}
+
+// fetchCpuRequest checks for the magic CPUREQUEST variable set as a fallback
+// by Infrared.  If this is set, we use it and the multiplier to set our
+// GOMAXPROCS as a fallback.
+func fetchCPURequest() (int, error) {
+	if v, ok := os.LookupEnv("CPUREQUEST"); ok {
+		req, err := strconv.Atoi(v)
+		if err != nil {
```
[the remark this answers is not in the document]
fishy: I think this comment still applies. currently we do different fallbacks between the two cases.

runtimebp/cpu.go
```diff
@@ -127,11 +131,83 @@ func defaultMaxProcsFormula(n float64) int {
...
+	}
+	return 0, fmt.Errorf("CPUREQUEST unset")
+}
+
+// GOMAXPROCS sets runtime.GOMAXPROCS with the formula,
 // in bound of [min, max].
 //
 // Currently the default formula is NumCPU() rounding up.
 func GOMAXPROCS(min, max int) (oldVal, newVal int) {
+	if req, err := fetchCPURequest(); err == nil {
```
kylelemons, line 205: (re: the above, you're using `err == nil` here and don't care about the error, so swapping it to a boolean is clearer)

kylelemons (changes requested):
thanks for starting this ball rolling! thoughts and comments below

isugimpy:
Did some significant changes here to make things more correct and follow the feedback.  Also added a bit of testing.

runtimebp/cpu.go
```diff
@@ -127,11 +134,108 @@ func defaultMaxProcsFormula(n float64) int {
...
+		_, limitSet = numCPUCgroupsV1(buf)
+		if limitSet == nil {
+			if req, ok := fetchCPURequest(); ok {
+				oldVal = runtime.GOMAXPROCS(req)
+				return oldVal, req
+			}
+		}
+	}
+	if _, ok := os.LookupEnv("BASEPLATE_GOMAXPROCSMULT"); ok {
+		return GOMAXPROCSwithFormula(min, max, scaledMaxProcsFormula)
```
fishy, line 237: since we already read and multiplied the multiplier in `fetchCPURequest`, do we really need this fallback? what do we multiply it with without the requests env? 1?

if that's the case, it's better to call the formula directly (vs. using `GOMAXPROCSwithFormula`, which would call `NumCPU` and try to read the cgroup limits and print error logs again).

we want to first refactor the part of `GOMAXPROCSwithFormula` to a helper function:

```
func setGOMAXPROCSwithBoundary(n, min, max int) (oldVal, newVal int) {
	newVal = boundNtoMinMax(n, min, max)
	oldVal = runtime.GOMAXPROCS(newVal)
	return oldVal, newVal
}
```

then use that directly:
```suggestion
		return setGOMAXPROCSwithBoundary(scaledMaxProcsFormula(1), min, max)
```

runtimebp/cpu.go
```diff
@@ -127,11 +134,108 @@ func defaultMaxProcsFormula(n float64) int {
...
+	}
 	return GOMAXPROCSwithFormula(min, max, defaultMaxProcsFormula)
```
fishy, line 239: and here should be:
```suggestion
	return setGOMAXPROCSwithBoundary(defaultMaxProcsFormula(1), min, max)
```
or maybe use `minGOMAXPROCS` instead of `defaultMaxProcsFormula(1)`?

fishy (changes requested):

runtimebp/cpu_test.go
```diff
@@ -185,3 +185,55 @@ func TestBoundNtoMinMax(t *testing.T) {
...
+		},
+	}
+
+	for label, data := range cases {
+		t.Run(label, func(t *testing.T) {
+			if data.N == "" {
+				os.Unsetenv("BASEPLATE_CPU_REQUEST")
+			} else {
+				os.Setenv("BASEPLATE_CPU_REQUEST", data.N)
+				defer os.Unsetenv("BASEPLATE_CPU_REQUEST")
```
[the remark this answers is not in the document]
isugimpy: Ooh, that's neat.

runtimebp/cpu.go
```diff
@@ -5,11 +5,16 @@ import (
 	"io"
 	"math"
 	"os"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
 )
 
+const defaultMultiplier = 2
```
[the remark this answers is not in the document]
konradreiche: Respectively this one could just be called `fallbackNumCPU`?

runtimebp/cpu.go
```diff
@@ -127,11 +134,108 @@ func defaultMaxProcsFormula(n float64) int {
...
+func scaledMaxProcsFormula(n float64) int {
+	// If we have a request below 1, round up so we don't end up single-threaded
+	if n < 1 {
+		n = 1
+	}
+	i := fetchMaxProcsMult()
+	return int(math.Ceil(n * float64(i)))
+}
+
+func fetchMaxProcsMult() int {
```
konradreiche, line 146:
```suggestion
func fetchMaxProcsMultiplier() int {
```
:no_bell: shorter environment makes sense but maybe to make it easier to understand that it is a factor.

runtimebp/cpu.go
```diff
@@ -127,11 +134,108 @@ func defaultMaxProcsFormula(n float64) int {
...
+			defaultMultiplier,
+			err,
+		)
+		intVal = defaultMultiplier
+	}
+	if intVal <= 0 {
+		intVal = 1
+	}
+	return intVal
+}
```
konradreiche, line 166:
```suggestion
	v, ok := os.LookupEnv("BASEPLATE_GOMAXPROCSMULT")
	if !ok {
	  return defaultMultiplier
	}
	multiplier, err := strconv.Atoi(v)
	if err != nil {
		fmt.Fprintf(
			os.Stderr,
			"runtimebp.fetchMaxProcsMult: GOMAXPROCSMULT is set to %q, which is invalid, falling back to %d to ensure parallelism: %v",
			v,
			defaultMultiplier,
			err,
		)
		return defaultMultiplier
	}
	if multiplier <= 0 {
		multiplier = 1
	}
	return multiplier
}
```
:no_bell: just an idea to make it more explicit when default is returned vs something that has been and parsed.

runtimebp/cpu.go
```diff
@@ -127,11 +134,108 @@ func defaultMaxProcsFormula(n float64) int {
...
+			os.Stderr,
+			"runtimebp.fetchMaxProcsMult: GOMAXPROCSMULT is set to %q, which is invalid, falling back to %d to ensure parallelism: %v",
+			v,
+			defaultMultiplier,
+			err,
+		)
+		intVal = defaultMultiplier
+	}
+	if intVal <= 0 {
+		intVal = 1
```
konradreiche, line 163: Shouldn't this also just fallback to the default multiplier 2 here since an invalid value was set by the environment?

konradreiche (changes requested):

fishy:
Sorry that things are a bit messy. I think we should take a step back to look at how things _should behave_ from API perspective.

We have 4 APIs touched here, [`InitFromConfig`](https://pkg.go.dev/github.com/reddit/baseplate.go/runtimebp#InitFromConfig), [`NumCPU`](https://pkg.go.dev/github.com/reddit/baseplate.go/runtimebp#NumCPU), [`GOMAXPROCS`](https://pkg.go.dev/github.com/reddit/baseplate.go/runtimebp#GOMAXPROCS), and [`GOMAXPROCSwithFormula`](https://pkg.go.dev/github.com/reddit/baseplate.go/runtimebp#GOMAXPROCSwithFormula):

1. `InitFromConfig`: this is the one used in `baseplate.New`, so we just want it to behave exactly hat we planned (first check `GOMAXPROCS` env var, then limit, then request*multiplier, then 2)
2. `NumCPU`: fallback to 2 certainly doesn't make sense here. fallback to request probably makes sense, but fallback to request*multiplier likely doesn't make sense.
3. `GOMAXPROCS`: this should be same as `InitFromConfig` except the `GOMAXPROCS` env var part, and the fallback to 2 part (that is handled by the min passed into `GOMAXPROCS`)
4. `GOMAXPROCSwithFormula`: this can get tricky, because we actually want different formulas in different cases in `GOMAXPROCS` now. we could just mark it as deprecated and make it behave exactly the same as `GOMAXPROCS` (ignore the formula passed in)

kylelemons (commented):
The various ways things are computed and fallbacks and such is straining my ability to hold it all in my head at once, and I think it might be confusing for users as well.

One thing that might be useful here is to try to collect all of the "fallback" logic into the same place.  For example:

```
func targetCPU(min, max int) (newVal int) {
  defer func() {
    if newVal < min { newVal = min }
    if newVal > max { newVal = max }
  }()
  
  cpu := defaultCPU
  if c, ok := cpuFromEnv(); ok {
    cpu = c
  } else c, ok := cpuFromLimits(); ok {
    cpu = c
  } else c, ok := cpuFromRequests(); ok {
    cpu = c
  }

  mult := defaultCPUMultiplier
  if m, ok := multFromEnv(); ok {
    mult = m
  }

  return int(math.Ceil(cpu*mult))
}

func GOMAXPROCS(min, max int) (oldVal, newVal int) {
  newVal = targetCPU(min, max)
  oldVal = runtime.GOMAXPROCS(newVal)
  return
}
```

Now, with the above, `BASEPLATE_GOMAXPROCS_MULT` will also apply to limits, which might not be what we want.  So, a further suggestion would be to have scoped multipliers:

```
GOMAXPROCS will set the maximum number of OS threads executing Go code within the given bounds
using one of the following conditions (whichever is found first):

* The GOMAXPROCS environment variable
* The Linux container CPU limit (multiplied by BASEPLATE_CPU_LIMIT_SCALE if present)
* The Infrared-provided container CPU request from the BASEPLATE_CPU_REQUEST environent variable (multiplied by BASEPLATE_CPU_REQUEST_SCALE if present, multiplied by 2 otherwise)
* The default value of 2

After the request is resolved as above, the value will be clamped to the given inclusive range `[min, max]`.
```

I don't see any actual uses of GOMAXPROCSWithFormula, so I kinda think we should just print out a warning and have it call GOMAXPROCS and mark it as Deprecated.

kylelemons (changes requested):
just RCing to make harold happy

mterwill:
Superseded by #547.

PR #547, opened 2022-07-28
runtimebp: New GOMAXPROCS heuristic

mterwill:
## 💸 TL;DR
Recent baseplate and internal infrastructure changes have led us to revisit our GOMAXPROCS strategy. We will now set GOMAXPROCS (if not specified) based on Kubernetes CPU requests (rather than limits) with a configurable scale multiplier. 

## 📜 Details
Continuation of https://github.com/reddit/baseplate.go/pull/546.

See: [Design Doc](https://docs.google.com/document/d/10Wd96OT0WpcSStb9GFIoMRM8UTSLtj-xHx439-LgEOg/edit)

I am working on a corresponding Infrared change to plumb through `BASEPLATE_CPU_REQUEST` to all containers and will not merge this until that is deployed.

⚠️ This is a behavior change for services running outside Reddit, and for services running inside Reddit that set their CPU limit != request (we audited this in the design doc).

Unit tests build confidence in the logic; metrics can help us build confidence and tune the heuristic.

### DIFF
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -1,9 +1,7 @@
 package runtimebp
 
 import (
-	"fmt"
-	"math"
-	"os"
+	"github.com/reddit/baseplate.go/runtimebp/internal/maxprocs"
 )
 
 // Config is the configuration struct for the runtimebp package.
@@ -20,21 +18,8 @@ type Config struct {
 	} `yaml:"numProcesses"`
 }
 
-// InitFromConfig sets GOMAXPROCS using the given config.
-//
-// NOTE: If the GOMAXPROCS environment variable is set,
-// this function will skip setting GOMAXPROCS,
-// even if the environment variable is set to some bogus value
-// (in that case it will be set to number of physical CPUs).
-func InitFromConfig(cfg Config) {
-	if v, ok := os.LookupEnv("GOMAXPROCS"); ok {
-		fmt.Fprintf(
-			os.Stderr,
-			"runtimebp.InitFromConfig: GOMAXPROCS environment variable is set to %q, skipping setting GOMAXPROCS.",
-			v,
-		)
-	} else {
-		prev, current := GOMAXPROCS(1, math.MaxInt)
-		fmt.Fprintf(os.Stderr, "GOMAXPROCS: Old: %d New: %d\n", prev, current)
-	}
+// InitFromConfig sets GOMAXPROCS based on an overridable heuristic described in maxprocs.
+// N.B. It does NOT respect the passed-in config.
+func InitFromConfig(_ Config) {
+	maxprocs.Set()
 }
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -18,7 +18,15 @@ type Config struct {
 	} `yaml:"numProcesses"`
 }
 
-// InitFromConfig sets GOMAXPROCS based on an overridable heuristic described in maxprocs.
+// InitFromConfig configures the runtime's GOMAXPROCS using the following heuristic:
+//   1. If $GOMAXPROCS is set, it relinquishes control to the Go runtime.
+//      This should cause the runtime to respect this value directly.
+//   2. If $BASEPLATE_CPU_REQUEST is unset/invalid, it relinquishes control to automaxprocs, minimum 2.
+//      See https://pkg.go.dev/go.uber.org/automaxprocs for specific behavior.
+//   3. Otherwise, $BASEPLATE_CPU_REQUEST is multiplied by $BASEPLATE_CPU_REQUEST_SCALE
+//      (or defaultCPURequestScale) to compute the new GOMAXPROCS, minimum 2.
+//
+// InitFromConfig also exports several metrics to facilitate further tuning/analysis.
 //
 // NOTE: It does NOT respect the passed-in config.
 func InitFromConfig(_ Config) {
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -19,12 +19,13 @@ type Config struct {
 }
 
 // InitFromConfig configures the runtime's GOMAXPROCS using the following heuristic:
-//   1. If $GOMAXPROCS is set, it relinquishes control to the Go runtime.
-//      This should cause the runtime to respect this value directly.
-//   2. If $BASEPLATE_CPU_REQUEST is unset/invalid, it relinquishes control to automaxprocs, minimum 2.
-//      See https://pkg.go.dev/go.uber.org/automaxprocs for specific behavior.
-//   3. Otherwise, $BASEPLATE_CPU_REQUEST is multiplied by $BASEPLATE_CPU_REQUEST_SCALE
-//      (or defaultCPURequestScale) to compute the new GOMAXPROCS, minimum 2.
+//
+// 1. If $GOMAXPROCS is set, it relinquishes control to the Go runtime.
+//    This should cause the runtime to respect this value directly.
+// 2. If $BASEPLATE_CPU_REQUEST is unset/invalid, it relinquishes control to automaxprocs, minimum 2.
+//    See https://pkg.go.dev/go.uber.org/automaxprocs for specific behavior.
+// 3. Otherwise, $BASEPLATE_CPU_REQUEST is multiplied by $BASEPLATE_CPU_REQUEST_SCALE
+//    (or defaultCPURequestScale) to compute the new GOMAXPROCS, minimum 2.
 //
 // InitFromConfig also exports several metrics to facilitate further tuning/analysis.
 //
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -19,7 +19,8 @@ type Config struct {
 }
 
 // InitFromConfig sets GOMAXPROCS based on an overridable heuristic described in maxprocs.
-// N.B. It does NOT respect the passed-in config.
+//
+// NOTE: It does NOT respect the passed-in config.
 func InitFromConfig(_ Config) {
 	maxprocs.Set()
 }
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -23,7 +23,12 @@ import (
 // or if there's no limit set in cgroup,
 // it will fallback to runtime.NumCPU() instead.
 //
-// Deprecated: NumCPU is deprecated. Instead, tune GOMAXPROCS as described in runtimebp.InitFromConfig.
+// Depending on your application, $BASEPLATE_CPU_REQUEST may also be helpful.
+// Infrared sets $BASEPLATE_CPU_REQUEST to the container's Kubernetes CPU
+// _request_ (rather than _limit_ as this function returns), rounded up to the
+// nearest whole CPU.
+//
+// To tune GOMAXPROCS, see runtimebp.InitFromConfig.
 func NumCPU() float64 {
 	// Big enough buffer to read the numbers in the files wholly into memory.
 	buf := make([]byte, 1024)
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -22,6 +22,8 @@ import (
 // If the current process is not running inside a container,
 // or if there's no limit set in cgroup,
 // it will fallback to runtime.NumCPU() instead.
+//
+// Deprecated: NumCPU is deprecated. See runtimebp/internal/maxprocs.
 func NumCPU() float64 {
 	// Big enough buffer to read the numbers in the files wholly into memory.
 	buf := make([]byte, 1024)
@@ -131,12 +133,16 @@ func defaultMaxProcsFormula(n float64) int {
 // in bound of [min, max].
 //
 // Currently the default formula is NumCPU() rounding up.
+//
+// Deprecated: GOMAXPROCS is deprecated. See runtimebp/internal/maxprocs.
 func GOMAXPROCS(min, max int) (oldVal, newVal int) {
 	return GOMAXPROCSwithFormula(min, max, defaultMaxProcsFormula)
 }
 
 // GOMAXPROCSwithFormula sets runtime.GOMAXPROCS with the given formula,
 // in bound of [min, max].
+//
+// Deprecated: GOMAXPROCSwithFormula is deprecated. See runtimebp/internal/maxprocs.
 func GOMAXPROCSwithFormula(min, max int, formula MaxProcsFormula) (oldVal, newVal int) {
 	newVal = boundNtoMinMax(formula(NumCPU()), min, max)
 	oldVal = runtime.GOMAXPROCS(newVal)
--- a/runtimebp/cpu.go
+++ b/runtimebp/cpu.go
@@ -23,7 +23,7 @@ import (
 // or if there's no limit set in cgroup,
 // it will fallback to runtime.NumCPU() instead.
 //
-// Deprecated: NumCPU is deprecated. See runtimebp/internal/maxprocs.
+// Deprecated: NumCPU is deprecated. Instead, tune GOMAXPROCS as described in runtimebp.InitFromConfig.
 func NumCPU() float64 {
 	// Big enough buffer to read the numbers in the files wholly into memory.
 	buf := make([]byte, 1024)
@@ -134,7 +134,7 @@ func defaultMaxProcsFormula(n float64) int {
 //
 // Currently the default formula is NumCPU() rounding up.
 //
-// Deprecated: GOMAXPROCS is deprecated. See runtimebp/internal/maxprocs.
+// Deprecated: GOMAXPROCS is deprecated. Instead, tune GOMAXPROCS as described in runtimebp.InitFromConfig.
 func GOMAXPROCS(min, max int) (oldVal, newVal int) {
 	return GOMAXPROCSwithFormula(min, max, defaultMaxProcsFormula)
 }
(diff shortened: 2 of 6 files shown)


isugimpy (approved):
This looks sane to me as written, ignoring my concerns with the design itself.

runtimebp/config.go
```diff
@@ -20,21 +18,8 @@ type Config struct {
...
-			os.Stderr,
-			"runtimebp.InitFromConfig: GOMAXPROCS environment variable is set to %q, skipping setting GOMAXPROCS.",
-			v,
-		)
-	} else {
-		prev, current := GOMAXPROCS(1, math.MaxInt)
-		fmt.Fprintf(os.Stderr, "GOMAXPROCS: Old: %d New: %d\n", prev, current)
-	}
+// InitFromConfig sets GOMAXPROCS based on an overridable heuristic described in maxprocs.
+// N.B. It does NOT respect the passed-in config.
```
kylelemons, line 22:
```suggestion
//
// Note: It does NOT respect the passed-in config.
```

runtimebp/cpu.go
```diff
@@ -22,6 +22,8 @@ import (
 // If the current process is not running inside a container,
 // or if there's no limit set in cgroup,
 // it will fallback to runtime.NumCPU() instead.
+//
+// Deprecated: NumCPU is deprecated. See runtimebp/internal/maxprocs.
```
kylelemons, line 26: Routing to an internal package is a bit confusing -- callers can't _use_ the internal package, so do you mean "see the docs in"?  If so, maybe the relevant docs should be in this package instead.

Either way, you can actually link to pkg.go.dev if you want since this is open source

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,111 @@
+// Package maxprocs optimizes GOMAXPROCS for Infrared environments.
+// It attempts to balance application performance with cluster-level resource utilization.
+// Applications may tune GOMAXPROCS as necessary (see Set).
+// The defaults in this package are subject to change.
+package maxprocs
+
```
kylelemons, line 6: maybe include a note that the "canonical" version of this is not here?

```suggestion

// Note: this is manually copied from internal baseplate/v2/internal/maxprocs/maxprocs.go

```

kylelemons (approved):

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,111 @@
...
+	envCPURequestScale := &floatEnv{key: "BASEPLATE_CPU_REQUEST_SCALE", gauge: mEnvCPURequestScale}
+
+	for _, env := range []*floatEnv{envGOMAXPROCS, envCPURequest, envCPURequestScale} {
+		env.raw, env.present = os.LookupEnv(env.key)
+		env.val, _ = strconv.ParseFloat(env.raw, 64)
+		env.gauge.WithLabelValues(strconv.FormatBool(env.present)).Set(env.val)
+	}
+
+	setBy := setByGOMAXPROCS
+	defer func() { initialGOMAXPROCS.WithLabelValues(setBy).Set(currentGOMAXPROCS()) }()
```
fishy, line 77: style nits:
1. prefer `With` over `WithLabelValues`
2. making this one-line seems a bit excessive to me.

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,111 @@
...
+//
+// Set also exports several metrics to facilitate further tuning/analysis.
+func Set() {
+	envGOMAXPROCS := &floatEnv{key: "GOMAXPROCS", gauge: mEnvGOMAXPROCS}
+	envCPURequest := &floatEnv{key: "BASEPLATE_CPU_REQUEST", gauge: mEnvCPURequest}
+	envCPURequestScale := &floatEnv{key: "BASEPLATE_CPU_REQUEST_SCALE", gauge: mEnvCPURequestScale}
+
+	for _, env := range []*floatEnv{envGOMAXPROCS, envCPURequest, envCPURequestScale} {
+		env.raw, env.present = os.LookupEnv(env.key)
+		env.val, _ = strconv.ParseFloat(env.raw, 64)
```
fishy, line 72: is there a good reason to ignore error here?

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	setBy := setByGOMAXPROCS
+	defer func() { initialGOMAXPROCS.WithLabelValues(setBy).Set(currentGOMAXPROCS()) }()
+
+	if envGOMAXPROCS.present {
+		return // let Go runtime handle it
+	}
+
+	if !envCPURequest.present {
```
fishy, line 83: it's ok to prefer cpu requests * scale over cpu limits, but completely ignore cpu limits is not right.

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,111 @@
...
+	runtime.GOMAXPROCS(int(
+		math.Max(
+			2, // to ensure some minimal parallelism
+			math.Ceil(envCPURequest.val*scale),
+		),
+	))
+}
+
+func currentGOMAXPROCS() float64 {
+	return float64(runtime.GOMAXPROCS(0))
```
fishy, line 110: this feels like an abuse of the sideeffect of `runtime.GOMAXPROCS` to me. there must be a better way of getting this info?

fishy (changes requested):

runtimebp/internal/maxprocs/maxprocs.go
mterwill: This is the only way to access the current GOMAXPROCS value as far as I know.

kylelemons: There isn't.  This is exactly how you are intended to get GOMAXPROCS.

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,111 @@
...
+//   2. If $BASEPLATE_CPU_REQUEST is unset/invalid, Set relinquishes control to the Go runtime.
+//      This should cause the runtime to use the detected number of CPUs on this machine.
+//   3. Otherwise, $BASEPLATE_CPU_REQUEST is multiplied by $BASEPLATE_CPU_REQUEST_SCALE
+//      (or defaultCPURequestScale) to compute the new GOMAXPROCS, minimum 2.
+//
+// Set also exports several metrics to facilitate further tuning/analysis.
+func Set() {
+	envGOMAXPROCS := &floatEnv{key: "GOMAXPROCS", gauge: mEnvGOMAXPROCS}
+	envCPURequest := &floatEnv{key: "BASEPLATE_CPU_REQUEST", gauge: mEnvCPURequest}
+	envCPURequestScale := &floatEnv{key: "BASEPLATE_CPU_REQUEST_SCALE", gauge: mEnvCPURequestScale}
```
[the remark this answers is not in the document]
kylelemons: If it can be done without globals, it should be done without globals.

runtimebp/internal/maxprocs/maxprocs.go
mterwill: Added a comment

kylelemons (approved):

runtimebp/internal/maxprocs/maxprocs.go
mterwill: `WithLabelValues` seems fine here to me as there is only one label and it's unlikely to change. Broke out into separate lines.

runtimebp/internal/maxprocs/maxprocs.go
mterwill: Added logic to fall back on https://github.com/uber-go/automaxprocs/

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,111 @@
...
+var (
+	mEnvGOMAXPROCS      = promauto.NewGaugeVec(prometheus.GaugeOpts{Name: "baseplate_go_env_gomaxprocs"}, []string{"present"})
+	mEnvCPURequest      = promauto.NewGaugeVec(prometheus.GaugeOpts{Name: "baseplate_go_env_baseplate_cpu_request"}, []string{"present"})
+	mEnvCPURequestScale = promauto.NewGaugeVec(prometheus.GaugeOpts{Name: "baseplate_go_env_baseplate_cpu_request_scale"}, []string{"present"})
+
+	initialGOMAXPROCS = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "baseplate_go_initial_gomaxprocs",
+	}, []string{"set_by"})
+
+	_ = promauto.NewGaugeFunc(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
mterwill: Credit @kylelemons for telling me about this!

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,111 @@
...
+	defaultCPURequestScale = 1.5
+
+	setByGOMAXPROCS = "gomaxprocs"
+	setByRequest    = "cpu_request"
+)
+
+var (
+	mEnvGOMAXPROCS      = promauto.NewGaugeVec(prometheus.GaugeOpts{Name: "baseplate_go_env_gomaxprocs"}, []string{"present"})
+	mEnvCPURequest      = promauto.NewGaugeVec(prometheus.GaugeOpts{Name: "baseplate_go_env_baseplate_cpu_request"}, []string{"present"})
+	mEnvCPURequestScale = promauto.NewGaugeVec(prometheus.GaugeOpts{Name: "baseplate_go_env_baseplate_cpu_request_scale"}, []string{"present"})
```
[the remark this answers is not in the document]
mterwill: 1. Will add help to metrics, good catch, thanks!
2. Is that documented in Prometheus best practices? I've never heard that before.

runtimebp/cpu.go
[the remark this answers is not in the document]
mterwill: What is the use case beyond setting GOMAXPROCS?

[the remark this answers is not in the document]
fishy: for example, they can use this info to decide the size of a worker pool.

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,129 @@
...
+	//
+	// Extremely performance-sensitive services should set scale to 1.0 to
+	// ensure consistent performance.
+	//
+	// This value was NOT chosen scientifically and IS subject to change.
+	defaultCPURequestScale = 1.5
+
+	setByGOMAXPROCS   = "gomaxprocs"
+	setByRequest      = "cpu_request"
+	setByAutomaxprocs = "automaxprocs"
```
fishy, line 43: change this to setByLimits

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,129 @@
...
+	}()
+
+	if envGOMAXPROCS.present {
+		return // let Go runtime handle it
+	}
+
+	if !envCPURequest.present {
+		setBy = setByAutomaxprocs
+		setWithAutomaxprocs()
+		return
```
fishy, line 102: instead of using automaxprocs, just move the unexported functions in `runtimebp` here to read cpu limits from cgroup v1 and v2, and use that info. `runtimebp.NumCPU` can also call those functions and since this is `internal` we don't need to worry about exporting them and adding new APIs.

fishy (changes requested):

runtimebp/internal/maxprocs/maxprocs.go
mterwill: I'd like to keep this implementation as close to baseplate v2 as possible, and we're not going to pull in the runtimebp functions there as automaxprocs nicely encapsulates and maintains this difficult-to-test logic.

runtimebp/internal/maxprocs/maxprocs.go
mterwill: I debated that as well. I think we should leave it as-is in case the behavior of automaxprocs changes in the future.

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if envCPURequest.val <= 0 {
+		// This should always be valid positive float in infrared-deployed applications.
+		fmt.Fprintf(os.Stderr, "maxprocs: $BASEPLATE_CPU_REQUEST=%q, want positive float. Falling back to Go's default", envCPURequest.raw)
+		return
+	}
```
pacejackson, line 120: https://github.com/reddit/baseplate.go/pull/547/files#diff-f30af23b7ae4538cec8fe53a3dd75227fb07edf28ca76ccac9031b683b6fb6b9R21-R27

says that this will be handled by `automaxprocs` but it doesn't look like that's what's going on here?

pacejackson (changes requested):

runtimebp/internal/maxprocs/maxprocs.go
mterwill: You're totally right, good catch – thank you!

pacejackson (approved):

runtimebp/config.go
```diff
@@ -20,21 +18,17 @@ type Config struct {
...
 }
 
-// InitFromConfig sets GOMAXPROCS using the given config.
+// InitFromConfig configures the runtime's GOMAXPROCS using the following heuristic:
+//   1. If $GOMAXPROCS is set, it relinquishes control to the Go runtime.
+//      This should cause the runtime to respect this value directly.
+//   2. If $BASEPLATE_CPU_REQUEST is unset/invalid, it relinquishes control to automaxprocs, minimum 2.
+//      See https://pkg.go.dev/go.uber.org/automaxprocs for specific behavior.
+//   3. Otherwise, $BASEPLATE_CPU_REQUEST is multiplied by $BASEPLATE_CPU_REQUEST_SCALE
+//      (or defaultCPURequestScale) to compute the new GOMAXPROCS, minimum 2.
```
fishy, line 27: nit: this will be fixed in go 1.19, but for now we do need blank lines between the ordered list items to have them rendered correctly on pkgsite. we also need to unindent them otherwise they'll be rendered as blockquotes.

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// Set configures the runtime's GOMAXPROCS using the following heuristic:
+//   1. If $GOMAXPROCS is set, Set relinquishes control to the Go runtime.
+//      This should cause the runtime to respect this value directly.
+//   2. If $BASEPLATE_CPU_REQUEST is unset/invalid, Set relinquishes control to automaxprocs, minimum 2.
+//      See https://pkg.go.dev/go.uber.org/automaxprocs for specific behavior.
+//   3. Otherwise, $BASEPLATE_CPU_REQUEST is multiplied by $BASEPLATE_CPU_REQUEST_SCALE
+//      (or defaultCPURequestScale) to compute the new GOMAXPROCS, minimum 2.
```
fishy, line 87: same here (exported identifiers in internal packages are still rendered ok pkgsite)

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,142 @@
...
+//
+// Set also exports several metrics to facilitate further tuning/analysis.
+func Set() {
+	envGOMAXPROCS := &floatEnv{key: "GOMAXPROCS", gauge: mEnvGOMAXPROCS}
+	envCPURequest := &floatEnv{key: "BASEPLATE_CPU_REQUEST", gauge: mEnvCPURequest}
+	envCPURequestScale := &floatEnv{key: "BASEPLATE_CPU_REQUEST_SCALE", gauge: mEnvCPURequestScale}
+
+	for _, env := range []*floatEnv{envGOMAXPROCS, envCPURequest, envCPURequestScale} {
+		env.raw, env.present = os.LookupEnv(env.key)
+		env.val, _ = strconv.ParseFloat(env.raw, 64) // error not needed as we check val > 0 and save raw
```
fishy, line 97: 🔕 the comment is not always true? the `>0` check is not handled in this for loop. outside of this for loop, there's <=0 check for `envCPURequest` and >0 check for `envCPURequestScale`, but I don't see any check for `envGOMAXPROCS`?

how about this: we change `present` to 3 states: true, not_a_number (or maybe not_valid), false?

fishy (approved):

runtimebp/config.go
mterwill: Thanks!

runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -0,0 +1,156 @@
...
+		Help: "Value of the GOMAXPROCS environment variable at startup. 0 if not a number",
+	}, []string{"status"})
+	mEnvCPURequest = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "baseplate_go_env_baseplate_cpu_request",
+		Help: "Value of the BASEPLATE_CPU_REQUEST environment variable at startup. 0 if not a number",
+	}, []string{"status"})
+	mEnvCPURequestScale = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "baseplate_go_env_baseplate_cpu_request_scale",
+		Help: "Value of the BASEPLATE_CPU_REQUEST_SCALE environment variable at startup. 0 if not a number",
+	}, []string{"status"})
```
fishy, line 58: 🔕 I'm wondering if it makes more sense to combine this to a single gauge with an additional env_var label. depending on how likely/unlikely we are going to handle more env vars or change env var names in the future.

fishy (approved):

PR #549, opened 2022-08-03
baseplate: Add StopDelay to Config

fishy:
This helps mitigate against a race condition from kubernetes between
draining the traffic from the pod and sending the pod SIGTERM.

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -17,6 +17,11 @@ import (
 	"github.com/reddit/baseplate.go/tracing"
 )
 
+const (
+	// DefaultStopDelay is the default StopDelay to be used in Serve.
+	DefaultStopDelay = 1 * time.Second
+)
+
 // Configer defines the interface that allows you to extend Config with your own
 // configurations.
 type Configer interface {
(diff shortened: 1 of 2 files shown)


baseplate.go
```diff
@@ -17,6 +17,11 @@ import (
 	"github.com/reddit/baseplate.go/tracing"
 )
 
+const (
+	// DefaultStopDelay is the default StopDelay to be used in Serve.
+	DefaultStopDelay = time.Second
```
nanassito, line 22: I know it doesn't change anything but it is easier to read IMO
```suggestion
	DefaultStopDelay = 1 * time.Second
```

nanassito (approved):

baseplate.go
kylelemons: +1

batchcloser/closers.go
```diff
@@ -49,7 +49,7 @@ type BatchCloser struct {
 func (bc *BatchCloser) Close() error {
 	var errs errorsbp.Batch
 	for _, closer := range bc.closers {
-		errs.AddPrefix(fmt.Sprintf("%#v", closer), closer.Close())
+		errs.AddPrefix(fmt.Sprintf("%T", closer), closer.Close())
```
kylelemons, line 52: 🔕 (this is a bit out of scope, and it is a somewhat major reduction in the verbosity of the prefix, might be worth not doing this in this PR at least)

kylelemons (approved):

batchcloser/closers.go
fishy: Not doing this will cause a race condition in the `timestampCloser` used in tests, which seems to be a new issue in go 1.19 (we don't see this race condition from before)

I'll add a note in the PR description/commit message.

fishy: ok I reverted this and implemented `GoString` on `timestampCloser` instead.

kylelemons (approved):
Aha! I'd just mention the 1.19 fix when you squash+merge

fishy:
> Aha! I'd just mention the 1.19 fix when you squash+merge

the fix is purely in the unit test code so I don't think it's worth mentioning.

pacejackson (approved):

PR #552, opened 2022-08-04
feat(prom): Add Prom metrics closer to the BP.spec for Redis clients

nanassito:
Closes # 

## 💸 TL;DR
This adds new Prometheus metrics that are closer to the BP.spec than existing ones.

## 📜 Details
Unfortunately there are some information that isn't available so I left these labels out.

No idea how to test

## ✅ Checks

### DIFF
--- a/redis/cache/redispipebp/monitored.go
+++ b/redis/cache/redispipebp/monitored.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/reddit/baseplate.go/prometheusbp"
 	"github.com/reddit/baseplate.go/redis/cache/redisx"
+	"github.com/reddit/baseplate.go/redis/internal/redisprom"
 	"github.com/reddit/baseplate.go/tracing"
 )
 
@@ -60,13 +61,40 @@ func (s MonitoredSync) Do(ctx context.Context, cmd string, args ...interface{}) 
 		s.Name+".do",
 		tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
 	)
+	command := extractCommand(cmd)
+	active := redisprom.ActiveRequests.With(prometheus.Labels{
+		redisprom.ClientNameLabel: s.Name,
+		redisprom.CommandLabel:    command,
+		redisprom.DatabaseLabel:   "", // We don't have that info
+		redisprom.TypeLabel:       "", // We don't have that info
+		redisprom.DeploymentLabel: "", // We don't have that info
+	})
+	active.Inc()
 	defer func(start time.Time) {
+		active.Dec()
 		err := redis.AsError(result)
+		durationSeconds := time.Since(start).Seconds()
 		promHistogram.With(prometheus.Labels{
 			labelSlug:    s.Name,
-			labelCommand: extractCommand(cmd),
+			labelCommand: command,
 			labelSuccess: prometheusbp.BoolString(err == nil),
-		}).Observe(time.Since(start).Seconds())
+		}).Observe(durationSeconds)
+		redisprom.LatencySeconds.With(prometheus.Labels{
+			redisprom.ClientNameLabel: s.Name,
+			redisprom.CommandLabel:    command,
+			redisprom.SuccessLabel:    prometheusbp.BoolString(err == nil),
+			redisprom.DatabaseLabel:   "", // We don't have that info
+			redisprom.TypeLabel:       "", // We don't have that info
+			redisprom.DeploymentLabel: "", // We don't have that info
+		}).Observe(durationSeconds)
+		redisprom.RequestsTotal.With(prometheus.Labels{
+			redisprom.ClientNameLabel: s.Name,
+			redisprom.CommandLabel:    command,
+			redisprom.SuccessLabel:    prometheusbp.BoolString(err == nil),
+			redisprom.DatabaseLabel:   "", // We don't have that info
+			redisprom.TypeLabel:       "", // We don't have that info
+			redisprom.DeploymentLabel: "", // We don't have that info
+		}).Inc()
 		span.FinishWithOptions(tracing.FinishOptions{
 			Ctx: ctx,
 			Err: err,
@@ -83,17 +111,43 @@ func (s MonitoredSync) Send(ctx context.Context, r redis.Request) (result interf
 		s.Name+".send",
 		tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
 	)
+	command := extractCommand(r.Cmd)
+	if command == "" {
+		command = "send"
+	}
+	active := redisprom.ActiveRequests.With(prometheus.Labels{
+		redisprom.ClientNameLabel: s.Name,
+		redisprom.CommandLabel:    command,
+		redisprom.DatabaseLabel:   "", // We don't have that info
+		redisprom.TypeLabel:       "", // We don't have that info
+		redisprom.DeploymentLabel: "", // We don't have that info
+	})
+	active.Inc()
 	defer func(start time.Time) {
+		active.Dec()
 		err := redis.AsError(result)
-		cmd := extractCommand(r.Cmd)
-		if cmd == "" {
-			cmd = "send"
-		}
+		durationSeconds := time.Since(start).Seconds()
 		promHistogram.With(prometheus.Labels{
 			labelSlug:    s.Name,
-			labelCommand: cmd,
+			labelCommand: command,
 			labelSuccess: prometheusbp.BoolString(err == nil),
-		}).Observe(time.Since(start).Seconds())
+		}).Observe(durationSeconds)
+		redisprom.LatencySeconds.With(prometheus.Labels{
+			redisprom.ClientNameLabel: s.Name,
+			redisprom.CommandLabel:    command,
+			redisprom.SuccessLabel:    prometheusbp.BoolString(err == nil),
+			redisprom.DatabaseLabel:   "", // We don't have that info
+			redisprom.TypeLabel:       "", // We don't have that info
+			redisprom.DeploymentLabel: "", // We don't have that info
+		}).Observe(durationSeconds)
+		redisprom.RequestsTotal.With(prometheus.Labels{
+			redisprom.ClientNameLabel: s.Name,
+			redisprom.CommandLabel:    command,
+			redisprom.SuccessLabel:    prometheusbp.BoolString(err == nil),
+			redisprom.DatabaseLabel:   "", // We don't have that info
+			redisprom.TypeLabel:       "", // We don't have that info
+			redisprom.DeploymentLabel: "", // We don't have that info
+		}).Inc()
 		span.FinishWithOptions(tracing.FinishOptions{
 			Ctx: ctx,
 			Err: redis.AsError(result),
@@ -105,13 +159,22 @@ func (s MonitoredSync) Send(ctx context.Context, r redis.Request) (result interf
 
 // SendMany wraps s.Sync.SendMany in a client span.
 func (s MonitoredSync) SendMany(ctx context.Context, reqs []redis.Request) (results []interface{}) {
-	const cmd = "send-many"
+	const command = "send-many"
 	span, ctx := opentracing.StartSpanFromContext(
 		ctx,
-		s.Name+"."+cmd,
+		s.Name+"."+command,
 		tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
 	)
+	active := redisprom.ActiveRequests.With(prometheus.Labels{
+		redisprom.ClientNameLabel: s.Name,
+		redisprom.CommandLabel:    command,
+		redisprom.DatabaseLabel:   "", // We don't have that info
+		redisprom.TypeLabel:       "", // We don't have that info
+		redisprom.DeploymentLabel: "", // We don't have that info
+	})
+	active.Inc()
 	defer func(start time.Time) {
+		active.Dec()
 		var err error
 		if len(results) > 0 {
 			first := redis.AsError(results[0])
@@ -132,11 +195,28 @@ func (s MonitoredSync) SendMany(ctx context.Context, reqs []redis.Request) (resu
 				}
 			}
 		}
+		durationSeconds := time.Since(start).Seconds()
 		promHistogram.With(prometheus.Labels{
 			labelSlug:    s.Name,
-			labelCommand: cmd,
+			labelCommand: command,
 			labelSuccess: prometheusbp.BoolString(err == nil),
-		}).Observe(time.Since(start).Seconds())
+		}).Observe(durationSeconds)
+		redisprom.LatencySeconds.With(prometheus.Labels{
+			redisprom.ClientNameLabel: s.Name,
+			redisprom.CommandLabel:    command,
+			redisprom.SuccessLabel:    prometheusbp.BoolString(err == nil),
+			redisprom.DatabaseLabel:   "", // We don't have that info
+			redisprom.TypeLabel:       "", // We don't have that info
+			redisprom.DeploymentLabel: "", // We don't have that info
+		}).Observe(durationSeconds)
+		redisprom.RequestsTotal.With(prometheus.Labels{
+			redisprom.ClientNameLabel: s.Name,
+			redisprom.CommandLabel:    command,
+			redisprom.SuccessLabel:    prometheusbp.BoolString(err == nil),
+			redisprom.DatabaseLabel:   "", // We don't have that info
+			redisprom.TypeLabel:       "", // We don't have that info
+			redisprom.DeploymentLabel: "", // We don't have that info
+		}).Inc()
 		span.FinishWithOptions(tracing.FinishOptions{
 			Ctx: ctx,
 			Err: err,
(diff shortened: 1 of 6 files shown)


redis/cache/redispipebp/monitored.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/reddit/baseplate.go/prometheusbp"
 	"github.com/reddit/baseplate.go/redis/cache/redisx"
+	"github.com/reddit/baseplate.go/redis/redisprom"
```
fishy: 🔕 as I don't feel strongly, but I would prefer to just define the metrics following the spec twice, in each client packages, instead of providing a package just for the metric definitions.

but if we are going to provide a package, it probably should be `internal` (either `baseplate.go/internal/redisprom`, or `baseplate.go/redis/internal/redisprom`)

redis/cache/redispipebp/monitored.go
```diff
@@ -60,19 +61,40 @@ func (s MonitoredSync) Do(ctx context.Context, cmd string, args ...interface{})
 		s.Name+".do",
 		tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
 	)
+	command := extractCommand(cmd)
+	active := redisprom.ActiveRequests.With(prometheus.Labels{
+		// We don't have LblClientName, LblDatabase, LblType, and LblDeployment
+		redisprom.LblPool:    s.Name,
```
fishy: and we probably should just spell it fully as `LabelPool`. saving 2 characters really doesn't seem worth it.

redis/redisprom/metrics.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/reddit/baseplate.go/prometheusbp"
+)
+
+var (
+	// Baseplate.spec Prometheus metrics
+	// https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#redis
+
+	LblPool       = "redis_pool"        // The name of the pool being monitored
```
kylelemons, line 13: * Don't drop letters to save typing (Label not Lbl)
* These should be constants not variables.
* These should not be in a public package, make it `internal/redisprom`
* 🔕 Consider using a suffix (PoolLabel) not a prefix (LabelPool).
* 🔕 I recommend unit tests for metrics.

kylelemons (changes requested):

redis/cache/redispipebp/monitored.go
```diff
@@ -83,34 +105,59 @@ func (s MonitoredSync) Send(ctx context.Context, r redis.Request) (result interf
 		s.Name+".send",
 		tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
 	)
+	command := extractCommand(r.Cmd)
+	if command == "" {
+		command = "send"
+	}
```
fishy, line 111: should we just move this logic into `extractCommand`?

redis/cache/redispipebp/monitored.go
```diff
@@ -60,19 +61,40 @@ func (s MonitoredSync) Do(ctx context.Context, cmd string, args ...interface{})
 		s.Name+".do",
 		tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
 	)
+	command := extractCommand(cmd)
+	active := redisprom.ActiveRequests.With(prometheus.Labels{
+		// We don't have LblClientName, LblDatabase, LblType, and LblDeployment
```
fishy, line 66: I think `s.Name` is client name from spec?

redis/redisprom/metrics.go
nanassito: added all the easy changes, trying to figure out how all this is tested so I can add cases for the metrics

redis/db/redisbp/hooks.go
```diff
@@ -79,20 +82,54 @@ func (h SpanHook) startChildSpan(ctx context.Context, cmdName string) context.Co
 		name,
 		tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
 	)
+	redisprom.ActiveRequests.With(prometheus.Labels{
+		// We don't have LblClientName and LblDatabase
```
fishy, line 86: I thought h.ClientName is client name, and database can be deduced from the address (if there's no `/n` at the end of address it's `0` or empty, forgot which is the one defined in the spec)

redis/internal/redisprom/metrics.go
```diff
@@ -0,0 +1,45 @@
+package redisprom
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/reddit/baseplate.go/prometheusbp"
+)
+
+// Baseplate.spec Prometheus metrics
+// https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#redis
```
fishy, line 10: this is an internal link which we probably should not put into public code.

fishy (changes requested):

redis/cache/redispipebp/monitored.go
nanassito: moved to internal

redis/cache/redispipebp/monitored.go
nanassito: updated

redis/cache/redispipebp/monitored.go
nanassito: it isn't. `redis_client_name` is the name passed to the redis backend to identify the client. Here `s.Name` is the "slug" used to identity the client pool.

redis/db/redisbp/hooks.go
nanassito: I don't understand the second part of your statement.

As for `h.ClientName`, it refers to the pool and isn't being passed down to Redis to identity the client. So it is the pool.

nanassito:
added the new metrics to the existing tests

redis/internal/redisprom/metrics.go
nanassito: gone

redis/cache/redispipebp/monitored.go
fishy: someone can correct me if I'm wrong but I believe what `redis_client_name` means is the same as slug (we also have another spec change to rename all `_slug` labels to `_client_name`). redis does not really have `client name` in the wire protocol.

redis/db/redisbp/hooks.go
fishy: from my understanding (and I got that from the spec pr discussion with kyle and nate), redis' "database" is basically something like this:

"ip:port/n"

while by default n=0 and you can omit it as "ip:port" (which is the vast majority of our usages)

I _believe_ (but I could totally be wrong) that n must be an int (vs. free form string)

nanassito: is this the same thing as `redis.Options.DB` ?

fishy: looks so but I'm not 100% sure. you probably need someone more familiar with redis to confirm it 😅

redis/cache/redispipebp/monitored.go
```diff
@@ -60,19 +61,46 @@ func (s MonitoredSync) Do(ctx context.Context, cmd string, args ...interface{})
...
+			redisprom.TypeLabel:       "", // We don't have that info
+			redisprom.DeploymentLabel: "", // We don't have that info
+		}).Inc()
 		span.FinishWithOptions(tracing.FinishOptions{
 			Ctx: ctx,
 			Err: err,
 		}.Convert())
 	}(time.Now())
 
+	active.Inc()
```
kylelemons, line 103: 🔕  very minor nit, but I tend to have the "do" go before the corresponding deferred "undo"... `lock(); defer unlock()`, and `inc(); defer dec()` -- so I don't think it matters much if you don't want to go through all of these files and move them, but I think it would be very slightly clearer if you did

```
active.inc()
defer func(start time.Time) {
  active.dec()
  // ...
}
```

kylelemons (approved):

redis/cache/redispipebp/monitored.go
nanassito: updated

redis/db/redisbp/hooks.go
nanassito: > MAY be blank if not specified.

Can I just leave it blank ?

redis/cache/redispipebp/monitored.go
nanassito: updated

nanassito: removed because it breaks the expectation of the other commands

redis/db/redisbp/hooks.go
```diff
@@ -79,20 +82,54 @@ func (h SpanHook) startChildSpan(ctx context.Context, cmdName string) context.Co
 		name,
 		tracing.SpanTypeOption{Type: tracing.SpanTypeClient},
 	)
+	redisprom.ActiveRequests.With(prometheus.Labels{
+		redisprom.ClientNameLabel: h.ClientName,
+		redisprom.TypeLabel:       h.Type,
+		redisprom.CommandLabel:    cmdName,
+		redisprom.DeploymentLabel: h.Deployment,
+		redisprom.DatabaseLabel:   "", // We don't have that info
```
fishy, line 90: we can just add this field to `SpanHook` struct and get it, right?

redis/db/redisbp/hooks.go
```diff
@@ -79,20 +82,54 @@ func (h SpanHook) startChildSpan(ctx context.Context, cmdName string) context.Co
...
+		redisprom.CommandLabel:    command,
+		redisprom.DeploymentLabel: h.Deployment,
+		redisprom.SuccessLabel:    prometheusbp.BoolString(err == nil),
+		redisprom.DatabaseLabel:   "", // We don't have that info
+	}).Inc()
+	redisprom.ActiveRequests.With(prometheus.Labels{
+		redisprom.DeploymentLabel: h.ClientName,
+		redisprom.TypeLabel:       h.Type,
+		redisprom.CommandLabel:    command,
+		redisprom.ClientNameLabel: "", // We don't have that info
```
fishy, line 130: and we do have this?

redis/db/redisbp/hooks.go
```diff
@@ -79,20 +82,54 @@ func (h SpanHook) startChildSpan(ctx context.Context, cmdName string) context.Co
...
+	redisprom.RequestsTotal.With(prometheus.Labels{
+		redisprom.ClientNameLabel: h.ClientName,
+		redisprom.TypeLabel:       h.Type,
+		redisprom.CommandLabel:    command,
+		redisprom.DeploymentLabel: h.Deployment,
+		redisprom.SuccessLabel:    prometheusbp.BoolString(err == nil),
+		redisprom.DatabaseLabel:   "", // We don't have that info
+	}).Inc()
+	redisprom.ActiveRequests.With(prometheus.Labels{
+		redisprom.DeploymentLabel: h.ClientName,
```
fishy, line 127: typo/copy-pasta error?

redis/db/redisbp/monitored_client.go
```diff
@@ -21,11 +22,19 @@ type PoolStatser interface {
...
+		return "reddit"
+	}
+}
+
 // NewMonitoredClient creates a new *redis.Client object with a redisbp.SpanHook
 // attached that connects to a single Redis instance.
 func NewMonitoredClient(name string, opt *redis.Options) *redis.Client {
 	client := redis.NewClient(opt)
-	client.AddHook(SpanHook{ClientName: name})
+	client.AddHook(SpanHook{ClientName: name, Type: "standalone", Deployment: getDeploymentType(opt.Addr)})
```
fishy, line 37: I believe that https://pkg.go.dev/github.com/go-redis/redis/v8#Options.DB and https://pkg.go.dev/github.com/go-redis/redis/v8#FailoverOptions.DB are the database label we needed (we don't have it for cluster client, so just use empty strring/"0")

fishy (changes requested):

fishy:
Thank you for keeping the old metrics for now. Can you also add the new pool metrics from the spec to https://github.com/reddit/baseplate.go/blob/71c84dd732f550c1ed7a111f6c060ea978a4cc15/redis/db/redisbp/prometheus.go#L92-L134?

redis/db/redisbp/hooks.go
nanassito: indeed

redis/db/redisbp/monitored_client.go
nanassito: spec calls for empty string if we don't have it, but I'm updating the spanHook to pass it when we do have it

nanassito:
@fishy I wanted to PR to be only about the client metrics. Can we separate the pool metrics into another PR ?

redis/db/redisbp/monitored_client.go
```diff
@@ -97,7 +107,7 @@ func (cc *ClusterClient) Wait(ctx context.Context, args WaitArgs) (replicas int6
 // redisbp.SpanHook attached.
 func NewMonitoredClusterClient(name string, opt *redis.ClusterOptions) *ClusterClient {
 	client := redis.NewClusterClient(opt)
-	client.AddHook(SpanHook{ClientName: name})
+	client.AddHook(SpanHook{ClientName: name, Type: "cluster", Deployment: getDeploymentType(opt.Addrs[0])})
```
nanassito, line 110: Note that I don't have the DB here

redis/db/redisbp/hooks.go
nanassito: passed through when I have it

redis/cache/redispipebp/monitored.go
```diff
@@ -83,17 +111,43 @@ func (s MonitoredSync) Send(ctx context.Context, r redis.Request) (result interf
...
-		}
+		durationSeconds := time.Since(start).Seconds()
 		promHistogram.With(prometheus.Labels{
 			labelSlug:    s.Name,
-			labelCommand: cmd,
+			labelCommand: command,
 			labelSuccess: prometheusbp.BoolString(err == nil),
-		}).Observe(time.Since(start).Seconds())
+		}).Observe(durationSeconds)
+		active.Dec()
```
fishy, line 134: this should also be moved to be the first thing of the defer'd function, as [kyle's suggestion](https://github.com/reddit/baseplate.go/pull/552#discussion_r938997231)

redis/cache/redispipebp/monitored.go
```diff
@@ -132,11 +194,29 @@ func (s MonitoredSync) SendMany(ctx context.Context, reqs []redis.Request) (resu
...
 		}
+		durationSeconds := time.Since(start).Seconds()
 		promHistogram.With(prometheus.Labels{
 			labelSlug:    s.Name,
-			labelCommand: cmd,
+			labelCommand: command,
 			labelSuccess: prometheusbp.BoolString(err == nil),
-		}).Observe(time.Since(start).Seconds())
+		}).Observe(durationSeconds)
+		active.Dec()
```
fishy, line 203: same here

redis/cache/redispipebp/monitored.go
```diff
@@ -148,18 +228,44 @@ func (s MonitoredSync) SendMany(ctx context.Context, reqs []redis.Request) (resu
...
 	defer func(start time.Time) {
+		durationSeconds := time.Since(start).Seconds()
 		promHistogram.With(prometheus.Labels{
 			labelSlug:    s.Name,
-			labelCommand: cmd,
+			labelCommand: command,
 			labelSuccess: prometheusbp.BoolString(err == nil),
-		}).Observe(time.Since(start).Seconds())
+		}).Observe(durationSeconds)
+		active.Dec()
```
fishy, line 252: and here

redis/cache/redispipebp/monitored.go
```diff
@@ -193,18 +299,45 @@ func (s MonitoredScanIterator) Next() (results []string, err error) {
...
-		const cmd = "scanner-next"
+		durationSeconds := time.Since(start).Seconds()
 		promHistogram.With(prometheus.Labels{
 			labelSlug:    s.name,
-			labelCommand: cmd,
+			labelCommand: command,
 			labelSuccess: prometheusbp.BoolString(err == nil),
-		}).Observe(time.Since(start).Seconds())
+		}).Observe(durationSeconds)
+		active.Dec()
```
fishy, line 318: and here

redis/db/redisbp/monitored_client.go
fishy: I would slightly prefer to split the `SpanHook` fields into one-per-line and explicitly add `Database: "", // we don't have this info for cluster clients`.

fishy (approved):

redis/cache/redispipebp/monitored.go
nanassito: updated

nanassito:
@fishy I don't have permissions to merge this but if you think this is ready, please merge it

PR #553, opened 2022-08-15
feat(prom): Implement BP.spec metrics for the pool in redisbp.

nanassito:
## 💸 TL;DR
This updates redisbp to publish the metrics expected by baseplate.spec.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


nanassito:
cc @kylelemons @fishy

redis/db/redisbp/prometheus.go
```diff
@@ -131,4 +132,30 @@ func (e exporter) Collect(ch chan<- prometheus.Metric) {
...
+		prometheus.GaugeValue,
+		float64(stats.IdleConns),
+		e.name,
+	)
+	// ch <- prometheus.MustNewConstMetric(
+	// 	redisprom.PeakActiveConnectionsDesc,
+	// 	prometheus.GaugeValue,
+	// 	float64(xxx),
+	// 	e.name,
+	// )
```
nanassito, line 154: that one is missing at this point. It is a bit annoying to implement this I'll need to lock to prevent race conditions.

fishy: let's remove the commented out code and add a comment of `// TODO: implement redisprom.PeakActiveConnectionsDesc`

nanassito: I implemented it :p

nanassito:
I'm not super happy with the PeakConnection approach. It also doesn't work in the `MonitoredScanIterator` though it is possible to duplicate the code to make it work there too.

Do you prefer me removing support for that metric completely or going forward with the current approach?

redis/cache/redispipebp/monitored.go
```diff
@@ -54,6 +58,26 @@ func extractCommand(cmd string) string {
...
+	if err != nil {
+		// How should I handle failure here?
+		return 0
+	}
+	return m.Counter.GetValue()
+}
+
+func (s MonitoredSync) UpdatePeakConnection(active prometheus.Gauge) {
+	peak := redisprom.PeakActiveConnectionsGauge.WithLabelValues(s.Name)
+	current := readGaugeValue(active)
```
fishy, line 73: I don't think we should read the gauge value.

Let's use something like https://github.com/reddit/baseplate.go/pull/554 instead?

nanassito: sure I can move to that once it is merged

redis/cache/redispipebp/monitored.go
```diff
@@ -123,6 +148,7 @@ func (s MonitoredSync) Send(ctx context.Context, r redis.Request) (result interf
 		redisprom.DeploymentLabel: "", // We don't have that info
 	})
 	active.Inc()
+	s.UpdatePeakConnection(active)
```
pacejackson, line 151: I don't think this is really right, or at least it's not measuring connections. Redispipe uses a single connection per node and batches requests into a pipeline so when you send a new request, you aren't using another connection.

pacejackson (changes requested):

pacejackson:
Really none of the connection metrics apply to redispipe AFAIK

redis/cache/redispipebp/monitored.go
konradreiche: +1 

It could also be more than one connection depending on how the client is configured :thinking:

redis/cache/redispipebp/baseplate.go
```diff
@@ -59,17 +60,19 @@ func wrapSender(sender redis.Sender, args BaseplateRedisClientArgs) (s redisx.Sy
 	}
 	if len(args.Retry) != 0 {
 		s = MonitoredSync{
-			Sync: s,
-			Name: args.Name + ".attempt",
+			Sync:                s,
+			Name:                args.Name + ".attempt",
```
konradreiche, line 64: I know this has been here before, but don't we use `_` to separate names in Prometheus?

redis/cache/redispipebp/monitored.go
```diff
@@ -54,6 +58,26 @@ func extractCommand(cmd string) string {
 	return cmd
 }
 
+func readGaugeValue(g prometheus.Gauge) float64 {
+	var m *io_prometheus_client.Metric
+	err := g.Write(m)
+	if err != nil {
+		// How should I handle failure here?
```
konradreiche, line 65: I believe we would need to have a metric-related logger that is configurable. Is this the first time we are handling Prometheus related write errors? :thinking:

redis/internal/redisprom/pools.go
```diff
@@ -0,0 +1,43 @@
+package redisprom
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const RedisPoolLabel = "redis_pool"
```
konradreiche, line 8: We define it as a label here but is it never set with a value when emitting metrics?

konradreiche (changes requested):

redis/cache/redispipebp/baseplate.go
nanassito: in metrics name yes, but this end up in a label so we don't really care about the format

redis/internal/redisprom/pools.go
nanassito: is it on all the metrics defined below in this file.

or am I misunderstanding your comment?

redis/cache/redispipebp/monitored.go
nanassito: removed.

redis/cache/redispipebp/monitored.go
nanassito: Any pointers to how I would do that ?

redis/internal/redisprom/pools.go
konradreiche:
```suggestion
const redisPoolLabel = "redis_pool"
```
Does it need to be exported then, can we just define it locally here?

redis/cache/redispipebp/monitored.go
konradreiche: Good question. For example in the `filewatcher` package we pass in a `log.Wrapper`, see [here](https://github.com/reddit/baseplate.go/blob/master/filewatcher/filewatcher.go#L103)

konradreiche (approved):

redis/internal/redisprom/pools.go
nanassito: done

redis/db/redisbp/example_hooks_test.go
```diff
@@ -23,7 +25,16 @@ func ExampleSpanHook() {
 		baseClient redis.Client
 	)
 	// Add the Hook onto baseClient
-	baseClient.AddHook(redisbp.SpanHook{ClientName: "redis"})
+	baseClient.AddHook(redisbp.SpanHook{
+		ClientName: "redis",
+		PromActive: &prometheusbp.HighWatermarkGauge{
+			HighWatermarkValue:   &prometheusbp.HighWatermarkValue{},
```
fishy, line 31: nit: prefer `new()`, also making this a pointer was a mistake, if I made this a non-pointer then we won't need this line.

redis/db/redisbp/example_hooks_test.go
```diff
@@ -6,7 +6,9 @@ import (
 	"github.com/go-redis/redis/v8"
 	"github.com/opentracing/opentracing-go"
 
+	"github.com/reddit/baseplate.go/internal/prometheusbp"
```
fishy, line 9: also this is an example so services actually don't have access to those `internal` packages 😅

fishy (approved):

redis/db/redisbp/example_hooks_test.go
nanassito: right but then what are they supposed to do ? Not instrument ?

fishy: they can instrument other things they want to instrument (for example, we can define a customized prometheus metric in this example and instrument that as the example)

nanassito: oh yeah I think I misread this file. I'll remove changes completely from here.

fishy (approved):

PR #554, opened 2022-08-16
internal/prometheusbp: Add HighWatermarkValue and HighWatermarkGauge

fishy:
Benchmark result:

    $ go test -bench .
    goos: linux
    goarch: amd64
    pkg: github.com/reddit/baseplate.go/internal/prometheusbp
    cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    BenchmarkHighWatermarkValue-12          24221392                48.11 ns/op            0 B/op          0 allocs/op
    PASS
    ok      github.com/reddit/baseplate.go/internal/prometheusbp    1.221s

cc @nanassito

### DIFF
--- a/internal/prometheusbp/doc.go
+++ b/internal/prometheusbp/doc.go
@@ -1,0 +1,3 @@
+// Package prometheusbp provides common code used by Baseplate.go regarding
+// Prometheus metrics.
+package prometheusbp
--- a/internal/prometheusbp/high_watermark.go
+++ b/internal/prometheusbp/high_watermark.go
@@ -1,0 +1,111 @@
+package prometheusbp
+
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// HighWatermarkValue implements an int64 gauge with high watermark value.
+type HighWatermarkValue struct {
+	lock sync.RWMutex
+	curr int64
+	max  int64
+}
+
+// Inc increases the gauge value by 1.
+func (hwv *HighWatermarkValue) Inc() {
+	hwv.lock.Lock()
+	defer hwv.lock.Unlock()
+
+	hwv.curr++
+	if hwv.curr > hwv.max {
+		hwv.max = hwv.curr
+	}
+}
+
+// Dec decreases the gauge value by 1.
+func (hwv *HighWatermarkValue) Dec() {
+	hwv.lock.Lock()
+	defer hwv.lock.Unlock()
+
+	hwv.curr--
+}
+
+// Set sets the current value, and updates high watermark if needed.
+func (hwv *HighWatermarkValue) Set(v int64) {
+	hwv.lock.Lock()
+	defer hwv.lock.Unlock()
+
+	hwv.curr = v
+	if hwv.curr > hwv.max {
+		hwv.max = hwv.curr
+	}
+}
+
+// Get gets the current gauge value.
+func (hwv *HighWatermarkValue) Get() int64 {
+	hwv.lock.RLock()
+	defer hwv.lock.RUnlock()
+
+	return hwv.curr
+}
+
+// Max returns the max gauge value (the high watermark).
+func (hwv *HighWatermarkValue) Max() int64 {
+	hwv.lock.RLock()
+	defer hwv.lock.RUnlock()
+
+	return hwv.max
+}
+
+func (hwv *HighWatermarkValue) getBoth() (curr, max int64) {
+	hwv.lock.RLock()
+	defer hwv.lock.RUnlock()
+
+	return hwv.curr, hwv.max
+}
+
+// HighWatermarkGauge implements a prometheus.Collector that reports up to 2
+// gauges backed by a HighWatermarkValue.
+type HighWatermarkGauge struct {
+	*HighWatermarkValue
+
+	// Optional gauge to report the current value when scraped
+	CurrGauge            *prometheus.Desc
+	CurrGaugeLabelValues []string
+
+	// Optional gauge to report the max value when scraped
+	MaxGauge            *prometheus.Desc
+	MaxGaugeLabelValues []string
+}
+
+// Describe implements prometheus.Collector.
+func (hwg HighWatermarkGauge) Describe(ch chan<- *prometheus.Desc) {
+	// All metrics are described dynamically.
+}
+
+// Collect implements prometheus.Collector.
+func (hwg HighWatermarkGauge) Collect(ch chan<- prometheus.Metric) {
+	curr, max := hwg.HighWatermarkValue.getBoth()
+
+	if hwg.CurrGauge != nil {
+		ch <- prometheus.MustNewConstMetric(
+			hwg.CurrGauge,
+			prometheus.GaugeValue,
+			float64(curr),
+			hwg.CurrGaugeLabelValues...,
+		)
+	}
+
+	if hwg.MaxGauge != nil {
+		ch <- prometheus.MustNewConstMetric(
+			hwg.MaxGauge,
+			prometheus.GaugeValue,
+			float64(max),
+			hwg.MaxGaugeLabelValues...,
+		)
+	}
+}
+
+var _ prometheus.Collector = HighWatermarkGauge{}
--- a/internal/prometheusbp/high_watermark_test.go
+++ b/internal/prometheusbp/high_watermark_test.go
@@ -1,0 +1,125 @@
+package prometheusbp_test
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/reddit/baseplate.go/internal/prometheusbp"
+)
+
+func TestHighWatermarkValue(t *testing.T) {
+	for _, c := range []struct {
+		label string
+		get   int64
+		max   int64
+		setup func(*prometheusbp.HighWatermarkValue)
+	}{
+		{
+			label: "inc-once",
+			get:   1,
+			max:   1,
+			setup: func(hwv *prometheusbp.HighWatermarkValue) {
+				hwv.Inc()
+			},
+		},
+		{
+			label: "inc-inc-dec",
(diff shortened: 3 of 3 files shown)


prometheusbp/high_watermark_gauge.go
```diff
@@ -0,0 +1,58 @@
+package prometheusbp
+
+import (
+	"sync"
+)
+
+// HighWatermarkGauge implements an int64 gauge with high watermark value.
+type HighWatermarkGauge struct {
+	lock sync.RWMutex
```
fishy, line 9: benchmark result with `RWMutex`:
```
goos: linux
goarch: amd64
pkg: github.com/reddit/baseplate.go/prometheusbp
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkHighWatermarkGauge-12          20724597                55.75 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/reddit/baseplate.go/prometheusbp     1.221s
```

benchmark result if we use `Mutex` instead:
```
goos: linux
goarch: amd64
pkg: github.com/reddit/baseplate.go/prometheusbp
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkHighWatermarkGauge-12          12081015               103.0 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/reddit/baseplate.go/prometheusbp     1.351s
```

prometheusbp/high_watermark.go
```diff
@@ -0,0 +1,111 @@
+package prometheusbp
+
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// HighWatermarkValue implements an int64 gauge with high watermark value.
```
kylelemons, line 9: I think this exposes too much of this API surface and makes it too hard to use.

I would use prometheus.Histogram as a guide here.  It's exposed as a single API but when collected it produces more than one metric with a fixed naming scheme.

Expose:
* `func NewHighWaterMarkGauge(HighWatermarkGaugeArgs) *HighWatermarkGauge`
* `func NewHighWaterMarkGaugeVec(HighWatermarkGaugeArgs, labels) *HighWatermarkGaugeVec`
* `type HighWatermarkGaugeVec`
  * With / WithLabels
* `type HighWatermarkGauge`
  * Add / Sub / Set
  * Inc / Dec
  * Collect

kylelemons (changes requested):

prometheusbp/high_watermark.go
fishy: I would strongly prefer to separate the API for the value and api for the actual prometheus metrics (the 2 gauges). There's no guarantee that we would always define the same set of labels for the 2 gauges, so an API like `prometheus.Histogram` can be very brittle.

As for `Add`/`Sub`, I just don't really see a use case for them yet. If we do have the use case for them later we can add them when then.

prometheusbp/high_watermark.go
```diff
@@ -0,0 +1,111 @@
...
+func (hwv *HighWatermarkValue) getBoth() (curr, max int64) {
+	hwv.lock.RLock()
+	defer hwv.lock.RUnlock()
+
+	return hwv.curr, hwv.max
+}
+
+// HighWatermarkGauge implements a prometheus.Collector that reports up to 2
+// gauges backed by a HighWatermarkValue.
+type HighWatermarkGauge struct {
```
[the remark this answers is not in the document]
fishy: So the actual gauge api is mostly just a sugar for the common use cases we have in the spec implementations. It's a very think wrapper around the value api that's trivial to implement, so I initially did not include this part, but then thought that it's probably better to save some repetitions.

What do you think if I move the gauge api to `internal` and only leave the value api as public?

prometheusbp/high_watermark.go
kylelemons: > There's no guarantee that we would always define the same set of labels for the 2 gauges

I would approach this from the _other_ side.  Why would we _not_ want them to have the same labels?  Unlike a histogram where you may want a counter with additional labels, we're not talking about exploding cardinality here, and it would be pretty confusing to have different labels for the instant and high watermark metrics.

prometheusbp/high_watermark.go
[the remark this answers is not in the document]
kylelemons: I think the valuable thing to export is an easy-to-use API.  If this is just used for internal things, then maybe we don't expose any of it.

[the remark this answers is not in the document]
fishy: ok moved both to internal.

internal/prometheusbp/high_watermark_test.go
```diff
@@ -0,0 +1,125 @@
...
+			hwv.Max()
+		},
+	}
+	hwv := new(prometheusbp.HighWatermarkValue)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			n := int(atomic.AddInt64(&counter, 1)) % len(operations)
```
kylelemons, line 121: 🔕 I feel like having a goroutine-local counter in a var is fine -- they don't need to rotate through the operations in lockstep for it to make sense.  That will remove the atomic overhead, which is non-negligible when you're benchmarking a mutex

kylelemons (approved):

internal/prometheusbp/high_watermark_test.go
fishy: good point. will do.

internal/prometheusbp/high_watermark.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// HighWatermarkValue implements an int64 gauge with high watermark value.
+type HighWatermarkValue struct {
+	lock sync.RWMutex
```
fishy, line 11: Benchmark result with `RWMutex`:
```
$ go test -bench .
goos: linux
goarch: amd64
pkg: github.com/reddit/baseplate.go/internal/prometheusbp
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkHighWatermarkValue-12          25154269                47.51 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/reddit/baseplate.go/internal/prometheusbp    1.250s
```

benchmark result with `Mutex`:
```
$ git diff
diff --git a/internal/prometheusbp/high_watermark.go b/internal/prometheusbp/high_watermark.go
index ccfa193..6f12a0e 100644
--- a/internal/prometheusbp/high_watermark.go
+++ b/internal/prometheusbp/high_watermark.go
@@ -8,7 +8,7 @@ import (
 
 // HighWatermarkValue implements an int64 gauge with high watermark value.
 type HighWatermarkValue struct {
-       lock sync.RWMutex
+       lock sync.Mutex
        curr int64
        max  int64
 }
@@ -45,23 +45,23 @@ func (hwv *HighWatermarkValue) Set(v int64) {
 
 // Get gets the current gauge value.
 func (hwv *HighWatermarkValue) Get() int64 {
-       hwv.lock.RLock()
-       defer hwv.lock.RUnlock()
+       hwv.lock.Lock()
+       defer hwv.lock.Unlock()
 
        return hwv.curr
 }
 
 // Max returns the max gauge value (the high watermark).
 func (hwv *HighWatermarkValue) Max() int64 {
-       hwv.lock.RLock()
-       defer hwv.lock.RUnlock()
+       hwv.lock.Lock()
+       defer hwv.lock.Unlock()
 
        return hwv.max
 }
 
 func (hwv *HighWatermarkValue) getBoth() (curr, max int64) {
-       hwv.lock.RLock()
-       defer hwv.lock.RUnlock()
+       hwv.lock.Lock()
+       defer hwv.lock.Unlock()
 
        return hwv.curr, hwv.max
 }
$ go test -bench .
goos: linux
goarch: amd64
pkg: github.com/reddit/baseplate.go/internal/prometheusbp
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkHighWatermarkValue-12          15068430                76.76 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/reddit/baseplate.go/internal/prometheusbp    1.244s
```

internal/prometheusbp/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package prometheusbp provides common code used by Baseplate.go regarding
```
konradreiche, line 1:
```suggestion
// Package prometheusbp provides common code used by baseplate.go regarding
```
:no_bell: is the official name all lowercase?

internal/prometheusbp/high_watermark.go
konradreiche, line 11:
```suggestion
	mu sync.RWMutex
```
Is mu to refer to the lock more common?

konradreiche (approved):
Could we add some comments in code or the git commit that explains what a high watermark value is?

internal/prometheusbp/high_watermark.go
fishy: why do you want to use **mu** to refer to lock instead of using **lock** to refer to lock though :)

konradreiche: For no good reason other than convention. I see it mostly being called `mu` in Go https://go.dev/tour/concurrency/9
Maybe because `lock.RLock()` reads a bit repetitive :shrug:

fishy: `lock` is noun and `Lock`/`RLock` is verb :)

internal/prometheusbp/doc.go
fishy: it's not :)

PR #555, opened 2022-08-16
Add a global registry that forces a baseplate_go="v0" constant label

kylelemons:
## 💸 TL;DR
We want to allow linking in both versions of baseplate into the same binary to permit transitional applications.

This PR adds a constant label `{baseplate_go="v0"}` to all exported promauto metrics.  This will allow both libraries to expose metrics with spec-compliant names and labels without conflict, but it will require some amount of coordination to keep the labels and help strings in sync.

### DIFF
--- a/log/wrapper.go
+++ b/log/wrapper.go
@@ -289,7 +289,7 @@ type Counter interface {
 // baseplate.ParseConfigYAML call, for example:
 //
 //	// a global variable
-//	var tracingFailures = promauto.With(internalv2compat.GlobalRegistry).NewCounter(prometheus.CounterOpts{
+//	var tracingFailures = promauto.NewCounter(prometheus.CounterOpts{
 //	  Namespace: "myservice",
 //	  Subsystem: "tracing",
 //	  Name:      "failures_total",
--- a/log/wrapper.go
+++ b/log/wrapper.go
@@ -289,7 +289,7 @@ type Counter interface {
 // baseplate.ParseConfigYAML call, for example:
 //
 //	// a global variable
-//	var tracingFailures = promauto.NewCounter(prometheus.CounterOpts{
+//	var tracingFailures = promauto.With(internalv2compat.GlobalRegistry).NewCounter(prometheus.CounterOpts{
 //	  Namespace: "myservice",
 //	  Subsystem: "tracing",
 //	  Name:      "failures_total",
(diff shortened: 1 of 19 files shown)


log/wrapper.go
```diff
@@ -289,7 +289,7 @@ type Counter interface {
 // baseplate.ParseConfigYAML call, for example:
 //
 //	// a global variable
-//	var tracingFailures = promauto.NewCounter(prometheus.CounterOpts{
+//	var tracingFailures = promauto.With(internalv2compat.GlobalRegistry).NewCounter(prometheus.CounterOpts{
```
fishy, line 292: this example certainly should not add that :)

thriftbp/server.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/reddit/baseplate.go"
 	"github.com/reddit/baseplate.go/errorsbp"
+
```
fishy, line 11: nit: this blank line will cause weird sorting issues in the future (e.g. people might be adding another import after `errorsbp`)

fishy (changes requested):

fishy:
does this mean people would need to add `{baseplate_go="v0", ...}` on their counters otherwise they will get double value if that's something emitted by both?

pacejackson (approved):

log/wrapper.go
kylelemons: whoops, I did a find/replace and missed a few that shouldn't be covered, thanks

thriftbp/server.go
kylelemons: interesting, I'm not sure what put that here.  It might've been goimports, which is how I got the import added after the find/replace.  Fixed, thanks.

kylelemons:
> does this mean people would need to add {baseplate_go="v0", ...} on their counters otherwise they will get double value if that's something emitted by both?

I'm not sure what you mean; this allows two metrics to coexist within the same binary from different libraries, which should only apply to common libraries, not to user code.  A service that imports v0 and v2 might get two copies of the `initial_gomaxprocs` varable, but will not get two copies of their own metrics.

fishy (approved):
:fish: after you fixed the tests

ghirsch-reddit (approved):

mterwill (approved):

PR #557, opened 2022-08-19
Cancel serve externally with call context's cancel. 

VVildVVolf:
Closes # 

## 💸 TL;DR

This helps to stop baseplate.Serve externally, as well as catching os signals.

## 📜 Details
[Design Doc]()

[Jira]()

`go test -run StopByCancel`

## ✅ Checks

- [ ] CI tests (if present) are passing (not sure how to check)

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -209,7 +209,12 @@ func Serve(ctx context.Context, args ServeArgs) error {
 			select {
 			case <-ctx.Done():
 				// The context timed-out or was cancelled so use that error.
-				err = fmt.Errorf("baseplate: context cancelled while waiting for server.Close(). %w", ctx.Err())
+				switch ctx.Err() {
+				case context.DeadlineExceeded:
+					err = fmt.Errorf("baseplate: context cancelled with timeout while waiting for server.Close(). %w", ctx.Err())
+				default:
+					err = fmt.Errorf("baseplate: context cancelled externally. %w", ctx.Err())
+				}
 			case e := <-closeChannel:
 				// server.Close() completed and passed its result to closeChannel, so
 				// use that value.
--- a/baseplate_test.go
+++ b/baseplate_test.go
@@ -235,7 +235,7 @@ func TestServeStopByCancel(t *testing.T) {
 	var resultError error
 	go func() {
 		// Let server be closed with context.
-		time.Sleep(time.Second*5)
+		time.Sleep(time.Second * 5)
 		if resultError == nil {
 			return
 		}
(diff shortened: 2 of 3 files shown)


kylelemons (changes requested):
Can you either sign the CLA or ping me in slack if you're an employee?

(you got rid of the PR boilerplate, btw, which has a number of instructions)

baseplate.go
```diff
@@ -209,7 +209,12 @@ func Serve(ctx context.Context, args ServeArgs) error {
 			select {
 			case <-ctx.Done():
 				// The context timed-out or was cancelled so use that error.
-				err = fmt.Errorf("baseplate: context cancelled while waiting for server.Close(). %w", ctx.Err())
+				switch ctx.Err() {
+				case context.DeadlineExceeded:
+					err = fmt.Errorf("baseplate: context cancelled with timeout while waiting for server.Close(). %w", ctx.Err())
+				default:
+					err = fmt.Errorf("baseplate: context cancelled externally. %w", ctx.Err())
```
ghirsch-reddit, line 216: nit, the original wording seemed fine for the default case
```suggestion
					err = fmt.Errorf("baseplate: context cancelled while waiting for server.Close(). %w", ctx.Err())
```

ghirsch-reddit (dismissed):
It's not clear to me that this is accomplishing what you want it to, since it just adds a new error message in that one case, and canceling a context was already possible without this change.

But it seems like a fine change to the error messages, so approving.

VVildVVolf:
@kylelemons the header is updated according to  PR boilerplate. I checked some previous PRs, seems like them, but please let me know if I need to improve something else

baseplate.go
VVildVVolf: Actually, the initial "while waiting for server.Close()" [seems like] meant that the exception was thrown, because of closing operation was happening too long. The runtimebp.HandleShutdown function notified ShutdownHandler only if runtimebp.HandleShutdown catch os.Signal - this let's ShutdownHandler write to the chanel, which runtimebp.HandleShutdown is waiting for to return. That's why the PR firstly handle the ctx.Done() instead of ignoring. The granulation of exception is minor in this case - just to check actual reason of stoping timeout|deadline

kylelemons:
Had a chat offline with @VVildVVolf -- we're going to close this in favor of the lifecycle improvements in v2, and revisit this PR if we find that v0 services can run into this case in the wild.  Since we don't control all callers of HandleShutdown, for example, we can't be fully confident that this change will have the desired and safe effect.

PR #558, opened 2022-08-23
bug(prom) | Re-add the global constant label that got dropped on 9814ca

nanassito:
## 💸 TL;DR
Previous changes went in concurrently and we ended up missing a label. This adds it back.

## ✅ Checks

### DIFF
--- a/redis/internal/redisprom/pools.go
+++ b/redis/internal/redisprom/pools.go
@@ -3,12 +3,15 @@ package redisprom
 import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	//lint:ignore SA1019 This library is internal only, not actually deprecated
+	"github.com/reddit/baseplate.go/internalv2compat"
 )
 
 const redisPoolLabel = "redis_pool"
 
 var (
-	MaxSizeGauge = promauto.NewGaugeVec(
+	MaxSizeGauge = promauto.With(internalv2compat.GlobalRegistry).NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "redis_client_max_size",
 			Help: "configured maximum number of clients to keep in the pool (for showing % used in dashboards)",


mterwill (approved):

redis/internal/redisprom/pools.go
```diff
@@ -3,12 +3,13 @@ package redisprom
 import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/reddit/baseplate.go/internalv2compat"
```
fishy, line 6: you have to add `//lint:ignore SA1019` before to silence staticcheck. example: https://github.com/reddit/baseplate.go/blob/9814cae3aaf1c265c18461680590a766e390d5a6/kafkabp/prometheus.go#L6

fishy (changes requested):

kylelemons (approved):

redis/internal/redisprom/pools.go
```diff
@@ -3,12 +3,15 @@ package redisprom
 import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+
```
fishy, line 6: and please don't add the extra blank line, which would cause sorting issues :)

nanassito: go fmt just re-adds it every time

redis/internal/redisprom/pools.go
```diff
@@ -3,12 +3,15 @@ package redisprom
...
+	//lint:ignore SA1019 This library is internal only, not actually deprecated
```
fishy, line 7: @kylelemons I think we should move `GlobalRegistry` to an actual internal package from `internalv2compat`, as it's not needed outside of this repo (unlike other things in `internalv2compat`), so that we don't have to do this every time it's used.

`internal/prometheusbp` is a good place for it.

redis/internal/redisprom/pools.go
fishy: I'm pretty sure that's `goimport` not `go fmt` (a lot of IDEs uses `goimport` instead of `go fmt`)

nanassito: sure either way it just comes back on its own using standard tooling

fishy: `goimport` is not "standard tooling" :)

fishy:
Replaced by #559

PR #559, opened 2022-08-23
Move GlobalRegistry from internalv2compat to internal/prometheusbp

fishy:
This is not really needed outside of Baseplate.go (unlike other stuff in
internalv2compat) so we can move it to an actual internal package to
stop fighting with staticcheck.

Also rename internal/prometheusbp to internal/prometheusbpint to avoid
the need to rename import when using it.

Also some fixups from previous commits:

1. We missed moving redisprom's MaxSizeGauge to GlobalRegistry
2. We missed moving custom exporters to GlobalRegistry
3. We didn't actually register the PromActive with prometheus register
4. SpanHook.PromActive is not something people can initialize outside of
   Baseplate.go, so unexport it.
5. Also add enforcement of prometheusbpint.GlobalRegistry to linters.sh


grpcbp/prometheus.go
```diff
@@ -6,8 +6,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 
-	//lint:ignore SA1019 This library is internal only, not actually deprecated
-	"github.com/reddit/baseplate.go/internalv2compat"
+	internal_prometheusbp "github.com/reddit/baseplate.go/internal/prometheusbp"
```
kylelemons, line 9: The style established by the standard library is that you don't name anything with underscores, even package names.  Either `internalprometheusbp` or `prometheusbpint` or something like that.

internal/prometheusbp/registry.go
```diff
@@ -1,4 +1,4 @@
-package internalv2compat
+package prometheusbp
```
kylelemons, line 1: is this a new package?  It looks like it but I can't quite tell from the UI.

If so, I would strongly recommend renaming this since this PR has already shown us that this is going to cause conflicts.

Suggested names: `internalprom` `promint`

https://go.dev/blog/package-names

redis/db/redisbp/hooks.go
```diff
@@ -33,7 +33,8 @@ type SpanHook struct {
 	Type       string
 	Deployment string
 	Database   string
-	PromActive *internal_prometheusbp.HighWatermarkGauge
+
+	promActive *internal_prometheusbp.HighWatermarkGauge
```
kylelemons, line 37: 🔕 hmm, this is a bit unfortunate.

kylelemons (approved):

internal/prometheusbp/registry.go
fishy: it's not new, but since it's internal we can still rename it.

renamed to `prometheusbpint`.

redis/db/redisbp/hooks.go
fishy: if we really have the need for this raised in the future we can define an interface for it and export it again. I think currently making the whole `SpanHook` exported was a mistake but that requires a breaking change to fix so ¯\\\_(ツ)\_/¯

kylelemons (approved):

nanassito (approved):

pacejackson (approved):

PR #560, opened 2022-08-23
Ensure const metrics are applied with a lint check

kylelemons:
A few metrics that use `prometheus.Register` were missed in https://github.com/reddit/baseplate.go/pull/555 .  This adds them and adds a lint check to prevent future misses.


mterwill (approved):

fishy:
Replaced by #559

PR #561, opened 2022-08-26
filewatcher: Implement support for directories

fishy:
This replaces the filewatcher part of changes of
https://github.com/reddit/baseplate.go/pull/472.

### DIFF
--- a/filewatcher/filewatcher.go
+++ b/filewatcher/filewatcher.go
@@ -5,26 +5,32 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
+	"sync"
 	"sync/atomic"
 	"time"
 
-	"gopkg.in/fsnotify.v1"
+	"github.com/fsnotify/fsnotify"
 
 	"github.com/reddit/baseplate.go/internal/limitopen"
 	"github.com/reddit/baseplate.go/log"
 )
 
+// DefaultFSEventsDelay is the default FSEventsDelay used when creating a new
+// FileWatcher.
+const DefaultFSEventsDelay = 1 * time.Second
+
 // DefaultPollingInterval is the default PollingInterval used when creating a
-// new file watcher.
+// new FileWatcher.
 const DefaultPollingInterval = 30 * time.Second
 
-// FileWatcher loads and parses data from a file and watches for changes to that
-// file in order to refresh it's stored data.
+// FileWatcher loads and parses data from a file or directory, and watches for
+// changes in order to refresh its stored data.
 type FileWatcher interface {
 	// Get returns the latest, parsed data from the FileWatcher.
-	Get() interface{}
+	Get() any
 
 	// Stop stops the FileWatcher.
 	//
@@ -37,7 +43,7 @@ type FileWatcher interface {
 }
 
 // InitialReadInterval is the interval to keep retrying to open the file when
-// creating a new file watcher, when the file was not initially available.
+// creating a new FileWatcher, when the file was not initially available.
 //
 // It's intentionally defined as a variable instead of constant, so that the
 // caller can tweak its value when needed.
@@ -56,7 +62,29 @@ const (
 //
 // Please note that Parser should always return the consistent type.
 // Inconsistent type will cause panic, as does returning nil data and nil error.
-type Parser func(f io.Reader) (data interface{}, err error)
+type Parser func(f io.Reader) (data any, err error)
+
+// A DirParser is a callback function that will be called when the watched
+// directory has its content changed or is read for the first time.
+//
+// Please note that a DirParser must return a consistent type.
+// Inconsistent types will cause a panic,
+// as does returning nil data and nil error.
+//
+// Use WrapDirParser to wrap it into a Parser to be used with FileWatcher.
+type DirParser func(dir fs.FS) (data any, err error)
+
+// WrapDirParser wraps a DirParser for a directory to a Parser.
+//
+// When using FileWatcher to watch a directory instead of a single file,
+// you MUST use WrapDirParser instead of any other Parser implementations.
+func WrapDirParser(dp DirParser) Parser {
+	return func(r io.Reader) (data any, err error) {
+		path := string(r.(fakeDirectoryReader))
+		dir := os.DirFS(path)
+		return dp(dir)
+	}
+}
 
 // Result is the return type of New. Use Get function to get the actual data.
 type Result struct {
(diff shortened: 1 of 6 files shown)


filewatcher/filewatcher.go
```diff
@@ -56,7 +62,28 @@ const (
 //
 // Please note that Parser should always return the consistent type.
 // Inconsistent type will cause panic, as does returning nil data and nil error.
-type Parser func(f io.Reader) (data interface{}, err error)
+type Parser func(f io.Reader) (data any, err error)
+
+// A DirParser is a callback function to be called the watched directory has its
```
kylelemons, line 67:
```suggestion
// A DirParser is a callback function that will be called when the watched directory has its
```

filewatcher/filewatcher.go
```diff
@@ -102,22 +127,38 @@ func (r *Result) watcherLoop(
...
 		if err != nil {
-			logger.Log(context.Background(), "filewatcher: parser error: "+err.Error())
+			logger.Log(context.Background(), err.Error())
 		} else {
 			r.data.Store(&atomicData{
 				data:  d,
 				mtime: mtime,
 			})
+			// remove all previously watched files
+			for _, path := range watchedFiles {
```
kylelemons, line 148: loop through `watcher.WatchedFiles()` or whatever instead?

filewatcher/filewatcher.go
```diff
@@ -161,24 +202,18 @@ func (r *Result) watcherLoop(
...
-						path,
-						err,
-					))
-					continue
+			case fsnotify.Create, fsnotify.Write, fsnotify.Rename, fsnotify.Remove:
+				// Use fsEventDelay to avoid calling forceReload repetively when a burst
+				// of fs events happens (for example, when multiple files within the
+				// directory changed).
+				if timer == nil {
+					timer = time.AfterFunc(fsEventsDelay, forceReload)
```
kylelemons, line 213: there's still a race condition here it looks like... if it takes longer than 200ms for the files to get written, and then you have an error watching the file(s), I think it can cause the watcher to effectively "forget" about the files forever.  It doesn't currently look to me like the polling interval will recover, though I may be missing code that would make it do so.

kylelemons (changes requested):

filewatcher/filewatcher.go
fishy: that's actually add by a fork of the original `fsnotify` package we were using, but since the original one is already stopped developing it's time to switch to the fork.

filewatcher/filewatcher.go
fishy: we only remove and readd watched files after parser returned success, we also log and ignore all failures of adding files to watch, so failure to watch one file would not affect other files (but yes that particular file could be forgotten forever).

>if it takes longer than 200ms for the files to get written

The intended usage of filewatcher is really for those rename swaps, not really open the files one-by-one and write the updates (we can probably add some more documents to clarify on that). and if someone really is using it to watch directories that's updated one-by-one, the delay is still configurable so they can configure it to be a larger number to mitigate those risks.

filewatcher/filewatcher.go
kylelemons: :ohno:

filewatcher/filewatcher.go
kylelemons: So the problem that I forsee is that this is used to monitor a CSI driver mount point, and those currently (and for a long time since they haven't even fixed it yet) rm the files before writing the new ones, it seems _very_ easy for a lot of or large files to exceed 200ms.

kylelemons (approved):

filewatcher/filewatcher.go
fishy: yea, the k8s implementation is not great :/

do you want to make the default longer? I do not want to make it too long mainly to not make the `secrets` test to sleep for too long (we currently don't expose the api to customize it so it has to use the default one). but maybe 1s is not too bad?

ghirsch-reddit (approved):

filewatcher/filewatcher.go
```diff
@@ -56,7 +62,28 @@ const (
...
 // Please note that Parser should always return the consistent type.
 // Inconsistent type will cause panic, as does returning nil data and nil error.
-type Parser func(f io.Reader) (data interface{}, err error)
+type Parser func(f io.Reader) (data any, err error)
+
+// A DirParser is a callback function that will be called when the watched
+// directory has its content changed or is read for the first time.
+//
+// Please note that DirParser should always return the consistent type.
+// Inconsistent type will cause panic, as does returning nil data and nil error.
```
pacejackson, line 71:
```suggestion
// Please note that a DirParser must return a consistent type.
// Inconsistent types will cause a panic, as does returning nil data and nil error.
```

Why would a nil error cause a panic?

filewatcher/filewatcher.go
```diff
@@ -102,22 +127,36 @@ func (r *Result) watcherLoop(
...
 		} else {
 			r.data.Store(&atomicData{
 				data:  d,
 				mtime: mtime,
 			})
+			// remove all previously watched files
+			for _, path := range watcher.WatchList() {
+				watcher.Remove(path)
+			}
+			// then readd all new files to watch
```
pacejackson, line 150:
```suggestion
			// then read all new files to watch
```

filewatcher/filewatcher.go
```diff
@@ -249,9 +282,66 @@ type Config struct {
 	// the parent directory will be remount upon change
 	// (for example, k8s ConfigMap).
 	PollingInterval time.Duration `yaml:"pollingInterval"`
+
+	// Optional, the delay between fs events and actually read and parse the
+	// changes.
```
pacejackson, line 287:
```suggestion
	// Optional, the delay between receiving the fs events and actually reading and parsing the
	// changes.
```

filewatcher/filewatcher.go
```diff
@@ -249,9 +282,66 @@ type Config struct {
...
+	//
+	// It's used to avoid short burst of fs events (for example, when watching a
```
pacejackson, line 289:
```suggestion
	// It's used to avoid short bursts of fs events (for example, when watching a
```

pacejackson (approved):
Just some comments about comments

filewatcher/filewatcher.go
fishy: That's the caveat from using [`atomic.Value`](https://pkg.go.dev/sync/atomic#Value.Store):

>Store sets the value of the Value to x. All calls to Store for a given Value must use values of the same concrete type. Store of an inconsistent type panics, as does Store(nil).

We can bypass this by doing something similar to 316e8443312ff705e8215947cb34297d81c1bc34, but I don't think we should do that unless absolutely necessary.

PR #562, opened 2022-08-26
Allow changing the implementation of Interface at run-time

kylelemons:
This is primarily necessary during tests, as most production code won't need to change,
but this is based on an internal detail of atomic.Value, which asserts the underlying
type of the value being Store'd is always identical.  Since it doesn't see Interface,
it only sees interface{} (aka any), it actually cares about the concrete type.

There are various ways we can resolve this, covered in the benchmark,
and with the current compiler using atomic.Value with a concrete struct
type seems to be the winner.  This is faster than the older interface version
because the type assertion to a concrete type is faster than the type assertion
to an interface type.

### DIFF
--- a/ecinterface/global_test.go
+++ b/ecinterface/global_test.go
@@ -85,6 +85,7 @@ func BenchmarkAtomics(b *testing.B) {
 			b.Fatalf("this is just to avoid eliding the call")
 		}
 	})
+	/* If you have a go1.18 compiler, you can throw this one in for fun:
 	b.Run("atomicPointer", func(b *testing.B) {
 		var global atomic.Pointer[Interface]
 		var impl Interface = nop{}
@@ -101,4 +102,5 @@ func BenchmarkAtomics(b *testing.B) {
 			b.Fatalf("this is just to avoid eliding the call")
 		}
 	})
+	*/
 }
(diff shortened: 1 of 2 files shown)


fishy (approved):

ecinterface/global_test.go
```diff
@@ -0,0 +1,100 @@
...
+			if global.Load().(current).Interface != nil {
+				count++
+			}
+		}
+		if count != b.N {
+			b.Fatalf("this is just to avoid eliding the call")
+		}
+	})
+	b.Run("atomicPointer", func(b *testing.B) {
+		var global atomic.Pointer[Interface]
```
fishy, line 85: you need to drop support to go 1.18 to use this 😅

ecinterface/global_test.go
```diff
@@ -85,6 +85,7 @@ func BenchmarkAtomics(b *testing.B) {
 			b.Fatalf("this is just to avoid eliding the call")
 		}
 	})
+	/* If you have a go1.18 compiler, you can throw this one in for fun:
```
fishy, line 88: this is actually go 1.19+ only, not 1.18+

fishy (approved):

PR #564, opened 2022-09-13
Update default Go metrics

SuperQ:
Closes # 

## 💸 TL;DR
Re-register the Go metrics collector with new baseplate defaults that include Go `runtime/metrics` scheduler metrics.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


internal/admin/server_test.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	got := []string{}
+	for _, r := range result {
+		got = append(got, r.GetName())
+	}
+
+	if !reflect.DeepEqual(got, expectedMetrics) {
+		t.Errorf("got %v, want %v", got, expectedMetrics)
+	}
```
pacejackson, line 27:
```suggestion
	if diff := cmp.Diff(got, expectedMetrics); diff != " {
		t.Errorf("registered metrics mismatch (-want +got):\n%s", diff)
	}
```

pacejackson (approved):
This looks fine to me, I would want at least one of the other approvers to sign off as well since I'm not really familiar with the metrics side of things.

SuperQ:
Yes, somehow when I hit the "re-request review", github removed @kylelemons and @fishy. :confused:

internal/admin/server.go
```diff
@@ -28,6 +34,12 @@ func init() {
 	Mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
 
 	Mux.Handle("/metrics", promhttp.Handler())
+
+	// Unregister the default GoCollector.
+	prometheus.Unregister(collectors.NewGoCollector())
```
fishy, line 39: :no_bell: why do we have to do this :sigh:, why isn't that the default?

internal/admin/server_go119_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"go_memstats_mcache_inuse_bytes",
+	"go_memstats_mcache_sys_bytes",
+	"go_memstats_mspan_inuse_bytes",
+	"go_memstats_mspan_sys_bytes",
+	"go_memstats_next_gc_bytes",
+	"go_memstats_other_sys_bytes",
+	"go_memstats_stack_inuse_bytes",
+	"go_memstats_stack_sys_bytes",
+	"go_memstats_sys_bytes",
+	"go_sched_gomaxprocs_threads",
```
fishy, line 33: :no_bell: this one seems to be the diff between 1.18 and 1.19?

fishy (approved):

internal/admin/server.go
SuperQ: Because the default registry includes some default settings for Go metrics to make it easy for end users.

Customizing the default registry is a bit awkward. We could avoid this by completely eliminating use of the default registry in baseplate.go. But, that means we would need to have user switch to `promauto.With(reg)...`.

So, we do a little bit of dance to get around it.

internal/admin/server_go119_test.go
SuperQ: Yup, new feature in Go 1.19.

internal/admin/server.go
fishy: we are actually already using `promauto.With(reg)` because of https://github.com/reddit/baseplate.go/commit/f68284f21f52d479de33e969eb09b3270cf1dfa5

SuperQ: Yea, I'm not sure that was the right thing to do. :-/

internal/admin/server_go119_test.go
kylelemons: I think I would skip the build tags and just do

```
expectedMetrics := ...

if !strings.HasPrefix(runtime.Version(), "go1.18") {
  expectedMetrics = append(...)
}
```

Or you could use https://pkg.go.dev/golang.org/x/mod/semver if you want to be a bit more precise.

kylelemons (approved):

internal/admin/server_go119_test.go
SuperQ: So, I thought about this a bit.
* Go doesn't actually use semver, so `semver.Compare()` is going to be messy.
* The build tags match Go version >=, so its a bit future-proof .

I think we need to keep this testing the way it is. (This is how upstream Prometheus client_golang does testing)

SuperQ:
Should we merge this?

PR #565, opened 2022-09-14
feat(secrets): support for csi directories

Tediferous:
Closes # 

## 💸 TL;DR

Allow baseplate go to read secrets from a vault csi directory

[Jira]()

## ✅ Checks

### DIFF
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -267,12 +267,12 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
 	return secretsValidate(secretsDocument)
 }
 
-// NewDirSecrets parses a directory and returns its secrets
-func NewDirSecrets(dir fs.FS) (*Secrets, error) {
+// FromDir parses a directory and returns its secrets
+func FromDir(dir fs.FS) (*Secrets, error) {
 	secretsDocument := Document{
 		Secrets: make(map[string]GenericSecret),
 	}
-	secretsDocument, err := csiPathParser(dir)
+	secretsDocument, err := walkCSIDirectory(dir)
 	if err != nil {
 		return nil, err
 	}
@@ -321,7 +321,8 @@ func secretsValidate(secretsDocument Document) (*Secrets, error) {
 	return secrets, nil
 }
 
-func csiPathParser(dir fs.FS) (Document, error) {
+// walkCSIDirectory parses a directory for vault secrets and merges them into one object
+func walkCSIDirectory(dir fs.FS) (Document, error) {
 	secretsDocument := Document{
 		Secrets: make(map[string]GenericSecret),
 	}
@@ -344,11 +345,14 @@ func csiPathParser(dir fs.FS) (Document, error) {
 			var secretFile CSIFile
 			err = json.NewDecoder(file).Decode(&secretFile)
 			if err != nil {
-				return err
+				return fmt.Errorf("decoding %q: %w", path, err)
 			}
 			secretsDocument.Secrets[path] = secretFile.Secret
 			return nil
 		},
 	)
-	return secretsDocument, err
+	if err != nil {
+		return Document{}, fmt.Errorf("Error during walkCSIDirectory: %w", err)
+	}
+	return secretsDocument, nil
 }
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -29,7 +29,7 @@ func (s Secret) IsEmpty() bool {
 	return len(s) == 0
 }
 
-// CSIFile represent the raw parsed object of a file made by the Vault CSI provider
+// CSIFile represents the raw parsed object of a file made by the Vault CSI provider
 type CSIFile struct {
 	Secret GenericSecret `json:"data"`
 }
@@ -287,8 +287,7 @@ func secretsValidate(secretsDocument Document) (*Secrets, error) {
 		credentialSecrets: make(map[string]CredentialSecret),
 		vault:             secretsDocument.Vault,
 	}
-	err := secretsDocument.Validate()
-	if err != nil {
+	if err := secretsDocument.Validate(); err != nil {
 		return nil, err
 	}
 	for key, secret := range secretsDocument.Secrets {
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -1,11 +1,11 @@
 package secrets
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
 	"io/fs"
-	"os"
 
 	"github.com/reddit/baseplate.go/errorsbp"
 )
@@ -338,12 +338,13 @@ func csiPathParser(dir fs.FS) (Document, error) {
 				return nil
 			}
 			// parse file
-			file, err := os.Open(path)
+			file, err := fs.ReadFile(dir, path)
 			if err != nil {
 				return err
 			}
+			reader := bytes.NewReader(file)
 			var secretFile CSIFile
-			err = json.NewDecoder(file).Decode(&secretFile)
+			err = json.NewDecoder(reader).Decode(&secretFile)
 			if err != nil {
 				return err
 			}
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/fs"
+	"os"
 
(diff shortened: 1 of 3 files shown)


secrets/secrets.go
```diff
@@ -298,3 +322,34 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+		".",
+		func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if d.IsDir() {
+				return nil
+			}
+			// parse file
+			file, err := os.Open(path)
```
fishy, line 341: you actually need to use `fs.Open` instead of `os.Open` (the path is relative inside the fs)

fishy: also you did not close the file :)

secrets/secrets.go
```diff
@@ -298,3 +322,35 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+		".",
+		func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if d.IsDir() {
+				return nil
+			}
+			// parse file
+			file, err := fs.ReadFile(dir, path)
```
fishy, line 341: read the file fully into memory then use `bytes.NewReader` is very inefficient. just use `d.Open`, defer close after error check, would be much better.

secrets/store.go
```diff
@@ -41,13 +43,22 @@ func NewStore(ctx context.Context, path string, logger log.Wrapper, middlewares
 	store := &Store{
 		secretHandlerFunc: nopSecretHandlerFunc,
 	}
+	parser := store.parser
```
fishy, line 46: nit: move this to right before line 53 instead.

fishy (changes requested):
can you please also add some unit tests for the directory case?

ghirsch-reddit (approved):
Agreed with Yuxuan's comment, but otherwise lgtm

konradreiche (approved):

secrets/store_test.go
```diff
@@ -77,6 +137,29 @@ func TestGetSimpleSecret(t *testing.T) {
...
+				}
+				if !reflect.DeepEqual(secret, tt.expected) {
+					t.Fatalf("expected %+v, actual: %+v", tt.expected, secret)
+				}
+			},
+		)
+	}
+	for _, tt := range tests {
+		t.Run(
+			tt.name,
```
fishy, line 155: this will run the subtests with the same name as from line 132. which will make the output very confusing.

either construct the name to add `-file`/`-dir` suffix (e.g. make line 132 `tt.name + "-file"` and here `tt.name + "-dir"`, or run another level of subtests with names `file`/`dir`.

I would strongly prefer to add another level of subtests. you can also make it another level of for loop to reuse the majority of the code:

```go
	for _, tt := range tests {
		t.Run(
			tt.name,
			func(t *testing.T) {
				for label, path := range map[string]string{
					"file": tmpPath,
					"dir": dirCSI,
				} {
					t.Run(label, func(t *testing.T) {
						store, err := secrets.NewStore(context.Background(), path, log.TestWrapper(t))
						if err != nil {
							t.Fatal(err)
						}
						t.Cleanup(func() { store.Close() })

						secret, err := store.GetSimpleSecret(tt.key)
						if tt.expectedError == nil && err != nil {
							t.Fatal(err)
						}
						if tt.expectedError != nil && err.Error() != tt.expectedError.Error() {
							t.Fatalf("expected error %v, actual: %v", tt.expectedError, err)
						}
						if !reflect.DeepEqual(secret, tt.expected) {
							t.Fatalf("expected %+v, actual: %+v", tt.expected, secret)
						}
					})
				}
			},
		)
	}
```

fishy (changes requested):

secrets/store_test.go
Tediferous: you are correct

fishy (approved):

secrets/secrets.go
```diff
@@ -258,16 +264,33 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
 		return nil, err
 	}
 
-	err = secretsDocument.Validate()
+	return secretsValidate(secretsDocument)
+}
+
+// NewDirSecrets parses a directory and returns its secrets
+func NewDirSecrets(dir fs.FS) (*Secrets, error) {
```
mterwill, line 271: This stutters – `secrets.NewDirSecrets`. Perhaps `secrets.FromDir`?

secrets/secrets.go
```diff
@@ -258,16 +264,33 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+
+func secretsValidate(secretsDocument Document) (*Secrets, error) {
 	secrets := &Secrets{
 		simpleSecrets:     make(map[string]SimpleSecret),
 		versionedSecrets:  make(map[string]VersionedSecret),
 		credentialSecrets: make(map[string]CredentialSecret),
 		vault:             secretsDocument.Vault,
 	}
+	err := secretsDocument.Validate()
+	if err != nil {
```
mterwill, line 291:
```suggestion
	if err := secretsDocument.Validate(); err != nil {
```

secrets/secrets.go
```diff
@@ -298,3 +321,35 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+			file, err := dir.Open(path)
+			if err != nil {
+				return err
+			}
+			defer file.Close()
+
+			var secretFile CSIFile
+			err = json.NewDecoder(file).Decode(&secretFile)
+			if err != nil {
+				return err
```
mterwill, line 348: 🔕 bare returns like this can make errors difficult to trace back if they don't contain sufficient context. I'm not going to audit everything in here, but something to keep in mind for the future. Instead:

a) if the error message contains sufficient context, add a comment to let code reviewers and future readers know that you thought about this:

```go
return err // contains context
```

or, b) wrap the error with a helpful message:

```go
return fmt.Errorf("decoding %q: %s", path, err)
```

secrets/secrets.go
```diff
@@ -298,3 +321,35 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
...
+			}
+			secretsDocument.Secrets[path] = secretFile.Secret
+			return nil
+		},
+	)
+	return secretsDocument, err
```
mterwill, line 354: Unless you explicitly need the return value in the error case, it's idiomatic to return `nil, err` in the multi-return error case. The error and the return value are generally considered to be mutually exclusive and this pattern makes the intended use explicit and lowers overall cognitive overhead. (In this case you'd also have to adjust the signature to return a `*Document`)

secrets/store_test.go
```diff
@@ -37,12 +37,72 @@ const specificationExample = `
...
+
+var someAPIKey = `
+{
+  "request_id": "1afc3036-2282-d483-c2d4-6d483efdf16c",
+  "lease_id": "",
+  "lease_duration": 2764800,
+  "renewable": false,
+  "data": {
+    "type": "simple",
+    "value": "Y2RvVXhNMVdsTXJma3BDaHRGZ0dPYkVGSg==",
```
mterwill, line 63: Sanity check: this isn't a real secret, right? Looks more random than the rest.

secrets/secrets.go
```diff
@@ -298,3 +321,35 @@ func NewSecrets(r io.Reader) (*Secrets, error) {
 	}
 	return secrets, nil
 }
+
+func csiPathParser(dir fs.FS) (Document, error) {
```
mterwill, line 325: 🔕 

```suggestion
// walkCSIDirectory ...
func walkCSIDirectory(dir fs.FS) (*Document, error) {
```

mterwill (approved):
A few style/readability nits, otherwise LGTM.

secrets/store_test.go
Tediferous: I dont know, I copied this from the other example secret that this test was already using

secrets/secrets.go
fishy: agreed but we should use `%w` for `err`.

secrets/store_test.go
fishy: I think it's a valid secret that's only used in unit tests. @pacejackson might know more of the origin of it 😅 

I'm 99% sure it's not something we used in prod.

fishy:
Oh sorry I didn't realize that you already got approvals from all reviewers. gonna merge this.

PR #567, opened 2022-09-21
enable using the baseplate.spec default latency buckets

ghirsch-reddit:
## 💸 TL;DR

[Baseplate.spec](https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#latency-histograms) defines a set of buckets that should be used for latency histograms. This PR replaces the old default buckets with the ones defined in the spec

## 📜 Details
Design Doc: N/A

Jira: N/A

## ✅ Checks

~~- [ ] Contributor License Agreement (CLA) completed if not a Reddit employee~~


prometheusbp/metrics_buckets_spec.go
```diff
@@ -0,0 +1,28 @@
+//go:build metrics_spec
+// +build metrics_spec
```
fishy, line 2: I don't really think we should use a build tag for this.

this means services need to change their build process to opt-in for the things defined in the spec. if we really want to use a build tag it should be flipped: that services need to opt-in to keep their dashboards from being broken (if they actually hardcoded the current buckets in their dashboards).

but really what this potentially breaks are dashboards using the old buckets for apdex scores. a communication on the release is better than doing all these hassle.

prometheusbp/metrics_buckets_spec.go
```diff
@@ -0,0 +1,28 @@
...
+
+package prometheusbp
+
+// DefaultBuckets is the default bucket values for a prometheus histogram metric.
+// These match the spec defined at https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#latency-histograms
+var DefaultBuckets = []float64{
```
fishy, line 8: we also realized that `DefaultBuckets` is a bad name, because this really is just `DefaultLatencyBuckets`, and should not be used on any histogram that's not latency.

I would suggest define the new buckets as `DefaultLatencyBuckets` instead, and make `DefaultBuckets` just an alias to it but also mark it as deprecated:
```go
// Deprecated: Please use DefaultLatencyBuckets instead for latencies, and define your own buckets for non-latency histograms
var DefaultBuckets = DefaultLatencyBuckets
```

fishy (changes requested):

prometheusbp/metrics_buckets_old.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// DefaultBuckets is the default bucket values for a prometheus histogram metric.
+var DefaultBuckets = prometheus.ExponentialBuckets(0.0001, 2.5, 14) // 100us ~ 14.9s
+
+// KafkaBuckets is the set of bucket values for Kafka-related prometheus histogram metrics.
+var KafkaBuckets = prometheus.ExponentialBucketsRange(1e-4, 10, 10) // 100us - 10s
```
konradreiche, line 14:
```suggestion
var (
	// DefaultBuckets is the default bucket values for a prometheus histogram metric.
	DefaultBuckets = prometheus.ExponentialBuckets(0.0001, 2.5, 14) // 100us ~ 14.9s
	// KafkaBuckets is the set of bucket values for Kafka-related prometheus histogram metrics.
	KafkaBuckets = prometheus.ExponentialBucketsRange(1e-4, 10, 10) // 100us - 10s
)
```
If I am not mistaken we follow the group similar values style guideline here: https://github.com/uber-go/guide/blob/master/style.md#group-similar-declarations

konradreiche (changes requested):
Just a style nit :ok_hand:

prometheusbp/metrics_buckets_spec.go
kylelemons: I'm on the fence here.  This is definitely a behavior change and it could break people, and so I'm somewhat inclined to not make it the default, but I could easily be convinced that we should do so, and just try to communicate to folks what is happening so they can set the build tag when they update if they know their dashboards will care.

I think the build tag is the right approach for allowing either opt-in or opt-out, whichever we go with.

prometheusbp/metrics_buckets_spec.go
kylelemons: This sounds like an improvement to me, though without going through all of the uses of DefaultBuckets I'm not sure it needs to be in the same PR.

prometheusbp/metrics_buckets_spec.go
```diff
@@ -0,0 +1,28 @@
...
+	0.005_000,  // 5ms
+	0.010_000,  // 10ms
+	0.025_000,  // 25ms
+	0.050_000,  // 50ms
+	0.100_000,  // 100ms
+	0.250_000,  // 250ms
+	0.500_000,  // 500ms
+	1.000_000,  // 1s
+	5.000_000,  // 5s
+	15.000_000, // 15s (fastly timeout)
```
kylelemons, line 22: We should remove the reference to fastly from the open source version imo.

prometheusbp/metrics_buckets_spec.go
```diff
@@ -0,0 +1,28 @@
+//go:build metrics_spec
+// +build metrics_spec
+
+package prometheusbp
+
+// DefaultBuckets is the default bucket values for a prometheus histogram metric.
```
kylelemons, line 6: (nit) Use paragraph breaks

```suggestion
// DefaultBuckets is the default bucket values for a prometheus histogram metric.
//
```

kylelemons (approved):

prometheusbp/metrics_buckets_old.go
```diff
@@ -0,0 +1,14 @@
+//go:build !metrics_spec
```
kylelemons, line 1: actually we probably want to be extra explicit with the build tags, and the Go ones are all one word things like netgo, so maybe:

```suggestion
//go:build oldprombuckets
```

prometheusbp/metrics_buckets_spec.go
ghirsch-reddit: Hmm, I suppose that if the default is the new behavior and users have to explicitly opt out, then they have to do an explicit action either way (add the build tag or update their dashboards).

So I think my instinct is opt-in > no build tag > opt-out

prometheusbp/metrics_buckets_spec.go
ghirsch-reddit: Makes sense, I'll update that. I won't update usage across everything, but I'll at least add a commit to this PR with a find/replace.

prometheusbp/metrics_buckets_spec.go
ghirsch-reddit: Not sure if there's a preferred style between this (with more breaks) and Konrad's comment https://github.com/reddit/baseplate.go/pull/567#discussion_r976845341

prometheusbp/metrics_buckets_spec.go
konradreiche: Build tags come with a variety of issues, it's not only updating your build process but also local development, tooling having issues to detect it for linting, etc.

prometheusbp/metrics_buckets_spec.go
konradreiche: I think you can do both :slightly_smiling_face: I might have squashed my suggestion too much.

konradreiche (approved):

prometheusbp/metrics_buckets_old.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	// DefaultLatencyBuckets is the default bucket values for a prometheus histogram metric.
+	DefaultLatencyBuckets = prometheus.ExponentialBuckets(0.0001, 2.5, 14) // 100us ~ 14.9s
+	// KafkaLatencyBuckets is the set of bucket values for Kafka-related prometheus histogram metrics.
+	KafkaLatencyBuckets = prometheus.ExponentialBucketsRange(1e-4, 10, 10) // 100us - 10s
+	// Deprecated: Please use DefaultLatencyBuckets instead for latencies, and define your own buckets for non-latency histograms
+	DefaultBuckets = DefaultLatencyBuckets
```
fishy: this is the same between the 2 cases from build tags so it can be moved back to `metrics.go`.

prometheusbp/metrics_buckets_spec.go
```diff
@@ -0,0 +1,31 @@
+//go:build newprombuckets
+// +build newprombuckets
+
+package prometheusbp
+
+var (
+	// DefaultLatencyBuckets is the default bucket values for a prometheus histogram metric.
```
fishy:
```suggestion
	// DefaultLatencyBuckets is the default bucket values for a prometheus latency histogram metric.
```

same for the other file.

fishy (changes requested):
I would still prefer no build tag > opt-out > opt-in.

really services using apdex score dashboards with those auto-generated-from-exponential buckets are rare cases. with opt-in the majority of services will not be using the new buckets at all because they don't know they need to opt-in.

also opt-in means the documentation rendered on pkg.go.dev will be the old one, which would be misleading.

prometheusbp/metrics.go
```diff
@@ -1,12 +1,32 @@
...
+		0.250_000,  // 250ms
+		0.500_000,  // 500ms
+		1.000_000,  // 1s
+		5.000_000,  // 5s
+		15.000_000, // 15s
+		30.000_000, // 30s
+	}
+	// KafkaLatencyBuckets is the set of bucket values for Kafka-related prometheus histogram metrics.
+	// These match the spec defined at https://github.snooguts.net/reddit/baseplate.spec/blob/master/component-apis/prom-metrics.md#latency-histograms
+	KafkaLatencyBuckets = DefaultLatencyBuckets
```
mterwill, line 25: Why export this as its own thing?

ghirsch-reddit: Oops, you're right. Holdover from the previous version that was split up (see the difference here https://github.com/reddit/baseplate.go/pull/567/commits/c32094adc176978980e65384418122bfd92ebc28)

prometheusbp/metrics.go
```diff
@@ -1,11 +1,30 @@
...
+		0.500_000,  // 500ms
+		1.000_000,  // 1s
+		5.000_000,  // 5s
+		15.000_000, // 15s
+		30.000_000, // 30s
+	}
 
-// DefaultBuckets is the default bucket values for a prometheus histogram metric.
-var DefaultBuckets = prometheus.ExponentialBuckets(0.0001, 2.5, 14) // 100us ~ 14.9s
+	// Deprecated: Please use DefaultLatencyBuckets instead for latencies, and define your own buckets for non-latency histograms
```
fishy, line 25: nit: I would tweak the wording slightly here:
```suggestion
	// Deprecated: Please use DefaultLatencyBuckets instead for latencies if the buckets suit your needs, or define your own buckets.
```

fishy (approved):

PR #568, opened 2022-09-24
Allow most numeric types for experiment targeting

konradreiche:
## 💸 TL;DR

The experiment targeting implementation is only working for the type `int` and `float64`. For inputs of type `int64`, `float32`, `uint`, etc. the node automatically evaluates to false. Instead, allow for most of Go's numeric types.

## 📜 Details

We had a bug where experiment targeting failed silently because an app build number was passed in as an `int64`. This change might be excessive but I also don't necessarily see why this should be applicable to a wider range of types. Alternatively, we could fire a log event for any unmapped type. 

See tests added.

## ✅ Checks

### DIFF
--- a/experiments/targeting.go
+++ b/experiments/targeting.go
@@ -189,7 +189,8 @@ func NewEqualNode(inputNodes map[string]interface{}) (Targeting, error) {
 func (n *EqualNode) Evaluate(inputs map[string]interface{}) bool {
 	candidateValue := inputs[n.fieldName]
 	switch n := candidateValue.(type) {
-	case int, float64:
+	case int, float64, int8, int16, int32, int64, uint, uint8, uint16, uint32,
+		uint64, float32:
 		candidateValue = json.Number(fmt.Sprintf("%v", n))
 	}
 	for _, value := range n.acceptedValues {
@@ -291,7 +292,27 @@ func (n *ComparisonNode) Evaluate(inputs map[string]interface{}) bool {
 	case int:
 		return n.comparer(float64(cv), value)
 	case float64:
-		n.comparer(cv, value)
+		return n.comparer(cv, value)
+	case int8:
+		return n.comparer(float64(cv), value)
+	case int16:
+		return n.comparer(float64(cv), value)
+	case int32:
+		return n.comparer(float64(cv), value)
+	case int64:
+		return n.comparer(float64(cv), value)
+	case uint:
+		return n.comparer(float64(cv), value)
+	case uint8:
+		return n.comparer(float64(cv), value)
+	case uint16:
+		return n.comparer(float64(cv), value)
+	case uint32:
+		return n.comparer(float64(cv), value)
+	case uint64:
+		return n.comparer(float64(cv), value)
+	case float32:
+		return n.comparer(float64(cv), value)
 	}
 	return false
 }
--- a/experiments/targeting_test.go
+++ b/experiments/targeting_test.go
@@ -612,6 +612,156 @@ func TestComparisonNodeBadInput(t *testing.T) {
 	}
 }
 
+func TestNumberTypes(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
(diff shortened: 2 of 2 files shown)


experiments/targeting.go
```diff
@@ -291,7 +292,27 @@ func (n *ComparisonNode) Evaluate(inputs map[string]interface{}) bool {
 	case int:
 		return n.comparer(float64(cv), value)
 	case float64:
-		n.comparer(cv, value)
+		return n.comparer(cv, value)
```
konradreiche, line 295: Looks like this was broken for `float64` too since there was no `return` statement.

experiments/targeting.go
```diff
@@ -189,7 +189,8 @@ func NewEqualNode(inputNodes map[string]interface{}) (Targeting, error) {
 func (n *EqualNode) Evaluate(inputs map[string]interface{}) bool {
 	candidateValue := inputs[n.fieldName]
 	switch n := candidateValue.(type) {
-	case int, float64:
+	case int, float64, int8, int16, int32, int64, uint, uint8, uint16, uint32,
+		uint64, float32:
```
pacejackson, line 193:
```suggestion
	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64,
		float32, float64:
```

a little tidier

pacejackson: Alternatively, would this work?

```suggestion
	case constraints.Integer, constraints.Float:
```

https://pkg.go.dev/golang.org/x/exp/constraints

experiments/targeting.go
pacejackson: Can you reorganize this so `float64` is after `float32` so it flows up naturally?

pacejackson (changes requested):
Just some style suggestions

experiments/targeting.go
fishy: `constraints` probably won't work but I'm not 100% sure.

if not using `constraints`, please add an extra blank line after this line: https://github.com/reddit/baseplate.go/blob/master/Style.md#blank-line-to-highlight-change-of-logical-level

fishy (approved):
agree with andrew's suggestion on ordering. otherwise LGTM.

experiments/targeting_test.go
```diff
@@ -612,6 +612,156 @@ func TestComparisonNodeBadInput(t *testing.T) {
...
+func TestNumberTypes(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name      string
+		targeting []byte
+		input     any
+	}{
+		{
+			name:      "gt-node-int",
```
ghirsch-reddit, line 624: 🔕 same thing about ordering here. Also it seems like you could probably simplify the test to have just an input of the following, and then just construct the two `[]byte` below since they're always the same.
```
{
  name: "node-***",
  input: ***(5),
}
```

ghirsch-reddit (approved):

experiments/targeting.go
kylelemons, line 193: If you're going to do linebreaks, do it to group things together logically, not based on length:

```suggestion
	case
		float32, float64,
		int, int8, int16, int32, int64,
		uint, uint8, uint16, uint32, uint64:
```

experiments/targeting_test.go
```diff
@@ -612,6 +612,156 @@ func TestComparisonNodeBadInput(t *testing.T) {
 	}
 }
 
+func TestNumberTypes(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name      string
+		targeting []byte
```
kylelemons, line 620: make this a string so you don't need `[]byte(` everywhere

experiments/targeting_test.go
kylelemons: I actually prefer having both the input and the output contained within each test case.

kylelemons (approved):

PR #570, opened 2022-09-27
Upgrade thrift to 0.17.0

fishy:
This allows us to remove tDuplicateToProtocol and getClientError implementation as they are in thrift v0.17.0 now.

There's no breaking changes between 0.16.0 and 0.17.0 regarding go library/compiler.


konradreiche (approved):

thriftbp/client_middlewares.go
```diff
@@ -156,6 +157,7 @@ func BaseplateDefaultClientMiddlewares(args DefaultClientMiddlewareArgs) []thrif
 		}),
 		PrometheusClientMiddleware(args.ServiceSlug),
 		BaseplateErrorWrapper,
+		thrift.ExtractIDLExceptionClientMiddleware,
```
kylelemons, line 160: This doesn't really document how it works.  It shows a bit of generated code, but not how it's an error.

https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#ExtractExceptionFromResult

Can you just type assert the `error` to the exception type now?  I don't see `Error() string` in the generated code but github is kinda fighting me onit.

kylelemons (approved):

thriftbp/client_middlewares.go
fishy: I think you are looking at the wrong doc :) This is the doc of this middleware: https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#ExtractIDLExceptionClientMiddleware

Basically, for client middlewares, the function signature is defined by `TClient.Call`: https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#TClient.Call

Withouth this middleware, when the server throws an exception defined in the IDL (for example `baseplate.Error`), the `err` return of `Call` would be nil, and the exception is part of the `result`.

This middleware basically extracts exceptions out of `result` and put them into `err` returns, so for all other middlewares before them (here they are `BaseplateErrorWrapper`. `PrometheusClientMiddleware`. etc), they would get the non-nil `err` directly without the need of `ExtractExceptionFromResult` (but I still kept `ExtractExceptionFromResult` call via `getClientError` on some of the middlewares that absolutely need it because there's no easy guarantee that the user always construct the middleware chain in the correct order. when they are in the correct order the `getClientError` would return quickly because `err` is already non-nil).

And yes with the correct order other middlewares can simply type assert the `error` if that's what you are asking.

PR #572, opened 2022-10-04
Update gRPC metrics to match v2 and spec

kylelemons:
To use v0 and v2 in the same binary, they require identical labels.  The v2 client currently is set up to match the spec, and so update the v0 middleware to do the same for legacy users of this middleware.

### DIFF
--- a/grpcbp/prometheus.go
+++ b/grpcbp/prometheus.go
@@ -55,6 +55,7 @@ var (
 	serverActiveRequestsLabels = []string{
 		serviceLabel,
 		methodLabel,
+		typeLabel,
 	}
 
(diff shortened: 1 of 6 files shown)


konradreiche (approved):

fishy (approved):

grpcbp/prometheus.go
```diff
@@ -55,6 +55,7 @@ var (
 	serverActiveRequestsLabels = []string{
 		serviceLabel,
 		methodLabel,
+		typeLabel,
```
mterwill, line 58: FWIW I don't actually see the `type` label in the spec for `grpc_server_active_requests` and `grpc_client_active_requests`, we should add it? (This does appear to match v2 though)

mterwill (approved):

PR #573, opened 2022-10-24
[fix] `httpbp` middleware doesn't flush chunked responses

adamthesax:
Closes # 

## 💸 TL;DR

`httpbp` middleware, in particular `RecordStatusCode`  and `PrometheusServerMetrics` don't work with chunked HTTP responses because they wrap `http.ResponseWriter` and do not implement `http.Flush`. This updates our response wrapping within in the Middleware to support both the `http.Flush` and `http.Hijacker` interface.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/httpbp/middlewares.go
+++ b/httpbp/middlewares.go
@@ -537,9 +537,9 @@ func allowFlushHijack(original, rw http.ResponseWriter) http.ResponseWriter {
 	case isFlusher && isHijacker:
 		return &wrappedFlushHijacker{rw, flusher, hijacker}
 	case isFlusher:
-		return &wrappedFlusher{rw, flusher}
+		return allowFlush(original, rw)
 	case isHijacker:
-		return &wrappedHijacker{rw, hijacker}
+		return allowHijack(original, rw)
 	default:
 		return rw
 	}
--- a/httpbp/middlewares.go
+++ b/httpbp/middlewares.go
@@ -394,9 +394,10 @@ func recordStatusCode(counters counterGenerator) Middleware {
 	return func(name string, next HandlerFunc) HandlerFunc {
 		counter := counters.Counter("baseplate.http." + name + ".response")
 		return func(ctx context.Context, w http.ResponseWriter, r *http.Request) (err error) {
-			wrapped := &statusCodeRecorder{ResponseWriter: w}
+			rec := &statusCodeRecorder{ResponseWriter: w}
+			wrapped := allowFlushHijack(w, rec)
 			defer func() {
-				code := wrapped.getCode(err)
+				code := rec.getCode(err)
 				counter.With("status", statusCodeFamily(code)).Add(1)
 			}()
 
@@ -453,9 +454,11 @@ func PrometheusServerMetrics(_ string) Middleware {
 			}
 			serverActiveRequests.With(activeRequestLabels).Inc()
 
-			wrapped := &responseRecorder{ResponseWriter: w}
+			rec := &responseRecorder{ResponseWriter: w}
+			wrapped := allowFlushHijack(w, rec)
+
 			defer func() {
-				code := errorCodeForMetrics(wrapped.responseCode, err)
+				code := errorCodeForMetrics(rec.responseCode, err)
 				success := isRequestSuccessful(code, err)
 
 				labels := prometheus.Labels{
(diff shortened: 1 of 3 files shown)


httpbp/middlewares.go
```diff
@@ -510,3 +513,50 @@ func (rr *responseRecorder) WriteHeader(code int) {
 	rr.ResponseWriter.WriteHeader(code)
 	rr.responseCode = code
 }
+
+// wrappers
```
kylelemons, line 517:
```suggestion
```

kylelemons (changes requested):
Thanks!

I think the only thing I would request is to make a unit test that makes a dummy server and makes sure that the handler can be hijacked and (separately) flushed.

pacejackson (approved):

httpbp/middlewares.go
```diff
@@ -510,3 +513,49 @@ func (rr *responseRecorder) WriteHeader(code int) {
...
+		return &wrappedFlushHijacker{rw, flusher, hijacker}
+	case isFlusher:
+		return allowFlush(original, rw)
+	case isHijacker:
+		return allowHijack(original, rw)
+	default:
+		return rw
+	}
+
+}
```
fishy, line 547: 🔕 it's manage-able as we currently have 2 optional interfaces, it will become a nightmare if there's a 3rd one added in the future. but I also can't find a better way to do this.

fishy (approved):

httpbp/middlewares_test.go
```diff
@@ -323,3 +326,88 @@ func TestSupportedMethods(t *testing.T) {
...
+
+	t.Run("hijackable-response", func(tt *testing.T) {
+		r := httptest.NewRequest(http.MethodGet, "/test", nil)
+		w := &hijackableResponseRecorder{httptest.NewRecorder(), false}
+		args.EndpointRegistry.ServeHTTP(w, r)
+
+		if !w.Hijacked {
+			t.Error("expected http response to be hijacked")
+		}
+	})
```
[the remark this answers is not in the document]
adamthesax: In [5ad3f18](https://github.com/reddit/baseplate.go/pull/573/commits/5ad3f180281e2a85770a8145bb35ac2b0f1a8d2b) I added test cases that should cover:
* non-Flusher/non-Hijacker
* Flusher
* Hijacker
* Hijacker and Flusher

Let me know if there are any other cases you can think of!

httpbp/middlewares.go
adamthesax: Yeah this approach is definitely verbose but I can't think of a better one. The only other interface on `http` that a `http.ResponseWriter` could implement would be [`http.Pusher`](https://pkg.go.dev/net/http#Pusher). I don't think it's worth supporting it now as I don't think we have any use case driving it, but that would be a 3rd interface that may come up in the future.

fishy (approved):

httpbp/middlewares.go
kylelemons: This is a quietly known issue amongst the ecosystem, and the approach in the generic case seems to be code generation.

https://www.doxsey.net/blog/fixing-interface-erasure-in-go/

kylelemons (approved):
I personally prefer the previous approach.  It's true that Pusher is missed, but it has the benefit that the type names wouldn't be atrocious :)

That said, I have no issue with keeping this.  The approach is sound, and it's more complete.

httpbp/response_wrappers.go
```diff
@@ -0,0 +1,33 @@
+// DO NOT EDIT.
```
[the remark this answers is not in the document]
fishy: +1.

I would say it's better to just hand write this code:
```go
type optionalResponseWriter int

const (
  _ optionalResponseWriter 1 << iota
  flusher
  hijacker
  pusher
)

func wrapResponseWriter(...) ... {
  var f optionalResponseWriter
  ...

  switch f {
  case 0:
    ...
  case flusher:
    ...
  case flusher | hijacker:
    ...
  ...
}
```

fishy (changes requested):
I don't have a strong preference between this and the previous approach, but if we pick this approach, I'd prefer to hand writing the code over using codegen.

httpbp/response_wrappers.go
```diff
@@ -0,0 +1,76 @@
...
+type optionalResponseWriter uint64
+
+const (
+	flusher optionalResponseWriter = 1 << iota
+	hijacker
+	pusher
+)
+
+func wrapResponseWriter(orig, wrapped http.ResponseWriter) http.ResponseWriter {
+	var f optionalResponseWriter
```
fishy, line 16: nit: maybe consider using a different name for `f` so you don't have to make flusher one `fl` that's inconsistent with the others :)

fishy (approved):

fishy (approved):

PR #576, opened 2022-10-27
httpbp: Set http.Server.ErrorLog

fishy:
It's used by upstream server to log some of the kinda-breaking issues,
for example [1]. Since the majority of the http servers at Reddit are
public facing, when those happens it's usually just user messing with
us, not really that the http client we control doing things wrong. This
gives us a way to suppress those logs, and also emit counters to better
track how many times those happened.

This also makes the upstream http server to use the same json logger by
zap as the rest of our code, at warning level.

[1]: https://github.com/golang/go/issues/25192#issuecomment-992276264

~~(I still need to write tests for it, but send it out first to get some early eyes/feedback)~~ done

### DIFF
--- a/httpbp/error_logger.go
+++ b/httpbp/error_logger.go
@@ -1,0 +1,56 @@
+package httpbp
+
+import (
+	"log"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	"github.com/reddit/baseplate.go/internal/prometheusbpint"
+)
+
+var httpServerLoggingCounter = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
+	Name: "httpbp_server_upstream_issue_logs_total",
+	Help: "Number of logs emitted by stdlib http server regarding an upstream issue",
+}, []string{"upstream_issue"})
+
+// This is a special zapcore used by stdlib http server to handle error logging.
+type wrappedCore struct {
+	zapcore.Core
+
+	counter25192  prometheus.Counter
+	suppress25192 bool
+}
+
+func (w wrappedCore) Check(entry zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
+	// Example message:
+	//     URL query contains semicolon, which is no longer a supported separator; parts of the query may be stripped when parsed; see golang.org/issue/25192
+	if strings.Contains(entry.Message, "golang.org/issue/25192") {
+		w.counter25192.Inc()
+		if w.suppress25192 {
+			// drop the log
+			return ce
+		}
+	}
+
+	return w.Core.Check(entry, ce)
+}
+
+func httpServerLogger(base *zap.Logger, suppress25192 bool) (*log.Logger, error) {
+	return zap.NewStdLogAt(base.WithOptions(
+		zap.Fields(zap.String("from", "http-server")),
+		zap.WrapCore(func(c zapcore.Core) zapcore.Core {
+			return wrappedCore{
+				Core: c,
+
+				suppress25192: suppress25192,
+				counter25192: httpServerLoggingCounter.With(prometheus.Labels{
+					"upstream_issue": "25192",
+				}),
+			}
+		}),
+	), zapcore.WarnLevel)
+}
--- a/httpbp/error_logger_test.go
+++ b/httpbp/error_logger_test.go
@@ -1,0 +1,75 @@
+package httpbp
+
+import (
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	"github.com/reddit/baseplate.go/prometheusbp/promtest"
+)
+
+func initBaseLogger(w io.Writer) *zap.Logger {
+	// Mostly copied from zap.NewExample, to make the log deterministic.
+	encoderCfg := zapcore.EncoderConfig{
+		MessageKey:     "msg",
+		LevelKey:       "level",
+		NameKey:        "logger",
+		EncodeLevel:    zapcore.LowercaseLevelEncoder,
+		EncodeTime:     zapcore.ISO8601TimeEncoder,
+		EncodeDuration: zapcore.StringDurationEncoder,
+	}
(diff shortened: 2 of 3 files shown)


httpbp/server.go
```diff
@@ -221,9 +235,21 @@ func NewBaseplateServer(args ServerArgs) (baseplate.Server, error) {
 		return nil, err
 	}
 
+	logger, err := httpServerLogger(args.UpstreamIssueLoggingSampleRates)
+	if err != nil {
+		// Should not happen, but if it really happens, we just fallback to stdlib
+		// logger, which is not that big a deal either.
```
ghirsch-reddit, line 241: Is this because we treat a nil logger as the stdlib logger somewhere? Where does that happen?

httpbp/error_logger.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/reddit/baseplate.go/randbp"
+)
+
+var httpServerLoggingCounter = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
+	Name: "httpbp_server_upstream_issue_logs_total",
+	Help: "Number of logs emitted by stdlib http server regarding an upstream issue",
+}, []string{"upstream_issue"})
+
+type issueConfig struct {
+	rate    float64
```
ghirsch-reddit, line 25:
```suggestion
	sampleRate    float64
```

httpbp/error_logger.go
```diff
@@ -0,0 +1,80 @@
...
+	counter prometheus.Counter
+}
+
+// This is a special zapcore used by stdlib http server to handle error logging.
+type wrappedCore struct {
+	zapcore.Core
+
+	issues map[int64]issueConfig
```
ghirsch-reddit, line 33: Do we need either this to be `map[int64]*issueConfig` or the counter in issueConfig to be a `*prometheus.Counter`? Or is it fine to increment using copies?

ghirsch-reddit (changes requested):
RC'ing to appease Harold. I would love to see an example usage as well, if that's not covered by the tests you add

httpbp/server.go
fishy: https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/net/http/server.go;l=3226-3232

it's also documented on [`http.Server.ErrorLog`](https://pkg.go.dev/net/http#Server.ErrorLog)

httpbp/error_logger.go
fishy: `prometheus.Counter` is an interface. it's almost never a good idea to get the pointer to an interface.

fishy:
>I would love to see an example usage as well

There's not much to example. Users can supply the map config on how much they want to suppress the logs. If they don't change any code beyond updating baseplate.go, we fallback to 100% sample, so they get the free counter and the log will be in proper json format (instead of using stdlib logger).

httpbp/error_logger_test.go
```diff
@@ -0,0 +1,92 @@
...
+		logger.Printf("Hello, golang.org/issue/25192!")
+		if got, want := sb.String(), `{"level":"warn","msg":"Hello, golang.org/issue/25192!","from":"http-server"}`; strings.TrimSpace(got) != want {
+			t.Errorf("Got %q, want %q", got, want)
+		}
+
+		sb.Reset()
+		logger.Printf("Hello, world!")
+		if got, want := sb.String(), `{"level":"warn","msg":"Hello, world!","from":"http-server"}`; strings.TrimSpace(got) != want {
+			t.Errorf("Got %q, want %q", got, want)
+		}
```
ghirsch-reddit, line 90: 🔕 just a nit that you do this differently on line 63

ghirsch-reddit (approved):

httpbp/error_logger_test.go
fishy: that's intentional. it's a long string so I actually prefer to have it in its own line as a `const`. but in this case I need it twice so a `const` no longer works.

httpbp/error_logger.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (w wrappedCore) Check(entry zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
+	// Example message:
+	//     URL query contains semicolon, which is no longer a supported separator; parts of the query may be stripped when parsed; see golang.org/issue/25192
+	if strings.Contains(entry.Message, "golang.org/issue/25192") {
+		cfg := w.issues[25192]
```
kylelemons, line 42: I don't think the map is really helping you here.  Since the error message is hard-coded and we are hard coding the supported IDs and so on, I would make a field for each one and just do
```
w.issue25192.Inc()
```

etc.

httpbp/error_logger_test.go
```diff
@@ -0,0 +1,92 @@
...
+		logger.Printf("Hello, golang.org/issue/25192!")
+		logger.Printf("Hello, golang.org/issue/25192!")
+		if str := sb.String(); strings.TrimSpace(str) != "" {
+			t.Errorf("Expected logs being suppressed, got %q", str)
+		}
+
+		sb.Reset()
+		logger.Printf("Hello, world!")
+		const want = `{"level":"warn","msg":"Hello, world!","from":"http-server"}`
+		if got := sb.String(); strings.TrimSpace(got) != want {
```
kylelemons, line 64: 🔕 cmp.Diff works well for strings like this, it calls out the part that changed instead of making a human squint at two almost identical lines

httpbp/server.go
```diff
@@ -157,6 +157,21 @@ type ServerArgs struct {
...
+	// emitted by stdlib http.Server regarding an upstream issue.
+	//
+	// The key is the upstream issue id.
+	// Currently the only supported one is 25192 [1],
+	// but the supported issue ids might expand in the future.
+	//
+	// When a supported issue id is not in the map,
+	// currently the default sample rate is 1
+	// (100% sampled, e.g. not dropping those logs).
+	// This might be changed to 0 in the future.
```
kylelemons, line 171: 🔕 I'm in kindof a YAGNI mode here... this is a lot of complexity to carry around for something we might not need.

kylelemons (changes requested):

httpbp/error_logger_test.go
fishy: the problem with `cmp.Diff` is that you can't declare and assign all of got, want, diff in the same `if` statement :)

kylelemons: Yeah, that is definitely a bummer.  I do
```
if diff := cmp.Diff(got, want); diff != "" {
  t.Errorf("<context> (-got +want)\n%s", diff)
}
```

which 🤷

kylelemons (approved):

konradreiche (approved):

PR #577, opened 2022-10-28
thriftbp: Match thrift client pool metrics to baseplate spec

fishy:
Temporarily emit both new and old metrics, we will remove the old metrics after the next release (so the next release will have both).

While I'm here, also:

1. Remove statsd metrics from thrift client pool
2. Change redis client pool metric name following spec change

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -6,11 +6,11 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strconv"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/avast/retry-go"
-	"github.com/go-kit/kit/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/reddit/baseplate.go/breakerbp"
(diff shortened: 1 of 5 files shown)


thriftbp/client_pool.go
```diff
@@ -611,13 +601,19 @@ func (p *clientPool) pooledCall(ctx context.Context, method string, args, result
 	return client.Call(ctx, method, args, result)
 }
 
-func (p *clientPool) getClient() (Client, error) {
+func (p *clientPool) getClient() (_ Client, err error) {
+	defer func() {
+		clientPoolGetsCounter.With(prometheus.Labels{
+			"thrift_pool":    p.slug,
+			"thrift_success": strconv.FormatBool(err == nil),
```
fishy, line 608: note: the spec only defined `thrift_pool` label, but I think `thrift_success` label here would be really useful.

kylelemons (approved):

thriftbp/client_pool.go
kylelemons: What does "success" mean here though, succesfully getting a client from the pool?

fishy: Yes. do you have a better name to suggest?

thriftbp/prometheus.go
```diff
@@ -238,25 +275,52 @@ var (
...
+	idle := float64(e.pool.NumAllocated())
+
 	// MustNewConstMetric would only panic if there's a label mismatch, which we
 	// have a unit test to cover.
+	ch <- prometheus.MustNewConstMetric(
+		legacyClientPoolActiveConnectionsDesc,
+		prometheus.GaugeValue,
+		float64(active),
+		e.slug,
+		e.slug,
```
pacejackson, line 298: why is e.slug here twice? same below.

pacejackson (approved):

thriftbp/prometheus.go
fishy: we are emitting both the old `thrift_client_name` and new `thrift_pool` labels.

PR #578, opened 2022-11-01
Export the versions of Go modules linked into the binary

kylelemons:
## 💸 TL;DR
Backport of module metrics from v2.0.0

## ✅ Checks

- [ ] ~Contributor License Agreement (CLA) completed if not a Reddit employee~

### DIFF
--- a/baseplate.go
+++ b/baseplate.go
@@ -11,7 +11,6 @@ import (
 	"github.com/reddit/baseplate.go/batchcloser"
 	"github.com/reddit/baseplate.go/configbp"
 	"github.com/reddit/baseplate.go/ecinterface"
-	"github.com/reddit/baseplate.go/internal/prometheusbpint"
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/metricsbp"
 	"github.com/reddit/baseplate.go/runtimebp"
@@ -310,7 +309,7 @@ func New(ctx context.Context, args NewArgs) (context.Context, Baseplate, error) 
 	bp := impl{cfg: cfg, closers: batchcloser.New()}
 
 	if info, ok := debug.ReadBuildInfo(); ok {
-		prometheusbpint.RecordModuleVersions(info)
+		metricsbp.RecordModuleVersions(info)
 	} else {
 		log.C(ctx).Warn("baseplate.New: unable to read build info to export dependency metrics")
 	}
--- a/baseplate.go
+++ b/baseplate.go
@@ -13,6 +13,7 @@ import (
 	"github.com/reddit/baseplate.go/ecinterface"
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/metricsbp"
+	"github.com/reddit/baseplate.go/prometheusbp"
 	"github.com/reddit/baseplate.go/runtimebp"
 	"github.com/reddit/baseplate.go/secrets"
 	"github.com/reddit/baseplate.go/tracing"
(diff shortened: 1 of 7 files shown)


internal/prometheusbpint/modules_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+func TestBuildInfoMetrics(t *testing.T) {
+	info := &debug.BuildInfo{
+		GoVersion: "go1.18.3",
+		Path:      "example.com/path/to/main/module",
+		Main: debug.Module{
+			Path:    "example.com/path/to/main/module",
+			Version: "(devel)",
+		},
+		Deps: []*debug.Module{{
```
fishy, line 20: nit: consider adding a case of `replaced=true`?

internal/prometheusbpint/modules.go
```diff
@@ -0,0 +1,31 @@
...
+	Name: "baseplate_go_modules",
+	Help: "Export the version information for included Go modules, and whether the module is the 'main' module or a 'dependency'.  Always 1",
+}, []string{"go_module", "module_role", "replaced", "module_version"})
+
+func RecordModuleVersions(info *debug.BuildInfo) {
+	record := func(role string, mod *debug.Module) {
+		goModules.With(prometheus.Labels{
+			"go_module":      mod.Path,
+			"module_role":    role,
+			"replaced":       strconv.FormatBool(mod.Replace != nil),
```
fishy, line 21: 🔕 I think overall we want to avoid using single word labels, as the whole label indices are global so two unrelated metrics emitting with the same label (e.g. conflicted labels) will cause searching by label slower. I'd suggest use `module_replaced` instead.

fishy (approved):

internal/prometheusbpint/modules.go
kylelemons: hmm, makes it a not-strict backport but I can make the change to v2 as well.

internal/prometheusbpint/modules.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var goModules = promauto.With(GlobalRegistry).NewGaugeVec(prometheus.GaugeOpts{
+	Name: "baseplate_go_modules",
+	Help: "Export the version information for included Go modules, and whether the module is the 'main' module or a 'dependency'.  Always 1",
+}, []string{"go_module", "module_role", "module_replaced", "module_version"})
+
+func RecordModuleVersions(info *debug.BuildInfo) {
+	record := func(role string, mod *debug.Module) {
```
ghirsch-reddit, line 17: Out of curiosity, what's the reason for defining this inside rather than at the top-level?

ghirsch-reddit (approved):

internal/prometheusbpint/modules.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var goModules = promauto.With(GlobalRegistry).NewGaugeVec(prometheus.GaugeOpts{
+	Name: "baseplate_go_modules",
+	Help: "Export the version information for included Go modules, and whether the module is the 'main' module or a 'dependency'.  Always 1",
+}, []string{"go_module", "module_role", "module_replaced", "module_version"})
+
+func RecordModuleVersions(info *debug.BuildInfo) {
```
fishy, line 16: 🔕 also, do we want to consider moving this to `prometheusbp` so some more customized services (that don't use `baseplate.New`) can also use it?

konradreiche: +1 I think it would be useful to not lock this behind `internal`

konradreiche (changes requested):

baseplate.go
```diff
@@ -310,7 +309,7 @@ func New(ctx context.Context, args NewArgs) (context.Context, Baseplate, error)
 	bp := impl{cfg: cfg, closers: batchcloser.New()}
 
 	if info, ok := debug.ReadBuildInfo(); ok {
-		prometheusbpint.RecordModuleVersions(info)
+		metricsbp.RecordModuleVersions(info)
```
fishy, line 312: you want `prometheusbp` instead of `metricsbp` 😅

konradreiche (approved):

fishy (approved):

PR #581, opened 2022-11-01
thriftbp: Metrics cleanup

fishy:
Remove statsd metrics, and rename some prometheus metrics to comply with prometheus naming convention. For renamed prometheus metrics, temporarily emitting both old and new, but mark old ones as deprecated to be removed later.

Also remove some statsd only features:

1. CountedServerTransport is marked as deprecated and no longer used by other code.
2. In ReportPayloadSizeMetrics server middleware, the rate arg is deprecated and the prometheus metrics is always reported.
3. ClientPoolConfig.MetricsTags & ClientPoolConfig.ReportPoolStats are deprecated. The pool stats are always reported as prometheus metrrics.

### DIFF
--- a/thriftbp/prometheus.go
+++ b/thriftbp/prometheus.go
@@ -167,11 +167,17 @@ var (
 		methodLabel,
 	}
 
-	panicRecoverCounter = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
+	// TODO: Remove after the next release (v0.9.12)
+	legacyPanicRecoverCounter = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
 		Namespace: promNamespace,
 		Subsystem: subsystemServer,
 		Name:      "panic_recover_total",
-		Help:      "The number of panics recovered from thrift server handlers",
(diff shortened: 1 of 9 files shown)


thriftbp/prometheus.go
```diff
@@ -167,11 +167,17 @@ var (
...
+	legacyPanicRecoverCounter = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
 		Namespace: promNamespace,
 		Subsystem: subsystemServer,
 		Name:      "panic_recover_total",
-		Help:      "The number of panics recovered from thrift server handlers",
+		Help:      "Deprecated: Use thriftbp_server_recovered_panics_total instead",
+	}, panicRecoverLabels)
+
+	panicRecoverCounter = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
+		Name: "thriftbp_server_recovered_panics_total",
```
fishy, line 179: technically (according to some dictionaries) "panic" is uncountable as a noun, but I think we pretty much use it as countable noun in go.

ghirsch-reddit (approved):
Probably good to get a ✅ from someone more familiar with the code, but this all looks great to me! Always great to see us ✂️ and clean up old stuff

kylelemons (approved):

PR #583, opened 2022-11-04
maxprocs: Print newline after log

mterwill:
## 💸 TL;DR

Since we're not using a real logger here (artifact of previous implementation, I _think_ because logging isn't initialized when this is called – but if that's not the case we can update to use zap) we weren't logging a newline. This was not caught by tests because the maxprocs library doesn't bake in any testing affordances so we had to stub it out.

## 📜 Details
Before:
```
maxprocs: Updating GOMAXPROCS=8: determined from CPU quotapanic: runtime error: invalid memory address or nil pointer dereference
```

After:
```
maxprocs: Updating GOMAXPROCS=8: determined from CPU quota
panic: runtime error: invalid memory address or nil pointer dereference
```

Will test in a sample service.

## ✅ Checks

### DIFF
--- a/runtimebp/internal/maxprocs/maxprocs.go
+++ b/runtimebp/internal/maxprocs/maxprocs.go
@@ -74,7 +74,7 @@ var (
 		ubermaxprocs.Set(
 			ubermaxprocs.Min(2),
 			ubermaxprocs.Logger(func(s string, i ...interface{}) {
-				fmt.Fprintf(os.Stderr, s, i...) // contains context
+				fmt.Fprintf(os.Stderr, s+"\n", i...) // contains context
 			}),
 		)
 	}
--- a/runtimebp/internal/maxprocs/maxprocs.go
+++ b/runtimebp/internal/maxprocs/maxprocs.go
@@ -74,7 +74,8 @@ var (
 		ubermaxprocs.Set(
 			ubermaxprocs.Min(2),
 			ubermaxprocs.Logger(func(s string, i ...interface{}) {
-				fmt.Fprintf(os.Stderr, s+"\n", i...) // contains context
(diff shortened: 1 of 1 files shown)


runtimebp/internal/maxprocs/maxprocs.go
```diff
@@ -74,7 +74,7 @@ var (
 		ubermaxprocs.Set(
 			ubermaxprocs.Min(2),
 			ubermaxprocs.Logger(func(s string, i ...interface{}) {
-				fmt.Fprintf(os.Stderr, s, i...) // contains context
+				fmt.Fprintf(os.Stderr, s+"\n", i...) // contains context
```
fishy, line 77: 🔕 string concatenation is actually quite inefficient especially when `s` is long (because it needs to make a bigger string and copy `s` over), so just do `fmt.Fprintln(os.Stderr, "")` after this line is probably better.

mterwill, line 77: Another concern of not using a real logging library is that this could get interleaved, but (a) this happens early in setup, so there shouldn't be any other goroutines running and (b) these log messages are small, should always be smaller than PIPE_BUF, so the OS guarantees they'll be atomic.

fishy (approved):
>I think because logging isn't initialized when this is called

yes that's the reason.

runtimebp/internal/maxprocs/maxprocs.go
mterwill: good point!

mterwill: To my comment above I guess this does open up a possibility the output will get interleaved, but I don't think we'll be running any other goroutines at the time this code runs.

pacejackson (approved):

PR #585, opened 2022-11-16
secrets: Fix bug about not wait for secrets.json to become available

fishy:
When we added support for CSI directories, we introduced a bug that we no longer wait for secrets.json to become available, because os.Stat call would fail with file not exist error immediately.

Since in CSI directories case the directory would be guaranteed to be available immediately (unlike the vault case), assume file not exist errors from os.Stat as using file instead of directory, and let filewatcher.New to handle the waiting as before. Also add a test for it.

While I'm here, also remove the usages of os.IsNotExist and os.IsExist follow the docs.

### DIFF
--- a/secrets/store_test.go
+++ b/secrets/store_test.go
@@ -3,6 +3,7 @@ package secrets_test
 import (
 	"context"
 	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


pacejackson (approved):

secrets/store_test.go
```diff
@@ -490,3 +491,27 @@ func TestAddMiddleware(t *testing.T) {
 		)
 	}
 }
+
+func TestNewStoreWaitBeforeAvailable(t *testing.T) {
+	const (
+		writeDelay = time.Second / 5
+		timeout    = time.Second * 5
```
kylelemons, line 498:
```suggestion
		writeDelay = 200*time.Millisecond
		timeout    = 5*time.Second
```

kylelemons (approved):

PR #587, opened 2022-11-30
Drop support for go 1.18

fishy:
Use the new atomic package API introduced in go 1.19:

* Replace atomic.Value with atomic.Pointer
* Replace atomic.Add*/Store*/Load*/etc. with new Int*/Uint* atomic types
* Except the atomic.Value used in filewatcher, as to take full advantage of that we would need to make breaking API changes to filewatcher

### DIFF
--- a/kafkabp/consumer.go
+++ b/kafkabp/consumer.go
@@ -71,13 +71,14 @@ type consumer struct {
 	cfg ConsumerConfig
 	sc  *sarama.Config
 
-	consumer           atomic.Value // sarama.Consumer
-	partitions         atomic.Value // []int32
-	partitionConsumers atomic.Value // []sarama.PartitionConsumer
+	consumer           atomic.Pointer[sarama.Consumer]
+	partitions         atomic.Pointer[[]int32]
+	partitionConsumers atomic.Pointer[[]sarama.PartitionConsumer]
 
-	closed          int64
-	consumeReturned int64
-	offset          int64
+	closed          atomic.Int64
+	consumeReturned atomic.Int64
+
+	offset int64
 
 	wg sync.WaitGroup
 }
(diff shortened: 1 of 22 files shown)


kafkabp/consumer.go
```diff
@@ -124,25 +125,10 @@ func newTopicConsumer(cfg ConsumerConfig, sc *sarama.Config) (Consumer, error) {
...
-	pc, _ := kc.partitionConsumers.Load().([]sarama.PartitionConsumer)
-	return pc
-}
-
 // reset recreates the consumer and assigns partitions.
 func (kc *consumer) reset() error {
-	if c := kc.getConsumer(); c != nil {
-		if err := c.Close(); err != nil {
+	if c := kc.consumer.Load(); c != nil && *c != nil {
+		if err := (*c).Close(); err != nil {
```
kylelemons, line 131: 🔕 The change here definitely feels worse.  `(*c)`s and `&c` both? It looks wrong, at least.

As long as the type safety ensures that people don't mess it up, I'm probably fine with it, but it just doesn't seem great.

tracing/tags.go
```diff
@@ -11,8 +9,19 @@ var alwaysIncludeAllowList = []string{
...
+// really guarantee atomicy because a slice is a container and its content can
+// be modified (e.g. replace individual strings inside). A copy of the slice is
+// required to guarantee atomicy.
+//
+// But here, when storing it we always create a new string (via
+// generateAllowList), and the read function (getAllowList) is unexported and we
+// never mutate the returned result, so skip copying is OK.
+//
+// But if any of those assumptions are broken in the future, we can not storing
+// the atomic.Pointer[[]string] naively, and would need to either make a copy or
```
kylelemons, line 22: 🔕 What you could also do is store an `atomic.Pointer[string]` and do a split whenver you need to access it.  Splitting a string into a slice is quite fast, and in some cases you might even get escape analysis to notice the slice isn't leaked.

(only works if there is a safe way to use some delimiter without messing up the tags)

kylelemons (approved):

kafkabp/consumer.go
fishy: it's kind of a compromise between excessive pointer (dereference) vs. type assertions to deal with `any`, as `atomic.Pointer` forces pointer even if the interface itself is already a pointer. the main problem is basically that there's not a way to say "pointer-like" in go's type system (e.g. make it happy to accept a pointer that implements an interface).

between the two I think type assertions are worse, but maybe that's just me :)

fishy: I can keep the helper function to return a (possibly nil) interface instead, so the code here won't look as bad ¯\\\_(ツ)\_/¯

fishy: https://github.com/reddit/baseplate.go/pull/587/commits/fe2a4e1dc4ec2cdce465e1c98673c55eababe8b1

pacejackson (approved):
I think the commit message you want is "Drop support for go 1.18", "drop support **to**" sounds like you are dropping the support down to 1.18 rather than removing it.

fishy:
Made a slight word tweak on the comments because the original one has some grammar errors :) merging it.

PR #588, opened 2022-12-16
secrets: Fix vault CSI support

fishy:
In Vault CSI the secret files are actually written under `..data` subdirectory which is also a symlink, and fs.WalkDir will skip symlinks unless they are the root.

So walk under `..data` instead, and also strip the prefix from the path key. Also add a test to test the rotation case and some improvements on existing tests.


secrets/secrets.go
```diff
@@ -323,19 +324,30 @@ func secretsValidate(secretsDocument Document) (*Secrets, error) {
...
+		// ref: https://pkg.go.dev/sigs.k8s.io/secrets-store-csi-driver/pkg/util/fileutil#AtomicWriter.Write
+		k8sSubdirectory = "..data"
+	)
 	secretsDocument := Document{
 		Secrets: make(map[string]GenericSecret),
 	}
 	err := fs.WalkDir(
 		dir,
-		".",
+		k8sSubdirectory,
```
kylelemons, line 337: 🔕 Probably worth a quick stat to make sure it exists, and if it doesn't, return an actionable error ("dir %q does not appear to be the root of a Vault CSI mount point" or something).

Alternately you can probably check the error from WalkDir to see if it's a PathError that's NotExists and matches the k8sSubdirectory.

secrets/store.go
```diff
@@ -41,6 +42,11 @@ type Store struct {
 // Context should come with a timeout otherwise this might block forever, i.e.
 // if the path never becomes available.
 func NewStore(ctx context.Context, path string, logger log.Wrapper, middlewares ...SecretMiddleware) (*Store, error) {
+	return newStoreOverrideFSEventsDelay(ctx, filewatcher.DefaultFSEventsDelay, path, logger, middlewares...)
```
kylelemons, line 45: 🔕 normally when a public constructor calls an internal one, I just make it the corresponding name (`newStore`), even if it has different args.  The suffixes are useful for when there is more than one of something imo

kylelemons (approved):

secrets/secrets.go
fishy: [that's basically the first thing `fs.WalkDir` does](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/io/fs/walk.go;l=110), and the error returned will be returned to the caller on https://github.com/reddit/baseplate.go/blob/2dd3ab82ca91dabd8b2aead5574f57ec30fb2b69/secrets/secrets.go#L355, so that should be good enough? I don't feel like we should add a special stat call to this :)

kylelemons: My concern is the error message, given that we're pointing WalkDir at a directory that the user did not specify directly for an "implementation detail" reason.  `os.Stat("/path/to/your/dir/..data"): file does not exist` is going to be a very opaque error message, and it kinda looks like someone forgot a slash after a `..`.

However, `dir "/path/to/your/dir" does not appear to be the root of a Vault CSI mount point: os.Stat("/path/to/your/dir/..data"): file does not exist` or something is going to be much easier to use to figure out what you need to do -- i.e. point it at a Vault CSI mount point, not any old dir.

kylelemons (approved):

secrets/secrets.go
fishy: Done in latest fixup commit.

secrets/secrets.go
```diff
@@ -321,21 +323,47 @@ func secretsValidate(secretsDocument Document) (*Secrets, error) {
...
 				return err
 			}
 			if d.IsDir() {
 				return nil
 			}
+			relPath, err := filepath.Rel(k8sSubdirectory, path)
+			if err != nil {
+				// Should not happen as this means path is not under k8sSubdirectory,
+				// but just in case
+				return nil
```
konradreiche, line 365: If we handle it for _just in case_, we should also just return the actual error, shouldn't we?

konradreiche (approved):

secrets/secrets.go
fishy: that would actually stop the walking and cause the whole parser to fail, which has bigger consequences, so probably not. but I don't feel too strongly either way

mterwill (approved):

kylelemons (approved):

PR #589, opened 2022-12-16
Create a secrets.Wrap API for wrapping alternate secrets implementations

kylelemons:
## 💸 TL;DR
This change allows creating an API-compatible Store object that delegates the calls to secret-getting functions to another store.

This is necessary because baseplate.go/v0 and edgecontext/v0 both hard-code *secrets.Store, and so it is not only a breaking change to replace them with an interface, it would require a lock-step change across two v0.x libraries, which is very likely to bite consumers.

## ✅ Checks

### DIFF
--- a/secret/secrets.go
+++ b/secret/secrets.go
@@ -1,0 +1,1 @@
+package secret
--- a/secrets/store.go
+++ b/secrets/store.go
@@ -3,6 +3,7 @@ package secrets
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"io/fs"
 	"os"
@@ -29,6 +30,8 @@ func nopSecretHandlerFunc(sec *Secrets) {}
 // memory so there's little performance impact to doing so and you will be sure
 // to always have the current version in the face of key rotation etc.
 type Store struct {
+	actual secretStore // if present, all methods are delegated to this interface
+
 	watcher filewatcher.FileWatcher
 
 	secretHandlerFunc SecretHandlerFunc
@@ -113,6 +116,9 @@ func (s *Store) getSecrets() *Secrets {
 //
 // Close doesn't return non-nil errors, but implements io.Closer.
 func (s *Store) Close() error {
+	if s.actual != nil {
+		return nil
+	}
 	s.watcher.Stop()
 	return nil
 }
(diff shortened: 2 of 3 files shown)


secret/secrets.go
```diff
@@ -0,0 +1 @@
+package secret
```
fishy, line 1: we already have https://github.com/reddit/baseplate.go/blob/master/secrets/doc.go so I don't think we need this file.

kylelemons:
Hmm, it looks like the edgecontext library expects middleware to work...

secrets/store.go
```diff
@@ -149,5 +167,21 @@ func (s Store) GetCredentialSecret(path string) (CredentialSecret, error) {
...
+// A secretStore is an interface that is satisfied by both v0 and v2 secrets.
+type secretStore interface {
+	GetSimpleSecret(path string) (SimpleSecret, error)
+	GetVersionedSecret(path string) (VersionedSecret, error)
+	GetCredentialSecret(path string) (CredentialSecret, error)
+}
+
+// Wrap returns a Store that forwards all secret retrieval operations to the given store.
+// The returned Store cannot use middleware or provide its Vault information.
+func Wrap(store secretStore) *Store {
```
fishy, line 185: this means we have to make a public api using an internal interface, just export `secretStore` interface?

I guess there's a possibility that we would add another common interface to both v0 and v2 and that will cause a breaking change to a public interface, but I think that possibility is pretty low at this point? if we really think that's possible then maybe just use `any` in the function signature and return `nil` if the `store` passed in does not satisfy the interface?

secrets/store.go
```diff
@@ -113,6 +116,9 @@ func (s *Store) getSecrets() *Secrets {
 //
 // Close doesn't return non-nil errors, but implements io.Closer.
 func (s *Store) Close() error {
+	if s.actual != nil {
+		return nil
```
fishy, line 120: just add `io.Closer` to the interface as well?

fishy (changes requested):

fishy:
> Hmm, it looks like the edgecontext library expects middleware to work...

ah yes that's true. we actually use middleware to decode the jwt token once vs. every time it's used.

secret/secrets.go
ghirsch-reddit: Purely for my own learning, what does having this (or the other) empty file accomplish?

kylelemons:
Closing in favor of https://github.com/reddit/baseplate.go/pull/588

PR #591, opened 2023-01-05
errorsbp: Update for go 1.20

fishy:
Go 1.20 is going to add native support for batched/multi errors, and we will be able to eventually deprecate errorsbp.Batch, but we can only do so after we dropped support for go 1.19.

Before that, make errorsbp.Batch play nicer with go 1.20 by:

* Implement the new `Unwrap() []error` optional interface.
* Make errorsbp.BatchSize also support other errors implementing the new optional interface.

### DIFF
--- a/errorsbp/batch_test.go
+++ b/errorsbp/batch_test.go
@@ -2,6 +2,7 @@ package errorsbp_test
 
 import (
 	"errors"
+	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


errorsbp/batch_test.go
```diff
@@ -265,3 +266,59 @@ func TestAddPrefix(t *testing.T) {
...
+			label: "batch-2",
+			want:  2,
+			err: func() error {
+				var batch errorsbp.Batch
+				batch.Add(errors.New("foo"))
+				batch.Add(errors.New("bar"))
+				return batch
+			}(),
+		},
+		// TODO: Add cases from errors.Join and fmt.Errorf once we drop support for
```
pacejackson, line 315: can you add a test case for the recursive case?

pacejackson (changes requested):

pacejackson (approved):

kylelemons (approved):
ah perfect, you already addressed the comment I was about to make 👍

ghirsch-reddit (approved):

PR #593, opened 2023-01-10
thriftbp: Force recoverPanik to be the last server middleware

fishy:
This was already done in httpbp. Without this, middlewares after it will get nil error when panic happens, which can result in inaccurate/misleading metrics or other bugs.

To do so:
* Mark the old RecoverPanic middleware as deprecated
* Unexport RecoverPanik to recoverPanik. This is a breaking change, but RecoverPanik was only added in beta releases, it's not in v0.9.11.

### DIFF
--- a/thriftbp/server_middlewares.go
+++ b/thriftbp/server_middlewares.go
@@ -67,8 +67,6 @@ type DefaultProcessorMiddlewaresArgs struct {
 // 4. ReportPayloadSizeMetrics
 //
 // 5. PrometheusServerMiddleware
-//
-// 6. RecoverPanik
 func BaseplateDefaultProcessorMiddlewares(args DefaultProcessorMiddlewaresArgs) []thrift.ProcessorMiddleware {
 	return []thrift.ProcessorMiddleware{
 		ExtractDeadlineBudget,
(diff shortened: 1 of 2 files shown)


thriftbp/server_middlewares.go
```diff
@@ -358,15 +355,21 @@ func tHeaderProtocol2String(proto thrift.THeaderProtocolID) string {
...
+// Deprecated: This is always added as the last server middleware (with a
+// different name) to ensure that other server middlewares get the correct error
+// if panic happened.
 func RecoverPanic(name string, next thrift.TProcessorFunction) thrift.TProcessorFunction {
-	return RecoverPanik(name, next)
+	return recoverPanik(name, next)
 }
 
-// RecoverPanik recovers from panics raised in the TProccessorFunction chain,
+// recoverPanik recovers from panics raised in the TProccessorFunction chain,
```
kylelemons, line 369: was this part of a tagged release?

kylelemons (approved):

thriftbp/server_middlewares.go
fishy: it's tagged in beta, not in the last non-beta release (v0.9.11)

pacejackson (approved):

ghirsch-reddit (approved):

PR #595, opened 2023-01-23
Make the HTTP size buckets usable

reptile-deer-66:
Current buckets are defined to be the same as the latency buckets, which don't go above ~15. While this works in the context of `seconds`, it doesn't work well for `bytes`. Those new buckets should allow us to track a wide range of request/response sizes.

### DIFF
--- a/httpbp/prometheus.go
+++ b/httpbp/prometheus.go
@@ -16,6 +16,22 @@ const (
 	endpointLabel   = "http_endpoint"
 )
 
+var ResponseSizeBuckets = []float64{
+	100,       // 100 B
+	500,       // 500 B
+	1 << 10,   // 1 KiB
+	4 << 10,   // 4 KiB
+	10 << 10,  // 10 KiB
+	50 << 10,  // 50 KiB
+	100 << 10, // 100 KiB
+	500 << 10, // 500 KiB
+	1 << 20,   // 1 MiB
+	5 << 20,   // 5 MiB
+	10 << 20,  // 10 MiB
+	50 << 20,  // 50 MiB
+	100 << 20, // 100 MiB
+}
+
 var (
 	serverLabels = []string{
 		methodLabel,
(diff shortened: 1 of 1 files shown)


httpbp/prometheus.go
```diff
@@ -16,6 +16,22 @@ const (
 	endpointLabel   = "http_endpoint"
 )
 
+var ResponseSizeBuckets = []float64{
```
ghirsch-reddit, line 19:
```suggestion
// should be kept in sync with the buckets defined in https://github.snooguts.net/reddit-go/httpbp/blob/main/httpbp/internal/httpmetrics/metrics.go
var ResponseSizeBuckets = []float64{
```

fishy, line 19: I'm not sure we want to export this and make this part of our API, better unexport it (use lower case `responseSizeBuckets`) instead.

also this is used in both request and response, so a better name would be `payloadSizeBuckets`.

ghirsch-reddit: 🔕 I think Yuxuan had also suggested not exporting the buckets 🤷

ghirsch-reddit (approved):
Thanks for fixing it!

fishy (changes requested):

httpbp/prometheus.go
reptile-deer-66: Updated!

httpbp/prometheus.go
```diff
@@ -16,6 +16,22 @@ const (
 	endpointLabel   = "http_endpoint"
 )
 
+var payloadSizeBuckets = []float64{
```
ghirsch-reddit, line 19: You missed the comment about adding a comment :)

reptile-deer-66: Ah man, I did.

fishy (approved):

httpbp/prometheus.go
reptile-deer-66: Take 3.

PR #596, opened 2023-01-24
thriftbp: Add retry to ttlClient.refresh

fishy:
This was discussed in the previous WG meeting, that we want ttlClient.refresh to retry to mitigate occasional dns blips.

### DIFF
--- a/thriftbp/ttl_client.go
+++ b/thriftbp/ttl_client.go
@@ -5,10 +5,12 @@ import (
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/avast/retry-go"
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/reddit/baseplate.go/prometheusbp"
 	"github.com/reddit/baseplate.go/randbp"
+	"github.com/reddit/baseplate.go/retrybp"
 )
 
 type ttlClientGenerator func() (thrift.TClient, thrift.TTransport, error)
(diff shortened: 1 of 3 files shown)


thriftbp/ttl_client.go
```diff
@@ -60,6 +73,9 @@ func (c *ttlClient) Close() error {
 		c.state <- state
 	}()
 	state.closed = true
+	if state.cancelRefresh != nil {
+		state.cancelRefresh()
```
kylelemons, line 77: 🔕 (optional) This doesn't join the background goroutine, so it could still be left running / hanging after Close.  (I don't think we guaranteed that before, it's a very useful thing to try to ensure)

kylelemons (approved):

kylelemons (approved):

thriftbp/ttl_client.go
fishy: A possible scenario is that `refresh` is currently blocking on dns, if we try to join here it will block `Close()` for a long time.

And `Close()` calls can often happen in hot path (e.g in a client call, we get a client out of the pool, found out that it's broken, so we `Close` it before opening a new one), which means this can unintentionally block on hot path.

konradreiche (approved):

thriftbp/ttl_client.go
```diff
@@ -97,15 +114,59 @@ func (c *ttlClient) IsOpen() bool {
 
 // refresh is called when the ttl hits to try to refresh the connection.
 func (c *ttlClient) refresh() {
-	client, transport, err := c.generator()
-	if err != nil {
+	// set up ctx for retry cancellation
+	var ctx context.Context
+	var closeCalled bool
+	{
```
ghirsch-reddit, line 120: Does this bracket notation (without a conditional or loop or function or anything) mean something? Haven't seen it before

thriftbp/ttl_client.go
```diff
@@ -97,15 +114,59 @@ func (c *ttlClient) IsOpen() bool {
...
+			client, transport, err = c.generator()
+			if err != nil {
+				ttlClientReplaceCounter.With(prometheus.Labels{
+					clientNameLabel: c.slug,
+					successLabel:    prometheusbp.BoolString(false),
+				}).Inc()
+			}
+			return err
+		},
+		retry.Attempts(ttlClientRefreshAttempts),
```
ghirsch-reddit, line 156: I'm not familiar with `retrybp`, but just confirming it is intentional to use this `retry.Attempts` from the avast package? If so might be worth replicating it in retrybp, but that's sort of separate from this PR

thriftbp/ttl_client.go
```diff
@@ -114,6 +175,7 @@ func (c *ttlClient) refresh() {
 	defer func() {
 		c.state <- state
 	}()
+	state.cancelRefresh = nil
```
ghirsch-reddit, line 178: Is it a problem that we deferred calling this earlier in the function?

thriftbp/ttl_client.go
fishy: it's to limit the scope of local variables, so I don't have to change line 174 from `:=` to `=`.

that being said it actually make more sense to convert this to a lambda with defer instead, so I'll do that.

thriftbp/ttl_client.go
fishy: retrybp is the baseplate wrapper on retry, we don't need to replicate everything when we are not really adding anything.

thriftbp/ttl_client.go
ghirsch-reddit: Ah cool, that's nice to know about!

ghirsch-reddit (approved):

PR #597, opened 2023-01-30
Remove CPU request scale multiplier

SuperQ:
## 💸 TL;DR
Remove `BASEPLATE_CPU_REQUEST_SCALE` to keep users from misusing it. We
still have the `GOMAXPROCS` value as a direct escape hatch for standard
baseplate deployments.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/runtimebp/config.go
+++ b/runtimebp/config.go
@@ -24,8 +24,7 @@ type Config struct {
 //     This should cause the runtime to respect this value directly.
 //  2. If $BASEPLATE_CPU_REQUEST is unset/invalid, it relinquishes control to automaxprocs, minimum 2.
 //     See https://pkg.go.dev/go.uber.org/automaxprocs for specific behavior.
-//  3. Otherwise, $BASEPLATE_CPU_REQUEST is multiplied by $BASEPLATE_CPU_REQUEST_SCALE
-//     (or defaultCPURequestScale) to compute the new GOMAXPROCS, minimum 2.
+//  3. Otherwise, $BASEPLATE_CPU_REQUEST is multiplied by 1.5, minimum 2.
 //
 // InitFromConfig also exports several metrics to facilitate further tuning/analysis.
 //
--- a/runtimebp/internal/maxprocs/maxprocs.go
+++ b/runtimebp/internal/maxprocs/maxprocs.go
@@ -54,10 +54,6 @@ var (
 		Name: "baseplate_go_env_baseplate_cpu_request",
 		Help: "Value of the BASEPLATE_CPU_REQUEST environment variable at startup. 0 if not a number",
 	}, []string{"status"})
-	mEnvCPURequestScale = promauto.With(prometheusbpint.GlobalRegistry).NewGaugeVec(prometheus.GaugeOpts{
-		Name: "baseplate_go_env_baseplate_cpu_request_scale",
-		Help: "Value of the BASEPLATE_CPU_REQUEST_SCALE environment variable at startup. 0 if not a number",
-	}, []string{"status"})
 
 	initialGOMAXPROCS = promauto.With(prometheusbpint.GlobalRegistry).NewGaugeVec(prometheus.GaugeOpts{
 		Name: "baseplate_go_initial_gomaxprocs",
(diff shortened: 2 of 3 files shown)


SuperQ:
I thought about a few other options:
* Crash if the scale is too high.
* Set back to default if scale is too high.

I was also thinking about adding a max check to see if request*scale > runtime.NumCPU(). Would need some mocks to make testing easier.

mterwill (commented):
Is this really necessary? Setting GOMAXPROCS too high will have performance but not safety implications for a given application. There are logs + metrics around all of this; should be easy enough for a service owner to detect and correct a misconfiguration, especially as they're validating their change.

SuperQ:
The reason we found this was due to a service configuring the scale multiplier to 20 with a request of 10. This created 200 runable threads on a 96 cpu instance. Not only is this impacting the service, the number of threads was impacting other containers on the same system in a very negative way.

fishy:
> The reason we found this was due to a service configuring the scale multiplier to 20 with a request of 10. This created 200 runable threads on a 96 cpu instance. Not only is this impacting the service, the number of threads was impacting other containers on the same system in a very negative way.

But that's just a misconfiguration. Someone can also misconfigure request to 200 which will achieve the same thing. I don't think a subtle arbitrary limitation is a good way to handle misconfigurations.

mterwill:
Agree. If we're really concerned about this we could check in infrared and fail to render the manifest.

SuperQ:
What about a check to make sure `GOMAXPROCS` is never more than `runtime.NumCPU()`?

fishy (changes requested):
> What about a check to make sure `GOMAXPROCS` is never more than `runtime.NumCPU()`?

`runtime.NumCPU()` is out of developers' control (they have no control on which node the pod will be scheduled on, it could be anything between 4 cpu or 128 cpu node), so this will introduce random failures on deploys.

SuperQ:
Yes, that's what I'm saying. We can clamp things to the `runtime.NumCPU()`. It's not going to cause a failure, it's just going to avoid spinning up 200 threads on a 4 CPU node.

fishy:
that goes back to my original point that the same pod deployed to different nodes will have different performance (for example, for services requesting 4*2, will have NumCPU capped at 4 when it's deployed to a small node, but have the full 8 when it's deployed to a bigger node).

kylelemons:
The default for GOMAXPROCS, if memory serves, was 2*CPUs+1 for a long time,
so I'm not sure that the CPU count makes sense either as an upper bound.

I think this is something that should be detected and addressed at a
different layer, maybe monitoring, and ideally based on whatever metric is
correlated with the performance loss (CPU contention? Context switches?),
since this is only one way we can hit such cases.

SuperQ:
@kylelemons From my looking at `runtime/os_linux.go`, it hasn't changed since at least 2014 (when it was converted from C to Go).

We added sched latency metrics, which will be available once v0.9.12 is released. That should give us a good idea if we're running into contention there.

I'm fine closing this, and leaving it for other layers to address. But I still think we should consider having GOMAXPROCS limited to NumCPU. From what I've seen, the go scheduler is not tuned or going to operate sanely with more POSIX threads to run on than real CPUs available to a process.

asandoval:
Reopening per internal conversations. Consensus is mounting that there are effectively few, if any, practical scenarios where tuning outside of this guidance is advised.

SuperQ:
I am thinking about changing this proposal to remove the `BASEPLATE_CPU_REQUEST_SCALE` multiplier. If users need an escape hatch, they can directly set `GOMAXPROCS`.

mterwill (changes requested):
The scale parameter facilitates tuning GOMAXPROCS when used with a vertical pod autoscaler.

SuperQ:
Yes, and we don't want users to touch the scale value, as they don't know what to do with it. It's not something we want to allow tuning of.

kylelemons:
We have customers who require this already, and who do know what to do with
it.

BigValen:
Could we make it harder to misinterpret ? Like support keywords, like { HIGH_IO=2.5, DEFAULT=2, CPU_INTENSIVE=1.5 } or something ? Bounds checking ?

kylelemons:
This has exceeded the discussion that can be reasonably conducted as part of a pull request.

If you want to proceed with a change here, please create a decision doc to gather input and establish consensus, and create a meeting if we can't come to one.

PR #598, opened 2023-01-31
clientpool: Handle a race condition of used after closed

fishy:


ghirsch-reddit (approved):

fishy:
actually this only handles get after close but not release after close, which would still panic, so closing as it goes not really improve things.

PR #600, opened 2023-02-06
tracing: Do not auto register ErrorReporterCreateServerSpanHook

fishy:
Currently tracing.InitFromConfig, which is called from baseplate.New, auto registers ErrorReporterCreateServerSpanHook. With the prometheus metrics we already have very good visibility on the errors happened on thrift/grpc/http servers, and those additional sentry reports mostly just adds noise and cost without being useful. Also in the current setup there's not really an easy way for service owners to unregister ErrorReporterCreateServerSpanHook after they called baseplate.New.

As a result, flip it to opt-in. If a service really want it, they can register it themselves in their main function, after baseplate.New call.

### DIFF
--- a/tracing/config.go
+++ b/tracing/config.go
@@ -76,18 +76,10 @@ type Config struct {
 	TestOnlyMockMessageQueue mqsend.MessageQueue `yaml:"-"`
 }
 
-// InitFromConfig initializes the global tracer using the given Config and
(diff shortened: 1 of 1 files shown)


konradreiche (approved):
This is great, I had a custom initializer to opt-out.

pacejackson (approved):

ghirsch-reddit (approved):

PR #602, opened 2023-02-16
Make `thriftbp.NewBaseplateClientPool()` behave more like the `grpcbp` one

nanassito:
## 💸 TL;DR
Changing the behavior of the pool initialization to that it retries to open connections until the required number is opened instead of failing on the first issue.

## 📜 Details
Prior to this PR the behavior of the thriftbp pool initialization was as followed:
* Try up to `initialConnections` times to open a connection, but bail out at the first error.
* If the number of connections that ended up being opened is lower than `requiredInitialConnections` then crash.
* Otherwise proceed despite having less than `initialConnections` connections opened.

This lead to reliability issues for services as the pods sometimes end up starting with very few connections ready. These connections are opened inline with the requests but that adds extra latency and opportunity for failures.
In my opinion this also is a confusing behavior since we end up returning a lower number than configured by the user. It is explicit in the documentation but I still found more than a few people surprised by this behavior.

With this PR applied the new behavior is as followed:
* Open up to `requiredInitialConnections` and retry if necessary until either success or context error
* Open the rest of the `initialConnections` using the same behavior as before.

While not my ideal approach, this allows the user to easily increase the effectively-usable number on `requiredInitialConnections` while leaving the `initialConnection` behavior unchanged.

Rollout:
* Once merged I will need to work with econ-be to update their calls as they are the only ones directly impacted by the behavior change.
* Services will be able to start setting `requiredInitialConnections` to higher values than currently.

## ✅ Checks


mterwill:
(OOO)

clientpool/channel.go
```diff
@@ -19,25 +22,55 @@ var _ Pool = (*channelPool)(nil)
...
+				maxClients:     maxClients,
+			}, lastAttemptErr
+		}
+		c, err := opener()
+		if err == nil {
+			pool <- c
+			i++
+		} else {
+			lastAttemptErr = err
+			log.Warnf("clientpool: error creating required client (will retry): %w", err)
```
pacejackson, line 62: I don't know that I would add a `log.Warn` here, I would worry about this causing a lot of log spam if there's an issue?

thriftbp/client_pool_test.go
```diff
@@ -79,6 +80,116 @@ func TestCustomClientPool(t *testing.T) {
...
+				i := 0
+				return func() (string, error) {
+					i += 1
+					var err error
+					if i == 1 {
+						err = fmt.Errorf("something broke")
+					}
+					return ln.Addr().String(), err
+				}
+			}(),
```
pacejackson, line 118: 🔕 Is there any other way for us to fake these errors other than hacking them into the AddressGenerator? This isn't _really_ the kind of error we would expect to get since most of these just return a static string (we added it to be compatible with some old ads code).

Not a blocker, but it might be nice to have the errors be closer to errors that we would actually expect/not rely on something that is really just a backwards compatibility hack.

clientpool/channel.go
```diff
@@ -19,25 +22,55 @@ var _ Pool = (*channelPool)(nil)
...
+		}
+	}
+	lastAttemptErr = nil
+
+	for i := requiredInitialClients; i < bestEffortInitialClients; i++ {
 		c, err := opener()
 		if err != nil {
-			finalErr = fmt.Errorf(
+			lastAttemptErr = fmt.Errorf(
 				"clientpool: error creating client #%d/%d: %w",
```
pacejackson, line 71:
```suggestion
				"clientpool: error creating best-effort client #%d/%d: %w",
```

pacejackson (approved):

clientpool/channel.go
nanassito: @fishy really wanted a way to make the error clear to users.

thriftbp/client_pool_test.go
nanassito: I don't know, that's how every other test I found does it so I used the same approach

kylelemons (approved):
I've been swamped, from a cursory review this looks fine.  We might want to discuss _always_ trying at least initialConnection times (i.e. not bailing on the first failure) as well, since I think that is an innocuous behavior change, but we can do that separately.

clientpool/channel.go
nanassito: added at 2/s

PR #603, opened 2023-02-17
errorsbp: Add Prefix

fishy:
This replaces errorsbp.Batch.AddPrefix for services already on go 1.20
to migrate from errorsbp.Batch to errors.Join, and help our eventual
deprecation of errorsbp.Batch when we drop support to go 1.19.

Currently a typical use case with errorsbp.Batch is like this:

    var batch errorsbp.Batch
    ...
    for _, work := range works {
      batch.AddPrefix(work.Name(), work.Do())
    }
    ...
    return batch.Compile()

With this change and go 1.20, they would change the code to something
like this:

    var errs []error
    ...
    for _, work := range works {
      errs = append(errs, errorsbp.Prefix(work.Name(), work.Do()))
    }
    ...
    return errors.Join(errs)

Without this change, they would need to add extra ifs to avoid producing
non-nil error if nothing failed:

    var errs []error
    ...
    for _, work := range works {
      if err := work.Do(); err != nil {
        errs = append(errs, fmt.Errorf("%s: %w", work.Name(), err))
      }
    }
    ...
    return errors.Join(errs)

Although errorsbp.Batch.AddPrefix does allow prefixing multiple errors
with the same prefix, which is not easy to do with this new function,
an internal code search shows that almost no one actually use AddPrefix
with multiple errors.

### DIFF
--- a/errorsbp/batch.go
+++ b/errorsbp/batch.go
@@ -146,7 +146,7 @@ func (be *Batch) AddPrefix(prefix string, errs ...error) {
 	}
 
 	appendSingle := func(err error) {
-		be.errors = append(be.errors, fmt.Errorf("%s: %w", prefix, err))
+		be.errors = append(be.errors, Prefix(prefix, err))
 	}
 
 	for _, err := range errs {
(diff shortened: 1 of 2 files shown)


fishy:
see also: https://go.dev/play/p/R7zUo9dwWew

ghirsch-reddit (approved):

konradreiche (approved):

pacejackson (approved):

kylelemons (approved):

PR #605, opened 2023-02-21
Upgrade golang.org/x/net

fishy:
This replaces https://github.com/reddit/baseplate.go/pull/604, as dependabot is not smart enough to actually run go mod tidy.


fishy:
Actually never mind. Github actions failed for other reasons.

PR #606, opened 2023-02-28
clientpool: Follow up on 6f97656

fishy:
The only reason the constructor in clientpool would return both non-nil Pool and error is to avoid making breaking changes. Since we already made a breaking change in 6f97656, it's better to always return nil Pool when returning an error.

Also fix %w used in logging, and add doccomment to the newly added functions in thriftbp.

### DIFF
--- a/clientpool/channel.go
+++ b/clientpool/channel.go
@@ -2,7 +2,6 @@ package clientpool
 
 import (
 	"context"
-	"fmt"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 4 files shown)


ghirsch-reddit (approved):
This all seems reasonable to me :shipit: 

With respect to both this and the other commit you linked to, since these are breaking changes do we need to do a major release (or since we're still on v0 at least a minor release)?

nanassito (approved):
that makes sense

clientpool/channel.go
```diff
@@ -11,23 +10,17 @@ import (
...
 var _ Pool = (*channelPool)(nil)
 
 // NewChannelPool creates a new client pool implemented via channel.
-//
-// Note that this function could return both non-nil Pool and error,
-// when we failed to create all asked bestEffortInitialClients.
-// In such case the returned Pool would have at least requiredInitialClients
-// clients we already established.
-func NewChannelPool(ctx context.Context, requiredInitialClients, bestEffortInitialClients, maxClients int, opener ClientOpener) (Pool, error) {
+func NewChannelPool(ctx context.Context, requiredInitialClients, bestEffortInitialClients, maxClients int, opener ClientOpener) (_ Pool, err error) {
```
konradreiche, line 23: Interesting, never thought wildcard can be used to have partially named result parameters.

konradreiche (approved):
:+1:

kylelemons (approved):
Can you add a blurb to the PR description for inclusion in a release note?

fishy:
> Can you add a blurb to the PR description for inclusion in a release note?

The actual breaking change happened in already merged and tagged as beta commit 6f97656. There's no breaking changes in this PR.

The only behavior change is that `clientpool.NewChannelPool` would no longer return both non-nil Pool and error. Which isn't something very noteworthy?

kylelemons (approved):
> The only behavior change is that clientpool.NewChannelPool would no longer return both non-nil Pool and error. Which isn't something very noteworthy?

It is a behavior change that calling code could have checked for, though they shouldn't have been reasonably able to rely on it, so it seems worth a mention.

fishy:
so for release note:

```
`clientpool.NewChannelPool` would no longer return both non-nil Pool and error.
```
:)

PR #607, opened 2023-03-01
kafkabp: Add ConsumePartitionFuncProvider to non-group consumers

fishy:
Currently when using non-group consumers, they always consume all
partitions from the topic.

This newly added ConsumePartitionFuncProvider allows the services to
optionally only consume from the partitions they are interested in. This
allows hugely reduced network traffic between the consumers and the
brokers when the service only need a subset of the partitions (for
example, the service itself is already sharded and the sharding is
aligned on kafka partitions).

### DIFF
--- a/kafkabp/config.go
+++ b/kafkabp/config.go
@@ -27,6 +27,10 @@ var (
 
 	// ErrOffsetInvalid is thrown when an invalid offset is specified.
 	ErrOffsetInvalid = errors.New("kafkabp: Offset is invalid")
+
+	// ErrNilConsumePartitionFunc is thrown when ConsumePartitionFuncProvider
+	// returns a nil ConsumePartitionFunc.
+	ErrNilConsumePartitionFunc = errors.New("kafkabp: ConsumePartitionFunc is nil")
 )
 
 // ConsumerConfig can be used to configure a kafkabp Consumer.
@@ -68,6 +72,23 @@ type ConsumerConfig struct {
 	//
 	// When GroupID is non-empty, Version must be at least "0.10.2.0".
 	GroupID string `yaml:"groupID"`
+
+	// Optional. This is only applicable when GroupID is empty string.
+	// When GroupID is empty, the configuration enables a TopicConsumer to read
+	// from all partitions of the given kafka stream.
+	// When ConsumePartitionFuncProvider is also specified, the TopicConsumer will
+	// only consume partitions that evaluates to true with the given
+	// predicate returned from ConsumePartitionFuncProvider.
+	// If ConsumePartitionFuncProvider is specified,
+	// it must return a non-nil predicate of type ConsumePartitionFunc,
+	// else the topic consumer will return error ErrNilConsumePartitionFunc.
+	//
+	// This function is called once per reset. The returned ConsumePartitionFunc
+	// is also called once per reset per partition.
+	// The API is designed to be two-layer so that it's possible to shift all
+	// heavylifting to ConsumePartitionFuncProvider to generate a []bool or set of
+	// ints, and make the returned ConsumePartitionFunc just do a simple lookup.
+	ConsumePartitionFuncProvider ConsumePartitionFuncProvider `yaml:"-"`
 
 	// Optional. The version of the kafka broker this consumer is connected to.
 	// In format of "0.10.2.0" or "2.4.0".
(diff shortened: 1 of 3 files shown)


kafkabp/consumer.go
```diff
@@ -214,10 +262,24 @@ func (kc *consumer) Consume(
...
+		partitionConsumers := make([]sarama.PartitionConsumer, 0, numPartitions)
+		partitionsToConsume := make([]bool, numPartitions)
+		partitionsSelector := kc.partitionSelectorProvider(numPartitions)
+		if partitionsSelector == nil {
+			return ErrNilPartitionSelector
+		}
+		for i := 0; i < numPartitions; i++ {
+			partition := partitions[i]
+			partitionsToConsume[partition] = partitionsSelector(partition)
+		}
```
konradreiche, line 276:
```suggestion
		for _, partition := range partitions {
			partitionsToConsume[partition] = partitionsSelector(partition)
		}
```

Maybe easier to read since we don't need the slice indexing here?

kafkabp/consumer.go
```diff
@@ -214,10 +262,24 @@ func (kc *consumer) Consume(
...
-		partitions := *kc.partitions.Load()
-		partitionConsumers := make([]sarama.PartitionConsumer, 0, len(partitions))
+		partitions := kc.getPartitions()
+		numPartitions := len(partitions)
+		partitionConsumers := make([]sarama.PartitionConsumer, 0, numPartitions)
+		partitionsToConsume := make([]bool, numPartitions)
+		partitionsSelector := kc.partitionSelectorProvider(numPartitions)
+		if partitionsSelector == nil {
+			return ErrNilPartitionSelector
+		}
```
konradreiche, line 272: Should this validation happen further up the line?

kafkabp/config.go
```diff
@@ -69,6 +72,15 @@ type ConsumerConfig struct {
...
 	GroupID string `yaml:"groupID"`
 
+	// Optional. This is only applicable when GroupID is empty string.
+	// When GroupID is empty, the configuration enables a TopicConsumer to read from
+	// all partitions of the given kafka stream. When PartitionsSelectorProvider is also specified
+	// the TopicConsumer will only consume partitions that evaluates to true with the given
+	// predicate returned from PartitionsSelectorProvider. If PartitionsSelectorProvider is specified,
+	// it must return a non-nil predicate of type PartitionsSelectorFunc, else the topic
+	// consumer will return an error of type ErrNilPartitionSelector
+	PartitionsSelectorProvider PartitionsSelectorProvider `yaml:"-"`
```
kylelemons, line 82: Why the two levels of indirection? Why not just `func(paritionID, numPartitions int) bool`

kylelemons (approved):

kafkabp/config.go
fishy: there could be expensive computation to generate `[]bool` (or a set of ints) from `numPartitions`, with `func(paritionID, numPartitions int) bool` you would need to use something like `sync.Once` to avoid doing that computation on each call. so tl;dr is that with two-layer approach we can shift all expensive work to this layer and make the next layer a simple lookup from either a slice or a set.

konradreiche (approved):

kafkabp/config.go
kylelemons: Makse sense.  If we are expecting to cover the use-case where the provider is expensive, we should be clear about how ofren / when the provider is called.  My assumption was that this would basically be called each time as `provider(shards)(num)`, and without evidence I think a user should assume that too.

kylelemons (approved):

kafkabp/config.go
```diff
@@ -74,20 +74,21 @@ type ConsumerConfig struct {
...
-	// the TopicConsumer will only consume partitions that evaluates to true with the given
-	// predicate returned from PartitionsSelectorProvider. If PartitionsSelectorProvider is specified,
-	// it must return a non-nil predicate of type PartitionsSelectorFunc, else the topic
-	// consumer will return error ErrNilPartitionsSelector.
+	// When GroupID is empty, the configuration enables a TopicConsumer to read
+	// from all partitions of the given kafka stream.
+	// When ConsumePartitionFuncProvider is also specified, the TopicConsumer will
+	// only consume partitions that evaluates to true with the given
+	// predicate returned from ConsumePartitionFuncProvider.
+	// If PartitionsSelectorProvider is specified,
```
kylelemons, line 82: missed an update here

kylelemons (approved):

PR #608, opened 2023-03-03
log: Auto wrap baseplate.Error in logs

fishy:
A pain point currently is that thrift generated baseplate.Error doesn't provide human readable message automatically since we are using optional fields, logging it will not produce human readable/meaningful info.

Although we provided WrapBaseplateError helper function, it has its own limitations, for example it's usually only safe to be used unconditionally on the client side, which limited its usefulness on the server side.

Add a wrapper to zapcore to auto wrap baseplate.Errors, so whenever they are logged via kv pairs they are always auto wrapped.


log/core_wrapper_test.go
```diff
@@ -33,17 +42,20 @@ func initCore(buf *bytes.Buffer) zapcore.Core {
...
 	t.Run("origin", func(t *testing.T) {
 		buf := new(bytes.Buffer)
 		core := initCore(buf)
 		logger := zap.New(core).Sugar()
 		log(logger)
 		actual := strings.TrimSpace(buf.String())
-		if actual != expectedOrigin {
-			t.Errorf("Expected log line %s, got %s", expectedOrigin, actual)
+		if !strings.HasPrefix(actual, expectedOriginPrefix) {
+			t.Errorf("Expected log line to start with `%s`, got %s", expectedOriginPrefix, actual)
```
kylelemons, line 58:
```suggestion
			t.Errorf("Expected log line to start with %q, got %q", expectedOriginPrefix, actual)
```

kylelemons (approved):

ghirsch-reddit (approved):
LGTM

Also a suggestion for future PRs like this where you move code, it can be convenient to make multiple commits where the first is literally just copy-pasting. That way it's quick for reviewers to first see that change, and to then look at the "actual changes" by filtering to just the other commits.

PR #613, opened 2023-04-11
expose dialer option in redis client config

marcoferrer:
Closes # 

## 💸 TL;DR

Exposing dialer option in redis cluster client config

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


pacejackson (changes requested):
Can't you just set the dialer you want on the returned `*redis.ClusterOptions`? That's part of why we return it like that, so you can customize it with things that aren't in the Config since the config is meant to be parsed from YAML.

marcoferrer:
Turns out that this isnt needed. Originally we were making this update because the client constructor only accepted the bp config struct. But that constructor isn't actually part of baseplate and is instead part of some other internal lib.

PR #620, opened 2023-06-14
[EX-608] Deprecate Experiments

mrlevitas:
## 💸 TL;DR
In the effort to migrate more services to new Decider experiments SDK, the old baseplate Experiments code is being marked deprecated.

## 📜 Details
[Jira](https://reddit.atlassian.net/browse/EX-608)

### DIFF
--- a/experiments/experiments.go
+++ b/experiments/experiments.go
@@ -16,7 +16,11 @@ import (
 
 	"github.com/gofrs/uuid"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
 	"github.com/reddit/baseplate.go/filewatcher"
+	"github.com/reddit/baseplate.go/internal/prometheusbpint"
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/timebp"
 )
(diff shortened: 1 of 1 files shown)


experiments/experiments.go
```diff
@@ -57,6 +57,8 @@ type Experiments struct {
 //
 // Context should come with a timeout otherwise this might block forever, i.e.
 // if the path never becomes available.
+//
+// Deprecated: baseplate.go/experiments is deprecated. Instead, use Decider API (github.snooguts.net/reddit-go/decider).
```
fishy, line 61: it's easier to just mark the whole package as deprecated, see https://github.com/reddit/baseplate.go/blob/b21d4c278843ac83543c84248aaced67aec7a06d/internalv2compat/compat.go#L6 for an example.

also I don't think we want to reference internal ghe from the open source code.

fishy (changes requested):

experiments/experiments.go
mrlevitas: ooh great catch, thanks!
And you're right, let me remove the reference.

mrlevitas: @fishy could you advisee on a separate question:
@kylelemons gave a great suggestion of putting in a prometheus metric into these deprecated `Variant()`/`Expose()` functions so we can see any stragglers still using the old code.

Could you point to an example of how I might be able to add those metrics? I'm not too familiar with the BP.go structure much : (

konradreiche: @mrlevitas you can check out `prometheus.go` in the `thriftbp` package for some examples: https://github.com/reddit/baseplate.go/blob/b21d4c278843ac83543c84248aaced67aec7a06d/thriftbp/prometheus.go

mrlevitas: Moved deprecation comment to package level & added prometheus counters.

mrlevitas:
strange..the branch is [4 commits ahead](https://github.com/reddit/baseplate.go/tree/depracate_experiments) but only 3 are shown here...

PR #621, opened 2023-06-26
Update prom metrics to match baseplate spec

aidaroff:
## 💸 TL;DR

This PR adds labels for prom metrics to match baseplate spec and updates metrics documentation where relevant. 

## 📜 Details
[Design Doc]()
1. `http_server_time_to_first_byte_seconds` -- description is incorrect. Should be “Time elapsed before first byte was sent". 
2. `http_client_latency_seconds` -- adds missing http_endpoint label
3. `http_client_requests_total` -- adds missing http_endpoint label
4. `http_client_active_requests` -- adds missing http_endpoint label 

[Jira]()

## ✅ Checks

### DIFF
--- a/httpbp/prometheus.go
+++ b/httpbp/prometheus.go
@@ -65,7 +65,7 @@ var (
 
 	serverTimeToFirstByte = promauto.With(prometheusbpint.GlobalRegistry).NewHistogramVec(prometheus.HistogramOpts{
 		Name:    "http_server_time_to_first_byte_seconds",
-		Help:    "Response size",
+		Help:    "Time elapsed before first byte was sent",
 		Buckets: prometheusbp.DefaultLatencyBuckets,
 	}, serverLabels)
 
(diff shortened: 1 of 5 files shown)


httpbp/prometheus.go
```diff
@@ -95,6 +95,7 @@ var (
 var (
 	clientLatencyLabels = []string{
 		methodLabel,
+		endpointLabel,
```
fishy, line 98: we are only adding the label to the metrics definition but never set it to non-empty?

fishy (changes requested):

httpbp/prometheus.go
aidaroff: I've added the metrics to the call sites with empty values

kylelemons (approved):
I assume this makes the `baseplate-v2-transition` test-case pass?  If so, ✅

PR #625, opened 2023-07-27
make secrets.Store middleware threadsafe

pacejackson:
## 💸 TL;DR
Guard secrets.Store middleware with a mutex to make adding new middlewares to an existing store threadsafe.

## 📜 Details
Sometimes we have to call `AddMiddleware` in a goroutine which can lead to race conditions with other calls to `AddMiddleware`. Adding a mutex around updating and calling the middleware functions protects us from this.

## ✅ Checks

### DIFF
--- a/secrets/testing.go
+++ b/secrets/testing.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+
 	"github.com/reddit/baseplate.go/filewatcher"
 )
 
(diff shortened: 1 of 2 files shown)


secrets/testing.go
```diff
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-
```
fishy, line 7: nit: don't remove this blank line

fishy (approved):

secrets/testing.go
pacejackson: Weird... not sure where that came from

fishy (approved):

PR #626, opened 2023-08-24
[RFC][Snoosweek] Support for DTab overrides

folone:
## 📚 Background reading
* [Linkerd doc](https://linkerd.io/1/advanced/dtabs/)
  * [Namers](https://github.com/linkerd/linkerd/blob/master/linkerd/docs/namer.md)
* [Finagle doc](https://twitter.github.io/finagle/guide/Names.html#interpreting-paths-with-delegation-tables)

## 💸 TL;DR
This PR demonstrates a simple support for system-wide request-level target override for services. I've tried my best to demonstrate how it would work in the following diagram:
Diagram

## 📜 Details
This is a demonstration PR, intended to elicit comments/opinions, not intended for merging as is.

This change does not exactly implement DTabs as described in the docs above – those require some internal knowledge of service/name resolver on behalf of the client constructor. Rather, this change is inspired by those documents. Which means the name "DTab" might not actually be very relevant here. Should this be called something like "ServiceAddressOverrides" instead?

### 💡 The general idea is as follows:
* We add a new request header that describes the desired overrides. In this implementation, the header (`X-DTabs`) takes the following form: `service.a.prod=>service.a.staging; service.b.prod=>service.b.staging;`. In this case we instruct the client to replace all the calls to `service.a.prod` with `service.a.staging` and all the calls to `service.b.prod` with `service.b.staging`. See the newly added tests for more context.
* The header gets propagated in the contexts, the same way we propagate other headers (such as the tracing headers).
* Whenever we encounter calls to `service.a.prod` or `service.b.prod` on our request path, and the above DTab is set, we reroute the requests to `service.a.staging`/`service.b.staging` correspondingly.

## 🧑‍🔬 User stories
* I'm a Product Manager, wanting to test out a feature that one of the engineers on my team is working on. Standing up a staged environment of everything is fine I guess. But I'd prefer that the engineer could stage just the services they are changing, and give me some headers I could plug into [ModHeader](https://chrome.google.com/webstore/detail/modheader-modify-http-hea/idgpnmonknjnojddfkpgkljpfnnfcklj), so I could use the actual product we're developing in its production environment (except for that one service).
* I'm an engineer building a [Snoosweek](https://www.reddit.com/r/RedditEng/comments/15r5oez/the_fall_2023_of_snoosweek/) feature. A bunch of the code I've written is in other teams' codebases, and lacks testing coverage (since it's a prototype, and I only had half a week to tie it all together). Reasonably enough, that other team won't let me merge/deploy the code to be able to demo it, even if I promise to hide everything behind an experiment and roll everything back next week.
* I'm an on-call engineer from a Core Services team during Snoosweek. Dozens of engineers have amazing ideas that involve code changes in my team's codebases. I'm happy to facilitate them demoing those changes, but a lot of these changes have merge conflicts with each other, and I don't exactly want all of those prototypes in our codebase during a code freeze. Wouldn't it be great if they could just stage their own versions of the service?

## 🔨 Improvements that can/should be made here
- [x] Creating new clients for dtabbed requests is much more expensive than just fetching a client from the pool. We could add an LRU cache for newly created client pools, to avoid re-creating them upon each call.
- [ ] DTabs should be passed around in a parsed form, rather than the string representation I have here, since this requires each successive service to re-parse the header.
- [ ] This PR only implements overriding target addresses for thrift clients. Complete implementation would involve supporting these overrides for all the clients baseplate has (e.g. gRPC).
- [ ] These headers should likely be filtered out from external (outside of Reddit) requests, to avoid leaking existing staged internal service addresses.
- [ ] Tests asserting new clients with address overrides get created when DTab headers are present.

## ✅ Checks

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/gofrs/uuid v3.2.0+incompatible
 	github.com/google/go-cmp v0.5.9
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
+	github.com/hashicorp/golang-lru/v2 v2.0.5
 	github.com/joomcode/errorx v1.0.3
 	github.com/joomcode/redispipe v0.9.4
 	github.com/opentracing/opentracing-go v1.2.0
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/avast/retry-go"
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/reddit/baseplate.go/breakerbp"
@@ -503,12 +504,16 @@ func newClientPool(
 		)
 	}
 
+	cache, err := lru.New[string, *clientPool](20)
+
 	// create the base clientPool, this is not ready for use.
 	pooledClient := &clientPool{
 		Pool: pool,
 
 		cfg:         cfg,
 		middlewares: middlewares,
+
+		cache: cache,
 	}
 	// finish setting up the clientPool by wrapping the inner "Call" with the
 	// given middleware.
@@ -558,6 +563,8 @@ type clientPool struct {
 
 	cfg         ClientPoolConfig
 	middlewares []thrift.ClientMiddleware
+
+	cache *lru.Cache[string, *clientPool]
 
 	wrappedClient thrift.TClient
 }
@@ -646,6 +653,13 @@ func (p *clientPool) getClient(fn func() (clientpool.Client, error)) (_ Client, 
 	return c.(Client), nil
 }
 
+/**
+ * Try to get a new client from DTabs.
+ * This method caches previously created clients in an LRU cache.
+ * If there is an override with the current address in the context,
+ * we try to get the client that has already been created from the cache.
+ * If none found, we create a new one.
+ */
 func (p *clientPool) fromDTabs(ctx context.Context, unparsed string, orElse func() (clientpool.Client, error)) func() (clientpool.Client, error) {
 	return func() (clientpool.Client, error) {
 		dtab, err := transport.ParseDTabs(unparsed)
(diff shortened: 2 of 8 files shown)


konradreiche (commented):
Interesting concept. Is the idea with DTabs, that more of the routing logic is moved to the client-side for specifying network or service hierarchy? :thinking:

thriftbp/client_pool.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/avast/retry-go"
+	lru "github.com/hashicorp/golang-lru/v2"
```
pacejackson, line 14: In the past, we've been hesitant to use this library because it included patented code:

https://github.com/hashicorp/golang-lru/blob/56a2dc02f65b563d946d7ce80b24f905f3519d84/arc/arc.go#L18

But it looks like they moved that into it's own package in v2.0.3

https://github.com/hashicorp/golang-lru/releases/tag/v2.0.3

so maybe it's ok now?

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/gofrs/uuid v3.2.0+incompatible
 	github.com/google/go-cmp v0.5.9
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
+	github.com/hashicorp/golang-lru/v2 v2.0.5
```
pacejackson, line 17: it looks like [v2.0.6](https://github.com/hashicorp/golang-lru/releases/tag/v2.0.6) has a bug-fix

kylelemons (changes requested):
Hi George!

This seems like a cool snoosweek project!  If you're interested in moving forward with this beyond Snoosweek, please write up a 1-pager or design doc and send it to us via Slack!  (We'll need to coordinate with Transport and the other baseplate implementations too)

folone:
@kylelemons @konradreiche @pacejackson Thanks for the feedback folks!
Looks like there might be some tentative interest in something like this. As Kyle suggested, I'll write up a design doc and try to get a slot in your office hours to discuss how feasible this is within Reddit.

fishy:
@folone This part won't really work in Reddit's environment:

>But I'd prefer that the engineer could stage just the services they are changing, and give me some headers I could plug into [ModHeader](https://chrome.google.com/webstore/detail/modheader-modify-http-hea/idgpnmonknjnojddfkpgkljpfnnfcklj), so I could use the actual product we're developing in its production environment (except for that one service)

Because it essentially requires traffic between production and staging clusters, which is not allowed.

Basically at Reddit stagings can only talk to stagings and productions can only talk to productions.

PR #628, opened 2023-09-08
Update go support to 1.20+1.21

fishy:
Also upgrade thrift to 0.19.0.

Also deprecate errorsbp.Batch.


httpbp/errors.go
```diff
@@ -771,9 +771,9 @@ func ClientErrorFromResponse(resp *http.Response) error {
...
 // keep-alive connections, so you should always defer it after checking error.
 func DrainAndClose(r io.ReadCloser) error {
-	var batch errorsbp.Batch
 	_, err := io.Copy(io.Discard, r)
-	batch.Add(err)
-	batch.Add(r.Close())
-	return batch.Compile()
+	return errors.Join(
+		errorsbp.Prefix("read fully", err),
+		errorsbp.Prefix("close", r.Close()),
```
pacejackson, line 777: ooc, why do we still need to use errorsbp here?

retrybp/retry_test.go
```diff
@@ -94,23 +94,18 @@ func TestDoBatchError(t *testing.T) {
...
-				"wrong number of errors in %v, expected %v",
-				batchErr.GetErrors(),
-				errList,
-			)
-
-			for _, err := range errList {
-				if !errors.Is(batchErr, err) {
-					t.Errorf("%v is not wrapped by %v", err, batchErr)
-				}
+	if got, want := errorsbp.BatchSize(err), len(errList); got != want {
```
pacejackson, line 97: is there something we can/should replace `errorsbp.BatchSize` with?

pacejackson (approved):

httpbp/errors.go
fishy: see https://github.com/reddit/baseplate.go/commit/d939c5374d5c8e453b1cbf5169706b9ac1d6991d

retrybp/retry_test.go
fishy: see https://github.com/reddit/baseplate.go/commit/1edc07c27eff4cfcd00a5ff7bbc69e9d7549d076

basically it's much easier to use than handle a few different cases yourself (check nil, check not a batch error, then try to unwrap to a batch and get the size).

kylelemons (approved):

PR #629, opened 2023-09-13
Add All Go metrics to admin endpoint default

SuperQ:
Closes # 

## 💸 TL;DR
Add All Go `runtime/metrics` to the default included metrics.

This will add several new metrics per process:
* `go_cgo_go_to_c_calls_calls_total`
* `go_cpu_classes_gc_mark_assist_cpu_seconds_total`
* `go_cpu_classes_gc_mark_dedicated_cpu_seconds_total`
* `go_cpu_classes_gc_mark_idle_cpu_seconds_total`
* `go_cpu_classes_gc_pause_cpu_seconds_total`
* `go_cpu_classes_gc_total_cpu_seconds_total`
* `go_cpu_classes_idle_cpu_seconds_total`
* `go_cpu_classes_scavenge_assist_cpu_seconds_total`
* `go_cpu_classes_scavenge_background_cpu_seconds_total`
* `go_cpu_classes_scavenge_total_cpu_seconds_total`
* `go_cpu_classes_total_cpu_seconds_total`
* `go_cpu_classes_user_cpu_seconds_total`
* `go_gc_cycles_automatic_gc_cycles_total`
* `go_gc_cycles_forced_gc_cycles_total`
* `go_gc_cycles_total_gc_cycles_total`
* `go_gc_gogc_percent`
* `go_gc_gomemlimit_bytes`
* `go_gc_heap_allocs_by_size_bytes`
* `go_gc_heap_allocs_bytes_total`
* `go_gc_heap_allocs_objects_total`
* `go_gc_heap_frees_by_size_bytes`
* `go_gc_heap_frees_bytes_total`
* `go_gc_heap_frees_objects_total`
* `go_gc_heap_goal_bytes`
* `go_gc_heap_live_bytes`
* `go_gc_heap_objects_objects`
* `go_gc_heap_tiny_allocs_objects_total`
* `go_gc_limiter_last_enabled_gc_cycle`
* `go_gc_pauses_seconds`
* `go_gc_scan_globals_bytes`
* `go_gc_scan_heap_bytes`
* `go_gc_scan_stack_bytes`
* `go_gc_scan_total_bytes`
* `go_gc_stack_starting_size_bytes`
* `go_godebug_non_default_behavior_execerrdot_events_total`
* `go_godebug_non_default_behavior_gocachehash_events_total`
* `go_godebug_non_default_behavior_gocachetest_events_total`
* `go_godebug_non_default_behavior_gocacheverify_events_total`
* `go_godebug_non_default_behavior_gotypesalias_events_total`
* `go_godebug_non_default_behavior_http2client_events_total`
* `go_godebug_non_default_behavior_http2server_events_total`
* `go_godebug_non_default_behavior_httplaxcontentlength_events_total`
* `go_godebug_non_default_behavior_httpmuxgo121_events_total`
* `go_godebug_non_default_behavior_installgoroot_events_total`
* `go_godebug_non_default_behavior_jstmpllitinterp_events_total`
* `go_godebug_non_default_behavior_multipartmaxheaders_events_total`
* `go_godebug_non_default_behavior_multipartmaxparts_events_total`
* `go_godebug_non_default_behavior_multipathtcp_events_total`
* `go_godebug_non_default_behavior_panicnil_events_total`
* `go_godebug_non_default_behavior_randautoseed_events_total`
* `go_godebug_non_default_behavior_tarinsecurepath_events_total`
* `go_godebug_non_default_behavior_tls10server_events_total`
* `go_godebug_non_default_behavior_tlsmaxrsasize_events_total`
* `go_godebug_non_default_behavior_tlsrsakex_events_total`
* `go_godebug_non_default_behavior_tlsunsafeekm_events_total`
* `go_godebug_non_default_behavior_x509sha1_events_total`
* `go_godebug_non_default_behavior_x509usefallbackroots_events_total`
* `go_godebug_non_default_behavior_x509usepolicies_events_total`
* `go_godebug_non_default_behavior_zipinsecurepath_events_total`
* `go_memory_classes_heap_free_bytes`
* `go_memory_classes_heap_objects_bytes`
* `go_memory_classes_heap_released_bytes`
* `go_memory_classes_heap_stacks_bytes`
* `go_memory_classes_heap_unused_bytes`
* `go_memory_classes_metadata_mcache_free_bytes`
* `go_memory_classes_metadata_mcache_inuse_bytes`
* `go_memory_classes_metadata_mspan_free_bytes`
* `go_memory_classes_metadata_mspan_inuse_bytes`
* `go_memory_classes_metadata_other_bytes`
* `go_memory_classes_os_stacks_bytes`
* `go_memory_classes_other_bytes`
* `go_memory_classes_profiling_buckets_bytes`
* `go_memory_classes_total_bytes`
* `go_sched_pauses_stopping_gc_seconds`
* `go_sched_pauses_stopping_other_seconds`
* `go_sched_pauses_total_gc_seconds`
* `go_sched_pauses_total_other_seconds`
* `go_sync_mutex_wait_total_seconds_total`

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


internal/admin/server_test.go
```diff
@@ -9,7 +9,22 @@ import (
 )
 
 var expectedMetrics = []string{
```
fishy, line 11: looks like on go 1.21 we got more than these: https://github.com/reddit/baseplate.go/actions/runs/6173805147/job/16756970260?pr=629

maybe instead of using diff for this test (equal), we only check that all the `expectedMetrics` are reported instead (minimal)?

fishy (changes requested):

internal/admin/server_test.go
SuperQ: The other thing we could do is create our own custom filter and only include the metrics we want.

```
  RedditMetricsGC = collectors.GoRuntimeMetricsRule{regexp.MustCompile(`^/gc/.*`)}
```

fishy: the new ones added in go 1.21 are also `go_gc_.*`, it just report more gc metrics in go 1.21 so that likely won't work,

internal/admin/server_121_test.go
```diff
@@ -0,0 +1,18 @@
+//go:build go1.21
+// +build go1.21
```
fishy, line 2: I do feel that those kind of compatibility hacks are hard to maintain (when we upgrade to 1.21+1.22 people may or may not remember to merge this file back into `server_test.go`), so if we expect this list to change frequently between versions, maybe a minimal check is better.

SuperQ: It's not terrible, Go treats `+build go1.21` as `>= 1.21`. So it's pretty easy to make it additive.

pacejackson: Could we just set the min version to 1.21?

internal/admin/server_test.go
SuperQ: I meant that we could customize this regexp, not use the default.

internal/admin/server_121_test.go
SuperQ: That would mean not supporting older compilers. The question is what does this project want to support for Go compiler compatibility.

Personally I recommend following [upstream Go on this](https://go.dev/doc/devel/release#policy).

kylelemons: I don't really think we need to merge them back into the list, that just seems like busywork.  This way it's documented which ones showed up in which version, which isn't really useless info even if it's about a version that's outside the support window.

internal/admin/server.go
```diff
@@ -26,6 +26,7 @@ const (
 var Mux = http.NewServeMux()
 
 var baseplateGoCollectors = collectors.WithGoCollectorRuntimeMetrics(
+	collectors.MetricsGC,
```
kylelemons, line 29: I didn't realize MetricsAll was an option; can we just use MetricsAll?

SuperQ: All is, IMO, a bit too noisy. I would prefer to be selective in what we enable.

SuperQ:
Updated now that 1.21 is the minimum version.

kylelemons (approved):

SuperQ:
Updated to include MetricsAll.

fishy (approved):
I'm still not sure that having an unit test to ensure that everything defined in upstream packages (go itself and prometheus) that we don't control is a good idea and scales well. I think the unit test should be changed to check for subset (makes sure a few key metrics are there) instead of equal.

PR #631, opened 2023-10-17
Implement streaming interceptors

RGood:
Closes # 

## 💸 TL;DR
* Implement streaming interceptors

## ✅ Checks


fishy (changes requested):
I don't think this is the direction we want to go. For what you need you should just use grpcbp v2 clients instead, which already have the streaming interceptors, and even if your service is on v0 using v2 clients are trivial (unlike switching to v2 grpcbp server).

RGood:
@fishy 

> I don't think this is the direction we want to go. For what you need you should just use grpcbp v2 clients instead, which already have the streaming interceptors, and even if your service is on v0 using v2 clients are trivial (unlike switching to v2 grpcbp server).

I really just need the server stream interceptors. I could take or leave the client interceptors

PR #633, opened 2023-11-08
tracing: skip reconfiguration if tracer already configured

redloaf:
## 💸 TL;DR
In baseplate.go v2 we configure a global opentracing tracer. Currently, calling `baseplate.New` will override the new tracer. This change prevents the override using a flag in `internalv2compat`.

## 📜 Details
If `internalv2compat.SetV2TracingEnabled(true)` is called before `InitGlobalTracer`, the opentracing global tracer will not be configured, and a message will be logged.

## ✅ Checks

### DIFF
--- a/internalv2compat/compat.go
+++ b/internalv2compat/compat.go
@@ -65,18 +65,18 @@ type IsThrift interface {
 	isThrift()
 }
 
-var (
-	v2TracingEnabledMutex sync.RWMutex
-	v2TracingEnabled      bool
-)
+var v2Tracing struct {
+	sync.Mutex
+	enabled bool
+}
 
 func SetV2TracingEnabled(enabled bool) {
-	v2TracingEnabledMutex.Lock()
-	defer v2TracingEnabledMutex.Unlock()
-	v2TracingEnabled = enabled
+	v2Tracing.Lock()
+	defer v2Tracing.Unlock()
+	v2Tracing.enabled = enabled
 }
 func V2TracingEnabled() bool {
-	v2TracingEnabledMutex.RLock()
-	defer v2TracingEnabledMutex.RUnlock()
-	return v2TracingEnabled
+	v2Tracing.Lock()
+	defer v2Tracing.Unlock()
+	return v2Tracing.enabled
 }
--- a/internalv2compat/compat.go
+++ b/internalv2compat/compat.go
@@ -8,6 +8,7 @@ package internalv2compat
 
 import (
 	"os"
+	"sync"
 	"sync/atomic"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


internalv2compat/compat.go
```diff
@@ -63,3 +64,19 @@ type IsHTTP interface {
 type IsThrift interface {
 	isThrift()
 }
+
+var (
+	v2TracingEnabledMutex sync.RWMutex
+	v2TracingEnabled      bool
```
fishy, line 70: use [`atomic.Bool`](https://pkg.go.dev/sync/atomic#Bool) instead?

tracing/tracer.go
```diff
@@ -82,6 +83,14 @@ type Tracer struct {
 // If it fails to do so, UndefinedIP will be used instead,
 // and the error will be logged if logger is non-nil.
 func InitGlobalTracer(cfg Config) error {
+	logger := cfg.Logger
+	if logger == nil {
+		logger = log.NopWrapper
+	}
+	if internalv2compat.V2TracingEnabled() {
+		logger(context.Background(), `v2 tracing is enabled; skipping v0 tracer configuration`)
```
fishy, line 91: nit:
```suggestion
	if internalv2compat.V2TracingEnabled() {
		cfg.Logger.Log(context.Background(), `v2 tracing is enabled; skipping v0 tracer configuration`)
```
see https://pkg.go.dev/github.com/reddit/baseplate.go/log#Wrapper.Log

internalv2compat/compat.go
```diff
@@ -63,3 +64,19 @@ type IsHTTP interface {
...
+func SetV2TracingEnabled(enabled bool) {
+	v2TracingEnabledMutex.Lock()
+	defer v2TracingEnabledMutex.Unlock()
+	v2TracingEnabled = enabled
+}
+func V2TracingEnabled() bool {
+	v2TracingEnabledMutex.RLock()
+	defer v2TracingEnabledMutex.RUnlock()
+	return v2TracingEnabled
+}
```
kylelemons, line 82: * I don't think you need an RWMutex here -- they are a bit slower, and are only really useful if you _need_ the multiple readers for throughput
* I often like combining the "locked thing" in with the mutex.

Together it could look like

```suggestion
var v2Tracing struct{
  sync.Mutex
	enabled bool
}

func SetV2TracingEnabled(enabled bool) {
	v2Tracing.Lock()
	defer v2Tracing.Unlock()
	v2Tracing.Enabled = enabled
}
func V2TracingEnabled() bool {
	v2Tracing.Lock()
	defer v2Tracing.Unlock()
	return v2Tracing.Enabled
}
```

kylelemons (approved):

tracing/tracer.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/opentracing/opentracing-go"
 
 	"github.com/reddit/baseplate.go/detach"
+	"github.com/reddit/baseplate.go/internalv2compat"
```
kylelemons, line 15: need a lint ignore for SA1019

tracing/tracer.go
```diff
@@ -102,13 +111,7 @@ func InitGlobalTracer(cfg Config) error {
 
 	tracer.sampleRate = cfg.SampleRate
 	tracer.useHex = cfg.UseHex
-
-	logger := cfg.Logger
-	if logger == nil {
-		logger = log.NopWrapper
-	}
```
fishy, line 109: and then revert this change.

here we decided intentionally to default to `NopWrapper` instead of `DefaultWrapper`

fishy (changes requested):

internalv2compat/compat.go
redloaf: Refactored.

internalv2compat/compat.go
redloaf: I considered it but picked a mutex because this should only be referenced during setup/init, and I feel like the benefit of a memory barrier outweighs the performance benefits of atomics.

fishy (approved):

PR #634, opened 2023-11-09
Add support for MinConnections to thriftbp.ClientPool

m3lawren:
## 💸 TL;DR

This adds the ability to specify a minimum number of connections to `thriftbp.ClientPool` and `clientpool.ChannelPool`. Having a maintained minimum pool size provides benefits for load balancing which can help even out CPU/RAM utilization on the server side.

## 📜 Details

We can't rely on `requiredInitialClients` (since it can fail startup) and if `bestEffortInitialClients` fails we end up with a partly preallocated pool, this provides a mechanism to maintain the pool size on an ongoing basis.

This is accomplished by spinning up an `ensureMinClients` goroutine per pool which will check the total number of connections (active+idle) and open more as needed to keep that amount >= the minimum.

It should be non-breaking as the default behaviour is to not have a minimum set and not spin up the `ensureMinClients` goroutine.

## ✅ Checks

### DIFF
--- a/clientpool/channel.go
+++ b/clientpool/channel.go
@@ -10,21 +10,33 @@ import (
 )
 
 type channelPool struct {
-	pool       chan Client
-	opener     ClientOpener
-	numActive  atomic.Int32
-	maxClients int
+	pool                   chan Client
+	opener                 ClientOpener
+	numActive              atomic.Int32
+	numTotal               atomic.Int32
+	backgroundTaskInterval time.Duration
+	minClients             int
+	maxClients             int
+	isClosed               atomic.Bool
 }
+
+const DefaultBackgroundTaskInterval = 5 * time.Second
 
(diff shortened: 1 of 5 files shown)


clientpool/channel.go
```diff
@@ -181,3 +220,24 @@ func (cp *channelPool) NumAllocated() int32 {
 func (cp *channelPool) IsExhausted() bool {
 	return cp.NumActiveClients() >= int32(cp.maxClients)
 }
+
+func (cp *channelPool) ensureMinClients() {
+	for !cp.isClosed.Load() {
+		time.Sleep(cp.backgroundTaskInterval)
```
fishy, line 226: create a [`time.Ticker`](https://pkg.go.dev/time#Ticker) to loop through instead of using sleep, you can use closed signal as the condition of termination of the loop.

PR #635, opened 2023-11-10
tracing: use opentracing.GlobalTracer instead of globalTracer

redloaf:
## 💸 TL;DR

There are some places where `globalTracer` is referenced instead of the `opentracing.GlobalTracer()`. If these ever differ (which is entirely possible), this breaks assumptions about how this library integrates with opentracing.

## 📜 Details
If `nil` is passed as the `tracer` to `newTrace` (which is incredibly common), use opentracing's global tracer instead of baseplate's.

## ✅ Checks


tracing/trace.go
```diff
@@ -49,16 +51,36 @@ type trace struct {
 	tags     map[string]string
 }
 
+func (t *trace) getTracer() *Tracer {
```
kylelemons, line 54: we don't use `get` prefixes, maybe `internalTracer()` or `internal()`?

tracing/trace.go
```diff
@@ -49,16 +51,36 @@ type trace struct {
...
 func newTrace(tracer *Tracer, name string) *trace {
+	var (
+		otTracer        opentracing.Tracer
+		traceID, spanID string
+	)
 	if tracer == nil {
-		tracer = &globalTracer
+		otTracer = opentracing.GlobalTracer()
+		traceID = globalTracer.newTraceID()
+		spanID = globalTracer.newSpanID()
```
kylelemons, line 72: 🤔 this seems wrong -- shouldn't it be `otTracer.NewTraceID()`? in which case can't these go after the if statement?  If not, then something subtle is happening here and we'll want a comment to explain

redloaf: `opentracing.Tracer` is an interface that has only `StartSpan`, `Inject`, and `Extract` methods. This function expects a `*tracing.Tracer`, but the global `opentracing.Tracer` may not be one. That's why it's relying on the `globalTracer`. I wrote it this way so that even if the `opentracing.GlobalTracer()` has been overridden, the IDs generated by baseplate will still conform to the specified configuration.

tracing/trace.go
```diff
@@ -159,7 +181,10 @@ func (t *trace) publish(ctx context.Context) error {
 	if !t.shouldSample() || t.tracer == nil {
 		return nil
 	}
-	return t.tracer.Record(ctx, t.toZipkinSpan())
+	if tracer := t.getTracer(); tracer != nil {
```
redloaf, line 184: This means if you override the global tracer with one that's not a `*Tracer`, no spans are ever published.

tracing/trace.go
kylelemons: Definitely worth a comment, then!

tracing/trace.go
fishy: aren't we actually trying to override it with a different tracer implementation? isn't it going to break a lot of things?

can you provide a high level explanation on how the library would work with the new tracer implementation?

redloaf: This is actually the behavior we desire. If you provide a tracer that is not a `*Tracer`, then `tracer.Record` is never called. However, presumably whatever tracer you do provide would record its spans and export them. Applying these changes makes this library compatible with other opentracing tracers, such as the opentracing bridge for opentelemetry, which will not be a `*Tracer` but would likely record and ship spans using an OTLP trace exporter.

tracing/span.go
```diff
@@ -161,7 +161,9 @@ func (s Span) StopTime() time.Time {
 // This uses the the logger provided by the underlying tracing.Tracer used to
 // publish the Span.
 func (s Span) logError(ctx context.Context, msg string, err error) {
-	s.trace.tracer.logger.Log(ctx, msg+err.Error())
+	if tracer := s.trace.internalTracer(); tracer != nil {
+		tracer.logger.Log(ctx, msg+err.Error())
+	}
```
kylelemons, line 166: in the interest of fallbacks, should this also do something if internalTracer isn't found?  like fall back to the global logger?

Why does this inject its own logger anyway? 🤔

tracing/trace.go
```diff
@@ -49,16 +51,41 @@ type trace struct {
...
 }
 
+func (t *trace) internalTracer() *Tracer {
+	if t == nil {
+		return nil
+	}
+	if tracer, ok := t.tracer.(*Tracer); ok && tracer != nil {
+		return tracer
+	}
+	return nil
```
kylelemons, line 61: don't return `nil` in-band, that's a recipe for nil pointer dereferences.  Return `*Tracer, bool`

tracing/trace.go
```diff
@@ -91,8 +118,8 @@ func (t *trace) toZipkinSpan() ZipkinSpan {
 	zs.Duration = timebp.DurationMicrosecond(end.Sub(t.start))
 
 	var endpoint ZipkinEndpointInfo
-	if t.tracer != nil {
-		endpoint = t.tracer.endpoint
+	if tracer := t.internalTracer(); tracer != nil {
```
kylelemons, line 121:
```suggestion
	if tracer, ok := t.internalTracer(); ok {
```

(see comment above about returning a bool too)

tracing/trace.go
```diff
@@ -49,16 +51,41 @@ type trace struct {
...
+	)
 	if tracer == nil {
-		tracer = &globalTracer
+		otTracer = opentracing.GlobalTracer()
+	
+		// opentracing.Tracer is an interface that has only StartSpan, Inject, and Extract methods.
+		// This function expects a *tracing.Tracer, but the global opentracing.Tracer may not be one.
+		// That's why it's relying on the globalTracer: even if the opentracing.GlobalTracer() has been overridden,
+		// the IDs generated by baseplate will still conform to the specified configuration.
+		traceID = globalTracer.newTraceID()
```
kylelemons, line 76: if we're not using globalTracer as a tracer, can we shield it somehow so that it only exposes the methods we want to be used on it, so it can't accidentally get missed again?

fishy (commented):
I think we want to take a step back here.

AFAIK we no longer have any infra to handle the zipkin tracing data any more, so the whole `tracing` package is not working. trying to keep it "working" isn't really worth the effort IMHO.

If the intention here is "let's make the `tracing` package interop with the new tracing infra and works to emit new traces", then we should redesign the API around how new tracing system works (with opentelemetry or whatever), and just mark whatever old API that no longer making sense as deprecated, instead of trying to "keep them working".

PR #637, opened 2023-12-21
V2 middleware injection

redloaf:
## 💸 TL;DR

Adds hooks for overriding tracing middleware.

## ✅ Checks

### DIFF
--- a/httpbp/middlewares.go
+++ b/httpbp/middlewares.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/reddit/baseplate.go/ecinterface"
 	"github.com/reddit/baseplate.go/errorsbp"
+	//lint:ignore SA1019 This library is internal only, not actually deprecated
 	"github.com/reddit/baseplate.go/internalv2compat"
 	"github.com/reddit/baseplate.go/log"
 	"github.com/reddit/baseplate.go/metricsbp"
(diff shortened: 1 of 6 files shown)


httpbp/middlewares.go
```diff
@@ -146,6 +147,12 @@ func httpErrorSuppressor(err error) bool {
 // NewBaseplateServer function which will automatically include InjectServerSpan
 // as one of the Middlewares to wrap your handlers in.
 func InjectServerSpan(truster HeaderTrustHandler) Middleware {
+	if mw := internalv2compat.V2TracingHTTPServerMiddleware(); mw != nil {
+		return func(name string, next HandlerFunc) HandlerFunc {
+			return next
+		}
+	}
```
kylelemons, line 154: To make it double clear that the middleware isn't being used here, you could:

```suggestion
	if internalv2compat.V2TracingHTTPServerMiddleware() != nil {
		// Skip span middleware, because v2 needs the endpoint.  The v2 middleware is injected at the endpoint level.
		return func(name string, next HandlerFunc) HandlerFunc {
			return next
		}
	}
```

internalv2compat/compat.go
```diff
@@ -80,3 +88,44 @@ func V2TracingEnabled() bool {
 	defer v2Tracing.Unlock()
 	return v2Tracing.enabled
 }
+func SetV2TracingThriftClientMiddleware(middleware thrift.ClientMiddleware) {
```
kylelemons, line 91: 🔕 I would put the customary spaces in between each function and the one above it

kylelemons (approved):

konradreiche (approved):

fishy (approved):

PR #638, opened 2024-01-05
Add config for the header use by the general ingress

nanassito:
## 💸 TL;DR
We now have a generic ingress for thrift traffic that routes the traffic based on a header. This means that looking at a service config we don't see what service the traffic is going to and that the application need to set the header manually unlike every other configuration bit.

So this PR adds a new configuration to set this header.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -434,7 +434,7 @@ func NewCustomClientPoolWithContext(
 	return newClientPool(ctx, cfg, genAddr, protoFactory, middlewares...)
 }
 
-const ThriftHostnameHeader = "Thrift-Hostname"
+const ThriftHostnameHeader = "thrift-hostname"
 
 func ThriftHostnameHeaderMiddleware(hostname string) thrift.ClientMiddleware {
 	return func(next thrift.TClient) thrift.TClient {
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -214,7 +214,7 @@ type ClientPoolConfig struct {
 
 	// The hostname to add as a "thrift-hostname" header.
 	//
-	// Optional. If empty, not "thrift-hostname" header will be sent.
+	// Optional. If empty, no "thrift-hostname" header will be sent.
 	ThriftHostnameHeader string `yaml:"thriftHostnameHeader"`
 }
 
@@ -436,7 +436,10 @@ func NewCustomClientPoolWithContext(
 
 const ThriftHostnameHeader = "thrift-hostname"
 
-func ThriftHostnameHeaderMiddleware(hostname string) thrift.ClientMiddleware {
+// thriftHostnameHeaderMiddleware adds a `thrift-hostname` header if one was
+// specified in the configuration.
+// This middleware is always added but will only add the header is necessary.
+func thriftHostnameHeaderMiddleware(hostname string) thrift.ClientMiddleware {
 	return func(next thrift.TClient) thrift.TClient {
 		return thrift.WrappedTClient{
 			Wrapped: func(ctx context.Context, method string, args, result thrift.TStruct) (thrift.ResponseMeta, error) {
@@ -525,7 +528,7 @@ func newClientPool(
 
 		slug: cfg.ServiceSlug,
 	}
-	middlewares = append(middlewares, ThriftHostnameHeaderMiddleware(cfg.ThriftHostnameHeader))
+	middlewares = append(middlewares, thriftHostnameHeaderMiddleware(cfg.ThriftHostnameHeader))
 
 	// finish setting up the clientPool by wrapping the inner "Call" with the
 	// given middleware.
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -440,7 +440,9 @@ func ThriftHostnameHeaderMiddleware(hostname string) thrift.ClientMiddleware {
 	return func(next thrift.TClient) thrift.TClient {
 		return thrift.WrappedTClient{
 			Wrapped: func(ctx context.Context, method string, args, result thrift.TStruct) (thrift.ResponseMeta, error) {
-				ctx = AddClientHeader(ctx, ThriftHostnameHeader, hostname)
+				if hostname != "" {
+					ctx = AddClientHeader(ctx, ThriftHostnameHeader, hostname)
+				}
 				return next.Call(ctx, method, args, result)
 			},
 		}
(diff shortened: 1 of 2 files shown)


thriftbp/client_pool.go
```diff
@@ -429,6 +434,19 @@ func NewCustomClientPoolWithContext(
 	return newClientPool(ctx, cfg, genAddr, protoFactory, middlewares...)
 }
 
+const ThriftHostnameHeader = "thrift-hostname"
```
fishy, line 437: is this intentionally lower-case?

for THeaders we use Camel-Case (see https://github.com/reddit/baseplate.go/blob/master/transport/headers.go), it will be better to keep it consistent.

thriftbp/client_pool.go
```diff
@@ -505,6 +523,11 @@ func newClientPool(
 
 		slug: cfg.ServiceSlug,
 	}
+
+	if cfg.ThriftHostnameHeader != "" {
```
fishy, line 527: instead of conditionally add the middleware, it might be better to always add the middleware but the middleware itself conditionally add the header (don't add it if it's empty string)

nanassito: sure

thriftbp/client_pool.go
nanassito: that's the value the ingress expects sadly.

@davinci26 can comment on whether this can be changed or not

davinci26: So according to the HTTP RFC header casing doesnt matter. @fishy does the same apply for `thrift`?

>    Each header field consists of a case-insensitive field name followed
   by a colon (":"), optional leading whitespace, the field value, and
   optional trailing whitespace.
   
   In any case `envoy` will lower case the headers when parsing the thrift payload so it doesn't care but it would be nice if we took a dependency on a thrift spec definition rather than an envoy implementation detail.
   
   
   https://github.com/envoyproxy/envoy/blob/12210e53db6cea79b58ee027f047faa356e1e392/source/extensions/filters/network/thrift_proxy/header_transport_impl.cc#L163

davinci26 (commented):
overall looks pretty sane to me. Is there a way we could add tests?

thriftbp/client_pool.go
```diff
@@ -211,6 +211,11 @@ type ClientPoolConfig struct {
 	//
 	// Optional. If this is empty, no "User-Agent" header will be sent.
 	ClientName string `yaml:"clientName"`
+
+	// The hostname to add as a "thrift-hostname" header.
+	//
+	// Optional. If empty, not "thrift-hostname" header will be sent.
+	ThriftHostnameHeader string `yaml:"thriftHostnameHeader"`
```
kylelemons, line 218:
```suggestion
	// The hostname to add as a "Thrift-Hostname" header.
	//
	// Optional. If empty, no "Thrift-Hostname" header will be sent.
	ThriftHostnameHeader string `yaml:"thriftHostnameHeader"`
```

kylelemons (approved):

thriftbp/client_pool.go
fishy: well this is not http header, it's thrift header.

envoy lower case all headers was a bug and fixed in latest version of envoy: https://github.com/envoyproxy/envoy/issues/20595

davinci26: > well this is not http header, it's thrift header.

I know that, this is what I am asking if you know. Does the thrift protocol state that thrift headers casing is important or not?

davinci26: even if this breaks the code pattern here. This is a header defined externally to baseplate.go so it should be fine to have different casing

thriftbp/client_pool.go
nanassito: well from the rest of the convo it looks like we'll keep the lower case in the end

nanassito:
Ideally I'd like to add tests but I'm not sure how. I haven't found a way to check the headers yet.

thriftbp/client_pool.go
```diff
@@ -429,6 +434,21 @@ func NewCustomClientPoolWithContext(
 	return newClientPool(ctx, cfg, genAddr, protoFactory, middlewares...)
 }
 
+const ThriftHostnameHeader = "thrift-hostname"
+
+func ThriftHostnameHeaderMiddleware(hostname string) thrift.ClientMiddleware {
```
fishy, line 439: Please write doccomments for this function

especially in the doccomment we should mention that if they already set `ThriftHostname` to non-empty string in the config then this middleware will be automatically applied in their client pool and they do not need to explicitly set/add this middleware again.

we do allow users to add additional middlewares via `NewBaseplateClientPool`/`NewBaseplateClientPoolWithContext`/`NewCustomClientPool`, so we want to avoid the situation of people adding it twice.

thriftbp/client_pool.go
fishy: but we still should `s/not/no/` :)

fishy (commented):
I still think if it's not too late we should change the header to Camel-Case, but that's not a blocking issue.

As for testing, the way to test headers is to:
1. start a server from a loopback port
2. create a client pool on the loopback port, with the header config
3. send a request
4. you can check the header on the server endpoint handler via [`thrift.GetHeader(ctx)`](https://pkg.go.dev/github.com/apache/thrift/lib/go/thrift#GetHeader)

thriftbp/client_pool.go
nanassito: I'm also thinking of not exporting it. That would reduce confusion

fishy: yea that works, too.

pacejackson (approved):

fishy (approved):

thriftbp/client_pool_test.go
```diff
@@ -190,6 +193,49 @@ func TestBehaviorWithNetworkIssues(t *testing.T) {
...
 }
 
+type thriftHostnameHandler struct {
+	server baseplate.Server
+}
+
+func (thriftHostnameHandler) IsHealthy(ctx context.Context, _ *baseplatethrift.IsHealthyRequest) (r bool, err error) {
+	value, ok := thrift.GetHeader(ctx, thriftbp.ThriftHostnameHeader)
+	if !ok {
+		return false, errors.New("did not find the thrift header")
```
fishy, line 203: I would slightly prefer to add `tb testing.TB` into `thriftHostnameHandler` struct and use `tb.Error()` instead, but this also works.

PR #639, opened 2024-01-08
kafkabp: Add NewConsumerWithConfigOverriders

fishy:
We occasionally have cases that services need to set some sarama config
that's not supported by kafkabp.ConsumerConfig, and there's currently no
way for them to do that besides forking kafkabp code locally. Provide
this as an escape hatch for easier iteration.

### DIFF
--- a/kafkabp/consumer.go
+++ b/kafkabp/consumer.go
@@ -114,9 +114,23 @@ type consumer struct {
 // topic. This implementation of Kafka consumer is suitable for use cases like
 // deliver config/data through Kafka to services.
 func NewConsumer(cfg ConsumerConfig) (Consumer, error) {
+	return NewConsumerWithConfigOverriders(cfg)
+}
+
+// SaramaConfigOverrider provides a way for users to override certain fields in
+// *sarama.Config generated from ConsumerConfig.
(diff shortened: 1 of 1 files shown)


kafkabp/consumer.go
```diff
@@ -113,12 +117,16 @@ type consumer struct {
 // - If GroupID is empty, it creates a consumer that has the whole view of the
 // topic. This implementation of Kafka consumer is suitable for use cases like
 // deliver config/data through Kafka to services.
-func NewConsumer(cfg ConsumerConfig) (Consumer, error) {
+func NewConsumer(cfg ConsumerConfig, opts ...SaramaConfigOverrider) (Consumer, error) {
```
fishy, line 120: @kylelemons technically this is a breaking change, but it should not break users' code in 99% of the cases.

if you want to be strict I can also add a `NewConsumerWithOverrider` function and keep this one intact.

pacejackson (approved):

kafkabp/consumer.go
fishy: Nevermind added `NewConsumerWithConfigOverriders` instead. I think that's also easier for us to track how many services are using it.

kylelemons (approved):

PR #640, opened 2024-01-16
Uprade gobreaker version to 0.5.0

sagynbek:
## 💸 TL;DR

Update gobreaker version from 0.4.1 to [0.5.0](https://github.com/sony/gobreaker/releases/tag/v0.5.0), which adds IsSuccess to Config

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/breakerbp/failure_ratio.go
+++ b/breakerbp/failure_ratio.go
@@ -76,6 +76,12 @@ type Config struct {
 
 	// Timeout is the duration of the 'Open' state. After an 'Open' timeout duration has passed, the breaker enters 'half-open' state.
 	Timeout time.Duration `yaml:"timeout"`
+
+	// IsSuccessful is called with the error returned from a request.
+	// If IsSuccessful returns true, the error is counted as a success.
+	// Otherwise the error is counted as a failure.
+	// If IsSuccessful is nil, default IsSuccessful is used, which returns false for all non-nil errors.
+	IsSuccessful func(err error) bool
 }
 
 // NewFailureRatioBreaker creates a new FailureRatioBreaker with the provided configuration.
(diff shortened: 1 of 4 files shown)


breakerbp/failure_ratio.go
```diff
@@ -76,6 +76,12 @@ type Config struct {
 
 	// Timeout is the duration of the 'Open' state. After an 'Open' timeout duration has passed, the breaker enters 'half-open' state.
 	Timeout time.Duration `yaml:"timeout"`
+
+	// IsSuccessful is called with the error returned from a request.
+	// If IsSuccessful returns true, the error is counted as a success.
+	// Otherwise the error is counted as a failure.
+	// If IsSuccessful is nil, default IsSuccessful is used, which returns false for all non-nil errors.
+	IsSuccessful func(err error) bool
```
fishy, line 84:
```suggestion
	IsSuccessful func(err error) bool `yaml:"-"`
```

fishy (commented):
The majority of our code using breaker is like this:
```go
resp, err := breaker.Execut(func() (any, error) {
  resp, err := actualCall(...)
  return resp, err
})
```
which means the feature `IsSuccessful` provided can easily be achieved via:

```go
var resp respType
var err error
breaker.Execut(func() (any, error) {
  resp, err = actualCall(...)
  if myIsSuccessful(err) {
    return resp, nil
  }
  return resp, err
})
```
so I don't think there's any urgency to merge this (as none of the use cases are actually blocked by this), I'd prefer to wait for some services actually try this to see how useful it is before proceeding.

sagynbek:
That's good workaround, it could be useful for this case: 

We get validation error, we don't want it to increase error count of Circuit breaker, but still want original `err` to exit function.

Counter argument might be, if it's validation error then probably `resp` is `nil` and we could use that to check? That would be right, but in general `IsSuccess` looks useful overall )) It's not urgent to merge, but might be included in next release?
cc: @fishy

kylelemons:
Have you tested this version in production yet?

kylelemons (commented):
You can use `go get` to update the dependency in your own application, you do not need to update it in baseplate in order for you to update.

As was mentioned above, it also doesn't seem like IsSuccessful is needed given the usual patterns of gobreaker, in which case you wouldn't even need to make your own copy of NewFailureRatioBreaker to test out the ability, but you could copy that function into your own repo it looks like to validate IsSuccessful is working as you expect.

And by production, I do mean production, not staging :D.

PR #641, opened 2024-01-24
thriftbp: Support native slog.LogValuer implementation for errors

fishy:
The next thrift release, 0.20.0, will add slog.LogValuer for all thrift generated errors [1] [2]. This prepares us so that WrapBaseplateError will keep the implementation if it's available to be used with slog.

This also drops support to go 1.20. 1.22 is already at rc2 today so I think it's time to do that.

[1]: https://github.com/apache/thrift/pull/2884
[2]: https://issues.apache.org/jira/browse/THRIFT-5745

### DIFF
--- a/internal/thriftint/baseplate_error.go
+++ b/internal/thriftint/baseplate_error.go
@@ -5,6 +5,7 @@ package thriftint
 import (
 	"errors"
 	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/apache/thrift/lib/go/thrift"
@@ -36,9 +37,14 @@ var (
 )
 
 type wrappedBaseplateError struct {
-	cause error
-	bpErr baseplateError
+	cause    error
+	bpErr    baseplateError
+	logValue slog.Value
 }
+
+var (
+	_ slog.LogValuer = wrappedBaseplateError{}
+)
 
 func (e wrappedBaseplateError) Error() string {
 	var sb strings.Builder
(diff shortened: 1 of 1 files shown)


pacejackson (approved):

konradreiche (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/reddit/baseplate.go
 
-go 1.20
+go 1.21
```
kylelemons, line 3: I think we should wait until 1.22 is fully released.  I don't expect 1.22 until next month, right?

fishy: I disagree that baseplate.go needs to follow go release cycle strictly. if some service is still using go 1.20, they are _very unlikely_ to upgrade baseplate.go version anyways, and if upgrading baseplate.go version breaks them it's time for them to upgrade to go 1.21 as well.

we certainly should not drop support to a go version as soon as the next version is released, but dropping support at the end of the lifecycle to a go version is appropriate. 1.20 is also unlike 1.19 that the next version has some performance regressions.

kylelemons: Well, the baseplate team _does_ want to follow the upstream Go release cycle :).  Maybe not "strictly," but definitely not _more_ strictly than the upstream.

fishy:
💇 @kylelemons go 1.22 is released today :)

fishy:
(will rerun github actions once the 1.22 image is ready)

kylelemons (approved):

pacejackson (approved):

PR #642, opened 2024-02-05
Introduce an error for trying to fetch a secret of the wrong type.

folone:
## 💸 TL;DR
This PR introduces a new error type for attempting to fetch a secret of the wrong type:
```
secrets: requested secret at path 'secret/myservice/external-account-key' of type 'simple' does not exist, but a secret of type 'versioned' does. Consider using the correct API to retrieve the secret
```

I've seen this issue recently, and had a few infra folks trying to debug why secrets are not being fetched. Turned out, I was using the wrong API (`GetSimpleSecret` instead of `GetVersionedSecret`). This change would've avoided this whole situation.

Added some unit tests.

## ✅ Checks

### DIFF
--- a/secrets/secrets_test.go
+++ b/secrets/secrets_test.go
@@ -118,3 +118,65 @@ func TestNewSecrets(t *testing.T) {
 		})
 	}
 }
+
+func TestSecretsWrongType(t *testing.T) {
+	rawSecrets := `
+		{
+			"secrets": {
+				"secret/myservice/external-account-key": {
+					"type": "versioned",
+					"current": "YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU=",
+					"previous": "aHVudGVyMg=="
+				},
+				"secret/myservice/some-api-key": {
+					"type": "simple",
+					"value": "Y2RvVXhNMVdsTXJma3BDaHRGZ0dPYkVGSg==",
+					"encoding": "base64"
+				},
+				"secret/myservice/some-database-credentials": {
+					"type": "credential",
+					"username": "spez",
+					"password": "hunter2"
+				}
+			},
+			"vault": {
+				"url": "vault.reddit.ue1.snooguts.net",
+				"token": "17213328-36d4-11e7-8459-525400f56d04"
+			}
+		}
+	`
+	buf := bytes.NewBuffer([]byte(rawSecrets))
+	secrets, err := NewSecrets(buf)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = secrets.GetSimpleSecret("secret/myservice/external-account-key")
+
+	if err == nil && err.Error() != "secrets: requested secret at path 'secret/myservice/external-account-key' of type 'simple' does not exist, but a secret of type 'versioned' does. Consider using the correct API to retrieve the secret" {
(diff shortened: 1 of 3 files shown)


secrets/secrets_test.go
```diff
@@ -118,3 +118,65 @@ func TestNewSecrets(t *testing.T) {
...
+	buf := bytes.NewBuffer([]byte(rawSecrets))
+	secrets, err := NewSecrets(buf)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = secrets.GetSimpleSecret("secret/myservice/external-account-key")
+
+	if err == nil || err.Error() != "secrets: requested secret at path 'secret/myservice/external-account-key' of type 'simple' does not exist, but a secret of type 'versioned' does. Consider using the correct API to retrieve the secret" {
```
fishy, line 157: while it's probably ok to have one test comparing `err.Error()` to make sure the text is up to expectation, for the majority of the tests, you should use `errors.As` and then compare the fields, over comparing the error text directly, in tests.

actually with the current implementation of `SecretWrongTypeError`, just use `errors.Is` in tests probably also works and is simpler.

secrets/secrets_test.go
```diff
@@ -118,3 +118,65 @@ func TestNewSecrets(t *testing.T) {
...
+	}
+
+	_, err = secrets.GetCredentialSecret("secret/myservice/some-api-key")
+	if err == nil || err.Error() != "secrets: requested secret at path 'secret/myservice/some-api-key' of type 'credential' does not exist, but a secret of type 'simple' does. Consider using the correct API to retrieve the secret" {
+		t.Fatalf("expected error %v, actual: %v", SecretWrongTypeError{
+			Path:         "secret/myservice/some-api-key",
+			DeclaredType: "credential",
+			CorrectType:  "simple",
+		}, err)
+	}
```
fishy, line 181: also please convert the 3 cases into table tests.

fishy (changes requested):
thanks for the PR!

secrets/secrets_test.go
```diff
@@ -118,3 +118,126 @@ func TestNewSecrets(t *testing.T) {
...
+		tt := tt // capture range variable for parallel testing
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			buf := bytes.NewBuffer([]byte(tt.input))
+			secrets, err := NewSecrets(buf)
+			if err != nil {
+				t.Fatal(err)
+			}
+			_, err = tt.function(secrets)
+			if tt.expectedError != nil && err.Error() != tt.expectedError.Error() {
```
fishy, line 238:
```suggestion
			if _, err := tt.function(secrets); !errors.Is(err, tt.expectedError) {
```
(`errors.Is` _should_ work with this implementation, but if it doesn't, then you need to change the type of `expectedError` to `SecretWrongTypeError`, and use something like this to compare:
```go
_, err = tt.function(secrets)
var swye SecretWrongTypeError
if errors.As(err, &swye) {
  // compare fields between swye and tt.expectedError
} else {
  t.Errorf("wrong error type...")
}
```
)

fishy (changes requested):

secrets/secrets_test.go
folone: Tested locally – worked 👍

fishy (approved):

PR #643, opened 2024-02-07
Update supported go versions to 1.21+1.22

fishy:
Split from https://github.com/reddit/baseplate.go/pull/641 to fix 1.22 specific problems.


internal/admin/server_test.go
```diff
@@ -50,12 +49,14 @@ func TestMetrics(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	got := []string{}
+	got := make(map[string]bool)
```
fishy, line 52: changed this to a subset check because https://github.com/reddit/baseplate.go/actions/runs/7807507022/job/21333706033?pr=641

konradreiche (approved):

internal/admin/server_test.go
pacejackson: why did this become an issue in 1.22?

pacejackson (approved):

internal/admin/server_test.go
fishy: 1.22 exposes more metrics and we were testing that they match exactly.

kylelemons: you could steal what we do in baseplate/v2:

https://sourcegraph.build.ue1.snooguts.net/github.snooguts.net/reddit-go/baseplate/-/tree/internal/baseint

kylelemons (approved):

internal/admin/server_test.go
fishy: yea but I don't think it's worth it to keep updating this list with every go release.

this is basically just testing https://github.com/reddit/baseplate.go/blob/49c0448b3fb3f584acde8c7ea83a7a2b1d4c8072/internal/admin/server.go#L28-L30, which basically delegates to prometheus lib code, making sure it _exactly matches_ a list doesn't really gives us more confidence compare to making sure it _contains a given subset_.

PR #646, opened 2024-04-08
Update thrift to v0.20.0

fishy:
Also add a test for 00f31e521d4e1975074a187f1851354d0532146d.


fishy:
Putting this on hold for now. I need to think more on how to handle the deprecation and the switching to slog.

kylelemons (approved):

konradreiche (approved):

PR #647, opened 2024-04-17
httpbp: Fix Retries middleware

fishy:
When we set GetBody in http.Request, it's expected that Body is also
set, add special handling in Retries to make sure we also set Body when
retrying when GetBody is also set before each retry attempt.

Also always clone the request before each retry attempt to avoid some
subtle errors, and skip the Retries middleware altogether if Body is set
but GetBody is not.

### DIFF
--- a/httpbp/client_middlewares.go
+++ b/httpbp/client_middlewares.go
@@ -2,7 +2,9 @@ package httpbp
 
 import (
 	"errors"
+	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"strconv"
 	"sync"
@@ -200,16 +202,36 @@ func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
 // Retries provides a retry middleware by ensuring certain HTTP responses are
 // wrapped in errors. Retries wraps the ClientErrorWrapper middleware, e.g. if
 // you are using Retries there is no need to also use ClientErrorWrapper.
-func Retries(limit int, retryOptions ...retry.Option) ClientMiddleware {
+func Retries(maxErrorReadAhead int, retryOptions ...retry.Option) ClientMiddleware {
 	if len(retryOptions) == 0 {
 		retryOptions = []retry.Option{retry.Attempts(1)}
 	}
 	return func(next http.RoundTripper) http.RoundTripper {
+		// include ClientErrorWrapper to ensure retry is applied for some HTTP 5xx
+		// responses
+		next = ClientErrorWrapper(maxErrorReadAhead)(next)
+
 		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+			if req.Body != nil && req.Body != http.NoBody && req.GetBody == nil {
+				slog.WarnContext(
+					req.Context(),
+					"Request comes with a Body but nil GetBody cannot be retried. httpbp.Retries middleware skipped.",
+					"req", req,
+				)
+				return next.RoundTrip(req)
+			}
+
 			err = retrybp.Do(req.Context(), func() error {
-				// include ClientErrorWrapper to ensure retry is applied for
-				// some HTTP 5xx responses
-				resp, err = ClientErrorWrapper(limit)(next).RoundTrip(req)
+				req = req.Clone(req.Context())
+				if req.GetBody != nil {
+					body, err := req.GetBody()
+					if err != nil {
+						return fmt.Errorf("httpbp.Retries: GetBody returned error: %w", err)
+					}
+					req.Body = body
+				}
+
+				resp, err = next.RoundTrip(req)
 				if err != nil {
 					return err
 				}
--- a/httpbp/client_middlewares_test.go
+++ b/httpbp/client_middlewares_test.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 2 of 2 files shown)


httpbp/client_middlewares.go
```diff
@@ -200,16 +201,23 @@ func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
...
 // wrapped in errors. Retries wraps the ClientErrorWrapper middleware, e.g. if
 // you are using Retries there is no need to also use ClientErrorWrapper.
-func Retries(limit int, retryOptions ...retry.Option) ClientMiddleware {
+func Retries(maxErrorReadAhead int, retryOptions ...retry.Option) ClientMiddleware {
 	if len(retryOptions) == 0 {
 		retryOptions = []retry.Option{retry.Attempts(1)}
 	}
 	return func(next http.RoundTripper) http.RoundTripper {
 		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+			if req.GetBody != nil {
```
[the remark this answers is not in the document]
fishy: that in most cases would result in 4xx errors anyways, I don't think we should try to get _too smart_.

maybe after the retry attempt, check if it's 4xx and the method is non-get and `GetBody` is nil, print a log message saying this is failed likely because you didn't set GetBody, but doing anything beyond the logging would be "too smart" in my books :)

also since `http.NewRequest`(-`WithContext`) auto populates `GetBody` with `*bytes.Buffer`/`*bytes.Reader`/`*strings.Reader` which should already covers like 95% of the cases how people give the request a body, I won't worry too much about that.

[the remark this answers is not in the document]
kylelemons: a few things:
* we shouldn't assume that sending a request without a body is even safe, let alone that we'll be able to distinguish the error the server will return
* failing to have a GetBody can fail _even before_ you have to do a retry, if the request redirects
* there are tons of instances of [&http.Request{](https://sourcegraph.build.ue1.snooguts.net/search?patternType=regexp&q=context%3Areddit+lang%3Ago+%26http.Request&groupBy=repo) that are not calling NewRequest.  Honestly I rarely call it, because I don't like handling the error.

I think the right thing to do is to skip the middleware (with a log message indicating why) if we're not going to be able to function properly, since we can't do the _better_ thing and return an error immediately (since that would be a breaking change).

httpbp/client_middlewares.go
```diff
@@ -200,16 +201,23 @@ func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
...
+				if err != nil {
+					return nil, fmt.Errorf("httpbp.Retries: GetBody returned error: %w", err)
+				}
+				req.Body = body
+			}
 			err = retrybp.Do(req.Context(), func() error {
 				// include ClientErrorWrapper to ensure retry is applied for
 				// some HTTP 5xx responses
-				resp, err = ClientErrorWrapper(limit)(next).RoundTrip(req)
+				resp, err = ClientErrorWrapper(maxErrorReadAhead)(next).RoundTrip(req)
```
kylelemons, line 220: also, shouldn't this be cloning the request?  or does ClientErrorWrapper do that?

httpbp/client_middlewares.go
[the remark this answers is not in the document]
fishy: I don't think there's anywhere in the RFC to say that a POST request must have a body. A server can be lenient on the method (want GET but don't complain if it's a POST instead) and the client can misuse POST instead of GET without a body. So being _too smart_ can break RFC compliant code (although definitely not ideal code).

But I guess rejecting requests with non-nil `Body` and nil `GetBody` is probably safe enough.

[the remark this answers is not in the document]
konradreiche: I struggle to draw the line here between useful defaults and not trying to be too clever.

At least I can see a few cases where POST and nil-body makes sense:

* REST endpoint to create an entity but the server side creates it with default attributes, no body needed 
* Upload of an empty file

While the latter maybe should be an error I'd think it's not for the middleware to decide that. 

But maybe Kyle knows of instances where this just leads to more headaches.

httpbp/client_middlewares.go
konradreiche: Looking over the fields of `http.Request` I can see there are a couple of fields including a channel for cancellation where it's unclear how it would behave if we don't always clone the request.

httpbp/client_middlewares_test.go
```diff
@@ -256,11 +257,22 @@ func TestClientErrorWrapper(t *testing.T) {
...
+		Unwrap() []error
+	}
+	if errors.As(err, &errs) {
+		return errs.Unwrap()
+	}
+	return []error{err}
+}
+
 func TestRetry(t *testing.T) {
-	t.Run("retry for timeout", func(t *testing.T) {
```
fishy, line 260: this subtest is removed because we don't really retry for timeouts (there's no timeout budget left for any retries), the expected attempt in this subtest is also 1.

httpbp/client_middlewares.go
```diff
@@ -200,16 +202,36 @@ func CircuitBreaker(config breakerbp.Config) ClientMiddleware {
...
 		retryOptions = []retry.Option{retry.Attempts(1)}
 	}
 	return func(next http.RoundTripper) http.RoundTripper {
+		// include ClientErrorWrapper to ensure retry is applied for some HTTP 5xx
+		// responses
+		next = ClientErrorWrapper(maxErrorReadAhead)(next)
+
 		return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+			if req.Body != nil && req.Body != http.NoBody && req.GetBody == nil {
+				slog.WarnContext(
```
kylelemons, line 216: 🔕 do you want to this as a once or do it on a rate limit?

kylelemons (approved):

httpbp/client_middlewares.go
fishy: this will only log once per request (it's not inside the retry loop), so if it's logged multiple times that means users have multiple requests constructed incorrectly, which I think it's warranted to log multiple times (for each incorrect request)

and we are using `slog`, so if users really want to rate-limit it/suppress it they can do so in their slog handler :)

wdyt?

konradreiche (approved):

httpbp/client_middlewares.go
kylelemons: 🤷

PR #649, opened 2024-05-03
thriftbp: Support unix domain socket on thrift clients

fishy:
Also update thrifttest to use UDS when possible.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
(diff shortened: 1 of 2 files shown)


thriftbp/thrifttest/server.go
```diff
@@ -131,20 +141,34 @@ func (s *Server) Close() error {
...
+	var socket *thrift.TServerSocket
+	var addr string
+	if cfg.TB != nil {
+		dir := cfg.TB.TempDir()
+		path := filepath.Join(dir, "socket")
+		socket = thrift.NewTServerSocketFromAddrTimeout(&net.UnixAddr{
+			Net:  "unix",
+			Name: path,
+		}, 0)
+		addr = "uds://" + path
```
fishy, line 153: I think this code will likely fail if ran on windows, but I don't think we care about windows users 🤷 

if we do care about them, then we can extract this into 2 files with build conditions.

konradreiche (approved):

kylelemons (approved):
Out of curiosity, what is the reason to prefer unix domain sockets over tcp for testing?  I've always preferred TCP when that's what will be used in real code; the two are very similar but do not behave identically and the underlying libraries may even have different explicit behaviors.

I don't have a strong opinion here of course.

fishy:
> Out of curiosity, what is the reason to prefer unix domain sockets over tcp for testing? I've always preferred TCP when that's what will be used in real code; the two are very similar but do not behave identically and the underlying libraries may even have different explicit behaviors.
> 
> I don't have a strong opinion here of course.

I think make the test as close to real production as possible makes sense. Reverted that part of the changes.

PR #650, opened 2024-05-06
thriftbp: Report request/response payload sizes from client side

fishy:
We currently already have a middleware to report that from the server side, but that's simulated by reconstruct THeaderProtocols and read/write to them, which will not include the header size and will not reflect compressions (zlib, etc.).

This client side reporting is done directly at the TTransport (TSocket) level, so it will reflect the exact number actually read/written at that level. If zlib is enabled in THeaderProtocol the size will reflect compressed payload. The only thing it will not reflect is potential SSL layer (e.g. when using TSSLSocket over TSocket, which we don't support yet).

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -555,13 +555,15 @@ func newClient(
 	genAddr AddressGenerator,
 	protoFactory thrift.TProtocolFactory,
 ) (*ttlClient, error) {
-	return newTTLClient(func() (thrift.TClient, thrift.TTransport, error) {
+	return newTTLClient(func() (thrift.TClient, *countingDelegateTransport, error) {
 		addr, err := genAddr()
 		if err != nil {
 			return nil, nil, fmt.Errorf("thriftbp: error getting next address for new Thrift client: %w", err)
 		}
 
-		transport := thrift.NewTSocketConf(addr, cfg)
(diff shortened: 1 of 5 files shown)


kylelemons (commented):
do we _want_ to make it "more accurate" than the server?  That might be confusing...

fishy:
> do we _want_ to make it "more accurate" than the server? That might be confusing...

why not? the server ones always come with the big caveat that they're estimations (see the doccomment [here](https://pkg.go.dev/github.com/reddit/baseplate.go/thriftbp#ReportPayloadSizeMetrics)), and also the client metrics will have different labels compared to the server ones.

we can _potentially_ do the same thing on the server side in the future (make the server one more accurate), but that requires open up new apis from upstream thrift (in order to do that we need both of these things at the same time: 1) access to the underlying `TTransport`; and 2) a clear boundary when a request is finished. we currently have apis for them separately but not an api with access to both).

konradreiche (approved):

kylelemons (approved):
the reason is that if you want to look at server bandwidth and see a mismatch between the client and the server, you might be rightfully confused.  Having different names for the metrics could help to clarify the difference.  I'm not sure how many people really look at bandwidth though, so very shrug.

PR #652, opened 2024-05-07
thriftbp: Fix a bug caused by bad merge from the previous PR

fishy:
Due to a bad merge between the previous 2 PRs, we are not really using the countingDelegateTransport in the client created by client pool, so the bytes are not counting and we always report 0 as the bytes read/written.

Rename the args inside newClient generator to fix the bug.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -573,26 +573,26 @@ func newClient(
 			return nil, nil, fmt.Errorf("thriftbp: error getting next address for new Thrift client: %w", err)
 		}
 
-		var transport thrift.TTransport
+		var raw thrift.TTransport
 		if path, ok := strings.CutPrefix(addr, "unix://"); ok {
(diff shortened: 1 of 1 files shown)


thriftbp/client_pool.go
```diff
@@ -573,26 +573,26 @@ func newClient(
...
+			TTransport: raw,
 		}
-		if err := cdt.Open(); err != nil {
+		if err := transport.Open(); err != nil {
 			return nil, nil, fmt.Errorf("thriftbp: error opening TSocket for new Thrift client: %w", err)
 		}
 
 		return thrift.NewTStandardClient(
 			protoFactory.GetProtocol(transport),
 			protoFactory.GetProtocol(transport),
```
fishy, line 594: the alternative fix is to use `cdt` over `transport` in these 2 lines, but I think renaming transport to raw and cdt to transport makes the code much clearer.

mathyourlife-reddit (approved):
The fix and rename sound good to me.

PR #654, opened 2024-05-22
Add filewatcher/v2

fishy:
Being the first lib implemented in baseplate.go and long predates generics, the API of filewatcher lacks type safety and can cause other annoyances.

Add filewatcher/v2 that provides:

* Generics/type safety
* Use io.Closer over Stop()
* Use With- options over a Config struct
* Use slog at error level over log wrapper

And change original filewatcher lib to a thin wrapper of v2.

This change intentionally does not change any of its users (experiments, secrets, etc.), as a proof that we did not break v0 API. The next change will switch them to use v2.


filewatcher/filewatcher.go
```diff
@@ -72,34 +64,27 @@ type Parser func(f io.Reader) (data any, err error)
...
+	return v2.WrapDirParser(dp)
 }
 
 // Result is the return type of New. Use Get function to get the actual data.
 type Result struct {
-	data atomic.Value
-
-	ctx    context.Context
-	cancel context.CancelFunc
+	result *v2.Result[any]
```
kylelemons, line 79: 🔕 value rather than pointer?

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+	var lock sync.Mutex
+	forceReload := func() {
+		// make sure we don't run forceReload concurrently
+		lock.Lock()
+		defer lock.Unlock()
+
+		d, mtime, files, err := openAndParse(path, parser, softLimit, hardLimit)
+		if err != nil {
+			slog.ErrorContext(r.ctx, "filewatcher: openAndParse returned error", "err", err)
+		} else {
```
kylelemons, line 134: 🔕  return in the if block so you can outdent the else?

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+			default:
+				// Ignore uninterested events.
+			case fsnotify.Create, fsnotify.Write, fsnotify.Rename, fsnotify.Remove:
+				// Use fsEventDelay to avoid calling forceReload repetively when a burst
+				// of fs events happens (for example, when multiple files within the
+				// directory changed).
+				if timer == nil {
+					timer = time.AfterFunc(fsEventsDelay, forceReload)
+				} else {
+					timer.Reset(fsEventsDelay)
```
kylelemons, line 198: reset is not safe to call if there is a chance that it might be expiring concurrently (which is basically always)

https://pkg.go.dev/time#Timer.Reset

```
if !t.Stop() {
	<-t.C
}
t.Reset(d)
```

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+				continue
+			}
+
+			reload()
+		}
+	}
+}
+
+// The actual data held in Result.data.
+type atomicData[T any] struct {
```
kylelemons, line 217: the name `atomicData` to me implies a type that handles the atomicity itself, rather than the data stored inside an atomic.Value.

I would probably call it `dataSnapshot` or something to imply that it's the pairing of a data and its mtime

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+}
+
+var (
+	_ FileWatcher[any] = (*Result[any])(nil)
+)
+
+// Config defines the config to be used in New function.
+//
+// Can be deserialized from YAML.
+type Config struct {
```
kylelemons, line 232: I wouldn't even have the Config in the v2 API.

You can have a `filewatcherv1.OptionsForV2(Config)` that returns a `filewatcherv2.Option` corresponding to whichever of its options were set to provide forward compatibility without having to carry forward the tightly coupled YAML

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+type opts struct {
+	fsEventsDelay       time.Duration
+	pollingInterval     time.Duration
+	initialReadInterval time.Duration
+
+	fileSizeLimit int64
+}
+
+// Option used in New.
+type Option func(*opts)
```
kylelemons, line 336: Whenever you have an `Option` type, we've come to the conclusion that you should always have a `WithOptions(...Option) Option` helper too.  It makes currying zero or more options into one much easier and avoids the need to append options (and potentially end up with aliasing bugs that can be super hard to track down).

For example, you can do `WithOptions(defaults(), providedOptions...)` to easily combine the default values (expressed as options) with the user-provided overrides.  It also replaces the loop in the constructor, which is now just `var opt Options; WithOptions(defaults(), providedOptions...)(&opt)`

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+}
+
+// New creates a new FileWatcher.
+//
+// If the path is not available at the time of calling,
+// it blocks until the file becomes available, or context is cancelled,
+// whichever comes first.
+//
+// When logger is non-nil, it will be used to log errors,
```
kylelemons, line 402: outdated notes about logger?

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+			return nil, fmt.Errorf(
+				"filewatcher: context canceled while waiting for file(s) under %q to load: %w, last err: %w",
+				path,
+				ctx.Err(),
+				lastErr,
+			)
+		}
+
+		var err error
+		data, mtime, files, err = openAndParse(path, parser, opt.fileSizeLimit, hardLimit)
```
kylelemons, line 437: risk of [leaking undefined behavior](https://reddit.atlassian.net/wiki/spaces/BAS/pages/2355396743/Go+Style+Notes#Leaking-Undefined-Behavior)

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+		if errors.Is(err, fs.ErrNotExist) {
+			lastErr = err
+			time.Sleep(opt.initialReadInterval)
```
kylelemons, line 440: debug log?

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+			)
+		}
+	}
+
+	res := new(Result[T])
+	res.data.Store(&atomicData[T]{
+		data:  data,
+		mtime: mtime,
+	})
+	res.ctx, res.cancel = context.WithCancel(context.Background())
```
kylelemons, line 468: why not use the input ctx?  or at least have a WithBackgroundContext option?  you might not even need your close (unless people _want_ to check its error) of you let the context contol its lifecycle

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+		opt.pollingInterval,
+		opt.fsEventsDelay,
+	)
+
+	return res, nil
+}
+
+// NewMockFilewatcher returns a pointer to a new MockFileWatcher object
+// initialized with the given io.Reader and Parser.
+func NewMockFilewatcher[T any](r io.Reader, parser Parser[T]) (*MockFileWatcher[T], error) {
```
kylelemons, line 485: I would put this in filewatchertest rather than in the main file

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+	data   atomic.Value
+	parser Parser[T]
+}
+
+// Update updates the data of the MockFileWatcher using the given io.Reader and
+// the Parser used to initialize the FileWatcher.
+//
+// This method is not threadsafe.
+func (fw *MockFileWatcher[T]) Update(r io.Reader) error {
+	data, err := fw.parser(r)
```
kylelemons, line 507: this is a [fake, not a mock](https://martinfowler.com/bliki/TestDouble.html)

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+type MockDirWatcher[T any] struct {
+	data   atomic.Value
+	parser DirParser[T]
+}
+
+// NewMockDirWatcher creates a MockDirWatcher with the initial data and the
+// given DirParser.
+//
+// It provides Update function to update the data after it's been created.
+func NewMockDirWatcher[T any](dir fs.FS, parser DirParser[T]) (*MockDirWatcher[T], error) {
```
kylelemons, line 536: same two notes -- should be `filewatchertest.NewFakeDirWatcher`

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+		if err != nil {
+			slog.ErrorContext(r.ctx, "filewatcher: openAndParse returned error", "err", err)
+		} else {
+			r.data.Store(&atomicData[T]{
+				data:  d,
+				mtime: mtime,
+			})
+			// remove all previously watched files
+			for _, path := range watcher.WatchList() {
+				watcher.Remove(path)
```
konradreiche, line 141: Do we care about logging the error here at all? Though I feel like it'll make it just unnecessarily noisy.

filewatcher/v2/filewatcher.go
konradreiche: This took me some time but now I'm fully sold on mock vs fakes.

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+			// guarantee which case would be picked, so do an additional ctx check
+			// here to make sure we don't spam the log with i/o errors (which mainly
+			// happen in tests)
+			if r.ctx.Err() != nil {
+				continue
+			}
+
+			switch ev.Op {
+			default:
+				// Ignore uninterested events.
```
konradreiche, line 190: I think we only ignore `fsnotify.Chmod`. Should we?

filewatcher/v2/filewatcher_test.go
```diff
@@ -0,0 +1,612 @@
...
+	if err := os.Mkdir(dir, 0777); err != nil {
+		t.Fatalf("Failed to create directory %q: %v", dir, err)
+	}
+	var parserCalled atomic.Int64
+	parser := filewatcher.WrapDirParser(func(dir fs.FS) (map[string]string, error) {
+		parserCalled.Add(1)
+		m := make(map[string]string)
+		if err := fs.WalkDir(dir, ".", func(path string, de fs.DirEntry, err error) error {
+			if err != nil {
+				return nil
```
konradreiche, line 368:
```suggestion
				return err
```
Unless it's intentionally

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+	// DefaultMaxFileSize is the default MaxFileSize used when it's <= 0.
+	//
+	// It's 10 MiB, with hard limit multiplier of 10.
+	DefaultMaxFileSize  = 10 << 20
+	HardLimitMultiplier = 10
+)
+
+// FileWatcher loads and parses data from a file or directory, and watches for
+// changes in order to refresh its stored data.
+type FileWatcher[T any] interface {
```
mathyourlife-reddit, line 44: It could be useful for users to have access to some timing data such as the last time the data was reloaded successfully.

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+		o.fsEventsDelay = delay
+	}
+}
+
+// WithPollingInterval sets the interval to check file changes proactively.
+//
+// Default to DefaultPollingInterval.
+// To disable polling completely, set it to a negative value.
+//
+// Without polling filewatcher relies solely on fs events from the parent
```
mathyourlife-reddit, line 356:
```suggestion
// Without polling, filewatcher relies solely on fs events from the parent
```

filewatcher/v2/filewatcher.go
```diff
@@ -0,0 +1,567 @@
...
+// If the path is not available at the time of calling,
+// it blocks until the file becomes available, or context is cancelled,
+// whichever comes first.
+//
+// When logger is non-nil, it will be used to log errors,
+// either returned by parser or by the underlying file system watcher.
+// Please note that this does not include errors returned by the first parser
+// call, which will be returned directly.
+func New[T any](ctx context.Context, path string, parser Parser[T], options ...Option) (*Result[T], error) {
+	opt := opts{
```
mathyourlife-reddit, line 407: Could rewrite using the default options and then appending user provided options such as:

```
opt := opts{}
// Default options
optList := []Option{
	WithFSEventsDelay(DefaultFSEventsDelay),
	WithPollingInterval(DefaultPollingInterval),
	WithInitialReadInterval(DefaultInitialReadInterval),
	WithFileSizeLimit(DefaultMaxFileSize),
}
// Default options get overridden by user provided options when processed in order.
optList = append(optList, options...)
```

or using the `WithOptions` note above from @kylelemons 

```
WithOptions(
	// Default options
	WithFSEventsDelay(DefaultFSEventsDelay),
	WithPollingInterval(DefaultPollingInterval),
	WithInitialReadInterval(DefaultInitialReadInterval),
	WithFileSizeLimit(DefaultMaxFileSize),
	// User options
	WithOptions(options...),
)
```

filewatcher/v2/filewatcher_test.go
fishy: it is intentional. in the lambda of `WalkDir` when you return non-nil error you stops the walk.

filewatcher/v2/filewatcher.go
fishy: the timer here was created via `time.AfterFunc`, if you look at its doccomment, it says:
>The returned Timer's C field is not used and will be nil.

filewatcher/v2/filewatcher.go
mathyourlife-reddit: The parser hook seems like a reasonable option. Thanks.

konradreiche (approved):
Nice, also gave it a test as a replacement for v1 in one of our libraries. Everything is working as expected.

filewatcher/filewatcher.go
kylelemons: hmm, it recognizes value types of `sync.Mutex` as long as you don't copy the containing struct... this doesn't seem like it should be different.  Can you create the struct without copying?

fishy: non-pointer type copy-protected fields are appropriate when you just use their zero values in the constructors directly (e.g. locks, zero atomic values, etc.), for this, since it's a wrapper from v2, when constructing it you need to get the data from v2 and feed into this type, so a copy will always happen.

kylelemons (approved):

filewatcher/filewatcher.go
```diff
@@ -188,35 +189,35 @@ func New(ctx context.Context, cfg Config) (*Result, error) {
 	return &Result{result: result}, nil
 }
 
+// MockFileWatcher is an implementation of FileWatcher that does not actually read
```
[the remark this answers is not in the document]
fishy: this is v1, renaming would be a breaking change.

filewatcher/filewatcher.go
```diff
@@ -225,29 +226,29 @@ func (fw *MockFileWatcher) Stop() {}
 // MockDirWatcher is an implementation of FileWatcher for testing with watching
 // directories.
 type MockDirWatcher struct {
```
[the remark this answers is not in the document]
fishy: same here

filewatcher/v2/filewatcher.go
fishy: ok `dataAt` it is then :)

filewatcher/v2/filewatcher.go
pacejackson: Probably a good idea to add a comment about this, I could see someone reading this code later and being confused.

fishy: I think the problem is that both `time.NewTimer` and `time.AfterFunc` creates a `*time.Timer`, but `time.Timer.Reset` has this caveat if it's created via `time.NewTimer` that you should do the additional check to avoid race condition, but that `Reset` itself is wrong (you just must always use it with some additional checks in certain conditions).

so overall I think the better approach is to be very careful to `time.NewTimer` instead of being careful to `time.Timer.Reset`. in most cases, you use `time.NewTimer` to do something after the timer fires, which is better done via `time.AfterFunc` which is much safer. or some other cases it might be able to be replaced by `time.NewTicker`. I'm not sure if there are some cases you should use `time.NewTimer` over the other 2 (maybe there are some very niche cases).

fishy: added comments linking back to this thread.

PR #655, opened 2024-05-22
experiments: Switch to filewatcher/v2

fishy:
Note: secrets package is not switching because filewatcher.MockFileWatcher is part of its API so switching to v2 will be a breaking change for secrets package.


fishy:
Based on https://github.com/reddit/baseplate.go/pull/654

konradreiche (approved):

pacejackson (approved):
cooooool I didn't know about this v2 package

PR #657, opened 2024-06-11
kafkabp: Support AWS IMDS v2 for rack id

fishy:
Also modernize the code a bit (use sync.OnceValues, etc.).

### DIFF
--- a/kafkabp/rack.go
+++ b/kafkabp/rack.go
@@ -5,6 +5,7 @@ import (
 	"encoding"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"strings"
 	"sync"
@@ -88,7 +89,7 @@ func FixedRackID(id string) RackIDFunc {
 // Default values for SimpleHTTPRackIDConfig.
 const (
 	SimpleHTTPRackIDDefaultLimit   = 1024
-	SimpleHTTPRackIDDefaultTimeout = time.Second
+	SimpleHTTPRackIDDefaultTimeout = 1 * time.Second
 )
 
 // SimpleHTTPRackIDConfig defines the config to be used in SimpleHTTPRackID.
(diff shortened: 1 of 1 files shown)


kafkabp/rack.go
```diff
@@ -126,55 +127,104 @@ func SimpleHTTPRackID(cfg SimpleHTTPRackIDConfig) RackIDFunc {
...
 // Global cache for AWSAvailabilityZoneRackID.
-var (
-	awsCachedRackID string
-	awsRackIDOnce   sync.Once
-)
+var awsRackID = sync.OnceValues(func() (string, error) {
+	const (
+		// References:
+		// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
+		// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
```
kylelemons, line 189: 🔕 is there a reference for the IMDSv2 / token API to add?

kafkabp/rack.go
```diff
@@ -126,55 +127,104 @@ func SimpleHTTPRackID(cfg SimpleHTTPRackIDConfig) RackIDFunc {
...
+		tokenURL = "http://169.254.169.254/latest/api/token"
+		azURL    = "http://169.254.169.254/latest/meta-data/placement/availability-zone"
+
+		timeout   = time.Second
```
kylelemons, line 193: 🔕 I think units should always come with a scalar too for readability, even if it's `1`

```suggestion
		timeout   = 1*time.Second
```

kafkabp/rack.go
```diff
@@ -126,55 +127,104 @@ func SimpleHTTPRackID(cfg SimpleHTTPRackIDConfig) RackIDFunc {
...
+		readLimit = 1024
+	)
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	token, err := func(ctx context.Context) (string, error) {
```
kylelemons, line 200: Same comments as the `rack` API: not sure I am seeing the reason for the anonymous function, and that header should be normalized

kafkabp/rack.go
```diff
@@ -194,18 +244,18 @@ const awsAZurl = "http://169.254.169.254/latest/meta-data/placement/availability
...
-				nil, // delegate, let it fallback to DefaultWrapper
-				awsRackFailure,
-			),
-		})()
-	})
-	return awsCachedRackID
+	id, err := awsRackID()
+	if err != nil {
+		awsRackFailure.Inc()
+		slog.Warn("Failed to get AWS availability zone as rack id", "err", err)
```
kylelemons, line 257: 🔕 is v0 ready for slog?

kylelemons (approved):

kafkabp/rack.go
fishy: I think we want to gradually shift v0 to use slog directly, which will use ctxlog if services have ctxlog set up in their main (as v0 cannot use ctxlog directly)

PR #659, opened 2024-08-15
Update supported go versions to 1.22+1.23

fishy:


internal/admin/server_test.go
```diff
@@ -50,7 +50,6 @@ var expectedMetrics = []string{
 	"go_godebug_non_default_behavior_http2client_events_total",
 	"go_godebug_non_default_behavior_http2server_events_total",
 	"go_godebug_non_default_behavior_installgoroot_events_total",
-	"go_godebug_non_default_behavior_jstmpllitinterp_events_total",
```
fishy, line 53: this seems to be no longer in go 1.23.

tracing/hooks_test.go
```diff
@@ -19,7 +19,7 @@ type CallContainer struct {
 func (c *CallContainer) AddCall(call string, fail bool) error {
 	c.Calls = append(c.Calls, call)
 	if fail {
-		return fmt.Errorf(call)
+		return errors.New(call)
```
kylelemons, line 22: does go vet call this out now?

kylelemons (approved):

tracing/hooks_test.go
fishy: staticcheck calls this out now: https://github.com/reddit/baseplate.go/actions/runs/10407002603/job/28821319767#step:6:7

fishy: the main point I think is the dynamic string part, so a string literal would still work, and if you use something like `fmt.Errorf("%s", call)` that likely also works.

fishy: which I think makes sense, as there's a possibility `call` contains some format verbs and then `fmt.Errorf(call)` would cause an error.

PR #662, opened 2024-09-20
kafkabp: Add AWSAvailabilityZoneIDRackID rack implementation

fishy:


pacejackson (approved):

kafkabp/rack.go
```diff
@@ -251,11 +266,50 @@ var awsRackID = sync.OnceValues(func() (string, error) {
...
+//
+// See [1] for differences between AWS availability zone and availability zone
+// id.
+//
+// [1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones
+func AWSAvailabilityZoneIDRackID() string {
+	id, err := awsZoneIDRackID()
+	if err != nil {
+		awsRackFailure.Inc()
+		slog.Warn("Failed to get AWS availability zone id as rack id", "err", err)
```
kylelemons, line 311: 🔕 LogDNA will not show the error message to users when the error is moved into the structured portion of the log.  I typically recommend having the error as part of the human portion instead, since error is not really an interesting thing to filter against or aggregate.

kylelemons (approved):

kafkabp/rack.go
fishy: my experience is the opposite, that with errors in kv pair makes it easier to find exact errors (it's probably not that helpful if you want to search for partial error messages) on logdna, for use cases like "how many time did this error happen, did we have a spike during this time window?".

with a better logging ui (like GCP's log explorer) you can add an kv value to the summary line in the ui if needed, which is the best of both worlds, but we just don't have that unfortunately.

PR #663, opened 2024-09-23
Update thrift to v0.21.0

fishy:
Also rerun thrift compiler v0.21.0 on baseplate.thrift.

Also add thriftbp.ClientPoolConfig.UseZlib, which was not possible before.


go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/reddit/baseplate.go
 
-go 1.22
+go 1.22.0
```
mathyourlife-reddit, line 3: this change necessary?

mathyourlife-reddit (approved):
LGTM. [Go changes](https://github.com/apache/thrift/blob/master/CHANGES.md#0210) seem fairly limited.

go.mod
fishy: I believe it is, as thrift declared `1.22.0` in v0.21.0 https://github.com/apache/thrift/blob/v0.21.0/go.mod#L3, and only `1.22.0` is considered matching that, and `1.22` does not.

it won't be required once we update to 1.23 here, and in the next version of thrift I'll also remember to remove the patch version.

thriftbp/client_pool.go
```diff
@@ -228,6 +228,12 @@ type ClientPoolConfig struct {
 	//
 	// Optional. If empty, no "thrift-hostname" header will be sent.
 	ThriftHostnameHeader string `yaml:"thriftHostnameHeader"`
+
+	// Uses zlib transform in thrift THeader connections between client and
```
kylelemons, line 232: is THeader just for headers, or for the whole connection?  `UseZlib` sounds like this is for the whole connection, but the docs make it sound like it's just for headers...

so maybe
* rename to `UseZlibForHeaders`, or
* update the comment to specify that this is for the whole connection

?

kylelemons (approved):

thriftbp/client_pool.go
fishy: `THeader` is a thrift TProtocol/TTransport (it's just the only TProtocol/TTransport that supports headers, thus the name)

so it's for the whole connection.

konradreiche (approved):

konradreiche (approved):

PR #665, opened 2024-10-07
httpbp/thriftbp: Tracing middlewares v2 compat fix

fishy:
For the following 3 tracing middlewares:

- thriftbp server
- thriftbp client
- httpbp client

The current behavior is to short-circuit to v2 compat middleware if injected, or fallback to v0 tracing implementation if not injected. Change the fallback logic to no-op, as the v0 tracing implementation is not really working any more and it will spam logs.

For httpbp server tracing middleware, remove it from defaults, as the current v2 compat implementation is to skip it when injected as the API is not sufficient for the v2 compat implementation to work.

This replaces and closes https://github.com/reddit/baseplate.go/pull/664.

### DIFF
--- a/httpbp/client_middlewares.go
+++ b/httpbp/client_middlewares.go
@@ -12,14 +12,12 @@ import (
 	"time"
 
 	"github.com/avast/retry-go"
-	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/reddit/baseplate.go/breakerbp"
 	//lint:ignore SA1019 This library is internal only, not actually deprecated
 	"github.com/reddit/baseplate.go/internalv2compat"
 	"github.com/reddit/baseplate.go/retrybp"
-	"github.com/reddit/baseplate.go/tracing"
 	"github.com/reddit/baseplate.go/transport"
 )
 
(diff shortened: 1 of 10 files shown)


httpbp/client_middlewares.go
```diff
@@ -266,28 +264,16 @@ func MaxConcurrency(maxConcurrency int64) ClientMiddleware {
...
-
-			defer func() {
-				span.FinishWithOptions(tracing.FinishOptions{
-					Ctx: req.Context(),
-					Err: err,
-				}.Convert())
-			}()
-			return next.RoundTrip(req.WithContext(ctx))
-		})
+		// no-op
```
kylelemons, line 275: Is this always broken-broken, or could there be folks who have figured out how to get this working or for whom the semi-broken traces are still useful?

fishy: if you define "always" as "since we retired the infra for tracing v0", then yes this is always broken :)

fishy: the longer answer is that this worked for v0 tracing but never worked for v2, the v2 tracing is done via the injected middlewares from the compat library, and recently since we enabled interop in other places, the fallback to v0 implementation (when v2 middleware is not injected) start to spam the logs.

fishy:
Link to httpbp server middleware code:
https://github.com/reddit/baseplate.go/blob/c7e065837fc5906607a660629fb6f9cb08e3247d/httpbp/middlewares.go#L150-L156

httpbp/client_middlewares.go
kylelemons: Thanks.  Can we emit a warning if this case gets hit to let folks know that v0 tracing middleware is no longer functional?  I doubt anyone will see it, but it feels like the least we can do 😅

kylelemons (approved):
OK by me, but wait for the actual baseplate team to sign off.

httpbp/client_middlewares.go
fishy: the point of this PR is to _not_ spam the logs, if we log every time we fallback to v0, then that defeats the purpose :)

trevorriles (approved):

mathyourlife-reddit (approved):
Definitely on-board with removing the outdated tracing that's spamming logs.  To @kylelemons 's point, leaving a sync.Once single log message could help users action on gradually improving the the invalid configuration to add the the v2 logging.  Not blocking the merge on that though.

fishy:
@mathyourlife-reddit ok once logging makes sense. Added. PTAL.

mathyourlife-reddit:
> @mathyourlife-reddit ok once logging makes sense. Added. PTAL.

Awesome! Thanks!

mathyourlife-reddit (approved):

PR #666, opened 2024-12-06
Add support for Thrift and HTTP client-side fault injection.

HiramSilvey:
## 💸 TL;DR
This PR introduces client-side fault injection capability to HTTP and Thrift clients. The behavior is controlled via new `X-Bp-Fault` headers, and does nothing if they aren't present, are invalid, or don't match the outgoing request.

## 📜 Details
If clients are currently passing matching `X-Bp-Fault` headers around and they happen to be valid according to this change, then unintended fault injection could take place. This is extremely unlikely to occur given how specific the headers and their valid values are, but I wanted to call it out as technically it could affect requests unintentionally.

Thorough unit tests were added across the common library and protocol-specific code.

## ✅ Checks


faults/common.go
```diff
@@ -0,0 +1,124 @@
...
+	"math/rand/v2"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type GetHeaderFn func(key string) string
+type ResumeFn func() (interface{}, error)
+type ResponseFn func(code int, message string) (interface{}, error)
+type SleepFn func(d time.Duration)
```
fishy, line 16: 1. please write doccomments for exported types
2. for new code, please use `any` over `interface{}`.

faults/common.go
```diff
@@ -0,0 +1,124 @@
...
+type randSingleton struct {
+	randInt *int
+}
+
+func (r randSingleton) getRandInt() int {
+	if r.randInt == nil {
+		*r.randInt = rand.IntN(100)
+	}
+	return *r.randInt
+}
```
fishy, line 28: use [`sync.OnceValue`](https://pkg.go.dev/sync#OnceValue) instead?

fishy: the current implementation is definitely not thread-safe.

faults/common.go
```diff
@@ -0,0 +1,124 @@
...
+
+func isSelected(percentageHeader string, GetHeaderFn func(key string) string, singleRand randSingleton) (bool, string) {
+	percentageStr := GetHeaderFn(percentageHeader)
+	if percentageStr == "" {
+		return true, ""
+	}
+	percentage, err := strconv.Atoi(percentageStr)
+	if err != nil {
+		return false, fmt.Sprintf("provided percentage \"%s\" is not a valid integer", percentageStr)
```
fishy, line 37: 1. use `%q` over `\"%s\"`
2. also wrap the original error?
```suggestion
		return false, fmt.Sprintf("provided percentage %q is not a valid integer: %w", percentageStr, err)
```

fishy: oh you are returning a string instead of error for the error. I don't really see a reason for that, please return an error instead.

faults/common.go
```diff
@@ -0,0 +1,124 @@
+// Package faults provides common headers and client-side fault injection functionality.
+package faults
```
fishy, line 2: is this package meant to be internal? do we expect packages outside of baseplate.go v0 to use it?

faults/common.go
```diff
@@ -0,0 +1,124 @@
...
+	RandInt *int
+	SleepFn *SleepFn
+}
+
+func InjectFault(params InjectFaultParams) (interface{}, error) {
+	serverAddress := params.GetHeaderFn(FaultServerAddressHeader)
+	shortAddress := params.Address
+	if i := strings.Index(params.Address, ".svc.cluster.local"); i != -1 {
+		shortAddress = params.Address[:i]
+	}
```
fishy, line 65: use [`strings.CutSuffix`](https://pkg.go.dev/strings#CutSuffix) instead?

your current implementation will also cut it if there's something after `.svc.cluster.local`, I'm not sure if that's intentional or not (what could be after it and do we really want to cut it in that case?)

httpbp/client_middlewares.go
```diff
@@ -15,6 +16,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/reddit/baseplate.go/breakerbp"
+	"github.com/reddit/baseplate.go/faults"
+
```
fishy, line 20: nit: don't add this extra blank line.

httpbp/client_middlewares.go
```diff
@@ -349,3 +355,47 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
+			resp, err := faults.InjectFault(faults.InjectFaultParams{
+				CallerName:   "httpbp.FaultInjection",
+				Address:      req.URL.Host,
+				Method:       strings.TrimPrefix(req.URL.Path, "/"),
+				AbortCodeMin: 100,
+				AbortCodeMax: 599,
+				GetHeaderFn:  faults.GetHeaderFn(req.Header.Get),
+				ResumeFn:     resumeFn,
+				ResponseFn:   responseFn})
```
fishy, line 397: [one-per-line](https://github.com/reddit/baseplate.go/blob/master/Style.md#one-per-line-style):
```suggestion
				ResponseFn:   responseFn,
			})
```

httpbp/client_middlewares.go
```diff
@@ -349,3 +355,47 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
+			return resp.(*http.Response), err
```
fishy, line 398: make `ResumeFn`, `ResponseFn`, `InjectFault`, and `InjectFaultParams` all generic with `[T any]` so we don't need to do this typecasting.

httpbp/client_middlewares_test.go
```diff
@@ -395,3 +398,137 @@ func TestCircuitBreaker(t *testing.T) {
...
+				req.Header.Set(faults.FaultAbortMessageHeader, tt.faultAbortMessageHeader)
+			}
+			if tt.faultAbortPercentageHeader != "" {
+				req.Header.Set(faults.FaultAbortPercentageHeader, tt.faultAbortPercentageHeader)
+			}
+
+			resp, err := client.Do(req)
+
+			if err != nil {
+				t.Log(string(debug.Stack()))
```
fishy, line 526: is this really necessary? at this point the stack to cause error to be non-nil is already gone anyways.

httpbp/client_middlewares.go
```diff
@@ -349,3 +355,47 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
+					Request:          req,
+					TLS:              req.TLS,
+				}, nil
+			})
+
+			resp, err := faults.InjectFault(faults.InjectFaultParams{
+				CallerName:   "httpbp.FaultInjection",
+				Address:      req.URL.Host,
+				Method:       strings.TrimPrefix(req.URL.Path, "/"),
+				AbortCodeMin: 100,
```
fishy, line 393: shouldn't min be 400?

thriftbp/client_middlewares.go
```diff
@@ -390,6 +401,39 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
 	}
 }
 
+func FaultInjectionClientMiddleware(address string) thrift.ClientMiddleware {
+	return func(next thrift.TClient) thrift.TClient {
+		return thrift.WrappedTClient{
+			Wrapped: func(ctx context.Context, method string, args, result thrift.TStruct) (thrift.ResponseMeta, error) {
```
fishy, line 407: we should short circuit this middleware if `address == ""`?

fishy (changes requested):

faults/common.go
HiramSilvey: Done.

faults/common.go
HiramSilvey: Updated!

faults/common.go
HiramSilvey: The current implementation is correct: Everything including and after `.svc.cluster.local` should be removed. `CutSuffix` doesn't allow for this functionality when the values after `.svc.cluster.local` are not known -- i.e. port values. I added a comment to clarify this.

faults/common.go
HiramSilvey: My understanding is that this implementation is separate from any other baseplate.go implementation, so I intend to duplicate a lot of this code exactly once for use in all non-v0 baseplate.go. Please let me know if that is not the recommended approach here.

httpbp/client_middlewares.go
HiramSilvey: `gofmt` is actually insisting this line is added -- maybe because the import beneath it has a comment above it? IMO we should just leave it as-is since `gofmt` recommends it.

httpbp/client_middlewares.go
HiramSilvey: Done!

httpbp/client_middlewares.go
HiramSilvey: Done! PTAL.

httpbp/client_middlewares_test.go
HiramSilvey: Fair enough. I liked some of the extra info it gave me when my tests failed, but I don't have a strong preference either way. Removed.

httpbp/client_middlewares.go
HiramSilvey: Originally I was thinking this could be flexible to allow other response codes, but it really doesn't make sense given the actual response returned when a fault is injected. Thanks for raising this. Updated to 400.

faults/common.go
fishy: internal allows you to use this package inside bp.go v0 but forbids it from being used outside of bp.go v0.

httpbp/client_middlewares.go
HiramSilvey: Ack. I assume this is OK to leave as-is since `goimports` recommends it? Or do you still want me to remove the line?

faults/common.go
```diff
@@ -0,0 +1,140 @@
...
+
+	serverMethod := params.GetHeaderFn(FaultServerMethodHeader)
+	if serverMethod != "" && serverMethod != params.Method {
+		return params.ResumeFn()
+	}
+
+	singleRand := randSingleton{
+		randInt: params.RandInt,
+	}
```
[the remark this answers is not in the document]
HiramSilvey: Updated. I was trying to be clever by _not_ calling `rand.IntN` when a random number is never actually used (i.e. if the percentage header is malformed), but I think this is a case where my solution just ended up being over-engineering. I've updated it to be simpler and just generate the random int once per invocation as recommended.

faults/common.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (r randSingleton) getRandInt() int {
+	if r.randInt == nil {
+		*r.randInt = rand.IntN(100)
+	}
+	return *r.randInt
+}
+
+func isSelected(percentageHeader string, GetHeaderFn func(key string) string, singleRand randSingleton) (bool, error) {
```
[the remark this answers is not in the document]
HiramSilvey: Refactored.

faults/common.go
HiramSilvey: Updated.

faults/common.go
fishy: the point is that the API surface of this package is overly complicated that we are not really comfortable commit to not making any breaking changes in the future, we don't have a no-breaking-change commitment for internal packages.

httpbp/client_middlewares.go
fishy: that's in violation of https://github.com/reddit/baseplate.go/blob/master/Style.md#import-groups.

faults/common.go
HiramSilvey: Ack, thanks. Yeah, the plan was to keep it internal since it's just referenced by the httpbp and thriftbp bp.go v0 packages. My background is in C++ so I'm still catching up on some of these Go-isms. Thanks for poking here -- moved!

httpbp/client_middlewares.go
HiramSilvey: My mistake. Fixed.

internal/faults/common.go
```diff
@@ -0,0 +1,135 @@
...
+	Address, Method            string
+	AbortCodeMin, AbortCodeMax int
+
+	GetHeaderFn GetHeaderFn
+	ResumeFn    ResumeFn[T]
+	ResponseFn  ResponseFn[T]
+
+	// Exposed for tests.
+	RandInt *int
+	SleepFn *SleepFn
```
fishy, line 57: for tests you don't need to export them, just don't use external test.

thriftbp/client_middlewares.go
```diff
@@ -390,6 +400,44 @@ func PrometheusClientMiddleware(remoteServerSlug string) thrift.ClientMiddleware
...
+				if address == "" {
+					return next.Call(ctx, method, args, result)
+				}
+
+				var getHeaderFn faults.GetHeaderFn = func(key string) string {
```
fishy, line 411: nit:
```suggestion
				getHeaderFn := func(key string) string {
```

fishy: same for the other 2 lambdas and the httpbp middleware.

internal/faults/common.go
```diff
@@ -0,0 +1,135 @@
...
+	serverAddress := params.GetHeaderFn(FaultServerAddressHeader)
+
+	// The short address should just be <service>.<namespace>, without the
+	// local cluster suffix or port. Non-cluster-local addresses are not
+	// shortened.
+	shortAddress := params.Address
+	if i := strings.Index(params.Address, ".svc.cluster.local"); i != -1 {
+		shortAddress = params.Address[:i]
+	}
+	if serverAddress == "" || serverAddress != shortAddress {
```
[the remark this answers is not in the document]
HiramSilvey: No, because if the request is not cluster-local then the address won't contain `.svc.cluster.local`, but we still want to be able to match on it if the fault header address is the same.

internal/faults/common.go
```diff
@@ -0,0 +1,135 @@
...
+func InjectFault[T any](params InjectFaultParams[T]) (T, error) {
+	serverAddress := params.GetHeaderFn(FaultServerAddressHeader)
+
+	// The short address should just be <service>.<namespace>, without the
+	// local cluster suffix or port. Non-cluster-local addresses are not
+	// shortened.
+	shortAddress := params.Address
+	if i := strings.Index(params.Address, ".svc.cluster.local"); i != -1 {
+		shortAddress = params.Address[:i]
+	}
```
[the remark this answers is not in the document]
HiramSilvey: Done.

internal/faults/common.go
HiramSilvey: Done.

thriftbp/client_middlewares.go
HiramSilvey: Done.

internal/faults/common_test.go
```diff
@@ -0,0 +1,415 @@
...
+		{
+			name:    "cluster local address",
+			address: "testService.testNamespace.svc.cluster.local:12345",
+			want:    "testService.testNamespace",
+		},
+		{
+			name:    "external address",
+			address: "foo.bar:12345",
+			want:    "foo.bar:12345",
+		},
```
fishy, line 31: this does not sound right to me. in one of the case we removed the `.svc.cluster.local` suffix and removed the port from the address; in the other case we kept the port (because it does not have `.svc.cluster.local` suffix).

so this function is inconsistent regarding whether it removes the port. this goes back to my question: https://github.com/reddit/baseplate.go/pull/666#discussion_r1874117167

HiramSilvey: That's fair. After thinking about it more, if we are simulating a server being down entirely anyways, then stripping the port on external addresses should be done as well since we don't necessarily want to treat different ports differently. Updated to strip on all addresses. PTAL.

HiramSilvey:
FYI I just updated to remove the single random integer implementation altogether as per https://github.com/grpc/proposal/blob/master/A33-Fault-Injection.md#evaluate-possibility-fraction.

httpbp/client_middlewares.go
```diff
@@ -349,3 +354,43 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
 }
+
+func FaultInjection() ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+			resumeFn := func() (*http.Response, error) {
+				return next.RoundTrip(req)
+			}
+			responseFn := func(code int, message string) (*http.Response, error) {
+				return &http.Response{
```
fishy, line 365: I'm not sure constructing `*http.Response` by hand like this is a good idea. cc @pacejackson as you wrote the majority of the httpbp implementation 😅

fishy (approved):
lgtm but please wait for others' reviews.

httpbp/client_middlewares.go
pacejackson: I didn't write the any of the `Client` though 😅

pacejackson: I think that was @konradreiche

HiramSilvey: Yeah, I originally wanted to emulate a server a closely as possible via `ResponseWriter`, but it's not designed to be used in a way that would allow me to not _actually_ send something and just extract the constructed raw `http.Response` from it. This was the next best thing that I could determine, though let me know if you know of a less manual solution.

pacejackson: It looks like thats [what ResponseRecorder does](https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/net/http/httptest/recorder.go;l=188) so I would imagine it's fine?l

pacejackson (approved):

HiramSilvey:
@fishy @pacejackson thanks for the reviews. I just made 1 small change based on the GRPC fault injection implementation: short circuit the injected delay if the request context is cancelled. Ref: https://github.com/grpc/grpc-go/blob/38a8b9a7057218c210a599d79f3e1d07f84d18c5/xds/internal/httpfilter/fault/fault.go#L196

I changed it to exit the fault injection middleware and continue processing as per normal. If you could take a quick look to just make sure that's OK that would be appreciated. Thanks!

HiramSilvey:
> lgtm but please wait for others' reviews.

@fishy it looks like @kylelemons won't be available to review this PR until the new year -- should this be blocked until then or is it OK to proceed with the 2 approvals I have now? If there is someone else I should get a sooner review from instead, let me know. Thanks!

HiramSilvey:
@kylelemons PTAL when you have a chance. Thanks!

HiramSilvey:
The changes here were merged as a part of https://github.com/reddit/baseplate.go/pull/673.

PR #668, opened 2025-01-09
thriftbp: Tweak client pool connection reuse logic

fishy:
Only exceptions defined in thrift IDL is safe for reuse (getting them means the roundtrip is done fully and the response is read fully). All other errors have different degree of risks for reusing.

### DIFF
--- a/thriftbp/client_pool.go
+++ b/thriftbp/client_pool.go
@@ -711,22 +711,14 @@ func shouldCloseConnection(err error) bool {
 	var te thrift.TException
 	if errors.As(err, &te) {
 		switch te.TExceptionType() {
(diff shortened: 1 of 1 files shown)


mathyourlife-reddit (commented):
Has this change been tested in a service yet to get a feel for how many more connections will be closed?

fishy:
@mathyourlife-reddit 

> Has this change been tested in a service yet to get a feel for how many more connections will be closed?

No tests done, but the number should be small, and for the majority of the connections that's reused before and will not be reused now, they really should not be reused (reusing before is a bug)

See https://issues.apache.org/jira/browse/THRIFT-5845 for more context.

mathyourlife-reddit (approved):
approving. Interested to hear how it looks when it goes out.

konradreiche (approved):

PR #669, opened 2025-01-09
Headerbp

pacejackson:
Closes # 

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks


httpbp/client_middlewares.go
```diff
@@ -88,6 +89,11 @@ func NewClient(config ClientConfig, middleware ...ClientMiddleware) (*http.Clien
 	if config.CircuitBreaker != nil {
 		defaults = append([]ClientMiddleware{CircuitBreaker(*config.CircuitBreaker)}, defaults...)
 	}
+
+	// only add the middleware to forward baseplate headers if the client is for internal calls
+	if config.InternalOnly {
+		defaults = append(defaults, ClientHeaderBPMiddleware(config.Slug))
+	}
```
pacejackson, line 96: Since we don't have a distinct "internal vs non-internal" client type, I was thinking it would probably be good to have some sort of flag that you have to manually set to have HTTP clients automatically propagate headers?

internal/headerbp/headerbp.go
```diff
@@ -0,0 +1,355 @@
...
+// ErrNewInternalHeaderNotAllowed is returned by a client when the call tries to set an internal header is not allowlisted
+var ErrNewInternalHeaderNotAllowed = fmt.Errorf("cannot send new internal headers on requests")
+
+// IsBaseplateHeader returns true if the header is for baseplate and should be propagated
+func IsBaseplateHeader(key string) bool {
+	if len(key) < len(headerPrefixLower) {
+		return false
+	}
+	prefix := key[:len(headerPrefixLower)]
+	return prefix == headerPrefixLower || prefix == headerPrefixCanonicalHTTP
```
fishy, line 32: nit: just use:
```suggestion
	return strings.HasPrefix(key, headerPrefixLower) || strings.HasPrefix(key, headerPrefixCanonicalHTTP)
```
?

this way you no longer rely on `headerPrefixCanonicalHTTP` and `headerPrefixLower` being the same length.

internal/headerbp/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	clientNameLabel   = "client_name"
+	headerNameLabel   = "header_name"
+	rpcTypeLabel      = "rpc_type"
+	serverMethodLabel = "server_method"
+	serviceLabel      = "service_name"
+)
+
+var (
+	clientHeadersRejectedTotal = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
+		Name: "baseplate_client_headers_rejected_total",
```
fishy, line 21: nit: `_rejected_headers_total`, the last word before total should be the unit in plural form ("headers")

internal/headerbp/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	}, []string{
+		rpcTypeLabel,
+		serviceLabel,
+		clientNameLabel,
+		clientMethodLabel,
+		headerNameLabel,
+	})
+
+	clientHeadersSentTotal = promauto.With(prometheusbpint.GlobalRegistry).NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "baseplate_client_headers_sent_total",
```
fishy, line 32: same comment here, `_sent_headers_total`.

internal/headerbp/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+		Buckets: []float64{1, 4, 8, 16, 32, 64, 128},
+	}, []string{
+		rpcTypeLabel,
+		serviceLabel,
+		clientNameLabel,
+		clientMethodLabel,
+	})
+
+	clientHeadersSentSize = promauto.With(prometheusbpint.GlobalRegistry).NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "baseplate_client_headers_sent_size_total",
```
fishy, line 43: `_sent_size_bytes`, `_total` is only for counters (this is a histogram), and there the unit is bytes.

internal/headerbp/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+		Buckets: []float64{1, 64, 128, 256, 512, 1024, 2048, 4096},
+	}, []string{
+		rpcTypeLabel,
+		serviceLabel,
+		clientNameLabel,
+		clientMethodLabel,
+	})
+
+	serverHeadersReceivedTotal = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
+		Name: "baseplate_service_headers_received_total",
```
fishy, line 54: `_received_headers_total`.

internal/headerbp/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+		Help: "Total number of internal headers that were automatically extracted by a server",
+	}, []string{
+		rpcTypeLabel,
+		serviceLabel,
+		serverMethodLabel,
+		headerNameLabel,
+	})
+
+	serverHeadersReceivedSize = promauto.With(prometheusbpint.GlobalRegistry).NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "baseplate_server_headers_received_size_total",
```
fishy, line 64: `_received_size_bytes`.

thriftbp/server_middlewares.go
```diff
@@ -450,3 +453,50 @@ func PrometheusServerMiddleware(method string, next thrift.TProcessorFunction) t
...
+				}
+
+				// check both the lower case and the http canonical case since thrift.GetHeader is case sensitive
+				var untrusted bool
+				if _, ok := thrift.GetHeader(ctx, headerbp.IsUntrustedRequestHeaderLower); ok {
+					untrusted = true
+				} else if _, ok := thrift.GetHeader(ctx, headerbp.IsUntrustedRequestHeaderCanonicalHTTP); ok {
+					untrusted = true
+				}
+				if untrusted {
```
fishy, line 474: I wonder if we should add a `x-bp-` version of the deadline propagation header, so that we can still propagate deadline even in a untrusted chain of calls?

thriftbp/server_middlewares.go
```diff
@@ -450,3 +453,50 @@ func PrometheusServerMiddleware(method string, next thrift.TProcessorFunction) t
...
+						}
+					}
+					if cleared {
+						ctx = thrift.SetReadHeaderList(ctx, readHeaderList)
+					}
+					return next.Process(ctx, seqID, in, out)
+				}
+
+				headers := headerbp.NewIncomingHeaders(
+					headerbp.WithThriftService("", name),
```
fishy, line 491: nit: for readability:
```suggestion
					headerbp.WithThriftService("" /* service */, name),
```
(and I assume we use `""` for service because we don't really have a notion of service name in thrift servers?)

thriftbp/server_middlewares.go
pacejackson: > I wonder if we should add a x-bp- version of the deadline propagation header, so that we can still propagate deadline even in a untrusted chain of calls?

I'm not sure I understand, this specifically would discard that header if the request was marked as untrusted? 😅

internal/headerbp/headerbp.go
pacejackson: the `strings.HasPrefix` approach is significantly slower (relatively speaking)

BenchmarkIsBaseplateHeader
BenchmarkIsBaseplateHeader/startswith
BenchmarkIsBaseplateHeader/startswith-16         	299018905	         3.799 ns/op	       0 B/op	       0 allocs/op
BenchmarkIsBaseplateHeader/==
BenchmarkIsBaseplateHeader/==-16                 	1000000000	         0.2518 ns/op	       0 B/op	       0 allocs/op
PASS

pacejackson: > this way you no longer rely on headerPrefixCanonicalHTTP and headerPrefixLower being the same length.

They must always be the same since they are just the same string with different cases

fishy (approved):

pacejackson:
> the test passed for go 1.22 but failed for go 1.23, I think that just means the test is flaky? I don't think you used anything that will cause a difference between go 1.22 and 1.23.

It does seem like the test is flaky, but I think it's flaky because it can't call the test service I'm setting up. Is there something I'm missing there?

httpbp/server_test.go
```diff
@@ -427,3 +433,283 @@ func TestPanicRecovery(t *testing.T) {
...
+			httpbp.ServerHeaderBPMiddleware("originHTTPBPV0"),
+		},
+	})
+	if err != nil {
+		t.Fatalf("failed to create test originServer: %v", err)
+	}
+	t.Cleanup(func() {
+		originServer.Close()
+	})
+	go originServer.Serve()
```
fishy, line 565: you can probably add a small sleep after this `go` to give it time to actually start the server, not sure if it's worth it though 🤷 looks like the failed test is just very unlucky.

httpbp/client_middlewares.go
```diff
@@ -349,3 +355,26 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
 		})
 	}
 }
+
+// ClientHeaderBPMiddleware is a middleware that forwards baseplate headers from the context to the outgoing request.
+//
+// If it detects any new baseplate headers set on the request, it will reject the request and return an error.
+func ClientHeaderBPMiddleware(client string) ClientMiddleware {
```
konradreiche, line 362:
```suggestion
func ClientHeaderBaseplateMiddleware(client string) ClientMiddleware {
```
:no_bell: maybe best spelled out since we have `Baseplate` more often in function or method names than `BP`.

httpbp/client_middlewares.go
```diff
@@ -349,3 +355,26 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+			for k := range req.Header {
+				if err := headerbp.CheckClientHeader(k,
+					headerbp.WithHTTPClient("", client, ""),
+				); err != nil {
+					return nil, err
+				}
```
konradreiche, line 370:
```suggestion
			for key := range req.Header {
				if err := headerbp.CheckClientHeader(
				  key,
					headerbp.WithHTTPClient("", client, ""),
				); err != nil {
					return nil, err
				}
```

:no_bell: I find this slightly more readable :slightly_smiling_face:

httpbp/middlewares.go
```diff
@@ -517,3 +518,48 @@ func (rr *responseRecorder) WriteHeader(code int) {
...
 }
+
+type untrustedHeadersKey struct{}
+
+func setUntrustedHeaders(ctx context.Context, h map[string]string) context.Context {
+	return context.WithValue(ctx, untrustedHeadersKey{}, h)
+}
+
+func GetUntrustedBaseplateHeaders(ctx context.Context) (map[string]string, bool) {
+	h, ok := ctx.Value(untrustedHeadersKey{}).(map[string]string)
```
konradreiche, line 529: I'm a big fan of always checking if the assertion was successful, though I got the impression that if we tightly control it, it's okay to not check to make the API more user-friendly :thinking:

internalv2compat/compat.go
```diff
@@ -137,3 +140,11 @@ func V2TracingHTTPServerMiddleware() func(name string, next http.Handler) http.H
 	defer v2Tracing.Unlock()
 	return v2Tracing.httpServerMiddleware
 }
+
+func SetV0BaseplateHeadersToContext(ctx context.Context, headers map[string]string) context.Context {
+	return headerbp.ToContext(ctx, headers)
+}
+
+func SetV2BaseplateHeadersLookup(setter func(context.Context, map[string]string) context.Context) {
+	headerbp.SetV2Context = setter
```
konradreiche, line 149: One thing I noticed with our `internal2compat` primarily sets on structs local to this package but then makes it accessible through a separate getter. Would it make sense to define this locally here too and then call `V2Context()` to get the function outside of this package?

konradreiche (changes requested):
Only requesting changes for the open internal2compat question/discussion since it seems like it won't clear my review queue unless I approve or request changes :sweat_smile:

httpbp/client_middlewares.go
pacejackson: `headerbp` is the name of the actual library so I figured it was better to keep it consistent with that?

internalv2compat/compat.go
pacejackson: I see, that makes sense, updated.

httpbp/middlewares.go
pacejackson: I don't think that really does anything here though

konradreiche (approved):
:shipit:

PR #672, opened 2025-01-13
Upgrade go-redis from v8 to v9 

kumaramit01:
## 💸 TL;DR
This PR updates the go-redis library to v9. The v9 go-redis library is backward compatible with Redis 6 and supports Redis 7+. The existing v8 library doesn't support Redis 7. The v8 library fails for `cluster slots`, `ping` and error messages.

The managed client implementation has a dependency on v8 go-redis, that required some minor changes. The config options have changed in v9 specifically a new config `ConnMaxLifetime` that   replaces `MaxConnAge`. 

We have tested go-redis with following redis version
| Redis Version | go-redis version | Supported|
|---------------|-----------------|-----------|
|  6.2.13             |       8.x                 |   Yes.         |
|  7.0.7                |       8.x                 |   No.          |
|  7.2.3                |       8.x                 |   No.          |
|  6.2.13             |       9.x                 |   Yes.         |
|  7.0.7                |       9.x                 |   Yes.         |
|  7.2.3                |       9.x                 |   Yes.         |

## ✅ Checks

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,6 @@ require (
 	github.com/fsnotify/fsnotify v1.5.4
 	github.com/getsentry/sentry-go v0.11.0
 	github.com/go-kit/kit v0.9.0
-	github.com/go-redis/redis/v8 v8.11.5
 	github.com/gofrs/uuid v3.2.0+incompatible
 	github.com/google/go-cmp v0.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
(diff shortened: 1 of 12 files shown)


fishy (changes requested):
this is actually a breaking change, as code using it will need to change their usage of `github.com/go-redis/redis/v8` into `github.com/redis/go-redis/v9`.

in general we just consider all the packages under `github.com/reddit/baseplate.go/redis/` as deprecated and tell people to use the new internal redisbp instead. we just didn't mark them as deprecated as that will also break the existing code.

kumaramit01:
Discussed offline -the support for v9 is via internal baseplate that has breaking changes.

PR #673, opened 2025-01-16
[REFACTORED] Add support for Thrift and HTTP client-side fault injection.

HiramSilvey:
## 💸 TL;DR
NOTE: This is a refactor of the approved PR https://github.com/reddit/baseplate.go/pull/666 based on offline feedback. In addition to being a refactor, this also moves the fault injection middleware to be the final middleware for each protocol, as it is meant to simulate the upstream service being unavailable.

This PR introduces client-side fault injection capability to HTTP and Thrift clients. The behavior is controlled via new `X-Bp-Fault` headers, and does nothing if they aren't present, are invalid, or don't match the outgoing request.

## 📜 Details
If clients are currently passing matching `X-Bp-Fault` headers around and they happen to be valid according to this change, then unintended fault injection could take place. This is extremely unlikely to occur given how specific the headers and their valid values are, but I wanted to call it out as technically it could affect requests unintentionally.

Thorough unit tests were added across the common library and protocol-specific code.

## ✅ Checks

### DIFF
--- a/httpbp/client_middlewares.go
+++ b/httpbp/client_middlewares.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"net/http"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -16,6 +17,7 @@ import (
 
 	"github.com/reddit/baseplate.go/breakerbp"
 	"github.com/reddit/baseplate.go/internal/headerbp"
+
 	//lint:ignore SA1019 This library is internal only, not actually deprecated
 	"github.com/reddit/baseplate.go/internalv2compat"
 	"github.com/reddit/baseplate.go/retrybp"
@@ -54,7 +56,12 @@ func (f roundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
 //
 // * PrometheusClientMetrics
 //
+// * clientFaultMiddleware
+//
 // ClientErrorWrapper is included as transitive middleware through Retries.
+//
+// IMPORTANT: clientFaultMiddleware MUST be the last middleware as it simulates
+// faults as if they originated from the upstream server.
 func NewClient(config ClientConfig, middleware ...ClientMiddleware) (*http.Client, error) {
 	if err := config.Validate(); err != nil {
 		return nil, err
@@ -94,7 +101,12 @@ func NewClient(config ClientConfig, middleware ...ClientMiddleware) (*http.Clien
 	if config.InternalOnly {
 		defaults = append(defaults, ClientBaseplateHeadersMiddleware(config.Slug))
 	}
+
 	middleware = append(middleware, defaults...)
+
+	// ensure client fault middleware is applied last
+	clientFaultMiddleware := NewClientFaultMiddleware(config.Slug)
+	middleware = append(middleware, clientFaultMiddleware.Middleware())
 
 	return &http.Client{
 		Transport: WrapTransport(&httpTransport, middleware...),
@@ -375,6 +387,42 @@ func ClientBaseplateHeadersMiddleware(client string) ClientMiddleware {
 				headerbp.WithHeaderSetter(req.Header.Set),
 			)
 			return next.RoundTrip(req)
+		})
+	}
+}
+
+// Middleware returns a middleware that injects faults into the outgoing HTTP
+// requests based on the X-Bp-Fault header values.
+func (c clientFaultMiddleware) Middleware() ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+			resume := func() (*http.Response, error) {
+				return next.RoundTrip(req)
+			}
+
+			abort := func(code int, message string) (*http.Response, error) {
+				return &http.Response{
+					Status:     http.StatusText(code),
+					StatusCode: code,
+					Proto:      req.Proto,
+					ProtoMajor: req.ProtoMajor,
+					ProtoMinor: req.ProtoMinor,
+					Header: map[string][]string{
+						// Copied from the standard http.Error() function.
+						"Content-Type":           {"text/plain; charset=utf-8"},
+						"X-Content-Type-Options": {"nosniff"},
+					},
+					Body:             http.NoBody,
+					ContentLength:    0,
+					TransferEncoding: req.TransferEncoding,
+					Request:          req,
+					TLS:              req.TLS,
+				}, nil
+			}
+
+			address := req.URL.Hostname()
+			method := strings.TrimPrefix(req.URL.Path, "/")
+			return c.injector.InjectWithAbortOverride(req.Context(), address, method, httpHeaders{req}, resume, abort)
 		})
 	}
 }
--- a/httpbp/faults.go
+++ b/httpbp/faults.go
@@ -1,0 +1,41 @@
+package httpbp
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/reddit/baseplate.go/internal/faults"
+)
+
+type clientFaultMiddleware struct {
+	injector faults.Injector[*http.Response]
+}
+
+// NewClientFaultMiddleware creates and returns a new client-side fault
+// injection middleware.
+//
+// This middleware injects faults into the outgoing HTTP requests based on the
+// X-Bp-Fault header values. If valid X-Bp-Fault values exist which are not
+// intended to be interpreted by this middleware, then unintended faults could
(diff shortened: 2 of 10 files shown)


internal/faults/common.go
```diff
@@ -0,0 +1,278 @@
...
+	protocolLabel      = "fault_protocol"
+	successLabel       = "fault_success"
+	delayInjectedLabel = "fault_injected_delay"
+	abortInjectedLabel = "fault_injected_abort"
+)
+
+var (
+	TotalRequests = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
+		Namespace: promNamespace,
+		Name:      "fault_requests_total",
```
fishy, line 34: for new code we don't recommend using `Namespace` or `Subsystem`, as that makes it much harder to search for the code emitting a metric, just put `faultbp_fault_requests_total` in the `Name` instead.

internal/faults/common.go
```diff
@@ -0,0 +1,278 @@
...
+)
+
+const (
+	promNamespace      = "faultbp"
+	clientNameLabel    = "fault_client_name"
+	serviceLabel       = "fault_service"
+	methodLabel        = "fault_method"
+	protocolLabel      = "fault_protocol"
+	successLabel       = "fault_success"
+	delayInjectedLabel = "fault_injected_delay"
```
fishy, line 27: this feels like the cardinality would be quite high? do we really need this one?

httpbp/client_middlewares.go
```diff
@@ -349,3 +354,37 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
 }
+
+func (c clientFaultMiddleware) Middleware() ClientMiddleware {
+	return func(next http.RoundTripper) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+			resumeFn := func() (*http.Response, error) {
+				return next.RoundTrip(req)
+			}
+
+			faultFn := func(code int, message string) (*http.Response, error) {
```
konradreiche, line 365:
```suggestion
			resume := func() (*http.Response, error) {
				return next.RoundTrip(req)
			}

			fault := func(code int, message string) (*http.Response, error) {
```

I would try to avoid suffixing it with the type for the same reason we don't suffix every variable with the type :slightly_smiling_face:

internal/faults/common.go
```diff
@@ -0,0 +1,278 @@
...
+	Lookup(ctx context.Context, key string) (string, error)
+}
+
+// ResumeFn is the function type to continue processing the protocol-specific
+// request without injecting a fault.
+type ResumeFn[T any] func() (T, error)
+
+// FaultFn is the function type to inject a protocol-specific fault with the
+// given code and message.
+type FaultFn[T any] func(code int, message string) (T, error)
```
konradreiche, line 62: Similarly here, maybe `type Resume` and `type Fault` is as clear. At least Resume already describe an action. Where as Fault should maybe be named something with `Inject`?

internal/faults/common.go
```diff
@@ -0,0 +1,278 @@
+// Package faults provides common headers and client-side fault injection
```
konradreiche, line 1: A better name than `common.go` could be `faults.go`. It's a nice convention to describe the main entry point for a package and `common` or `util` tend to get overused as catch-all file.

internal/faults/common.go
```diff
@@ -0,0 +1,278 @@
...
+	serviceLabel       = "fault_service"
+	methodLabel        = "fault_method"
+	protocolLabel      = "fault_protocol"
+	successLabel       = "fault_success"
+	delayInjectedLabel = "fault_injected_delay"
+	abortInjectedLabel = "fault_injected_abort"
+)
+
+var (
+	TotalRequests = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
```
konradreiche, line 32: Unexport unless needed outside of the package, which it probably shouldn't :slightly_smiling_face:

internal/faults/common.go
```diff
@@ -0,0 +1,278 @@
...
+			}
+
+			totalReqsCounter(true, true).Inc()
+			return faultFn(code, abortMessage)
+		}
+	}
+
+	totalReqsCounter(true, false).Inc()
+	return resumeFn()
+}
```
konradreiche, line 272: This function appears to be too long to easily glance over it. Can portions of it be factored into helper functions instead?

konradreiche (changes requested):

internal/faults/common.go
HiramSilvey: Done!

internal/faults/common.go
HiramSilvey: This is just a bool which informs whether any delay was added, so the cardinality should be very small.

internal/faults/common.go
HiramSilvey: Updated to `Resume` and `Abort`.

httpbp/client_middlewares.go
HiramSilvey: Done!

internal/faults/common.go
HiramSilvey: Done.

internal/faults/common.go
HiramSilvey: Done.

fishy (approved):

httpbp/client_middlewares.go
```diff
@@ -349,3 +354,37 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
 		})
 	}
 }
+
+func (c clientFaultMiddleware) Middleware() ClientMiddleware {
```
konradreiche, line 358: Can we add method comments to match the amount of documentation to other middlewares, though maybe with the bulk of the information on the `NewClientFaultMiddleware` constructor? Since you've already provided great detail in the description:

> If clients are currently passing matching X-Bp-Fault headers around and they happen to be valid according to this change, then unintended fault injection could take place. This is extremely unlikely to occur given how specific the headers and their valid values are, but I wanted to call it out as technically it could affect requests unintentionally.

httpbp/faults.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/reddit/baseplate.go/internal/faults"
+)
+
+type clientFaultMiddleware struct {
+	injector faults.Injector[*http.Response]
+}
+
+// NewClientFaultMiddleware creates and returns a new client-side fault
+// injection middleware.
+func NewClientFaultMiddleware(clientName string) clientFaultMiddleware {
```
konradreiche, line 16: Maybe here is the best place to add most of the documentation?

internal/faults/faults.go
```diff
@@ -0,0 +1,277 @@
...
+
+// Global variable to allow overriding in tests.
+var selected = func(percentage int) bool {
+	// Use a different random integer per feature as per
+	// https://github.com/grpc/proposal/blob/master/A33-Fault-Injection.md#evaluate-possibility-fraction.
+	return rand.IntN(100) < percentage
+}
+
+// Global variable to allow overriding in tests.
+var sleep = func(ctx context.Context, d time.Duration) error {
```
konradreiche, line 105: I don't know if we have another test that does something similar, but overall I would try to avoid adding global state for testing or adding code to the non-test file to make it testable for that matter.

It can make some of the code or tests harder to understand, but also potentially lead to flaky tests when it comes to parallel execution.

Would it be possible to move the function into the `Injector` type?

internal/faults/faults.go
```diff
@@ -0,0 +1,277 @@
...
+		return ctx.Err()
+	}
+	return nil
+}
+
+// Injector contains the data common across all requests needed to inject
+// faults on outgoing requests.
+type Injector[T any] struct {
+	clientName, callerName     string
+	abortCodeMin, abortCodeMax int
```
konradreiche, line 120: Formatting nit: each field on its own line unless I didn't see this in our style guide :eyes:

konradreiche (approved):
Looks great, a couple more asks but nothing blocking as far as I can tell :slightly_smiling_face:

internal/faults/faults.go
HiramSilvey: Done! This was a vestige from some iterations of the code. I agree moving it to the Injector type is much cleaner.

internal/faults/faults.go
HiramSilvey: oops! fixed

httpbp/client_middlewares.go
HiramSilvey: done!

PR #678, opened 2025-02-06
add headerbp signature validation/minting

pacejackson:
## 💸 TL;DR

We ran into some issues trying to rely just on the X-Rddt-Untrusted header being set on ingress and are switching over to having a signature on these headers for HTTP requests.

This does include some small breaking changes, but this had not been rolled out to any services yet.

### DIFF
--- a/headerbp/signing.go
+++ b/headerbp/signing.go
@@ -2,6 +2,7 @@ package headerbp
 
 import (
 	"bytes"
+	"cmp"
 	"context"
 	"fmt"
 	"slices"
@@ -18,7 +19,7 @@ const delimiter = rune(0)
 
 var signatureVersionPrefix = strconv.Itoa(signatureVersion)
 
-var ErrInvalidSignatureVersion = fmt.Errorf("invalid  version")
+var ErrInvalidSignatureVersion = fmt.Errorf("invalid version")
 
 type headerSignatureContextKey struct{}
 
@@ -67,14 +68,8 @@ func concatHeaders(b *bytes.Buffer, headerNames []string, getHeader func(string)
 			original:   k,
 		})
 	}
-	slices.SortFunc(manifest, func(i, j manifestEntry) int {
-		if i.normalized < j.normalized {
-			return -1
-		} else if i.normalized > j.normalized {
-			return 1
-		} else {
-			return 0
-		}
+	slices.SortFunc(manifest, func(left, right manifestEntry) int {
+		return cmp.Compare(left.normalized, right.normalized)
 	})
 
 	b.WriteString(versionPrefix)
@@ -109,12 +104,11 @@ func SignHeaders(
 	defer putBuffer(b)
 
 	concatHeaders(b, headerNames, getHeader, signatureVersionPrefix)
-	signature, err := signing.Sign(
-		signing.SignArgs{
-			Message:   b.Bytes(),
-			Secret:    signingSecret,
-			ExpiresIn: 5 * time.Minute,
-		})
+	signature, err := signing.Sign(signing.SignArgs{
+		Message:   b.Bytes(),
+		Secret:    signingSecret,
+		ExpiresIn: 5 * time.Minute,
+	})
 	if err != nil {
 		return "", fmt.Errorf("generating signature: %w", err)
 	}
--- a/headerbp/signing.go
+++ b/headerbp/signing.go
@@ -22,7 +22,7 @@ var ErrInvalidSignatureVersion = fmt.Errorf("invalid  version")
 
 type headerSignatureContextKey struct{}
 
-func setHeaderSignature(ctx context.Context, sig string) context.Context {
+func setSignatureOnContext(ctx context.Context, sig string) context.Context {
 	return context.WithValue(ctx, headerSignatureContextKey{}, sig)
 }
 
@@ -175,5 +175,5 @@ func VerifyHeaders(
 		return ctx, fmt.Errorf("verification error: %w", err)
 	}
 	ctx = setV2SignatureContext(ctx, signature)
-	return setHeaderSignature(ctx, signature), nil
+	return setSignatureOnContext(ctx, signature), nil
 }
--- a/headerbp/signing.go
+++ b/headerbp/signing.go
@@ -1,0 +1,179 @@
+package headerbp
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"slices"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/reddit/baseplate.go/secrets"
+	"github.com/reddit/baseplate.go/signing"
+)
+
+const delimiter = rune(0)
+
+var signatureVersionPrefix = strconv.Itoa(signatureVersion)
+
+var ErrInvalidSignatureVersion = fmt.Errorf("invalid  version")
+
+type headerSignatureContextKey struct{}
+
+func setHeaderSignature(ctx context.Context, sig string) context.Context {
+	return context.WithValue(ctx, headerSignatureContextKey{}, sig)
+}
+
+// HeaderSignatureFromContext gets the header signature from the context. This can be used in client middleware to propagate the
+// signature along with the headers if they are unchanged by the request.
+func HeaderSignatureFromContext(ctx context.Context) (string, bool) {
+	sig, ok := ctx.Value(headerSignatureContextKey{}).(string)
+	return sig, ok
+}
+
+var messageBufferPool = sync.Pool{
+	New: func() interface{} {
+		return new(bytes.Buffer)
+	},
+}
+
+func getBuffer() *bytes.Buffer {
+	buf := messageBufferPool.Get().(*bytes.Buffer)
+	buf.Reset()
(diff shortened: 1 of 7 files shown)


httpbp/client_middlewares.go
```diff
@@ -370,21 +372,54 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
-				headerbp.WithHeaderSetter(req.Header.Set),
+				headerbp.WithHeaderSetter(func(key, value string) {
+					if !hasSignature {
+						baseplateHeaders = append(baseplateHeaders, key)
+					}
+					req.Header.Set(key, value)
+				}),
 			)
+			// the !hasSignature check is redundant since we do not add to the baseplateHeaders list unless it is false, but
+			// it's here to make it clear that we only need to update the signature if there is no signature in the context
```
mathyourlife-reddit, line 412: 👍

httpbp/client_middlewares.go
```diff
@@ -370,21 +372,54 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
+			if len(baseplateHeaders) > 0 && !hasSignature {
+				_signature, err := headerbp.SignHeaders(req.Context(), *signingSecret, baseplateHeaders, req.Header.Get)
+				if err != nil {
+					return nil, fmt.Errorf("signing baseplate headers: %w", err)
+				}
+				signature = _signature
```
mathyourlife-reddit, line 418:
```suggestion
				if _signature, err := headerbp.SignHeaders(req.Context(), *signingSecret, baseplateHeaders, req.Header.Get); err != nil {
					return nil, fmt.Errorf("signing baseplate headers: %w", err)
				} else {
					signature = _signature
				}
```

🔕

httpbp/middlewares.go
```diff
@@ -530,14 +531,43 @@ func GetUntrustedBaseplateHeaders(ctx context.Context) (map[string]string, bool)
 	return h, ok
 }
 
+type SecretsStore interface {
```
mathyourlife-reddit, line 534: godoc

mathyourlife-reddit (approved):

headerbp/signing.go
```diff
@@ -0,0 +1,179 @@
...
+
+	"github.com/reddit/baseplate.go/secrets"
+	"github.com/reddit/baseplate.go/signing"
+)
+
+const delimiter = rune(0)
+
+var signatureVersionPrefix = strconv.Itoa(signatureVersion)
+
+var ErrInvalidSignatureVersion = fmt.Errorf("invalid  version")
```
fishy, line 21: nit: you have 2 spaces here (between `invalid` and `version`)

headerbp/signing.go
```diff
@@ -0,0 +1,179 @@
...
+		})
+	}
+	slices.SortFunc(manifest, func(i, j manifestEntry) int {
+		if i.normalized < j.normalized {
+			return -1
+		} else if i.normalized > j.normalized {
+			return 1
+		} else {
+			return 0
+		}
```
fishy, line 77: nit: this can be simplified into:
```suggestion
		return cmp.Compare(i.normalized, j.normalized)
```
(note: the stdlib cmp: https://pkg.go.dev/cmp)

also nit: for `slices` package usages I would use `left`/`right` over `i`/`j` as they are no longer indices, and `left`/`right` makes it much more obvious when you need to swap them for reverse order.

headerbp/signing.go
```diff
@@ -0,0 +1,179 @@
...
+	b := getBuffer()
+	defer putBuffer(b)
+
+	concatHeaders(b, headerNames, getHeader, signatureVersionPrefix)
+	signature, err := signing.Sign(
+		signing.SignArgs{
+			Message:   b.Bytes(),
+			Secret:    signingSecret,
+			ExpiresIn: 5 * time.Minute,
+		})
```
fishy, line 117: nit:
```suggestion
		},
	)
```

headerbp/signing.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return "", fmt.Errorf("generating signature: %w", err)
+	}
+
+	b.Reset()
+	b.WriteString(signatureVersionPrefix)
+	b.WriteRune('.')
+	b.WriteString(signature)
+	return b.String(), nil
```
fishy, line 126: for this usage (that you are producing a string at the end), `strings.Builder` actually performs better than `bytes.Buffer`, as `bytes.Buffer.String()` would need an O(N) copy to convert from `[]byte` to `string`.

with `strings.Builder` you no longer need a pool (`strings.Builder.Reset` basically discards everything and creates a new buffer, a side-effect of avoiding the O(N) copy).

headerbp/signing.go
```diff
@@ -0,0 +1,179 @@
...
+		return ctx, fmt.Errorf("%w: %w", ErrInvalidSignatureVersion, err)
+	}
+	if components.version != 1 {
+		return ctx, fmt.Errorf("%w: unsupported version number %d", ErrInvalidSignatureVersion, components.version)
+	}
+
+	b := getBuffer()
+	defer putBuffer(b)
+	concatHeaders(b, headerNames, getHeader, components.versionPrefix)
+	if err := signing.Verify(b.Bytes(), components.signature, verificationSecret); err != nil {
```
fishy, line 174: oh but you are also using `Bytes`, in that case I guess it makes sense to use `bytes.Buffer` with a pool.

(or you can use an interface to make `concatHeaders` accept both `*bytes.Buffer` and `*strings.Builder`, but  that's probably too overkill 😅)

httpbp/client_middlewares.go
```diff
@@ -370,21 +372,54 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
 // ClientBaseplateHeadersMiddleware is a middleware that forwards baseplate headers from the context to the outgoing request.
 //
 // If it detects any new baseplate headers set on the request, it will reject the request and return an error.
-func ClientBaseplateHeadersMiddleware(client string) ClientMiddleware {
+func ClientBaseplateHeadersMiddleware(client string, store SecretsStore, path string) ClientMiddleware {
+	getSigningSecret := func() *secrets.VersionedSecret {
+		secret, err := store.GetVersionedSecret(path)
+		if err != nil {
+			log.Errorf("Failed to get secret %q: %v", path, err)
```
fishy, line 379: for new code I'd probably prefer to avoid using the `log` library and use `slog.Error` directly instead. the adapter we use to use slog as the backend of zap have to add extra callstack layers making callsite and callstacks much less accurate.

fishy (approved):

headerbp/signing.go
pacejackson: Ya, I'm really just using it because we already have it. Switching to a `strings.Builder` adds an additional allocation over just re-using the buffer we already grabbed.

httpbp/client_middlewares.go
```diff
@@ -370,21 +371,58 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
...
 // If it detects any new baseplate headers set on the request, it will reject the request and return an error.
-func ClientBaseplateHeadersMiddleware(client string) ClientMiddleware {
+func ClientBaseplateHeadersMiddleware(client string, store SecretsStore, path string) ClientMiddleware {
+	getSigningSecret := func() *secrets.VersionedSecret {
+		secret, err := store.GetVersionedSecret(path)
+		if err != nil {
+			slog.Error(
+				"Failed to get secret:",
+				slog.String("path", path),
+				slog.String("err", err.Error()),
```
fishy, line 381: nit: slog does not force you to convert it to `slog.Attr` first, you can just use:
```suggestion
				"err", err,
```
or if you want to convert it to `slog.Attr` first, use `Any` over `String` for errors:
```suggestion
				slog.Any("err", err),
```
the reasons are:
1. if `err` is nil `err.Error` would panic (won't happen here though)
2. `err` might actually implemented [`slog.LogValuer`](https://pkg.go.dev/log/slog#LogValuer) to log more useful info via slog than `err.Error()` produces (we did implement that for all thrift idl exceptions, for example)

same for the other place.

httpbp/client_middlewares.go
```diff
@@ -370,21 +371,58 @@ func PrometheusClientMetrics(serverSlug string) ClientMiddleware {
 // ClientBaseplateHeadersMiddleware is a middleware that forwards baseplate headers from the context to the outgoing request.
 //
 // If it detects any new baseplate headers set on the request, it will reject the request and return an error.
-func ClientBaseplateHeadersMiddleware(client string) ClientMiddleware {
+func ClientBaseplateHeadersMiddleware(client string, store SecretsStore, path string) ClientMiddleware {
+	getSigningSecret := func() *secrets.VersionedSecret {
+		secret, err := store.GetVersionedSecret(path)
+		if err != nil {
+			slog.Error(
+				"Failed to get secret:",
```
fishy, line 379: also nit: for structured logging the `:` is less useful 😅
```suggestion
				"Failed to get secret",
```

fishy (approved):

PR #681, opened 2025-02-14
Add CredentialSecret methods

max-melentyev:
So that it can be used with an interface


pacejackson (commented):
For some context, making these have an interface makes me worried that we'll start passing them around when we really should be passing around loaders. The Secret objects themselves don't ever update, you have to re-load them each time to get updates.

PR #682, opened 2025-02-14
Update for go 1.24

fishy:
- Update go.mod to require 1.23+
- Update CI to run on go 1.23 & 1.24

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/reddit/baseplate.go
 
-go 1.22.0
+go 1.23
 
 require (
 	github.com/Shopify/sarama v1.29.1
--- a/internal/admin/server_test.go
+++ b/internal/admin/server_test.go
@@ -57,7 +57,6 @@ var expectedMetrics = []string{
 	"go_godebug_non_default_behavior_randautoseed_events_total",
 	"go_godebug_non_default_behavior_tarinsecurepath_events_total",
 	"go_godebug_non_default_behavior_tlsmaxrsasize_events_total",
-	"go_godebug_non_default_behavior_x509sha1_events_total",
 	"go_godebug_non_default_behavior_x509usefallbackroots_events_total",
 	"go_godebug_non_default_behavior_zipinsecurepath_events_total",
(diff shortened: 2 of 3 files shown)


internal/admin/server_test.go
```diff
@@ -57,7 +57,6 @@ var expectedMetrics = []string{
 	"go_godebug_non_default_behavior_randautoseed_events_total",
 	"go_godebug_non_default_behavior_tarinsecurepath_events_total",
 	"go_godebug_non_default_behavior_tlsmaxrsasize_events_total",
-	"go_godebug_non_default_behavior_x509sha1_events_total",
```
fishy, line 60: this is no longer available in go 1.24.

@kylelemons I think we should remove all the `godebug` ones here unless we really care about some of them.

mathyourlife-reddit: I don't see internal references to them in CS, and v2 only checks for a small subset of these for regressions.

mathyourlife-reddit (approved):

internal/admin/server_test.go
fishy: yes, every time we need to update go minor version this test will break, and I'll make the same suggestion of only check for a subset that we really care about, and I believe there are always some objections to that suggestion.

kylelemons (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/reddit/baseplate.go
 
-go 1.22.0
+go 1.23
```
kylelemons, line 3: 🔕 I think it wants 1.23.0 nowadays... I've run into a few issues with totally dropping the third dot, e.g. when things parse it out and expect e.g. `GOTOOLCHAIN=<that>` to work

kylelemons (approved):

go.mod
fishy: do you have an example of dropping patch release from `go` line breaks things?

fishy: will merge this as-is. we can add `.0` if this really breaks something (I don't think it will)

PR #683, opened 2025-02-18
remove the thrift headerbp client middleware from defaults

pacejackson:
## 💸 TL;DR
There seems to be something wrong with this where it's picking up headers that are being forwarded as ones that are being manually set by the caller.


ben-mays-rdt:
This is deployed out to 400 repos so it's probably not the happy path that triggered it. I suspect perhaps concurrency bug or re-entrant middleware loop in the client? We can debug as part of postmortem.

fishy (approved):

PR #684, opened 2025-02-19
Make headerbp client middleware safer

pacejackson:
After some investigation, we figured out that the cause of the error we encountered was due to the service wiring up the default client middleware manually and using the `NewBaseplateThriftPool` constructor, resulting in the default middleware running twice. This means that on the first pass, the middleware would add the header to the request. The second pass would then see this and fail the request. I made two changes to make this safer:

1. Added an idempotency check to the client middleware so it only runs once per call. Add an error log and a metric when we detect this happening.
2. Changed the client middlewares to instead drop new baseplate headers and log an error message rather than failing the call.

### DIFF
--- a/headerbp/headerbp.go
+++ b/headerbp/headerbp.go
@@ -3,6 +3,7 @@ package headerbp
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"strings"
 	"sync"
 )
@@ -127,8 +128,8 @@ func setHeadersOnContext(ctx context.Context, headers map[string]string) context
 	return context.WithValue(ctx, headersKey{}, headers)
 }
 
-// CheckClientHeader checks if the header is allowlisted and returns an error if it is not.
-func CheckClientHeader(name string, options ...CheckClientHeaderOption) error {
+// ShouldRemoveClientHeader checks if the header is allowlisted and returns if the header should be removed
+func ShouldRemoveClientHeader(name string, options ...CheckClientHeaderOption) bool {
 	cfg := &checkClientHeaders{}
 	WithCheckClientHeaderOptions(options...).ApplyToCheckClientHeaders(cfg)
 
@@ -140,9 +141,13 @@ func CheckClientHeader(name string, options ...CheckClientHeaderOption) error {
 			cfg.Method,
 			strings.ToLower(name),
 		).Inc()
-		return fmt.Errorf("%w, %q is not allowlisted", ErrNewInternalHeaderNotAllowed, name)
+		slog.Error(
+			"client header rejected",
+			"header", name,
+		)
+		return true
 	}
-	return nil
+	return false
 }
 
 type CheckClientHeaderOption interface {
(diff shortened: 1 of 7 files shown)


headerbp/metrics.go
```diff
@@ -17,6 +17,14 @@ const (
 )
 
 var (
+	clientMiddlewareIdempotencyCheckTotal = promauto.With(prometheusbpint.GlobalRegistry).NewCounterVec(prometheus.CounterOpts{
+		Name: "baseplate_client_middleware_idempotency_check_total",
```
mathyourlife-reddit, line 21: We might want to make this for other middleware as well. How about adding a "middleware" metric label?

headerbp/headerbp.go
```diff
@@ -319,14 +360,36 @@ func WithHeaderSetter(setter func(key, value string)) SetOutgoingHeadersOption {
...
 }
 
+type setOutgoingIdempotencyKey struct{}
+
+func HasSetOutgoingHeaders(ctx context.Context, options ...HasSetOutgoingHeadersOption) bool {
+	cfg := &hasSetOutgoingHeaders{}
+	WithHasSetOutgoingHeadersOptions(options...).ApplyToHasSetOutgoingHeaders(cfg)
+	hasSet, ok := ctx.Value(setOutgoingIdempotencyKey{}).(bool)
+	hasSet = hasSet && ok
+	if hasSet {
```
fishy, line 370: nit: as long as you are using two receiver type casting (which avoids panicking if the type mismatches), you don't have to check `ok`, as when `ok` is false (the type is not a bool), `hasSet` will always be false (the zero value of bool). so this can be simplified into:
```suggestion
	if hasSet, _ := ctx.Value(setOutgoingIdempotencyKey{}).(bool); hasSet {
```

headerbp/metrics.go
```diff
@@ -17,6 +17,15 @@ const (
...
```
fishy, line 21: this is a metrics inside package `headerbp`, but the name is a bit too broad for what it does. I'd probably rename it into (`checks` also should be in plural form):
```suggestion
		Name: "baseplate_headerbp_client_middleware_idempotency_checks_total",
```
but also, this is only emitted when check fails, not when you checks (you don't emit it when check passes), so the "unit" should not be "checks" but something like "failures":
```suggestion
		Name: "baseplate_headerbp_client_middleware_idempotency_check_failures_total",
```
re @mathyourlife-reddit: none of the other middlewares will cause harm when added twice (prometheus metrics can cause double counter when added twice, but that's the limit of what "harm" they can cause, and it's possible to add prometheus metric middleware multiple times with different names, for example with retries), so I don't foresee we'll add idempotency check to other middlewares.

mathyourlife-reddit: no other current middleware cause issues if added twice (except for prometheus), but others in the future might.  Having a more general metric with the label gives us some flexibility in alerting/graphs instead of needing to duplicate in the future.

pacejackson: I'm gonna keep it more generic but add `_failures`

fishy (approved):

headerbp/metrics.go
fishy: 🔕 with a generic name we probably should define this metric in a common package then.

but we don't currently have a common package appropriate for this, so I guess leave it here for now is fine, we can move it to a different package later if we ever need to use it for something not related to headerbp.

PR #686, opened 2025-02-20
add support for thrift service name in server traces

marcoferrer:
Closes # 

## 💸 TL;DR

Right now server trace spans for thrift report the service name as unknown. Because the thrift compiler doesn't generate schema metadata, we can fallback to relying on compiler conventions to source the service name defined in the schema.

## 📜 Details
[Design Doc]()

[Jira]()

## ✅ Checks

### DIFF
--- a/thriftbp/server.go
+++ b/thriftbp/server.go
@@ -2,6 +2,10 @@ package thriftbp
 
 import (
 	"context"
+	"fmt"
+	"path/filepath"
+	"reflect"
+	"strings"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
@@ -143,6 +147,7 @@ func NewBaseplateServer(
 	middlewares := BaseplateDefaultProcessorMiddlewares(
 		DefaultProcessorMiddlewaresArgs{
 			EdgeContextImpl:     bp.EdgeContextImpl(),
+			ServiceName:         GetThriftServiceName(cfg.Processor),
 			ErrorSpanSuppressor: cfg.ErrorSpanSuppressor,
 		},
 	)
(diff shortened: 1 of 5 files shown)


thriftbp/server.go
```diff
@@ -190,3 +195,19 @@ var (
...
+	if t == nil {
+		return "unknown"
+	}
+	value := t.Elem()
+	if value == nil {
+		return "unknown"
+	}
+
+	return fmt.Sprintf("%s.%s",
+		filepath.Base(value.PkgPath()),
```
fishy, line 210: why do we use `filepath` here? `filepath` is the one that uses `/` for *nix and `\` for windows, I don't think that's how go package path works. it should always be `/` (as go package path is kind of url based) so we should use `path` instead.

also do you have an example/test what this is supposed to get? I'm not sure why we infer this from go package path instead of set explicitly.

fishy (changes requested):

thriftbp/server.go
marcoferrer: Good call on filepath. Ill push that fix. I added a test to show what the expected output is. Im pulling the last part of the go path because that is usually defined by either go package set in the thrift file or the name of the thrift file itself. I then append the name of the processor with the processor suffix stripped since the thrift compiler matches that to the service schema name. Its not perfect. But I think its going to give us more consistent results. Especially since we can do something similar in the thrift client. This option should get us the closest to proper package and service names like grpc. But Im open to alternatives.

mathyourlife-reddit (approved):
thanks @marcoferrer

trevorriles (approved):

fishy (approved):

pacejackson (approved):

marcoferrer:
Rebased and updated the imple to use thrift method descriptors added in `reddit-go/thriftbp/pull/138`

pacejackson (approved):

PR #693, opened 2025-03-27
Use empty string as the fault metric method label for HTTP requests.

HiramSilvey:
In-path parameters were causing this label to explode in cardinality. Until a better solution is worked out, we will drop that label value entirely for HTTP requests. This PR updates the `Inject` function to take in both a `method` and `methodLabel` parameter, since while we don't want to output the label, we do still want the method to be available for matching against the fault configuration in case a very specific fault is desired. Note that matching against all methods is unchanged by just not including a method to match against in the fault configuration, as was always the case. The parameter list was additionally updated to be a struct as it was getting long and unwieldy.

### DIFF
--- a/httpbp/client_middlewares.go
+++ b/httpbp/client_middlewares.go
@@ -15,6 +15,7 @@ import (
 	"github.com/avast/retry-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/reddit/baseplate.go/headerbp"
+	"github.com/reddit/baseplate.go/internal/faults"
 	"github.com/reddit/baseplate.go/secrets"
 
 	"github.com/reddit/baseplate.go/breakerbp"
@@ -467,7 +468,15 @@ func (c clientFaultMiddleware) Middleware() ClientMiddleware {
 			address := req.URL.Hostname()
 			method := strings.TrimPrefix(req.URL.Path, "/")
 			header := httpHeader(req.Header)
-			return c.injector.InjectWithAbortOverride(req.Context(), address, method, &header, resume, abort)
+			return c.injector.Inject(req.Context(),
+				faults.InjectParameters[*http.Response]{
+					Address:     address,
+					Method:      method,
+					MethodLabel: "",
+					Headers:     &header,
+					Resume:      resume,
+					Abort:       abort,
+				})
 		})
 	}
 }
--- a/internal/faults/faults.go
+++ b/internal/faults/faults.go
@@ -21,8 +21,6 @@ import (
 const (
 	promNamespace      = "faultbp"
 	clientNameLabel    = "fault_client_name"
-	serviceLabel       = "fault_service"
-	methodLabel        = "fault_method"
 	protocolLabel      = "fault_protocol"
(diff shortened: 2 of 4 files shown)


mathyourlife-reddit (approved):
any dashboards/alerts/recording rules that'll need to be adjusted for the drop of these labels?

pacejackson (changes requested):
These seem like pretty standard metrics across baseplate so I'm surprised it caused issues here, is there anything different about these?

HiramSilvey:
> any dashboards/alerts/recording rules that'll need to be adjusted for the drop of these labels?

No, they're relatively new so I only had some draft dashboards using them so far.

HiramSilvey:
> These seem like pretty standard labels across baseplate so I'm surprised it caused issues here, is there anything different about these?

The method label in particular ended up using far too much memory in one particularly large service internally. There's some more information in a thread in the baseplate-go channel.

scotthew1 (approved):
@pacejackson the address/method are for the upstream, so this is causing problems with services calling many different endpoints (i.e. graphql)

pacejackson (changes requested):
@HiramSilvey @scotthew1 My concern is that this is being treated as an anomaly for this middleware when this is actually a common pattern within all of our client middlewares.

[headerbp](https://github.com/reddit/baseplate.go/blob/e78dc85026b4c62bf1da3b1619917bd1d4bb2b52/headerbp/metrics.go#L29-L60)
[thriftbp](https://github.com/reddit/baseplate.go/blob/e78dc85026b4c62bf1da3b1619917bd1d4bb2b52/thriftbp/prometheus.go#L69-L82)
[httpbp](https://github.com/reddit/baseplate.go/blob/e78dc85026b4c62bf1da3b1619917bd1d4bb2b52/httpbp/prometheus.go#L95-L132)

So I'm worried that there's either:

A) Something else wrong
B) We are just already sitting close to the edge and this is just what pushed it over
C) GQL is doing something else to mitigate this and we should probably be doing that

pacejackson (approved):

httpbp/client_middlewares.go
```diff
@@ -467,7 +468,15 @@ func (c clientFaultMiddleware) Middleware() ClientMiddleware {
 			address := req.URL.Hostname()
 			method := strings.TrimPrefix(req.URL.Path, "/")
 			header := httpHeader(req.Header)
-			return c.injector.InjectWithAbortOverride(req.Context(), address, method, &header, resume, abort)
+			return c.injector.Inject(req.Context(),
+				faults.InjectParameters[*http.Response]{
+					Address:     address,
+					Method:      method,
+					MethodLabel: "",
```
fishy, line 475: 🔕 this is under `internal` api so we can change this later, thus 🔕, but this is going to be a huge source of confusion in the future.

fishy (approved):

httpbp/client_middlewares.go
HiramSilvey: Ack. I'll think of a better long-term solution for a follow-up PR then.

PR #694, opened 2025-04-07
Update fsnotify

fishy:
And remove the workaround we added in e78dc85 as the issue was fixed upstream, but keep the tests we added in that commit.

While I'm here, also update golang.org/x/net to latest for security fixes.


kylelemons (approved):

mathyourlife-reddit (approved):
🐟

PR #695, opened 2025-04-22
OAM-8189 Add BucketingKey, BucketingValue to ExperimentEvent

julieahnt:
Closes # 

## 💸 TL;DR
Add general field for bucketing key and value instead of adding new rand unit directly as in `device_id`. Instead, we can use the new fields for any new rand units. 
e.g.
- BucketingKey: "post_id"
- BucketingValue: "1k3xqmf"

## 📜 Details
[Jira OAM-8189](https://reddit.atlassian.net/browse/OAM-8189)

## ✅ Checks

### DIFF
--- a/experiments/experiments.go
+++ b/experiments/experiments.go
@@ -399,7 +399,7 @@ type ExperimentEvent struct {
 	// from (optional).
 	DeviceID uuid.UUID
 	// BucketingKey is the key this experiment is being randomized on (optional).
-	BucketingKey string
+	BucketingField string
 	// BucketingValue is the value for the BucketingKey (optional).
 	BucketingValue string
 	// Experiment is the experiment of the applied treatment (required).


kylelemons (commented):
A few quick questions:
* Have you shared more details about this change in #baseplate-go?  There's not a lot to go on here.
* Have you sent a companion PR for the v2 version of this code?
* Do we even need this in baseplate.go v0, or can we put this in the v2 version only?

julieahnt:
Hi @kylelemons, thank you for the comment.

The PR was to add a new field to the event which we will be only used on [GQL](https://github.snooguts.net/reddit/reddit-service-graphql/blob/master/graphql-experiment/internal/service/experiments/expose.go#L66). I don't think there is any equivalent code in V2, However, I am also exploring to migrate GQL to use GRPC event collector, and that is why I mark this PR as draft.

PR #697, opened 2025-05-09
breakerbp: Add jitter to timeout

fishy:
A typical use case of circuit breaker would often have this scenario happen:

1. Client pods have circuit breaker set up for server, with same settings
2. Server goes into an outage
3. All client pods' circuit breakers tripped at roughly the same time
4. Server recovered from outage
5. Because all client pods' circuit breakers are tripped, server does not get a lot of request, its HPA is running near minimal
6. All client pods' circuit breakers' timeout happen at roughly the same time, they all start to send requests to the server again at roughly the same time
7. Server's HPA is not able to scale up quickly enough with the sudden influx of traffic, so the load crashes the server pods again

So, adding a jitter to the timeout will help spread the recovering traffic to the server, which should help server to recover more smoothly.

### DIFF
--- a/breakerbp/failure_ratio.go
+++ b/breakerbp/failure_ratio.go
@@ -2,7 +2,7 @@ package breakerbp
 
 import (
 	"context"
-	"fmt"
+	"log/slog"
 	"time"
 
 	"github.com/apache/thrift/lib/go/thrift"
@@ -12,11 +12,11 @@ import (
 
 	"github.com/reddit/baseplate.go/internal/prometheusbpint"
 	"github.com/reddit/baseplate.go/log"
+	"github.com/reddit/baseplate.go/randbp"
 )
 
 const (
-	promNamespace = "breakerbp"
-	nameLabel     = "breaker"
+	nameLabel = "breaker"
 )
 
 var (
@@ -25,9 +25,8 @@ var (
 	}
 
 	breakerClosed = promauto.With(prometheusbpint.GlobalRegistry).NewGaugeVec(prometheus.GaugeOpts{
-		Namespace: promNamespace,
-		Name:      "closed",
-		Help:      "0 means the breaker is currently tripped, 1 otherwise (closed)",
+		Name: "breakerbp_closed",
+		Help: "0 means the breaker is currently tripped, 1 otherwise (closed)",
 	}, breakerLabels)
 )
 
@@ -39,7 +38,7 @@ type FailureRatioBreaker struct {
 	name              string
 	minRequestsToTrip int
 	failureThreshold  float64
-	logger            log.Wrapper
+	logContext        context.Context
 }
 
 // Config represents the configuration for a FailureRatioBreaker.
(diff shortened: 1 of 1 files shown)


breakerbp/failure_ratio.go
```diff
@@ -85,12 +95,20 @@ func NewFailureRatioBreaker(config Config) FailureRatioBreaker {
...
-		logger:            config.Logger,
+	}
+	if config.LogContext != nil {
+		failureBreaker.logContext = config.LogContext
+	} else {
+		failureBreaker.logContext = context.Background()
+	}
+	jitterRatio := 0.5
+	if config.TimeoutJitterRatio != nil {
+		jitterRatio = *config.TimeoutJitterRatio
```
pacejackson, line 106: Should we maybe log a message if we get a jitter ratio that would be normalized (so if it's <= 0 or > 1)?

breakerbp/failure_ratio.go
```diff
@@ -76,6 +81,11 @@ type Config struct {
 
 	// Timeout is the duration of the 'Open' state. After an 'Open' timeout duration has passed, the breaker enters 'half-open' state.
 	Timeout time.Duration `yaml:"timeout"`
+
+	// TimeoutJitterRatio is the jitter ratio to be applied to Timeout.
+	//
+	// Optional, default is 0.5, means the actual timeout used can be Timeout+-50%.
+	TimeoutJitterRatio *float64 `yaml:"timeoutJitterRatio"`
```
pacejackson, line 88: 🔕 I don't think this really needs to be a pointer since `0` is normalized by [randbp.JitterRatio](https://github.com/reddit/baseplate.go/blob/4020824b36e41451704c4213f4af376aa65f9015/randbp/jitter.go#L14-L16) to be 1 anyways, we could probably just treat 0 as "not set"?

pacejackson (changes requested):
Just a couple small questions.

breakerbp/failure_ratio.go
fishy: A sane default is important :) If the jitter ratio is (near) 1, we could get to a very very short jittered timeout, which can be bad.

pacejackson: To be clear, what I meant was if it's set to `0` we treat that as `not set` and set it to the default value.

fishy: but comparing float numbers to 0 can be tricky, so pointer is better :)

pacejackson: Fair point 😅

pacejackson (approved):

breakerbp/failure_ratio.go
```diff
@@ -151,8 +192,13 @@ func (cb FailureRatioBreaker) stateChanged(name string, from gobreaker.State, to
 		nameLabel: cb.name,
 	}).Set(value)
 
-	message := fmt.Sprintf("circuit breaker %v state changed from %v to %v", name, from, to)
-	cb.logger.Log(context.Background(), message)
+	slog.InfoContext(
+		cb.logContext,
+		"circuit breaker state changed",
```
kylelemons, line 197: 🔕 snoolog is like LogDNA, it doesn't really display these well.  You'll just see a flood of "circuit breaker state changed" and will have to jump through hoops just to see "circuit breaker state changed from X to Y" which is what you basically want to see as a human. Can we format the log message so it has everything a human needs, and leave the kvs to be just for filtering?

(I'm not sure how much we want to care about this in infrared v0 for the debug case above, though, since we don't have a great way to save the allocs it would take to format it if it's not being printed.)

kylelemons (approved):

breakerbp/failure_ratio.go
fishy: when upstream is healthy the state would never change and this log won't appear at all. so when this log appears, even if it's changing from open to closed (recover), it means the state changed to open before so it's not exactly false alarm.

and have kv pairs of the from/to states make it much easier to search/filter.

PR #700, opened 2025-07-14
[Faults] Add support for matching on the host/authority header.

HiramSilvey:
The PRs adding this to bpv2 have already been approved, and this PR is just to get bpv0 up-to-date with those.

This introduces support for a new `h` parameter in the fault configuration header value. Specifically, if the `h` parameter is present, then the value of it will be checked against the host/authority header of the outgoing request before determining whether the request matches for fault injection.

Note that this is _not_ supported for gRPC due to the authority header not being exposed by client-middleware runtime. If you'd like more info on the workaround for gRPC, I'm happy to DM information on what we plan to do.

### DIFF
--- a/internal/faults/headers_test.go
+++ b/internal/faults/headers_test.go
@@ -63,6 +63,7 @@ func TestParseMatchingFaultHeader(t *testing.T) {
 		name             string
 		headerValue      string
 		canonicalAddress string
+		host             string
 		method           string
 		abortCodeMin     int
 		abortCodeMax     int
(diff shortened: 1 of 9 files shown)


thriftbp/client_middlewares.go
```diff
@@ -27,6 +27,10 @@ import (
 	"github.com/reddit/baseplate.go/transport"
 )
 
+// HostHeader is the Thrift equivalent of the HTTP1 Host header and HTTP2 :authority pseudo-header.
+// Exposed for tests.
+const HostHeader = "thrift-hostname"
```
fishy, line 32: we already have this defined: https://pkg.go.dev/github.com/reddit/baseplate.go/thriftbp#ThriftHostnameHeader

internal/faults/headers_test.go
```diff
@@ -272,12 +286,13 @@ func TestParsingFaultConfiguration(t *testing.T) {
 		},
 	}
 
+	testHost := ""
 	testMethod := ""
 	testAbortCodeMin, testAbortCodeMax := 0, 0
```
pacejackson, line 291:
```suggestion
	var testHost, testMethod string
	var testAbortCodeMin, testAbortCodeMax int
```

pacejackson (commented):
one last small style suggestion

PR #705, opened 2025-12-23
Differentiate servers running within a single runtime

gregbarasch:
## 💸 TL;DR
Our thrift server metrics assume that there will only be one server running within a single runtime. This is not always true. If a second server is running, the thrift server metrics will clash with one another. This PR adds a serverName label to our serverside metrics.

## 📜 Details
This update is backwards compatible, and notably should not increase cardinality in any significant way. Preexisting metric usage will emit "unknown" as the serverName, which will retain metric cardinality and will add a negligible amount of data for metric DBs. 

The API itself does not enable backward compatible attribute changes. We have a choice. We can: 
1. inject the serverName into the ctx on each request
2. Always call  GetThriftServiceName(cfg.Processor) -- needs testing
3. (The option chosen) Provide an alternative WithArgs function 

Commonly, we might make use of the optionals or builder pattern for forward looking, backwards compatible changes. Instead though we've opted to use a Cfg struct, as this matches patterns found elsewhere in the repo.  

Unit tests

## ✅ Checks

### DIFF
--- a/internal/prometheusbpint/spectest/spec.go
+++ b/internal/prometheusbpint/spectest/spec.go
@@ -180,7 +180,6 @@ func buildLabels(name, prefix, clientOrServer string) map[string]struct{} {
 //
 // active_requests metrics expect the following labels:
 //   - "method"
-
 func thriftSpecificLabels(name string) []string {
 
 	labelSuffixes := []string{"method"}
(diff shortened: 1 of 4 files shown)


konradreiche (commented):
This makes sense to me but I believe we don't have the equivalent in baseplate v2. I've added @kylelemons for additional perspective.

kylelemons (commented):
We can't add additional labels here without breaking compatibility with thriftbp/v2 via interop, and unfortunately such breakages cause _runtime_ metrics blackouts.

This should probably be brought up in #baseplate-spec to weigh the alternatives

